PR #90, opened 2014-08-01
First version of the variants sorting algorithm 

fnaum:
First version of the variants sorting algorithm (WIP), just to get some feedback and know if we are in the same page heading in the right direction.

Need to add more test cases to fully test it (i.e asymmetric variants list with the same weight)

The hashing we discussed for the ambiguous cases seems that it should not be necessary since the default sorting takes care of those cases and guarantees repetitive results regardless of the variants positions on the package.yaml 
I think we should guarantee the same result if the packages are changed in order (line in which they appear).. but if the contents are changed (position of the packages inside a variant, or new variants lines are added/removed) then we can not guaranteed (should not expect) the same result.


nerdvegas:
Superceded by #165. Let's move all convo there.

PR #121, opened 2014-10-31
Handling quoted values in shells

fnaum:
Sometimes we need to set values that contains quotes, and we don't want the shell to interpret them. So we need to escape them. 
with the old command, escaping like so will be good for bash but not for tcsh
` - export VARIABLE_WITH_QUOTES="loadPlugin(\"mayaPlugin\")"`

with  new commands
`setenv("VARIABLE_WITH_QUOTES", 'loadPlugin("mayaPlugin")')`
they need to be escaped in both tcsh and bash

Added some test cases to show this. Also fixed the other test so they really run in the shell. The **@shell_dependent** function decorator is only changing the value in the config, we need to create the shell with that value and also pass it to  `excecute_shell`

### DIFF
--- a/src/rez/tests/data/commands/packages/shelltest/1/package.yaml
+++ b/src/rez/tests/data/commands/packages/shelltest/1/package.yaml
@@ -1,0 +1,5 @@
+name: shelltest
+version: '1'
+
+commands:
+  - export VARIABLE_WITH_QUOTES="loadPlugin(\"mayaPlugin\")"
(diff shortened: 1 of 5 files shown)


nerdvegas:
Closing this, as it's basically a dup of " trying to fix escaping and quoting #134 "
And I have aa fix available for testing. Forgot to mention - I've also tested it on every shell type, tests all passing.

fnaum:
:+1:

fnaum:
:+1:

PR #134, opened 2014-12-09
trying to fix escaping and quoting

fnaum:
Hi Allan,

What I'm currently doing here is attempting to fix the escaping and quoting issue I mentioned before.

Basically I quote the whole string instead of the individual items when there is a list of values separated by an `env_sep`

so basically we end up with python strings like    `'"/path1:/path2"'   '"/path1;/path2"'` which are what we expect when they get interpreted by any shell

but with that change and adapting some test cases, there is one test case failing on `tests_context`. (probably the only piece of code calling `execute_command`) 

That fails  because the `PATH` end up like `'"/path1:/path2"'` with is totally fine if it gets interpreted by any shell but is breaks when is interpreted by python itself
The `subprocess` module does a naive split by `os.path.pathsep` ending up like
`"/path1`
`/path2"`

So these change does not fix the issue

### DIFF
--- a/src/rez/rex.py
+++ b/src/rez/rex.py
@@ -9,7 +9,7 @@ from string import Formatter
 from rez.system import system
 from rez.config import config
 from rez.exceptions import RexError, RexUndefinedVariableError
-from rez.util import AttrDictWrapper, shlex_join, expandvars
+from rez.util import AttrDictWrapper, shlex_join, expandvars, remove_enclosing_quotes
 from rez.vendor.enum import Enum
 
 
@@ -336,22 +336,18 @@ class ActionManager(object):
         if expanded_key in self.environ:
             env_sep = self._env_sep(expanded_key)
             self.actions.append(action(unexpanded_key, str(unexpanded_value)))
-            parts = self.environ[expanded_key].split(env_sep)
+            parts = remove_enclosing_quotes(self.environ[expanded_key]).split(env_sep)
 
-            unexpanded_values = env_sep.join(
-                addfunc(self._escape(unexpanded_value),
-                        [self._keytoken(expanded_key)]))
+            unexpanded_values = addfunc(self._escape(unexpanded_value), [self._keytoken(expanded_key)], env_sep)
 
-            expanded_values = env_sep.join(
-                addfunc(self._escape(expanded_value), parts))
+            expanded_values = addfunc(self._escape(expanded_value), parts, env_sep)
 
-            self.environ[expanded_key] = \
-                env_sep.join(addfunc(str(expanded_value), parts))
+            self.environ[expanded_key] = addfunc(str(expanded_value), parts, env_sep)
         else:
             self.actions.append(Setenv(unexpanded_key, str(unexpanded_value)))
             self.environ[expanded_key] = str(expanded_value)
-            unexpanded_values = self._expand(unexpanded_value)
-            expanded_values = self._expand(expanded_value)
+            unexpanded_values = self._escape(self._expand(unexpanded_value))
+            expanded_values = self._escape(self._expand(expanded_value))
             interpfunc = None
 
         applied = False
@@ -374,12 +370,27 @@ class ActionManager(object):
             self.interpreter.setenv(key, value)
 
     def prependenv(self, key, value):
-        self._pendenv(key, value, Prependenv, self.interpreter.prependenv,
-                      lambda x, y: [x] + y)
+        def prepend_quoted(x, y, sep=None):
+            if sep:
+                value_list = [x] + y
+                # remove the quotes of individual items on the list and return the joined string quoted
+                value_list = [remove_enclosing_quotes(x) for x in value_list]
+                return '"%s"' % sep.join(value_list)
+            else:
+                return [x] + y
+        self._pendenv(key, value, Prependenv, self.interpreter.prependenv, prepend_quoted)
 
     def appendenv(self, key, value):
-        self._pendenv(key, value, Appendenv, self.interpreter.appendenv,
-                      lambda x, y: y + [x])
+        def append_quoted(x, y, sep=None):
+            if sep:
+                value_list = y + [x]
+                # remove the quotes of individual items on the list and return the joined string quoted
+                value_list = [remove_enclosing_quotes(x) for x in value_list]
+                return '"%s"' % sep.join(value_list)
+            else:
+                return y + [x]
+
+        self._pendenv(key, value, Appendenv, self.interpreter.appendenv, append_quoted)
 
     def alias(self, key, value):
         key = str(self._format(key))
--- a/src/rez/rex.py
+++ b/src/rez/rex.py
@@ -373,9 +373,12 @@ class ActionManager(object):
         def prepend_quoted(x, y, sep=None):
             if sep:
                 value_list = [x] + y
-                # remove the quotes of individual items on the list and return the joined string quoted
-                value_list = [remove_enclosing_quotes(x) for x in value_list]
-                return '"%s"' % sep.join(value_list)
+                if sep != os.path.pathsep:
+                    # remove the quotes of individual items on the list and return the joined string quoted
+                    value_list = [remove_enclosing_quotes(x) for x in value_list]
+                    return '"%s"' % sep.join(value_list)
+                else:
+                    return sep.join(value_list)
             else:
                 return [x] + y
         self._pendenv(key, value, Prependenv, self.interpreter.prependenv, prepend_quoted)
@@ -384,9 +387,12 @@ class ActionManager(object):
         def append_quoted(x, y, sep=None):
             if sep:
                 value_list = y + [x]
-                # remove the quotes of individual items on the list and return the joined string quoted
-                value_list = [remove_enclosing_quotes(x) for x in value_list]
-                return '"%s"' % sep.join(value_list)
+                if sep != os.path.pathsep:
+                    # remove the quotes of individual items on the list and return the joined string quoted
+                    value_list = [remove_enclosing_quotes(x) for x in value_list]
+                    return '"%s"' % sep.join(value_list)
+                else:
+                    return sep.join(value_list)
             else:
                 return y + [x]
 
--- a/src/rez/tests/test_rex.py
+++ b/src/rez/tests/test_rex.py
@@ -86,8 +86,8 @@ class TestRex(TestBase):
                        Source('./script.src')],
                    expected_output = {
                        'FOO': 'foo',
-                       'A': os.pathsep.join(["/data","/tmp"]),
-                       'B': os.pathsep.join(["/tmp","/data"])})
+                       'A': '"%s"' % os.pathsep.join(["/data","/tmp"]),
+                       'B': '"%s"' % os.pathsep.join(["/tmp","/data"])})
 
     def test_2(self):
         """Test simple setenvs and assignments."""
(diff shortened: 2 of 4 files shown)


federicon:
Hi Allan, sorry for not replying earlier, I just came from  my longest holidays ever and probably the longest period without touching a computer :)

There were two problems that were explained [here] (https://github.com/nerdvegas/rez/pull/121) 

1.  The escaping on the plugin (_escape_string) does not get called when there is prepend/append  (instead of setenv)  
i.e  A package with variable like:

`export BLA=loadPlugin("mayaPlugin"):${BLA}`

ends up ( (in the context file)) unescaped,  like :

`setenv BLA loadPlugin("mayaPlugin")`

both **tcsh** and **bash** are broken.
2. The quoting fails when the env_var_separators is not os.path.pathsep (ie CMAKE_MODULE_PATH which  uses ';' instead of ':' ) It does quote the individual parts of a multi path variable instead of quoting the whole thing

i.e using
OLD COMMAND: 
`export CMAKE_MODULE_PATH=!ROOT!/cmake';'$CMAKE_MODULE_PATH`

ends with (in the context file)
`setenv CMAKE_MODULE_PATH "/film/tools/packages/stdlib/4.8.2/CentOS6.2_thru_7/cmake";${CMAKE_MODULE_PATH}`

Note the semicolon outside the quoting. Both **tcsh** and **bash** will fail evaluating this line. They will try to execute ${CMAKE_MODULE_PATH} 


To reproduce them: Use your latest **2.0** branch , create the following package 

    config_version : 0
    uuid: 97937642-3646-4966-ab9c-8f337bbbce6a
    name: aaa
    version: 1.0.0
    authors: [ rez ]
    description: test
    private_build_requires:
     - cmake-2.8
    commands:
     - export BLA=loadPlugin("mayaPlugin"):${BLA}

Do `rez env aaa` , and it  will unveil the issue number **1**

Create the following now

    config_version : 0
    uuid: 97937642-3646-4966-ab9c-8f337bbbce6b
    name: bbb
    version: 1.0.0
    authors: [ rez ]
    description: test2
    private_build_requires:
     - cmake-2.8
    commands:
    - export CMAKE_MODULE_PATH=!ROOT!/cmake';'$CMAKE_MODULE_PATH

Do `rez env bbb` , and it  will unveil the issue number **2**

nerdvegas:
Hey Fede, welcome back! I bet you're sick of holidays and raring to go on
more python development.

Ok three points:

1) Could we limit the discussion to Rex examples, rather than old-style
bash? I just want to avoid cases where there's actually a bug in the
old-to-new command conversion, as opposed to a bug in Rex itself - better
not two confuse these two separate issues, I think it will be easier for
all of us.

2) Do you know about the rez-interpret tool? It will interpret arbitrary
Rex code in a script, and convert to any target format - you don't have to
muck about with packages to test this stuff.

3) I'm moving onto a non-rez-related major project now so my time on Rez is
very restricted. I'm really hoping to have another developer take on the
Rez workload soon through work, but how soon and to what capacity is out of
my hands :(. This is a bummer I know. I will give as much help as I can in
the meantime though.

Thanks,
A

federicon:
Thanks Allan, I have an itching sensation since yesterday, Mark was suggesting I might have become allergic to work :)

Point 3 is a real bummer, I hope your new project is exciting... and I hope we and other people can contribute efforts to get rez 2 out soon.

I did not knew about the rez-interpret tool, that makes things much easy.

I think that the 2 issues are in the REX code nor in the old-to-new conversion:

Issue number 1 can be shown with the following code

`env.BLA.append('loadPlugin("mayaPlugin")')`

    >rez-interpret rexcode -f bash
    #!/bin/bash
    export BLA=loadPlugin("mayaPlugin")

    > rez-interpret rexcode -f tcsh
    #!/bin/tcsh
    setenv BLA loadPlugin("mayaPlugin")

As you can see no escaping happens.

If we set a  value before 
    env.BLA = "tmp"
    env.BLA.append('loadPlugin("mayaPlugin")')

Then we get the right escaping

    > rez-interpret rexcode -f tcsh
    #!/bin/tcsh
    setenv BLA "tmp"
    setenv BLA ${BLA}:"loadPlugin("\""mayaPlugin"\"")"

But we have many packages that assume the a path might have already be set so the go with 
    env.BLA.append('loadPlugin("mayaPlugin")')
    or
    export BLA=loadPlugin("mayaPlugin"):${BLA}

For issue number 2 use this rexcode

    env.CMAKE_MODULE_PATH = "/tmp"
    env.CMAKE_MODULE_PATH.append("/var/tmp")

    > rez-interpret rexcode -f tcsh
    #!/bin/tcsh
    setenv CMAKE_MODULE_PATH "/tmp"
    setenv CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH};"/var/tmp"

    > rez-interpret rexcode -f bash
    #!/bin/bash
    export CMAKE_MODULE_PATH="/tmp"
    export CMAKE_MODULE_PATH=${CMAKE_MODULE_PATH};"/var/tmp"

Both are incorrect, when the shell try to set that they will see the second line as two different command and try to execute /var/tmp

The expected result for these would be:
    
    setenv CMAKE_MODULE_PATH "/tmp"
    setenv CMAKE_MODULE_PATH" ${CMAKE_MODULE_PATH};/var/tmp"
and
    export CMAKE_MODULE_PATH="/tmp"
    export CMAKE_MODULE_PATH="${CMAKE_MODULE_PATH};/var/tmp"


Hope this is more clear now
Fede

fnaum:
Hey Allan,
   I guess you are busy with other non rez projects but....
   Did you have a change to look into this one? at least to see if you can repro what I reported?
Cheers
Fede

mstreatfield:
Hi Allan,

It would be great if we could get a bit of insight on this one to help us on our way.   I've spent most of the afternoon trying different combinations of rex commands with only limited success.  Which is quite frustrating given we already had a solution which seemed to work ok.

For issue 1,  I think I need this:

    appendenv("BLAH", expandable('loadPlugin("A")'))
    appendenv("BLAH", expandable('loadPlugin("B")'))
    appendenv("BLAH", expandable('loadPlugin("C")'))

However if BLAH isn't set before the first `appendenv` then the escaping doesn't happen.  I think this is line 353 and 353 of `rex.py` needing to call `self._escape` instead of `self._expand`.  Is it that simple?

Issue 2 I think is unresolvable in the current implementation.  It works fine if the default os separator is used but we have many situations where this is unfortunately not the case.  And indeed, even within Rez `CMAKE_MODULE_PATH` is wired to use `;` and so no longer works out of the box.  Are you not using the `CMAKE_MODULE_PATH` anywhere and so are unaffected by it?

Mark.

nerdvegas:
I need to look at this but I haven't had time. I was supposed to have a
developer looking into this but he keeps getting dragged onto other work -
hopeful that changes soon. We're lucky in that it is not affecting us
because we happen not to be building anything compiled via rez-build. But
I'm sure we will hit the problem as well.

The literal/expandable work was done because, even though the previous
escaping was working, it didn't leave room for supporting non-expanding
strings as well. As soon as that feature request comes up (and it's a
reasonable one imo) this expansion would have to have been reworked anyway.

I have some ideas on sorting this out. Bear in mind that expandable/literal
should be supported in *all* rex commands, not just set/append/prepend (for
eg, source etc). I can probably push to get a whole day of work in either
tomorrow or Monday, since this other developer has not been available and
Rez has been languishing. In any case as soon as I get the time this is
what I'll be looking at fixing. I'll keep you posted.

Thx
A




On Wed, Mar 11, 2015 at 11:19 PM, Mark Streatfield <notifications@github.com

chadrik:
I haven't had time to keep up with this thread, but in case it makes a difference, I want to point out that `CMAKE_MODULE_PATH` is not a real [cmake environment variable](http://www.cmake.org/Wiki/CMake_Useful_Variables#Environment_Variables).  It's provided by rez, and the reason that semi-colons are used is just a hack, so that the value can be passed directly to the cmake command, which uses semi-colons to separate array items (note, this behavior has been there since rez 1.0). 

So, if the semi-colons are causing a problem, I think we should just ditch support for them and make `CMAKE_MODULE_PATH` work like a normal env variable, which will require something like `value.replace(os.path.pathsep, ';')` before passing to cmake.

nerdvegas:
Hey Chad,

You're right, it's been like that for so long that I'd completely forgotten
that this isn't actually an environment variable.

So although it doesn't address the overall string escaping issue, I think
it would be reasonable to switch to a colon-separated CMAKE_MODULE_PATH and
do the string replacement in the cmake plugin, as Chad suggests. This is
actually what Rez-1 used to do, and in light of that fact that this is a
cmake variable and not an env-var that it expects, this is actually more
correct behaviour.

Mark, would this help in the interim? I daresay we could remove the pesky
special-case path separator support at the same time - unless someone out
there is actually relying on that.

Thx
A

mstreatfield:
> Mark, would this help in the interim?

No.  We have other environment variables that use separators other than `os.pathsep`.  This includes ` `, `,` and `;`.  It is not practical for us to update these variables to use `os.pathsep`, mostly because these variables include `:` as a value in the list and so can't be split on that character.

I am also not sure it's safe to assume that the escaping is consistent across all shells.  I didn't really look into this while implementing the Windows support (and don't have a depth of knowledge to know for sure), but it feels likely that other shells might require different escaping rules (such as Windows Command Prompt).  Therefore making it the sole responsibility of Rex is maybe not right.  For example, compare this on Windows Command Prompt:

    > set a="1 2 3"
    > echo %a%
    "1 2 3"
    > set a=1 2 3
    > echo %a%
    1 2 3

With the bash equivalent:

    > export a="1 2 3"
    > echo $a
    1 2 3
    > export a=1 2 3
    bash: export: '2': not a valid identifier
    bash: export: '3': not a valid identifier

Based on my understanding of the recent changes, Rex is always wrapping the string in `"` which results in different behaviour on Windows.  So the Windows shell implementation needs to retain some control over the processing.  But @chadrik, you seem to have the most Windows knowledge of all of us so can maybe tell me if I've over simplified that.

I keep getting confused between the different objects and code paths in Rex/Shells, but I think at the moment the shell plugins are passed an already escaped and separator joined string for (for example) the `setenv` method.  Perhaps they should be passed a higher level object (including the `literal` or `expandable` etc) and then handle the quoting itself based on what it's been given?

nerdvegas:
FYI, I have the go ahead to dedicate tomorrow to this. Will mail an update after that.

Wrt passing expandable/literal object to shell so expansion happens later - this is the same idea I have had - will investigate.

Thx
A

nerdvegas:
Hi all,

I've implemented a fix for this and am fairly confident it's working. You can find it in the "string_escape2" branch. Once it's gotten a few weeks of testing I'll be closing this PR. Thanks for your work on this Fede.

As suggested in previous comments, the fix does indeed involve passing EscapedString objects to the ActionInterpreter subclasses so they can decide how to best handle them (previously the ActionManager was doing this before passing it onto the interpreter).

You should be able to set/append/prepend any combination of expandable and literal strings with the new code.

fnaum:
awesome, we'll try to merge that branch and do more testing on our side

nerdvegas:
Cool, I'll close for now. Will reopen if there's a problem.

mstreatfield:
Hey @nerdvegas,

Thanks for this - I took a quick look at it today and it seems much better.  Although I think I am right in saying the rex unit tests are broken in this branch... the expected values for most of the rex commands are missing an `EscapedString` instance?  

Also, we have a unit test for building a package which contains a `!` requirement.  It looks like this isn't being escaped at the moment so is causing errors in bash, but I'm still looking at that.

But apart from that so far so good.

nerdvegas:
Oh yes, those UT bugs have been fixed in the issue_143 branch

PR #135, opened 2014-12-15
Changelog Maxsize Setting

mstreatfield:
Add a setting to `rezconfig` that allows the changelog to be trimmed to a maximum size.  This prevents overly large `release.yaml` files (7MB is our largest yet) which greatly increase the resolve time in `rez env`.


mstreatfield:
It looks like this has been incorporated in resources2 branch so is perhaps no longer required.

nerdvegas:
This is already implemented in resources2.

PR #138, opened 2014-12-16
rez yaml2py command

mstreatfield:
A small utility to convert a `package.yaml` file to `package.py` using the `package_maker` module.

When the resource system has been updated so it can write package files directly this code can be updated to use that.  In the meantime, this should do the trick.


mstreatfield:
Sure, no worries.  Will leave this for you to close if you decide against it.

nerdvegas:
There is an equivalent available in resources2.

PR #139, opened 2014-12-17
Build Variant Index Environment Variable

mstreatfield:
This change introduces an environment variable for the CMake build system specifying the index of the current variant being built.

We use this to ensure Doxygen documentation (which is quite large) is only built for a specific (generally first) variant without having to look for a specific package in the environment.


nerdvegas:
I've implemented this in resources2 just now, but with some changes.
Rather than adding the 'get_current_variant_index' function, I've updated the build system API so that current build variant is passed to BuildSystem.build(). This way we're not trying to infer the current variant from the resolved environment (an approach that is error-prone in the long term imo).

Also, I've set the variable REZ_BUILD_VARIANT_INDEX rather than REZ_BUILD_VARIANT_NUMBER, since this matches the language in the rest of the project.

It's tested and appears to work, so closing for now. Let me know if you hit any problems.

PR #140, opened 2014-12-17
Sort print_info on ResolvedContext

mstreatfield:
Add an argument to allow the output of `print_info` method on a `ResolvedContext` to be alphabetically sorted.

We discussed making this configurable so that the global behaviour can be set for `rez env`, as well as adding a command line option.  I haven't had time to to this, and most likely wont before Christmas.  This keeps current behaviour and gets us part of the way there - I'll try and finish it off with a new pull request next year.

### DIFF
--- a/src/rez/resolved_context.py
+++ b/src/rez/resolved_context.py
@@ -591,8 +591,13 @@ class ResolvedContext(object):
             d["removed_packages"] = removed_packages
         return d
 
-    def print_info(self, buf=sys.stdout, verbosity=0):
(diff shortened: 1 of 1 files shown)


nerdvegas:
I've added this to 2.0. Since I've only ever gotten complaints about the default ('source') order, I don't think anyone is going to be impacted much by changing the default ordering.

So, the default is now alphabetical, and this can be overridden using the new rez-context flag "--source-order".

PR #141, opened 2014-12-17
LOCAL_SYMLINK Option in CMake Macros

mstreatfield:
Add an option `LOCAL_SYMLINK` to the `rez_install_files` and `rez_install_dirs` macros to create symlinks.

When installing files, this will create a symlink to the original source file, instead of performing a copy.  This is useful for developing files that don't require a compilation step (and so don't need rebuilding) to enable a quicker iteration for the developer.  Example of such files are mel, otls and (somewhat unfortunately I think) python.

A central `rez release` overrides this behaviour - it is not possible to release a symlink.

### DIFF
--- a/src/rezplugins/build_system/cmake_files/InstallDirs.cmake
+++ b/src/rezplugins/build_system/cmake_files/InstallDirs.cmake
@@ -7,12 +7,15 @@
 # a .svn dir) are excluded. If this macro does not provide enough fine-grain control for
 # your needs, then you should use cmake's install(DIRECTORY) macro instead (and you should
 # use the values REZ_FILE_INSTALL_PERMISSIONS and REZ_EXECUTABLE_FILE_INSTALL_PERMISSIONS
-# to specify the permissions you want).
+# to specify the permissions you want).If LOCAL_SYMLINK is preset it would create a symlink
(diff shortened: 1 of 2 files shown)


nerdvegas:
I've merged this into resources2.

nerdvegas:
Hey @mstreatfield,

That was the code in your PR. But I've gone ahead and fixed it, as well as added LOCAL_SYMLINK support for rez_install_python.

Changes in latest resources2.

mstreatfield:
I believe the `rez_install_files` macro requires a similar update for the same reason.

I also discovered over the weekend that we had modified `rez_install_dirs` since the pull request was originally made.  We had to add an extra `/` in the path when making a symlink, so references to:

    install(CODE "execute_process(COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_INSTALL_PREFIX}${dest_dir})" )

Become 

    install(CODE "execute_process(COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_INSTALL_PREFIX}/${dest_dir})" )

This affects lines 62, 65 and 66.  Sorry, I'm not able to make another pull request for it at the moment to properly make those changes.  And I think because it was an old pull request (4 months) we forgot to keep it up-to-date).

PR #142, opened 2014-12-17
Targeted Resource Location

mstreatfield:
Some resources have a static pattern for the file being located (no wildcards).  This means we can look for them directly rather than looking at every child of the parent resource.  For things like `release.yaml` this can make quite a difference in resolution time.

I know you are working on the resource system anyway, but we've had this change locally for quite a while and it's working well.  Until the larger work is done, every little helps!


mstreatfield:
This is not required in the 'resources2' branch (I don't think).

nerdvegas:
Not required in resources2 branch, closing.

PR #144, opened 2014-12-17
Misc. Build Process Changes

mstreatfield:
+ Call `make ; make install` instead of just `make install`.  This is closer to Rez 1 behaviour and works around a problem in the ExternalProjectAdd macro of CMake.

+ Ensure install paths are always made.  This includes the `REZ_LOCAL_PACKAGES_PATH` value as well as the variant subfolder.  Previously this was taken care of by CMake macros when installing `package.yaml` files etc.  Now that is no longer the case not having them made can cause build issues.

### DIFF
--- a/src/rez/build_process.py
+++ b/src/rez/build_process.py
@@ -148,12 +148,13 @@ class StandardBuildProcess(BuildProcess):
     def release(self):
         assert(self.vcs)
         install_path = self.package.config.release_packages_path
+
+        if not os.path.exists(install_path):
+            os.mkdir(install_path)
+
         base_build_path = os.path.join(self.working_dir,
                                        self.package.config.build_directory,
                                        "release")
-
-        if not os.path.exists(install_path):
-            raise ReleaseError("Release path does not exist: %r" % install_path)
 
         print "Checking state of repository..."
         self.vcs.validate_repostate()
@@ -377,6 +378,9 @@ class LocalSequentialBuildProcess(StandardBuildProcess):
             if r.status != ResolverStatus.solved:
                 raise BuildContextResolveError(r)
 
+            if install and not os.path.exists(install_path):
+                os.makedirs(install_path)
+
             # run build system
             self._pr("\nInvoking %s build system..." % self.buildsys.name())
             ret = self.buildsys.build(r,
@@ -392,8 +396,6 @@ class LocalSequentialBuildProcess(StandardBuildProcess):
 
                 extra_files = ret.get("extra_files", []) + [rxt_file]
                 if install and extra_files:
-                    if not os.path.exists(install_path):
-                        os.makedirs(install_path)
                     for file in extra_files:
                         shutil.copy(file, install_path)
             else:
--- a/src/rezplugins/build_system/cmake.py
+++ b/src/rezplugins/build_system/cmake.py
@@ -154,8 +154,6 @@ class CMakeBuildSystem(BuildSystem):
         # assemble make command
         cmd = ["make"]
         cmd += (self.child_build_args or [])
-        if install and "install" not in cmd:
-            cmd.append("install")
 
         # execute make within the build env
         _pr("\nExecuting: %s" % ' '.join(cmd))
@@ -163,6 +161,16 @@ class CMakeBuildSystem(BuildSystem):
                                               block=True,
                                               cwd=build_path,
                                               actions_callback=callback)
+        if not retcode and install and "install" not in cmd:
+            cmd.append("install")
+
+            # execute make install within the build env
+            _pr("\nExecuting: %s" % ' '.join(cmd))
+            retcode, _, _ = context.execute_shell(command=cmd,
+                                                  block=True,
+                                                  cwd=build_path,
+                                                  actions_callback=callback)
+
         ret["success"] = (not retcode)
         return ret
 


mstreatfield:
Ok, so I might need a little further refactory.  Should the steps be:

    pre build hook
        make (clean)
    post build hook
    pre release hook
        make install
    post release hook

Or:

    pre build hook
        pre release hook
            make
            make install
        post release hook
    post build hook

nerdvegas:
The latter I would think, since in the case of your external projects, 'make' actually does do the install?

Would it be a bit less confusing if pre/post_build is only called for builds/local installs, and pre/post_release is only called when a release occurs (ie, pre/post_build is NOT called during a release at all)? I can see there being cases where users don't want a hook used during build to also happen during a release. It may mean a little redundancy for some users (they might have to configure the same hooks for build and release), but imo the extra flexibility is worth it.

nerdvegas:
To add to this:

The recent resources update (ready very soon) means that you can now install variants from one package path into another, via the API. The build/release process now uses this - instead of just copying the package.yaml into the install path, it actually installs the variant correctly - and this is what makes installing variants into existing packages now possible.

But what this also means is there will be cases where we might need a release-type-hook in situations that aren't actually a release or a build per se. Two cases I can think of - ingestion (what is currently 'rez-bind'), and Mark's example of 'promotion' of a beta package to a release.

So what I'm getting at is perhaps we need a third hook - pre/post_install let's say. We will need something like that at Method for example, because all our release package locations are write-protected, and we need to run a script to enable/disable writes.

If we're going to need this (and I'm thinking that we will) I think it will be too complicated to overlap any of these hooks. So then, *build is used with 'rez-build', *release with 'rez-release' and *install with 'rez-bind' and any other tool or process that does variant installation.

nerdvegas:
Ok so I've decided to dive in and rework buildprocess.py - I kinda have to as part of the resources update (due to unrelated issues). I'm going to switch it to a new plugin type, 'build_process'.

A problem with the release hooks currently is that they only trigger before all variants are released, and after. This is going to be problematic when we add other types of build processes, such as remote - we probably need the release hook to trigger before and after *every* variant release, separately.

Rather than ditching pre/post_release though, we should add another - pre_variant_release  /post_variant_release - and leave *release as is. So rather than what we have now:

pre_release
  build variant 0
  release variant 0
  build variant 1
  release variant 1
post_release

Instead we have:

pre_release
  pre_variant_release
    build variant 0
    release variant 0
  post_variant_release
  pre_variant_release
    build variant 1
    release variant 1
  post_variant_release
post_release

Thx
A

nerdvegas:
@mstreatfield - although now I seem to recall you may have actually needed this overlap for reasons I can't quite remember...

mstreatfield:
In the hooks, I don't think so... I did ask a few weeks ago if I could add
some that were missing (think they aren't quite symmetrical at the
moment).  But I can double check on Monday.

nerdvegas:
Hey @mstreatfield,

The build process has been rewritten in resources2 so I don't know that it's worth trying to merge this. But I still think there's the need for this functionality.

I have an idea on how we should do this. I'm not sure about forcing all installs to now equate to make; make install - this might be unexpected behaviour (consider that we have plugin support for different build system types now, not just cmake).

What if instead this behaviour is configurable? We could add an option 'build_on_install'. If true, rez would do both a non-installed build, and an installed build, when using "rez-build -i". Since you can also set config options in the package.py/yaml file now, setting this option to True would be done right there in the package definition. So, you would just set this on your ExternalProject packages, and other packages would be unaffected.

Thoughts? This would be a fairly trivial update I think.

mstreatfield:
Hey @nerdvegas,

> I have an idea on how we should do this. I'm not sure about forcing all installs to now equate to make; make install - this might be unexpected behaviour (consider that we have plugin support for different build system types now, not just cmake).

Yes - this thread was primarily from the perspective of a CMake build system using a Make child build system.  I think the *policy* of the CMake build system (when used with Make) should be to call `make ; make install`.  If a different build system or child build system of CMake is used then obviously the semantics of `make ; make install` either change or don't apply at all - they have their own policy.

I think up until now `rez-release` has been calling both `build` and `release` methods on the build process.  With my change above this resulted in the following for a release:

    cmake ; make
    cmake ; make ; make install

So make was called three times which is obviously bad.  What I think should happen (and what appears to be happening now if I combine my branch with your changes) is that release isn't calling `build` and so you get:

    cmake ; make ; make install

Which I think is the correct behaviour.   Personally I think this should be the default, but I guess a configuration option to turn it on wouldn't be the end of the world - aside from my `ExternalProject_Add` argument above (and the fact it matches Rez 1) I don't have anything better to offer!

This means the general pattern is:

    pre_release
        pre_variant_release
            release variant 0  # cmake ; make ; make install
        post_variant_release
        pre_variant_release
            release variant 1  # cmake ; make ; make install
        post_variant_release
    post_release

With other build systems replacing the `cmake` and `make` calls appropriate for their requirements.  

Also, slightly related to this thread, can the 'local' build process be renamed back to 'localSequentialBuild' (or something similar).  I think the original name was better, and gives a nice symmetry when adding a 'localParrallelBuild' (something we are looking at to build multiple variants at the same time locally).

nerdvegas:
Ok, let's just go with what we have at the moment then. Are you able to send a new PR to merge with resources2?

Re default build process, I'll rename it to "local_sequential" ('build' is redundant).

Thx
A

nerdvegas:
Actually - before I do that rename I think we need to consider something.

At the moment there's no way to select the build process you want for a given package (that just hasn't been done yet). So first question - should this be controllable via new args to rez-build, or by config, or both? I am thinking both.

So then there's the question, should local parallel build just be implemented as part of the existing LocalBuildProcess class? For example, it would be sensible for the user to be able to do something like this:

]$ rez-build --nthreads 4

..where 'local' is the default build process type, which has added the 'nthreads' arg and will switch behaviour accordingly. This seems more intuitive than:

]$ rez-build --mode=local_parallel --nthreads 4

In the same vein I would expect to be able to do this kind of thing:

]$ rez-build --mode=remote  # sequential
]$ rez-build --mode=remote --concurrent  # just use as many hosts as are available

The point is there's a 1-1 relationship between the plugin class name and what 'mode' will be exposed on the command line. If it does make sense to split the implementation into separate classes though, we could still just have a 'local.py', which would contain three classes - a LocalBuildProcess that just forwards on to the appropriate subclass based on the user's args, and LocalSequentialBuildProcess/LocalParallelBuildProcess subclasses.

mstreatfield:
]$ rez-build --nthreads 4

Yes, I agree that is nicer.  

I assume the unit by which work is divided is a variant (so `--nvariants 4`?) with build system and child build systems determining the number of threads for each build (with `-j` argument for Make for example).

I think the code for this (especially remote builds) might be such that multiple classes would be nicer/easier to maintain.

nerdvegas:
No I think --nthreads is right - you're saying you want to assign N threads to build N variants - if the build system then goes ahead and uses further threads within each of these then so be it. --nvariants sounds like I'm capping the number of variants I want to build.

Anyway I think we're a bit off topic on this PR. Are you ok sending another PR for resources2 when you're ready? I'll just close this one when that happens, thx. A

mstreatfield:
I'm going to close this pull request.  With the work in resources2 I don't think it is relevant, and I don't think an alternative pull request against resources2 is required.

PR #150, opened 2015-01-06
Make package callbacks available

alefeve:
Set package.py commands functions as "callbacks" parameter of ResolvedContext __init__ function.
We needed to have custom functions in packages, for deployment, for example.

### DIFF
--- a/src/rez/resolved_context.py
+++ b/src/rez/resolved_context.py
@@ -121,7 +121,8 @@ class ResolvedContext(object):
                  building=False, caching=None, package_paths=None,
                  add_implicit_packages=True, max_fails=-1, time_limit=-1,
                  callback=None, package_load_callback=None, max_depth=None,
-                 start_depth=None, buf=None):
+                 start_depth=None, buf=None,
+                 callbacks=["pre_commands", "commands", "post_commands"]):
         """Perform a package resolve, and store the result.
 
         Args:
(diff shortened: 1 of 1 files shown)


nerdvegas:
Hi Alex,

I will accept this feature in theory but there are too many outstanding issues. Also it will need to be merged with the resources2 branch but I think given the points below it will probably be easier to reimplement from scratch. Issues are:

1) 'callbacks' isn't really the right terminology. I would call these "exec_points".
2) This functionality should be exposed in the rez-env tool.
3) This functionality should be exposed in the GUI.
4) Rez *always* stores enough info into env-vars to be able to reproduce the same env again. Thus this info should be written to a REZ_USED_EXEC_POINTS var.
5) This info is not being stored into context files - the serialisation part is missing. This should be done carefully, ie `serialize_version` should be upped (to a new minor version) and to_dict/from_dict updated accordingly.
6) There is no safeguard against an exec point being selected that is not a function.

Lastly, I'm aiming to disallow arbitrary keys in the root of packages. This is for forwards compatibility reasons - a future rez release might introduce a new attribute that you're already using as a custom attrib in your studio - this would cause unavoidable compatibility problems. To get around this a 'custom' namespace is provided. So this feature needs to support the user providing a function in the 'custom' namespace, and using this as an exec point (I think dot-notation would be acceptable for this).

Hth
A

nerdvegas:
(7) This should also be available as a rezconfig setting.

Issue #55, opened 2014-05-06
Windows support


mottosso:
Hi,

I'd be interested in providing support for Windows. 

Having briefly looked through your manual and source-dived part of your 2.0 release, I've gathered that much of Rez is built upon bash, which I can imagine being a challenge if ported to cmd.exe.

Is there anything else you think I should be aware of? Which part of your codebase would you recommend I get started with?

### Edit

In an attempt to make this issue referable to newcomers and have them get up to speed more quickly, I'll collect some relevant links here.

* [Windows Fork](https://github.com/instinct-vfx/rez)
* [Rez Usergroup](https://groups.google.com/forum/#!forum/rez-config)

Best,
Marcus

mstreatfield:
Hello @nerdvegas, @mottosso, @instinct-vfx and @chadrik,

I got Windows and OSX support up and running at work on Friday and am now able to `rez env` on both operating systems.

For OSX there really wasn't much to do as it pretty much 'just worked' but I'll make a pull request with the changes next week.  I can also `rez build` on OSX (but have not tested `rez release` or other features).  Unit tests pass.

WIndows was a little tricker.  
* There are changes to the `install.py` script which are fairly trivial and I'll make a pull request for these next week too - mainly some path handling changes for virtualenv.
* I created a `cmd` shell plugin to work with the Command Prompt and to replace the existing `Windows` plugin.  Currently there is no shell detection so this is the default.  I'll include this in the pull request.  It doesn't include some of the more advanced features of Bash and SH plugins but gets us going.
* The API for shell plugins is really unclear.  It wasn't obvious what methods the plugin should implement, and the inheritance between existing plugins is broken in some cases.  `UnixShell` ended up being a better base class than the intended `Shell`.  I think it would be good to make a pass at refactoring this before looking at other Windows shells such as PowerShell.
* Coloured output doesn't work - the control characters are printed.  According to the `colorise` documentation we are doing right so not sure what is wrong there.
* There is a bug in the `config` module and the handling of `~` in config values related to Python's `os.path.expanduser` method (read the documentation on this method to see how it is implemented different across operating systems to better understand this problem).  When getting the implicit packages from the config file (which are all weak), the `~` is being expanded, so the implicit package becomes `C:\Users\platform-osx`.  This isn't a problem on Linux/OSX due to the different implementation of `os.path.expanduser`.  My current thinking is to update the `config` module so only parts of the schema defined as a path (not just `Str`) are expanded.  Thoughts? 
[...]

PR #159, opened 2015-03-09
Windows and OSX Support

mstreatfield:
This pull request introduces the changes required for Rez to work on OSX and Windows.  The main changes are:

* fix the install process.
* complete the Windows specitific `system` and `platform_` implementations.
* replace the `Windows` stub shell implementation with `cmd` for the Windows Command Prompt.
* move the `prompt` setting from `rezconfig` to the individual shell implementations (as it is shell specific).
* create a `rezconfig.py` for easier cross-platform configuration within the same Rez install.
* update the resource location code.
* update the unit tests.

A bit more detail...

* Colour in stdout/stderr messages is explicitly disabled.  Although it should work, it doesn't.  I have been unable to determine why (according to the Colorama documentation we are doing everything right).
* In the context of `config.py` the `os.path.expanduser` implementation is a bit dangerous.  This method is implemented differently on Windows which caused problems with weak references in the implicit packages list.  I have created a custom (although slightly limited) version of `expanduser` for this purpose which is sufficient for our needs.  This is documented.
* Creating `rezconfig.py` was necessary to be able to support cross-platform configuration in the same file, as suggested by Chad in issue #55 and inline with other open source projects.  Currently both the yaml and .py file are support (for backwards compatibility) but dropping the yaml longer term would reduce code complexity.
* Where Windows doesn't explicitly support a feature (such as `os.chmod` for forwarding scripts etc.) I've blcoked this out with a comment and `if/else`.  Perhaps there are other cross-platform solutions we can use here, but I don't know what they are.
* The changes to the resource system where subtle - essentially escaping a regex and removing a `to_posixpath` translation.  Otherwise this was the easiest thing I had to change (where I initially thought it the most complex).
* There is a method called `_keytoken` in Rex which I had to extend (with `get_key_topken`) and make shell specific as it was assuming `${THIS}` for variable reference.  I don't fully understand why it's necessary - especially in the python context - but it is.
* As the command prompt is shell specific I've moved this configuration from `rezconfig` into the plugins themselves.
* The inheritence between `Shell`, `UnixShell` and the other implementations is a bit broken - or at least the `Shell` API is not clearly defined.  I've tidied this up a little bit (but not very much).  More work could be done here to improve readability.  This area feels a bit brittle.
* No automatic shell detection is performed.  If you are on Windows you use the Windows Command Prompt and that's it.  This could be extended in the future, but as this is sufficient for my needs I've not investigated further/other shells.
* Generally the changes to the tests where to support `\` instead of '/' and are pretty straight forward.  The thing to remember here is the context - in some cases even if the test is running on Windows it's comparing against a hardcoded value in the fixtures and so `/` is required.  Therefore the changes might look inconsistent but they are correct.
* I'm a little disappointed with my changes to `test_shells.py` - I've had to skip a lot of the tests.  This is in part due to some tests exercising features of the Windows Command Prompt that aren't support.  More could be done here.
* Binding and forwarding scripts don't work.  This is due to the way Windows recognises executables (with the `PATHEXT` environment variable) and the lack of support for shebang lines.  It should be possible to support them, but I need to better understand how Windows handles this area.  Suggestions?
* The string escaping in Rex has caused some problems and is incompatible with the Windows Command Prompt.  In general I haven't handled escaping in this shell due to issues that Federico has raised.  But the current implementation seems a bit too much and you'll see I'm explicitly replacing the `"` in the `set` command.  Longer term this is probably incorrect - but I think changes to Rex are required first.

This branch was made from 2.0 and not resources2 - I realise this is probably an issue - however I had already started work before resources2 was created/finished.  The sheer number of changes in the resources2 branch (which go beyond directly improving the resource system) is going to make this hard to integrate.  I don't have time to tackle this immediately - but I do need Windows support now - so took the path of least resistance.


nerdvegas:
Closing - we have now merged this PR with resources2.

PR #161, opened 2015-03-17
Fixed IndentetionError caused by non-indented comments in def commands()...

holofermes:
...:This pull should resolve the indentation issue with comments.

### DIFF
--- a/src/rez/resolved_context.py
+++ b/src/rez/resolved_context.py
@@ -1409,8 +1409,13 @@ class ResolvedContext(object):
                 try:
                     executor.execute_code(commands.source)
                 except error_class as e:
-                    msg = "Error in commands in package %r:\n%s" % (pkg.uri, str(e))
-                    raise PackageCommandError(msg)
+                    try:
+                        if (type(e) == IndentationError):
+                            commands.maybe_indent()
+                            executor.execute_code(commands.source)
+                    except error_class as e:
+                        msg = "Error in commands in package %r:\n%s" % (pkg.uri, str(e))
+                        raise PackageCommandError(msg)
 
         _heading("post system setup")
         # append suite path if there is an active parent suite
--- a/src/rez/utils/data_utils.py
+++ b/src/rez/utils/data_utils.py
@@ -6,7 +6,6 @@ from collections import MutableMapping
 from inspect import getsourcelines
 from threading import Lock
 from textwrap import dedent
-import re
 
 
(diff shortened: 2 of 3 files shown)


src/rez/utils/data_utils.py
```diff
@@ -14,13 +13,29 @@ class SourceCode(object):
...
+        for i, line in enumerate(codelines):
+            if line.startswith('#'):
+                nextindex = i+1 if i < linescount else i-1
+                nextline = codelines[nextindex]
+                while nextline.startswith('#'):
+                    nextline = codelines[nextindex]
+                    nextindex = nextindex+1 if nextindex < linescount else nextindex-1
+                firstchar = len(nextline)-len(nextline.lstrip())
+                codelines[i] = '%s%s' % (nextline[:firstchar], line)
+        code = '\n'.join(codelines)
```
nerdvegas: may as well dedent after doing this - good chance this will give valid code, and will avoid an unnecessary IndentationError during command execution.

src/rez/resolved_context.py
```diff
@@ -1409,8 +1409,13 @@ def _minor_heading(txt):
                 try:
                     executor.execute_code(commands.source)
                 except error_class as e:
-                    msg = "Error in commands in package %r:\n%s" % (pkg.uri, str(e))
-                    raise PackageCommandError(msg)
+                    try:
+                        if (type(e) == IndentationError):
```
nerdvegas: ouch - this will cause non-indentation error to be silently skipped

src/rez/utils/data_utils.py
```diff
@@ -14,13 +13,29 @@ class SourceCode(object):
     def __init__(self, source):
         self.source = source.rstrip()
 
+    def maybe_indent(self):
```
nerdvegas: this doesn't describe what the function does

src/rez/utils/data_utils.py
```diff
@@ -14,13 +13,29 @@ class SourceCode(object):
...
+        if self.source and self.source[0] in (' ', '\t'):
+            # outer indent(s) follow, perhaps comments
+            self.source = "if True:\n" + self.source
```
nerdvegas: Probably not a good idea to change in place - SourceCode instances are stored in Packages/Variants, which are stored in a cache. Better to return a copy. This could cause bugs in threaded code.

PR #163, opened 2015-03-23
Install variants into packages additively

holofermes:
This changes should address the issue of locally installing all variants modifications to a package.


nerdvegas:
Closing, implemented same fix separately.

PR #172, opened 2015-04-13
make is_local return False by default, with real implementation on filesystem resources

elrond79:
since is_local is defined by whether or not it is found in the "local packages path", it seems a concept that really only makes sense for filesystem resources.  As source, made the default implementation of is_local always return False, and override this on the filesystem resources.

This is a potential fix to this issue: https://github.com/nerdvegas/rez/issues/162

...and it touches on some of the issues in this thread: https://groups.google.com/forum/#!topic/rez-config/5q1k0iJFGEs

However, it's still not an ideal solution, as it still makes the assumption that pkg.root is a directory (and runs os.path.exists on it).

Perhaps a better / more generic approach would be allow resources to provide aritrary "tags" with additional information about that particular resource, and then rez-context could display these tags.  In this particular case, filesystem resources could attach a "local" tag to packages it finds within the local packages path.

The "NOT FOUND" information could also be implemented as a tag... or, alternatively, we could provide resources with the ability to perform validation, in some way appropriate for that resource type, and return brief descriptions if problems are found.

### DIFF
--- a/src/rez/package_resources_.py
+++ b/src/rez/package_resources_.py
@@ -207,6 +207,14 @@ class PackageRepositoryResource(Resource):
         """
         raise NotImplementedError
 
+    @property
+    def is_local(self):
+        """Returns True if the variant is from a local package"""
+        # Base implementation returns False, because currently, a file is local
+        # only if it was found inside the directory defined by the
+        # local_packages_path - which means only FileSystem resources can be
+        # local... so override a "real" check for is_local in those subclasses
+        return False
 
 class PackageFamilyResource(PackageRepositoryResource):
     """A package family.
@@ -382,3 +390,8 @@ class VariantResourceHelper(VariantResource):
     def _load(self):
         # doesn't have its own data, forwards on from parent instead
         return None
+
+    @property
+    def is_local(self):
+        """Returns True if the family is from a local package"""
+        return self.parent.is_local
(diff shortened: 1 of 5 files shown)


nerdvegas:
Hey elrond79,

Just to clear things up - local packages aren't restricted to filesystem package repositories - although I don't see this being common (but who knows) theoretically the local package repository could be any repository type.

The fact that env-vars such as REZ_PACKAGES_PATH and LOCAL_PACKAGES_PATH refer to 'paths' at all is historical. REZ_PACKAGES_PATH is now able to contain package repository uri-type strings as well as file paths - for example, a theoretical database backend might have a location string "mongodb:host=svr,namespace=central", and this value could be listed in REZ_PACKAGES_PATH. I think LOCAL_PACKAGES_PATH hasn't been updated to behave the same way yet. Off the top of my head, is_local() implementation should be changed so that it simply checks that the location identifier string of its associated package repository matches the value of LOCAL_PACKAGES_PATH.

I'll take your PR though, and make the associated fixes at the same time.
Thx
A

elrond79:
Hmm... yeah, makes sense. I like the idea of using the repository to check whether the package is local.

You can go ahead and just reject this PR if you want - I actually just noticed a mistake anyway (packages_.py, PackageBaseResourceWrapper now has two definitions for "is_local").  Or, if you want, I can break out the changes to resolved_context.py into a separate pull request...

nerdvegas:
Hey elrond79,

I'm going to close this PR, but I'll incorporate your ResolvedContext changes separately, thx.

elrond79:
Cool, the changes made in your local_repo_fix branch do the trick! Deleted the branch from our repo in favor of your changes...

Was thinking about making a test for this - but not sure where that would go in the existing test structure.  What do you think of a new test_cli.py module?

PR #176, opened 2015-04-17
Windows and OSX Support For resources2

mstreatfield:
This pull request introduces the changes required for Rez to work on OSX and Windows for the resources2 branch. The main changes are:

* fix the install process.
* complete the Windows specitific system and platform_ implementations.
* replace the Windows stub shell implementation with cmd for the Windows Command Prompt.
* move the prompt setting from rezconfig to the individual shell implementations (as it is shell specific).
* create a rezconfig.py for easier cross-platform configuration within the same Rez install.
* update the unit tests.
* fix a few broken imports and misc. problems I encountered (which aren't directly Windows related).

See #159 for a more detailed discussion of some of these points as they are generally still valid in this new pull request (and I don't want to repeat them unnecessarily).


nerdvegas:
@ expanduser - I actually didn't know that ~foo will expand to HOME if 'foo' is a valid user! That's actually a bit flakey then - after all by coincidence you could have a package name that matches a user name, thus the weak reference '~jbloggs' will incorrectly expand to HOME. Perhaps we should change expanduser for all cases, and expand only if ~ is followed by os.sep - not perfect but probably fine for 99% of cases.

mstreatfield:
Yes I think that's probably ok. I'll do and update the pull request on
Monday.

It's hard to think of a case where you'd want a reference to a home
directory other than your own.

mstreatfield:
I have pushed a new commit which includes `rez.vendor.six` - Paul pointed out that this was missing in the mailing list.  I am only using one function from it (`exec`) so it might be cheaper to (if the license allows) just inline that function as opposed to the whole module.  Originally the memcache implementation we were using (for some internal work before resources2) had `six` as a dependency so we already had it around and using it in one more place wasn't such a big deal.

I checked `expanduser`.  I think in most other places I could see we are actually expanding a path so the behaviour of `os.path.expanduser` is correct - it's only in the context of `config.py` where we might be expanding a weak implicit package that it's an obvious issue (and this is where I fixed it originally).  So I haven't made any more changes in that respect.

mstreatfield:
I have discovered an interesting problem where the unit tests fail on OSX - they are running too quickly.  For example, on osx `test_7` in `test_packages` fails with the exception `Internal failure - expected installed variant`.  This test passes ok on Windows and Linux.

If it put a `time.sleep` call after each `variant.install` call in that test it starts passing again.  I believe there is some latency in the internal memcache key generation using `os.stat` where the `mtime` property is a little bit behind... rapid calls to `variant.install` are not seeing the updated `mtime` and so using (what is now) stale cache results.  This causes the test to fail.

I can put a sleep in the tests, but it feels a bit flaky.  5 tests are affected in this way.

nerdvegas:
Hey Mark,

I know about this already, the issue is not specific to osx. I've already started a memcached_fix branch a week or so and am in the process of fixing it.

Cheers
A

mstreatfield:
Ah ok. I'd not seen it at all before and yet today it has plagued me.

Is memcache used for all resources - even those in the local packages path?

nerdvegas:
Yeah it's used for all packages, including local. But it also checks for
any file changes as well, so even manually changing a local package won't
break it. I haven't had any issues until I hit the same problem as you
recently. It was unexpected because I am using the whole timestamp in the
memcached key - and that timestamp goes down to the millisecond so I didn't
expect a clash, but that does seem to be what's happening on occasion.

I think the fix is to disable the memcache during variant installation and
to clear any associated memcached entries at the same time. This is only a
problem when installing then retrieving the variant within a single process.

Just a note - the memcached_fix branch also changes the behaviour a little
so that timestamped resolves are also cached. These were being discarded
before - but with some changes coming up in our pipeline I have seen the
need for this.

Thx
A

nerdvegas:
Hey @mstreatfield,

I think your PR is missing the 'anti' package under test/data/builds/packages - it isn't there in your branch, and the builds UT fails.

Thx
A

mstreatfield:
Sorry, try now.

nerdvegas:
Merged, all UTs passing.
Thx
A

nerdvegas:
Hey @mstreatfield,

I haven't been able to look into this yet (and don't have a Windows box so probably can't test) but since the merge I've noticed this exception on cmd shell plugin load:

```
  File "/home/ajohns/tmp/rez_venv/local/lib/python2.7/site-packages/rez-2.0.BETA.30-py2.7.egg/rez/plugin_managers.py", line 120, in load_plugins
    module = loader.find_module(modname).load_module(modname)
  File "/usr/lib/python2.7/pkgutil.py", line 246, in load_module
    mod = imp.load_module(fullname, self.file, self.filename, self.etc)
  File "/home/ajohns/tmp/rez_venv/lib/python2.7/site-packages/rez-2.0.BETA.30-py2.7.egg/rezplugins/shell/cmd.py", line 15, in <module>
    class CMD(Shell):
  File "/home/ajohns/tmp/rez_venv/lib/python2.7/site-packages/rez-2.0.BETA.30-py2.7.egg/rezplugins/shell/cmd.py", line 19, in CMD
    executable = Shell.find_executable('cmd')
AttributeError: type object 'Shell' has no attribute 'find_executable'
```

Thx,
A

mstreatfield:
I moved the `find_executable` class method from `LinuxShell` to the `Shell` base class as it was required by the `cmd` implementation.  Other shells use it in a similar way as `cmd`.

As this is a class method being used by a class attribute it's possible Python is evaluating the attribute before the method has been bound, hence it can't be found.  I was getting a similar error message before I started the Windows integration but it seemed somewhat non-deterministic so I wasn't able to fully debug it.

Looking at the other shells, I updated those to use `executable = UnixShell.find_executable('bash')`, explicitly calling the method on the base class (which presumably has had the method bound).  It's a bit crusty either way.

Does `executable` have to be a class attribute - could it be a cached property?  I can't obviously see it being used as a class attribute anywhere.  This would also make it similar to the vcs plugins where they determine the executable at `__init__` time.

mstreatfield:
How would you like to proceed - make `executable` a cached property on the instance instead?

nerdvegas:
Hey Mark,

Sorry I'd meant to get back to this and it fell off my radar. I've fixed this in a recent push to resources2. As a bonus, the 'cmd' plugin doesn't appear with 'rez -i' on non-windows platforms now, rather than reporting failure due to not finding the 'cmd' executable.

mstreatfield:
@nerdvegas thanks.  What is the status of the `memcached_fix` for the `Internal failure - expected installed variant` test failure mentioned a few comments up the chain?  I've lost track of the various branches a little bit and I am still seeing that error when I run the tests.

nerdvegas:
That should be fixed, iirc I merged that into resources2 already.

Related to memcached - I'm fixing a separate issue now where too many
connections can be made to the cache server, but this isn't a big issue and
in the meantime you can use memcache's -c option to bump up the max
connection count. We only have so many because all our farm jobs perform
resolves. That should be pushed tomorrow sometime.

Thx
A

mstreatfield:
Hmmm ok I'll go back and take a look at that.  

r.e. memcache, yes we'd implemented something to force a disconnect after a solve already.  But your changes look like a cleaner solution so far.  However with the current head of resources2 the unit tests fail - the `timeout_cache` is trying to cache a list (which is unhashable) of the memcache server uris.  Have I just tried to pick it up too early?  For example:

```
ERROR: test_1 (rez.tests.test_solver.TestSolver)
Extremely basic solves involving a single package.
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/tests/test_solver.py", line 103, in test_1
    ["nada[]"])
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/tests/test_solver.py", line 49, in _solve
    s1, s2, s_perms = self._create_solvers(reqs)
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/tests/test_solver.py", line 29, in _create_solvers
    verbosity=Solver.max_verbosity)
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/solver.py", line 1659, in __init__
    phase = _ResolvePhase(self.request_list.requirements, solver=self)
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/solver.py", line 1036, in __init__
    scope = _PackageScope(package_request, solver=solver)
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/solver.py", line 842, in __init__
    package_request.name, package_request.range)
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/solver.py", line 1934, in _get_variant_slice
    range=range)
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/solver.py", line 819, in get_variant_slice
    variants = variant_list.get_intersection(range)
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/solver.py", line 479, in get_intersection
    for var in package.iter_variants():
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/packages_.py", line 159, in iter_variants
    for variant in repo.iter_variants(self.resource):
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rezplugins/package_repository/filesystem.py", line 438, in iter_variants
    for variant in self._get_variants(package_resource):
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/backport/lru_cache.py", line 74, in wrapper
    result = user_function(*args, **kwds)
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rezplugins/package_repository/filesystem.py", line 573, in _get_variants
    return [x for x in package_resource.iter_variants()]
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/package_resources_.py", line 296, in iter_variants
    num_variants = len(self._data.get("variants", []))
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/utils/data_utils.py", line 100, in __get__
    result = self.func(instance)
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/utils/resources.py", line 86, in _data
    return self._load() if self.schema else None
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rezplugins/package_repository/filesystem.py", line 139, in _load
    data = load_from_file(self.filepath, self.file_format)
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/serialise.py", line 52, in load_from_file
    return _load_from_file(**kwargs)
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/utils/memcached.py", line 269, in wrapper
    client = get_client()
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/utils/memcached.py", line 261, in get_client
    return get_memcached_client(servers, debug=debug)
  File "/scratch/markst/workspaces/rez/packages/rez/2.0.rc1.6.0/CentOS-6.2_thru_7/python-2.7/lib/python2.7/site-packages/rez-2.0.rc1.6.0-py2.7.egg/rez/utils/timeout_cache.py", line 59, in wrapper
    entry = entries.get(key)
TypeError: unhashable type: 'list'
```

nerdvegas:
Too early, sorry - expect a fix to be pushed today.

A

nerdvegas:
I've just pushed a fix, but it's worth commenting on the google group so please refer there for more details...

PR #178, opened 2015-04-21
Initial commit for exclude/include feature

holofermes:


chadrik:
@holofermes, it would be great to get a description of what this feature is all about.  It sounds useful.

nerdvegas:
Hey Chad,

holofermes aka Fabio is working at Method with us, so I can explain the idea here.

A big problem we have at the moment is that there's not an easy way to release a package for testing purposes, while at the same time making sure that production don't pick it up accidentally. Ie, we need to be able to release alpha/beta/etc packages. You can currently do that, kind of - you can rez-build -i to a custom path, rather than rez-release, and have your testers add this path to $REZ_PACKAGES_PATH - but in practice that's finnicky enough that people just don't do it.

The approach here is pretty simple. We provide a new 'exclude_packages' configurable setting, that's just a list of globbed strings. Packages that match this string are not visible. So by default, you might have this set to, let's say, ["*.dev"]. Thus, releasing the package foo-1.0.0.dev will not be seen by users. However, you can change this setting (via rezconfig file or new rez-env flag) to make those packages become visible. So, a tester might go "rez-env --exclude=  foo" in order to test said dev package, for example (note exclude is set to none here).

An include_packages setting is included also. If a package matches an exclusion pattern but also an inclusion pattern, it is included. This gives us a way to more finely control which dev packages are visible, rather than just globally turning them all on/off. So in the above example, the user testing foo would more appropriately run "rez-env --include="foo-*" foo.

As a bonus, this feature can also be used as a general blacklisting tool - if package bah-1.2.3 is known to be broken, just add it as a non-globbed path to 'exclude_packages'.

One question left to answer is, once you've released foo-1.0.0.dev and it's been tested and known to work, what happens? Do we re-release as foo-1.0.0 and delete foo-1.0.0.dev? Or do we "move" foo-1.0.0.dev to foo-1.0.0? I'm not sure about that yet - initially we can just re-release and delete the dev package (no extra work is needed to do this) but whether that will be satisfactory or not I'm not yet sure.

A

elrond79:
Awesome... Chad and I were just talking about the need for such a feature earlier this afternoon! It gives us a lot of what we want, but we have a few ideas to make it even more useful / flexible:

- using string matching is powerful, but it would also be nice to have a way to specify by version expressions: ie, --exclude='foo>3'.  The two could co-exist, as long as you have a way to specify whether you're looking for glob matching or version expressions - ie, we could do:
    ```--exclude='*.dev' --exclude_ver='foo>3'```
or
    ```--exclude='glob:*.dev' --exclude='ver:foo>3'```
One immediate benefit of this is it allows a way to mark / unmark something as "production-ready" WITHOUT having to re-release. Ie, if we're testing foo v4, we can simply release it, since we've excluded anything > 3; then, once we've deemed it ready, we update the global exclude setting to be 'foo>4'.

- Another benefit of allowing version expressions is you could potentially then allow packages themselves to set excludes.  ie, you might make it so that foo excludes bar>3.  This would work basically like a requirement (or, really, an anti-requirement), except that it's overrideable.  It becomes most powerful if we can set up excludes based on version - ie, the foo.py package might have:
    ```python
    excludes = {
        '3+<4': 'bar>5',
        '4+<5': 'bar>9',
    }
    ````
   For a more real life example, we could effectively make it so that a certain version of maya wouldn't use a version of mental ray above X... but the user could still override that, if necessary.

- Lastly, allowing version-excludes would provide an easy way for users to easily request excluded packages, by having the various command-line tools (or perhaps just rez-env?) set up "automatic includes" if a user requests a specific version.  That is, if a user wants to test foo-4, then instead of having to do
    ```rez-env --include=foo-4 foo-4```
they could just do 
    ```rez-env foo-4```
and rez-env would AUTOMATICALLY add the "--include=foo-4", since a specific version was requested.  This sort of automation would be handled within / reserved for the command-line tools, to avoid potential headaches, while still providing a friendly interface for us humans.

Thoughts?

nerdvegas:
Below.


On Tuesday, April 21, 2015, Paul Molodowitch <notifications@github.com>
wrote:

> Awesome... Chad and I were just talking about the need for such a feature
> earlier this afternoon! It gives us a lot of what we want, but we have a
> few ideas to make it even more useful / flexible:
>
>    -
>
>    using string matching is powerful, but it would also be nice to have a
>    way to specify by version expressions: ie, --exclude='foo>3'. The two could
>    co-exist, as long as you have a way to specify whether you're looking for
>    glob matching or version expressions - ie, we could do:
>    --exclude='*.dev' --exclude_ver='foo>3'
>    or
>    --exclude='glob:*.dev' --exclude='ver:foo>3'
>    One immediate benefit of this is it allows a way to mark / unmark
>    something as "production-ready" WITHOUT having to re-release. Ie, if we're
>    testing foo v4, we can simply release it, since we've excluded anything >
>    3; then, once we've deemed it ready, we update the global exclude setting
>    to be 'foo>4'.
>
> I like it. I don't think there will be any need to specify whether you're
globbing or using version ranges...  It'll probably be good enough to treat
as a version range if it can be parsed as such (chars such as * will not
parse using a Requirement class).

>
>    -
>    -
>
>    Another benefit of allowing version expressions is you could
>    potentially then allow packages themselves to set excludes. ie, you might
>    make it so that foo excludes bar>3. This would work basically like a
>    requirement (or, really, an anti-requirement), except that it's
>    overrideable. It becomes most powerful if we can set up excludes based on
>    version - ie, the foo.py package might have:
>
>    excludes = {
>        '3+<4': 'bar>5',
>        '4+<5': 'bar>9',
>    }
>
>    For a more real life example, we could effectively make it so that a
>    certain version of maya wouldn't use a version of mental ray above X... but
>    the user could still override that, if necessary.
>
>
A problem I see there is that you'd probably want the package excludes to
merge with the common excludes, but no other configure settings do that
(they just overwrite) and this might be difficult to do.

>
>    -
>    -
>
>    Lastly, allowing version-excludes would provide an easy way for users
>    to easily request excluded packages, by having the various command-line
>    tools (or perhaps just rez-env?) set up "automatic includes" if a user
>    requests a specific version. That is, if a user wants to test foo-4, then
>    instead of having to do
>    rez-env --include=foo-4 foo-4
>    they could just do
>    rez-env foo-4
>    and rez-env would AUTOMATICALLY add the "--include=foo-4", since a
>    specific version was requested. This sort of automation would be handled
>    within / reserved for the command-line tools, to avoid potential headaches,
>    while still providing a friendly interface for us humans.
>
>
This might be a step too far - after all, if I request foo-4, that doesn't
mean I want foo-4.0.1.dev. We could make the exception for a specific
version, ie foo==4 though, but that would be the only case I think.

nerdvegas:
Further to this - another useful type of exclusion I can see is timestamp-based. I can see this being quite useful for locking off shows and so on, but only for specific packages.
For example, imagine we could do this:

```
exclude_packages:
- *>1429728021
```
In other words, exclude all packages released after the given timestamp.

elrond79:
>     using string matching is powerful, but it would also be nice to have a
>     way to specify by version expressions: ie, --exclude='foo>3'. The two could
>     ...
>  I like it. I don't think there will be any need to specify whether you're
> globbing or using version ranges...  It'll probably be good enough to treat
> as a version range if it can be parsed as such (chars such as * will not
> parse using a Requirement class).

Awesome, even better...

>     Another benefit of allowing version expressions is you could
>     potentially then allow packages themselves to set excludes. ie, you might
>     make it so that foo excludes bar>3. This would work basically like a
>     requirement (or, really, an anti-requirement), except that it's
>     overrideable. It becomes most powerful if we can set up excludes based on
>     ...
> A problem I see there is that you'd probably want the package excludes to
> merge with the common excludes, but no other configure settings do that
> (they just overwrite) and this might be difficult to do.

Well, it's definitely a burn to have it work differently than other settings... but I still think it would be useful enough to make it worth doing anyway.  Since I'm imaging there just being one global (merged) set of includes / excludes, I'm hoping implementation shouldn't be too hard.  We could take a crack at it here, if you want.

>     Lastly, allowing version-excludes would provide an easy way for users
>     to easily request excluded packages, by having the various command-line
>     tools (or perhaps just rez-env?) set up "automatic includes" if a user
>     requests a specific version. That is, if a user wants to test foo-4, then
>     ...
> This might be a step too far - after all, if I request foo-4, that doesn't
> mean I want foo-4.0.1.dev. We could make the exception for a specific
> version, ie foo==4 though, but that would be the only case I think.

Hmm...

nerdvegas:
"""
    Another benefit of allowing version expressions is you could
    potentially then allow packages themselves to set excludes. ie, you might
    make it so that foo excludes bar>3. This would work basically like a
    requirement (or, really, an anti-requirement), except that it's
    overrideable. It becomes most powerful if we can set up excludes based on
    version - ie, the foo.py package might have:

    excludes = {
        '3+<4': 'bar>5',
        '4+<5': 'bar>9',
    }

    For a more real life example, we could effectively make it so that a
    certain version of maya wouldn't use a version of mental ray above X... but
    the user could still override that, if necessary.
"""

Actually I want to revisit this. I think there are two issues here:

1) This is not really related to the include/exclude feature. It's something different - it's overridable package requirements.

2) I think this is dangerous territory, and it muddies the waters between configuring packages for production use (ie, managing which versions of which packages are going to be used in show X) and package dependencies. Rez has always striven to keep these two concept separated - most environment management systems (that I've seen in vfx anyway) do not, and to their detriment IMO. Take your example - why should the maya package be specifying the mental ray version in the first place? You're trying to control package versions in the wrong place - this is a show configuration issue. A package's requirements are supposed to define which other packages they require and are compatible with - being able to override this doesn't make much sense.

You could argue that you might want to create a 'bundle' package - a package that just acts as a container for other packages, so you would want to override requirements for this package. I would say that by doing this you are trying to shoehorn a package into being something that it is not (and should not be).

Instead, what you might want (and what I've done in studios and am in the process of doing now) is your own configuration system, that supports hierarchical overrides and that is integrated with Rez. So, people use this configuration system to define what packages to use in what areas of production, and this data is then given to Rez to construct the appropriate environment. You might set a version of mental ray in some 'maya' context at the studio-wide level, and then override that mental ray version for a specific show.

I have often thought about writing an open source project just like this, that marries together overridable configuration management and rez-based package management. The problem I always run into is that this in itself is a big project, whos scope is package management and much more - it includes configuration management in general (we like to configure lots of things in vfx that are not software - environment variables, server locations, the list is endless), as well as configuring package versions. It'd be a great project, a very useful project, but a lot of work too.

elrond79:
I noticed I accidentally sent my earlier comment unfinished.  I was going to respond to this bit:

>     Lastly, allowing version-excludes would provide an easy way for users
>     to easily request excluded packages, by having the various command-line
>     tools (or perhaps just rez-env?) set up "automatic includes" if a user
>     requests a specific version. That is, if a user wants to test foo-4, then
>     instead of having to do
>     rez-env --include=foo-4 foo-4
>     they could just do
>     rez-env foo-4
>     and rez-env would AUTOMATICALLY add the "--include=foo-4", since a
>     specific version was requested. 
>     ...
> This might be a step too far - after all, if I request foo-4, that doesn't
> mean I want foo-4.0.1.dev. We could make the exception for a specific
> version, ie foo==4 though, but that would be the only case I think.

...but apparently hit send before I was finished.

In any case, was going to say that exact version would be better than nothing... but would still be inconvenient for artists, who probably won't remember that they need to set maya-2016.01, or foo-4.0.3.dev.  Possibly we could first try to resolve with full includes, and if that fails, try adding our auto includes.  That way, if these versions are available:

    foo-3.1.5
    foo-4.0.1
    foo-4.0.5.dev

...and they request "foo-4", they'll get foo-4.0.1; but if the ONLY foo-4 version is dev:

    foo-3.1.5
    foo-4.0.5.dev

...then they get foo-4.0.5.dev, which seems reasonable.

elrond79:
> Take your example - why should the maya package be specifying the mental ray version in the first place? You're trying to control package versions in the wrong place - this is a show configuration issue. A package's requirements are supposed to define which other packages they require and are compatible with - being able to override this doesn't make much sense.

Hmm... I see your point - it makes sense to keep package requirements and "preferences" separate.  So I'm ok with not getting per-package excludes.  I essentially just thought it was something that would be cool / could be handy.

> You could argue that you might want to create a 'bundle' package - a package that just acts as a container for other packages, so you would want to override requirements for this package. I would say that by doing this you are trying to shoehorn a package into being something that it is not (and should not be).

I agree that effort needs to be made to keep strict software requirements separate... but, as long as this tennant is not violated, why are packages that act as containers for other packages such a terrible idea? From the standpoint of looking at what the packages are actually doing, a "container" package behaves exactly the same way as other packages.  You could say that any package that has requirements is essentially a container package.

You even explicitly recommend doing such things in the original rez manual - you call them "bundles" there. I'm curious what made you change your mind on this point?
 
> Instead, what you might want (and what I've done in studios and am in the process of doing now) is your own configuration system, that supports hierarchical overrides and that is integrated with Rez. So, people use this configuration system to define what packages to use in what areas of production, and this data is then given to Rez to construct the appropriate environment. You might set a version of mental ray in some 'maya' context at the studio-wide level, and then override that mental ray version for a specific show.
> 
> I have often thought about writing an open source project just like this, that marries together overridable configuration management and rez-based package management. The problem I always run into is that this in itself is a big project, whos scope is package management and much more - it includes configuration management in general (we like to configure lots of things in vfx that are not software - environment variables, server locations, the list is endless), as well as configuring package versions. It'd be a great project, a very useful project, but a lot of work too.

I agree, that would be handy.  But until that sort of work is done, we're reliant on our current configuration system, which is heavily dependant on what you call "container" packages.

To elaborate - we do have a tool which sits on top of rez, but it's a pretty thin wrapper, and relies on rez to do most of the heavy lifting.  We define our "configuration" inside of special "luma" packages which accompany the "clean" software packages.  ie, we have both a "maya" package, which contains only the information strictly necessary to run a clean, non-luma version of maya, and a "luma_maya" package, which includes all the other stuff we need to run maya inside our pipe.  Our wrapper tool essentially just ensures that when "maya" is requested, we instead load "luma_maya" (which in turn requires normal "maya"), if we're in a "luma" context.

So, anyway - your comment that container packages are something that we "shouldn't" do has me slightly worried. Are there particular issues you see arising from such a system, or is it more of a conceptual purity issue?

nerdvegas:
I didn't mean to imply that container packages are bad, that isn't what I
meant. I only meant that using container packages with the idea that you
can override the requirements in said packages, is bad. Packages that just
contain other packages are fine. But as soon as you start trying to
override their requirements, you're gradually shifting towards trying to
use packages as some sort of configuration tool. I think therein lies the
danger - that is not what packages are. One of the central tenets of Rez
has been to completely separate package configuration from show
configuration, and I think it's very important to keep it this way.

We take a different approach to managing our environments - a separate
configuration system determines what contexts exist and what package
versions are within each. We can override this configuration data at
various levels (show, shot etc), apply timestamps (to lock down shows) and
so on. When we setshot, this config data is gathered up, and tools are
generated that boot us into the correct environment. So running 'maya'
would create the relevant context (which is hierarchically configurable)
and start maya within that environment.

Well actually, we don't do all that just yet, but we're moving towards
that. We're using suites just now, but they're a pain to maintain. We
haven't done the above approach before because large resolves can be
expensive and take some time, which would make tools slow to load. But the
resources2 branch includes various optiomisations, and memcached support,
which has a huge effect on resolve times and makes runtime resolves viable.

Hope that clarifies some things. Package/configuration in vfx is actually a
pretty interesting problem.
A

elrond79:
Gotcha, makes sense. Thanks for the clarification!

mstreatfield:
> As a bonus, this feature can also be used as a general blacklisting tool - if package bah-1.2.3 is known to be broken, just add it as a non-globbed path to 'exclude_packages'.

Although as this is in the main `rezconfig` it's not timestamped right?  So if "rolling back" to an earlier environment the `rezconfig` will remain unchanged blacklisting a package that wasn't blacklisted at the rolled back time.

We've had a bit of a pickle with blacklisting using a meta-package.  Would be good to formalise this as an actual concept a bit more concretely.  We are considering another workaround but a solution would be better... but that's another thread.

> One question left to answer is, once you've released foo-1.0.0.dev and it's been tested and known to work, what happens? Do we re-release as foo-1.0.0 and delete foo-1.0.0.dev? Or do we "move" foo-1.0.0.dev to foo-1.0.0? I'm not sure about that yet - initially we can just re-release and delete the dev package (no extra work is needed to do this) but whether that will be satisfactory or not I'm not yet sure.

Not sure how the version sorting handles this, but `foo-1.0.0.dev` really feels like `foo-1.0.0.rc.1` to me.  This a) means you can have multiple dev versions, b) `rc` is a bit more of a well known/understood pattern in software and c) logically follows to a `foo-1.0.0` that is a proper releasing, not requiring the `rc` packages to be deleted.

> Further to this - another useful type of exclusion I can see is timestamp-based. I can see this being quite useful for locking off shows and so on, but only for specific packages.

Where do you see the `exclude_packages` configuration living - in `rezconfig` still?  Does each project at Method have it's own `rezconfig` using the `REZ_CONFIG_FILE` variable?

nerdvegas:
Hey @mstreatfield,

1) No filters are not timestamped. Package blacklisting might be better served having this support, but that doesn't belong here - the fact you can use this filtering feature to blacklist packages is incidental, the main point is to provide a way to stage package releases. As you say this is a separate thread.

2) 'dev' was just an example, 'rc' is fine also. Rez will sort '1.0.0' < '1.0.0.rc.1' though. There's no way around this without really twisting up how versions work and making them too complicated (imo). You can think of versions as a kind of hierarchical numerical system, and you run into a bunch of paradoxical cases trying to support this. Other versioning schemes might support it, but they won't also treat versions as mathematical objects the way rez does - ie things that you can invert, intersect, union and so on. I get around this by instead doing "1.0.rc.1", "1.0.rc.2", ..., "1.0.0" - this works because 1.0.0 > 1.0.rc.

3) Yes the default filters are set in rezconfig, but you can override them when you create a ResolvedContext, and you can add to/replace them with rez-env also. We don't have a separate rezconfig for each show no, but that would work. I'm moving toward a new system based on live resolves, where the result is going to be pretty much the same though - some config data (not managed by rez) will be hierarchically merged, and some of this will control rez stuff like package requests.. and filters.

PR #188, opened 2015-04-30
Minor fixes

elrond79:
if high verbosity, will print traceback for package command execution errors
misc other small cleanups


src/rez/resolved_context.py
```diff
@@ -23,6 +23,7 @@
 from rez.vendor import yaml
```
nerdvegas: @elrond79, did you know that by setting 'config.catch_rex_errors' to False, you can see the trace on rex errors also?

elrond79:
hmm... no, I didn't know that.  That would have given me what I needed!  Oh
well... I supose it still makes makes sense to print when using "-vv"...

PR #194, opened 2015-05-08
selftest: enable specifying of specific test methods

elrond79:
Liked the changes that you made to use the test module docstring for the selftest-arg docstring, but in the shuffle the ability to specify specific exact tests to run got lost (ie, to test ONLY "rez.tests.test_build.TestBuild.test_builds")

This commit adds that back in, so hopefully we now have the best of all worlds: automated test discovery, nice-formatted help for module tests, and the ability to run exact tests.


nerdvegas:
My apologies, I'd though that --tests was only choosing a test down to the same as selecting --build etc. Will merge shortly.

PR #197, opened 2015-05-12
prepend rez path

elrond79:
fix for https://github.com/nerdvegas/rez/issues/196

### DIFF
--- a/src/rez/resolved_context.py
+++ b/src/rez/resolved_context.py
@@ -1445,8 +1445,8 @@ class ResolvedContext(object):
         # append system paths
         executor.append_system_paths()
 
-        # append rez path
-        executor.append_rez_path()
+        # prepend rez path
+        executor.prepend_rez_path()
 
     def _append_suite_paths(self, executor):
         from rez.suite import Suite
--- a/src/rez/rex.py
+++ b/src/rez/rex.py
@@ -1077,11 +1077,15 @@ class RexExecutor(object):
         paths_str = os.pathsep.join(paths)
         self.env.PATH.append(paths_str)
 
-    def append_rez_path(self):
-        """Append rez path to $PATH, so that rez cli tools are available in
+    def prepend_rez_path(self):
+        """Prepend rez path to $PATH, so that rez cli tools are available in
         the generated environment."""
+        # We prepend because we want to make sure that if, ie, the user calls
+        # a production-installed rez binary, but has another version of rez on
+        # their path, that it uses the production-installed binaries in any
+        # subprocesses
         if system.rez_bin_path:
-            self.env.PATH.append(system.rez_bin_path)
+            self.env.PATH.prepend(system.rez_bin_path)
 
     @classmethod
     def compile_code(cls, code, filename=None, exec_namespace=None):


nerdvegas:
Hey @elrond79,

Unfortunately prepending rez bin dir to PATH causes a different error.

It goes like this: It is common practice to use rez-gui via the "rezgui" rez package. It's done this way so that PyQt/PySide doesn't have to be installed into the rez installation's virtualenv (that's a pain, and most studios will already have pyqt/pyside Rez packages).

So, you can run rez-gui like so:
]$ rez-env rezgui pyside -- rez-gui

We happen to be using suites to manage our packages, and we have it set up so that 'rez-gui' is a suite tool, which launches into the same env you would get via the rez-env call above. The problem is, prepending rez bin path to PATH is causing the native rez 'rez-gui' to be visible, rather than the suite's 'rez-gui', causing rez-gui to crash (pyqt/pyside is not available).

In many setups this is going to be undesirable - people don't want to have to install pyqt/pyside a second time into rez's virtualenv, so they will want their own 'rez-gui' tool visible, rather than rez's.

Since the PATH prepend is done in Rex, there's not an easy way to perform a more complex action, such as "replace rez if already in path else append."

Not sure what to do about this just yet, but it's a showstopper right now and I've had to revert our installation.

nerdvegas:
Ok, so I've fixed this by leaving things as they are, but updating rez-gui so that if it fails due to missing PyQt/PySide, it searches PATH for a different rez-gui executable, and runs that instead.

It feels a little bit dirty, but it does work, so I'll leave it at that unless we hit any more problems..

elrond79:
Cools, sounds good...

nerdvegas:
Hmm. This isn't going to work in some cases. We're hitting issues where rez-gui does eventually run, but after bouncing around 4 different rez-gui invocations, which causes an unacceptably large delay.

I'm going to change this so that whether rez is prepended or appended becomes configurable. the default will be to prepend as that matches existing behaviour, so @elrond79 if you need the prepend then be sure to set the option as such after I do the push. Should be pretty soon.

mstreatfield:
It's been a while since I've used rez-gui, but would this be simpler if rez-gui wasn't installed as part of rez itself?

We have a tool (called dora) that serves a very similar purpose to rez-gui.  However it is a separate package that depends on rez (which we also have a package for) and PyQt etc.  To use it, someone must 'rez env' it as they would any other package.  As rez-gui is a layer on top of rez - it's a separate product almost - and not part of the core functionality, this feels acceptable.

nerdvegas:
Politics. To get that to happen I'd need to get the OK for open source development on another project at my place of work. Generally I agree with you though.

mstreatfield:
You wouldn't have to take it out of the rez repository necessarily (although I think we should).  Initially you could just change the way it's built/installed within rez.  So instead of being picked up by `install.py` it's has it's own (somewhere, somehow).

elrond79:
Perhaps we could leave it appending, and then in situations where rez is
calling out to it's own binaries (ie, bez), it calls it explicitly, using a
full path? Of course, this would mean finding such use cases... but
hopefully a search for subprocess would catch most.

Or, thinking along mark's lines, there could be an option on install to NOT
install rez-gui.

Just throwing these out there as ideas.. I'd actually be ok with the
configuration for prepend instead of append... though it feels odd to be
using that just to make a unittest work.

nerdvegas:
@elrond79, there are cases where an absolute path to rez binaries isn't possible. For example, tool wrappers creates by suites are shebanged with _rez_fwd, hard pathing this case would break tools with new rez releases, or leave them on an old release, and neither are an option.

I agree with Mark that rez-gui should be a separate project, for now the configurable option I've added gives us enough to get around any issue and I'm inclined to leave it that way for the moment.

A

elrond79:
Like I said, I'm fine with a configurable option.

However - think you may have misinterpreted my suggestion. I meant that,
instead of doing:

    subprocess.call(['bez'])

we instead do:

    subprocess.call([os.path.join(system.rez_bin_path, 'bez')])

So we could apply it to some calls and not others (ie, rez-gui, or
wrappers).  Sort of moot, though, since I also agree that making rez-gui
it's own project is a better long-term solution...

PR #198, opened 2015-05-13
Version normalize

elrond79:
This is a fix for https://github.com/nerdvegas/rez/issues/190

Currently, resource handle objects have no means for enforcing standardization, which can result in cache misses, since they're used as keys.  This is particularly problematic with memory resources, since a cache miss can mean not just inefficiency, but retrieving the wrong data.

This branch adds a mechanism for allowing different resource types to normalize / standardize their variables, and then uses that to standard the way the "version" variable is stored.

There's also a few minor test updates in this branch

### DIFF
--- a/src/rez/package_repository.py
+++ b/src/rez/package_repository.py
@@ -187,12 +187,55 @@ class PackageRepository(object):
         """
         return 0
 
-    def get_resource(self, resource_key, **variables):
+    def make_resource_handle(self, resource_key, **variables):
+        """Create a `ResourceHandle`
+
+        Nearly all `ResourceHandle` creation should go through here, because it
+        gives the various resource classes a chance to normalize / standardize
+        the resource handles, to improve caching / comparison / etc.
+        """
+        if variables.get('repository_type', self.name()) != self.name():
+            raise RuntimeError("repository_type mismatch - requested %r, "
+                               "repository_type is %r"
+                               % (variables['repository_type'], self.name()))
+
         variables["repository_type"] = self.name()
-        handle = ResourceHandle(resource_key, variables)
+
+        if variables.get('location', self.location) != self.location:
+            raise RuntimeError("location mismatch - requested %r, repository "
+                               "location is %r" % (variables['location'],
+                                                   self.location))
+        variables["location"] = self.location
+
+        resource_cls = self.pool.get_resource_class(resource_key)
+        variables = resource_cls.normalize_variables(variables)
+        return ResourceHandle(resource_key, variables)
+
+    def get_resource(self, resource_key, **variables):
+        """Get a resource.
+
+        Attempts to get and return a cached version of the resource if
+        available, otherwise a new resource object is created and returned.
+
+        Args:
+            resource_key (`str`):  Name of the type of `Resources` to find
+            variables: data to identify / store on the resource
+
+        Returns:
+            `PackageRepositoryResource` instance.
+        """
+        handle = self.make_resource_handle(resource_key, **variables)
         return self.get_resource_from_handle(handle)
 
     def get_resource_from_handle(self, resource_handle):
+        """Get a resource.
+
+        Args:
+            resource_handle (`ResourceHandle`): Handle of the resource.
+
+        Returns:
+            `PackageRepositoryResource` instance.
+        """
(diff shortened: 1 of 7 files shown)


nerdvegas:
Hey Paul,

I have some concerns about the changes made to resources.py... this is quite generic code that I have also used in some other projects, but the changes you've made mix in some concepts that I think don't belong in this module... for example, introducing 'location' and 'repository_type' as necessary properties. Also, you are correct that 'variables' needs to be normalised, but it seems to me more appropriate to do that outside of this module. Also, why was the 'resource_pool' argument to PackageRepository removed in favour of 'cache_size'? Now each repository has its own pool, whereas before I was (intentionally) sharing a single pool across all repositories (via PackageRepositoryManager) so that I can control the total number of resources resident in memory (via config.resource_caching_maxsize).

Would leaving resources.py as-is and normalising the resource variables outside fix the issue? I don't yet understand why the other changes were made.

Glad to take the unit test updates though, but they're mixed in with this PR.

Cheers
A

nerdvegas:
ps - disregard what I said about location/repository_type, I haven't touched this code in a while and I see this is relevant because they are properties inherent to the PackageRepository class. Mostly then my concern is about the changes to use of the resource pooling.

thx
A

elrond79:
> Also, why was the 'resource_pool' argument to PackageRepository removed in favour of 'cache_size'? Now each repository has its own pool, whereas before I was (intentionally) sharing a single pool across all repositories (via PackageRepositoryManager) so that I can control the total number of resources resident in memory (via config.resource_caching_maxsize).

Hmm, good point. This was done because I ran across an issue where I was getting cross-repository contamination due to cache hits - ie, I did something like repoA.get_package('foo'), and ended up getting the foo package from repoB.  And I couldn't think of a reason why you would want to get cross-repo cache hits, so the simplest solution seemed to be to move the pool onto the Repository. I hadn't considered the desire to have a single pool for easier control over max cache size...

However, now that ResourceHandles should always have the "location" and "repository_type" variables added, cross-repo cache-hits shouldn't be able to happen anymore, so I can move the pools back onto the PackageRepositoryManager.

> Would leaving resources.py as-is and normalising the resource variables outside fix the issue?

Sure. I put the normalization on the Resource class because I was thinking that any subclass with variables might potentially want to standardize those variables - but I don't have any objection to moving them up to a higher level.  Maybe PackageRepositoryResource?  As long as it's somewhere that all classes that get registered on ResourcePools inherit from.

> I don't yet understand why the other changes were made.

Well, as far as the remaining changes - these are mostly related towards trying to funnel all ResourceHandle creation through PackageRepository.make_resource_handle.  Why did I want to do that? 

Well, first, we ensure we always get the "location" and "repository_type" variables added (even more important if we go back to a shared pool).

Second, since I wanted to allow Resource subclasses to customize normalization, we need a way to figure out which Resource class is associated with a given resource key - and this information is effectively stored in the ResourcePool.resource_classes... which the repository has access to. So making all handle creation go through the repo seemed to make sense.

---------------------------------------

Anyway, for now, I'll make a version of these changes that leave ResourcePools on the PackageRepositoryManager, so you can see what that would look like... and I'll make a separate pull request for just the unittest stuff.

nerdvegas:
Thanks, that all makes sense I think. I will probably put off merging this until next week... we're in full production use now so I'm being pretty careful about introducing changes one step at a time, especially if they touch anything a bit deep like this does. All our render blades do resolves so there's the real chance of taking out the whole studio(!!).

elrond79:
So, made another commit, which fixes one of the ugly bits dealing with the changes I made to resources.py - basically, that Resource.handle relied on ._repository.make_resource_handle.  Though this technically works, since currently the only class that inherits from Package is PackageRepositoryResource, it was very conceptually unclean, since the base class shouldn't have any knowledge of / interaction with repositories.

So, instead, to ensure that all handles are constructed with normalized data, I decided to just validate the variables on Resource creation; this seems much more straightforward, and also ensures that a Resource's variables should always match it's handle's variables.

nerdvegas:
@elrond79, meant to respond some time ago. Thx, this is still on my radar, will probably leave for another week or two to let things stabilise, but will merge after that.
A

PR #204, opened 2015-05-22
Misc Improvements

mstreatfield:
Various changes based on modifications we've made locally that I think can be pushed back to the main resources2 branch.

In each case the commit should explain the change well enough I think.  None are more than a couple of lines.  Nearly all are changes we've been running with for a few weeks so I think they are safe enough.


bpabel:
Closing this PR. It's very old and it's merging into another old branch. If it's still relevant we can open another.

PR #207, opened 2015-06-02
Custom per-package version priority

elrond79:
This branch provides a way to specify a custom solver prioritization order for the versions of specific packages.

It can be thought of as similar to the REZ_RESOLVE_MODE setting from earlier versions of rez, except that instead of only allowing "latest" or "earliest," it allows custom orderings on a per-package basis.

It is also similar to package filters, in that it provides a means to alter which packages are picked when resolving solve, except that the packages are still available as an option.  This means that, for instance, if a deprioritized package is a requirement of another package, that other package can always still solve.  It also doesn't require explicit command line overrides if a user wishes to use a deprioritized package.

This make package priorities more suitable for use by end-users / artists, which are the primary target audience.  For instance, suppose version 3 of the Foo software package is released, but it's a major redesign, and some people love it, and some hate it. Package reprioritization will allow artists to use whichever package they feel most comfortable with.

For a simple example, consider you have a package foo with versions [1, 2, 3].  By default the prioritization (from highest to lowest) is [3, 2, 1].  But if you add this to your .rezconfig:

```yaml
version_priorities:
  foo: [2, 3, 1]
```

...then do "rez-env foo", you will get version 2.  Also, since non-matching packages are automatically appended to the priority list, the same thing could have been accomplished with:

```yaml
version_priorities:
  foo: [2]
```

See the notes in rezconfig.py for details of usage.

Lemme know what you think!

- Paul

PS - Assuming you're on board, I'm fine with this being scheduled as a post-2.0 feature

PPS - It does NOT change the sort order of versions in any other way - it only alters the prioritization used for picking packages when solving; so, for instance, if the priority mode is changed to (highest priority to lowest) [2, 3, 1], then version 2 will still match the version range 1+<3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,11 @@
 
 Rez is a cross-platform software package management API and set of tools. Rez can
 build and install packages, and resolve environments at runtime that use a dependency
-resolution algorithm to avoid version conflicts. The main tools are:
+resolution algorithm to avoid version conflicts. Both third party and internally
+developed packages can be made into Rez packages, and any kind of package (python,
+compiled, etc) is supported.
+
+The main tools are:
 
 * **rez-env**: Creates a configured shell containing a set of requested packages.
   Supports *bash*, *tcsh* and *cmd* (Windows), and can be extended to other shells.
--- a/src/rez/cli/python.py
+++ b/src/rez/cli/python.py
@@ -13,6 +13,7 @@ def setup_parser(parser, completions=False):
     parser.add_argument(
         "ARG", type=str, nargs='*',
         help='arguments to python script')
+    parser.add_argument('-c', help="python code to execute", dest='command')
 
     if completions:
         from rez.cli._complete_util import FilesCompleter
@@ -28,6 +29,9 @@ def command(opts, parser, extra_arg_groups=None):
 
     if opts.interactive:
         cmd.append("-i")
+
+    if opts.command:
+        cmd.extend(['-c', opts.command])
 
     if opts.FILE:
         cmd.append(opts.FILE)
--- a/src/rez/config.py
+++ b/src/rez/config.py
@@ -167,29 +167,6 @@ class OptionalDictOrDictList(Setting):
     _env_var_name = None
 
 
-class VersionPriorityDict(Setting):
-    schema = Or(And(None, Use(lambda x: {})),
-                Schema({
-                    str: And(
-                        [Or(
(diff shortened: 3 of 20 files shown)


nerdvegas:
Hey Paul,

Off the top of my head I don't see a problem with this in any technical
sense. The only issue I see when put into practice is that users will start
abusing the feature in order to avoid new package releases (from the
pipeline department for eg). I can see nasty cases arising where a studio
has a runtime resolving system (ie, resolves occur at tool load time),
where an artist reports a bug, only for the pipeline developer to discover
that they've effectively locked themselves to an earlier package version(s)
by changing this setting in their user .rezconfig file.

But, with great power etc. They can also use package_filter in the same
nefarious way, so hey.

Also, in Rez-1 you could change the resolve mode to earliest or latest as
you mention. Recently someone mentioned that a studio might actually be
interested in the 'earliest' option being returned. I wonder if it would be
useful not only to set the priority version order for a package, but to be
able to mark a package and set it to 'earliest' mode? Ie,
'version_priorities' could also be a string ('latest', 'earliest') as well
as a list. If this is the case, we should also add back in the global
earliest/latest setting. And perhaps 'version_priority' would be a slightly
more appropriate name.

Thoughts?
A

elrond79:
> Off the top of my head I don't see a problem with this in any technical
> sense. The only issue...
> But, with great power etc. They can also use package_filter in the same
> nefarious way, so hey.

True... perhaps we could have an option to disable / ignore the setting? Of course, then we would need a way to make sure THAT setting wasn't overridden - somehow marking a setting as non-overrideable?  Or just rely on the knowledge not filtering out...


> Also, in Rez-1 you could change the resolve mode to earliest or latest as
> you mention. Recently someone mentioned that a studio might actually be
> interested in the 'earliest' option being returned. I wonder if it would be
> useful not only to set the priority version order for a package, but to be
> able to mark a package and set it to 'earliest' mode? Ie,
> 'version_priorities' could also be a string ('latest', 'earliest') as well
> as a list. If this is the case, we should also add back in the global
> earliest/latest setting.

Sure, I think that would be easy...

> And perhaps 'version_priority' would be a slightly
> more appropriate name.

Can do...

elrond79:
Ok, changed the name to version_priority, and added ability to set a package's version_priority to latest (default) or earliest, as well as a custom list

elrond79:
https://github.com/nerdvegas/rez/pull/265

PR #208, opened 2015-06-03
Variant select mode

elrond79:
fixes a bug where variant_select_mode was always using intersection_priority mode, regardless of what the setting was

Also added some more clarification / examples to the description in rezconfig

The source of the problem was a comparison of a string to the VariantSelectMode enum, which always returns False. Would like to protect against future problems of this sort by modifying the __eq__ for the enum to either error or do a "real" comparison against strings and integers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,11 @@
 
 Rez is a cross-platform software package management API and set of tools. Rez can
 build and install packages, and resolve environments at runtime that use a dependency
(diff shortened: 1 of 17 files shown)


nerdvegas:
Thanks for spotting this. Will catch up with these recent misc PRs soon.
A

elrond79:
Cool. No rush...

elrond79:
https://github.com/nerdvegas/rez/pull/259

PR #210, opened 2015-06-10
Allow to rez-release without releasing tags

holofermes:
allow to rez-release without releasing the tag, and without getting the changelog.
Actually in the  code there is still no logic for not getting the changelog, so this needs to be implemented.

### DIFF
--- a/src/rez/build_process_.py
+++ b/src/rez/build_process_.py
@@ -17,7 +17,7 @@ def get_build_process_types():
 
 
 def create_build_process(process_type, working_dir, build_system, vcs=None,
-                         ensure_latest=True, verbose=False):
+                         ensure_latest=True, no_update_repo=False, verbose=False):
     """Create a `BuildProcess` instance."""
     from rez.plugin_managers import plugin_manager
     process_types = get_build_process_types()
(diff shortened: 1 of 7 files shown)


nerdvegas:
I've changed this a little.

Rather than a rez-release --no-update-repo option, there's now a --skip-errors option. The option is here at all really for the case where something is going wrong (probably do to with communicating with the repo) but a package really needs to be released. I figured that trying to tackle many different specific issues (can't write a tag; can't access repo; can't find an old commit etc) would result in lots of different options that are rarely used and are overall confusing. Instead, just one - tolerate any repo-related errors.

I've also added configurable behaviour so that rez-release can abort a package release, if that package version is already tagged on the repo (currently an existing tag does not cause a problem). This gives a way to coordinate package releases in a multi-site scenario, so that one site doesn't release the same version (but different commit) of the same package. See the comments for this new option ('check_tag' in rezconfig.py) for a longer description.

nerdvegas:
I've also just noticed that 'skip_errors' in the code would more appropriately be called 'skip_repo_errors', will make that change just now.

PR #217, opened 2015-06-25
Allow the repository for the release package to live at a higher level

elrond79:
This way, the repo can be at, for instance, the package family level, instead of at the versioned package level.  ie, right now, you are required to have:

my_package/1.1.0/.git
my_package/1.1.2/.git
my_package/2.0.0/.git
my_package/2.0.1/.git
my_package/3.5.1/.git

However, it's fairly frequent that you might have a change which you will need to be applied to all 5 versions.  In this case, you then need to also make 5 separate commits, which seems both more work, and organizes the repository information in a less useful way.  With this commit, you can instead have:

my_package/.git

...and make only one commit, containing the changes to all 5 versions.

(FYI, This pull request also contains the commit from my other hg_vcs pull request...)

### DIFF
--- a/src/rez/build_process_.py
+++ b/src/rez/build_process_.py
@@ -70,7 +70,7 @@ class BuildProcess(object):
         self.vcs = vcs
         self.ensure_latest = ensure_latest
 
-        if vcs and vcs.path != working_dir:
+        if vcs and vcs.pkg_root != working_dir:
             raise BuildProcessError(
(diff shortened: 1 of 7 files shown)


nerdvegas:
Great, it just so happens another studio will benefit from this feature
currently.

Cheers
A

elrond79:
Cool, good to hear...

nerdvegas:
Merged to resources2, not sure why github missed it.

PR #220, opened 2015-07-02
Update schema to 0.3.1 + Optional defaults

elrond79:
The latest version of schema has support for defaults for the Optional class.  Merged in the changes to the rez version of schema.

Also, to ease any such future merges of upstream schema changes, I created a "schema_official" branch (with tags indicating versions); the idea is that when a new version of schema is released that we want to pull into rez, you commit the new unaltered files on the schema_official branch, then merge that into the main branch.  

Since all of rez's changes are now in commits applied on top of the schema_official branch, we can leverage git's merging tools to make merging of these official changes easier.

Also fixed a few small issues with the schema module / tests.

### DIFF
--- a/src/rez/vendor/schema/schema.py
+++ b/src/rez/vendor/schema/schema.py
@@ -12,22 +12,12 @@ class SchemaError(Exception):
 
     @property
     def code(self):
-        # It appears that the uniq method is attempting to return a unique list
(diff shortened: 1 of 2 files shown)


nerdvegas:
Nice, thanks.

PR #226, opened 2015-07-20
checking if packages path item exists before call to get_repository

reedwade:
As described in issue219


reedwade:
um -- crap, sorry -- 225 not 219

ignore this, I'm going make a better named branch also

PR #227, opened 2015-07-21
checking if packages path item exists before call to get_repository

reedwade:
A fix for Issue 225.

Unsure if it would be better to catch this deeper in. It feels wrong for the package repository module to not deal better with this case.

### DIFF
--- a/src/rez/packages_.py
+++ b/src/rez/packages_.py
@@ -326,6 +326,8 @@ def iter_package_families(paths=None):
         `PackageFamily` iterator.
     """
     for path in (paths or config.packages_path):
+        if not os.path.isdir(path):
+            continue
         repo = package_repository_manager.get_repository(path)
         for resource in repo.iter_package_families():
             yield PackageFamily(resource)


nerdvegas:
Hi Reed,

Apologies but looking at this code once more has jogged my memory. This fix doesn't work. That's because 'path' is not necessarily a filepath - with recent changes, 'path' could be something else for other, non-filesystem-based package resources. It might be a database URI, for example. The name 'path' is a bit misleading and it's still there mostly for historical purposes.

Was there a bug associated with this fix? I looked at package_repository/filesystem.py again and I think this issue has already been dealt with - if the path does not exist, it just behaves as a package repository containing zero package families.

Please let me know if there's currently an issue and I will look into how to best fix it. In the meantime I'll close this PR.

Thanks,
A

reedwade:
makes sense

I've patched our local copy which is fine for now for our testing.

I'll have a poke around inside there to understand better.

reedwade:
oh, and the bug exhibits as a big chunk of python stack trace when you attempt auto-completion and if the package list contains non-existent directories (like ~/packages for me since I don't have one)

I'd copy in the error here but it's in a weird network I can't easily copy bits out of right now.

PR #234, opened 2015-08-13
Load from file fixes

elrond79:
some minor fixes for serialize.load_from_file:
removed the data_type kwarg, as it was non-functional and unused
added format_ and update_data_callback to the memcache key for _load_from_file (since they affect the results)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,11 @@
 
 Rez is a cross-platform software package management API and set of tools. Rez can
 build and install packages, and resolve environments at runtime that use a dependency
-resolution algorithm to avoid version conflicts. The main tools are:
+resolution algorithm to avoid version conflicts. Both third party and internally
+developed packages can be made into Rez packages, and any kind of package (python,
+compiled, etc) is supported.
+
+The main tools are:
 
 * **rez-env**: Creates a configured shell containing a set of requested packages.
   Supports *bash*, *tcsh* and *cmd* (Windows), and can be extended to other shells.
--- a/src/rez/cli/python.py
+++ b/src/rez/cli/python.py
@@ -13,6 +13,7 @@ def setup_parser(parser, completions=False):
     parser.add_argument(
         "ARG", type=str, nargs='*',
         help='arguments to python script')
+    parser.add_argument('-c', help="python code to execute", dest='command')
 
     if completions:
         from rez.cli._complete_util import FilesCompleter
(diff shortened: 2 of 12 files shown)


elrond79:
Third commit isn't really a fix, but sort of a feature - basically, I broke out a separate mem-cached read_file func, to avoid unnecessary file reads (ie, if kwargs to load_from_file change).

nerdvegas:
Hey Paul,

I'm going to incorporate this PR soon (a really interesting bug came up related to file loads). but I noticed something - now that you have two @memcached decorated functions - one calling the other - I think this might actually double the number of memcache entries for the same file, as well as doubling the amount of traffic to and from the memcached server.

elrond79:
@nerdvegas - Yeah, since we have nested @memcached functions, the initial call to _load_from_file does result in two interactions with with the memcached server.  Subsequent calls to _load_from_file, though, should simply hit the cache on the outer call, so traffic isn't doubled... unless, of course, most calls to _load_from_file are cache misses, in which case, the cache isn't doing much anyway.

You're right about efftictively doubling storage, however.  My thinking was that hitting the disk is so expensive that we'd want to avoid repeating it whenever possible - so, ie, if _load_from_file is called twice for the same file, but with different args, we'd want to avoid re-reading from disk.

If you think the extra storage cost is too high, though, I'm not opposed to getting rid of the read_file func.  Hopefully reading the same file with different args doesn't happen too often, and the cache around _load_from_file would protect us from having to do it too often.

Alternatively, we could also remove the caching from _load_from_file, and leave it on read_file.  If the processing cost of the stuff that _load_from_file does isn't too high, that might make sense... not sure how expensive that is, though.

nerdvegas:
Yeah I'm thinking we should keep it simple. Even though different args -
format, callback - could in theory change the results, in practice I know
all the cases where they are used and the extra overhead doesn't win us
anything.

I would sorely like to have just a memcached read_file() function, however
there is a gotcha. The reason update_data_callback is there at all is to
trim changelogs which for some reason, in some of our packages got
extremely large. If we just had read_file(), these extremely large package
definitions would end up in the cache, blowing away lots of other cache
entries (that's what was actually happening). It's annoying but at the
moment I can't see a way around it.

A

elrond79:
Hmm - so what you say about a memcached read_file makes sense... so I agree we should ditch the third commit.

But if I read you right, you're also advocating ditching the second commit - with the changes to the memcache key - which I think is dangerous. Even if currently it doesn't buy us anything, it's good protection against bugs in the future.

elrond79:
killed that read_file commit...

elrond79:
https://github.com/nerdvegas/rez/pull/257

PR #237, opened 2015-08-14
added VariantResourceHelper.variant_requires easy-access property

elrond79:
title says it all... needed to query this a few times in my own code, thought it would be handy / make things more readable


elrond79:
https://github.com/nerdvegas/rez/pull/256

PR #238, opened 2015-08-18
Cli callbacks

elrond79:
We wanted a way to run some checks before allowing rez-release to run, so I added a configurable callback system for cli commands

basically, you can set arbitrary python code to run before or after specific commands (and the pre-callbacks have a chance to abort)

### DIFF
--- a/src/rez/cli/_main.py
+++ b/src/rez/cli/_main.py
@@ -80,7 +80,10 @@ def _add_common_args(parser):
                              "successfully, and None if it did not; "
                              "error - None if the command completed "
                              "succesfully, and the exception instance if it "
-                             "did not")
+                             "did not. Note that for commands which can "
+                             "start interactive shells, such as rez-env, "
+                             "this command will not be run until after the "
+                             "shell exits")
 
 
 class InfoAction(_StoreTrueAction):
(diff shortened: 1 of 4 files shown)


nerdvegas:
Hey Paul, what about the release_hook plugin? This is specifically for pre-
and post- release operations. There's a 'command' type that allows you to
run any arbitrary command, with options to abort the release and so on. We
use this pretty extensively. Another hook - emailer - sends emails on
package release.

These hooks will work regardless of whether a release happens as a result
of someone using the rez-release tool, or whether a release occurs via code
(not something we're doing yet but who knows). Is there need for a general
callback system, especially if it only works with cli tools?

Thx
A

elrond79:
Hmm, didn't know about that plugin - I'll check it out when I get into work, but i'm guessing it would serve our needs.

We don't have a specific need for any other callback type - iIf you want to axe this PR, I'm fine with that.

nerdvegas:
I'll close it for now. Let me know if there are issues with the release hook at all. There's been talk in the past of extending it to support more hook points, such as pre/post-build, that is certainly an option also.

elrond79:
Cool... just got our callback working with the existing release_hooks.command... I like it, does exactly what we need!

nerdvegas:
Awesome, apologies it isn't documented anywhere yet. Starting on that soon.

A

mstreatfield:
> These hooks will work regardless of whether a release happens as a result
of someone using the rez-release tool, or whether a release occurs via code
(not something we're doing yet but who knows)

Just to pick up this comment... we do a release via code and not the normal `rez release` cli.

PR #240, opened 2015-08-22
release_vcs/svn: added get_current_revision, create_release_tag

elrond79:
just added some missing func to the svn release plugin, to make it minimally functional


elrond79:
https://github.com/nerdvegas/rez/pull/258

PR #241, opened 2015-08-22
Check to make sure that essential files (ie, package.py, etc) are tracked in revision control before release

elrond79:
It happened a couple times that I accidentally forgot to add the package.py + build files to the repository before doing a rez release.

This pull request makes it so that the release check ensures that essential files are actually tracked in the repo (as well as checking for outstanding changes) before allowing the commit to go through.  Essential files are defined as the package file, plus any files registered by the build system.

Support was added for hg, git, and svn, as well as tests for all 3 (which is also nice because, as far as I could tell, there were no tests previously for these revision control systems).  All tests are guarded by skip-condtions, that check that the appropriate tool is installed.

This pull request also merges in the svn_release PR I just made - if for some reason you want these changes, but not the ones in svn_release, just merge the commit before that merge, and it should be functional (just without svn support for checking if essential files are tracked)


elrond79:
https://github.com/nerdvegas/rez/pull/264

PR #247, opened 2015-09-17
Minor fixes

elrond79:
docstring update
fix for rez.config.Dict/OptionalDict 
use functools.wraps to preserve docstring/func name


elrond79:
https://github.com/nerdvegas/rez/pull/260

PR #248, opened 2015-09-22
In combined package yaml files, allow versions to be given with ints or floats

elrond79:
is particularly nice for yaml files, since you can now do, ie:
  versions: [1.5]
...instead of:
  versions: ['1.5']
...which feels more consistent, since you can also do:
  versions [1.5.3]


nerdvegas:
Hey Paul,

Float support for version numbers was actually removed because it causes a bug.

Specifically: 0.9 !< 0.10 in float, but should be as version number.

elrond79:
Ah, good point. I'll close this pull request

PR #249, opened 2015-09-22
Physical cpus

elrond79:
Allow setting of build_thread_count to either number of physical cores, or logical cores.
Will differ if hyperthreading is enabled, in which case logical_cores == 2 * physical_cores

My personal experience has been that setting to number of physical_cores leaves the system more responsive, at minimal loss of compilation time...


nerdvegas:
Closing, merged to master.

PR #251, opened 2015-10-01
rez-python: add support for "-c" cmdline arg to run arbitrary string

elrond79:
add support for running arbitary python strings with rez-python


nerdvegas:
Closing, merged to master.

PR #252, opened 2015-10-13
when using rez-build --ba pass the string to bez as well

holofermes:
like the title says.
Currently this change would make all the previous bez build fail, as the arguments to bez's build function has one more argument. Also it would be up to the implementer of rezbuild.py to re-build arguments as they would get as a string, ie build_args="foo=1 bar=cat".


nerdvegas:
Closing, merged with master.

PR #257, opened 2015-10-28
Load from file fixes

elrond79:
redo of:

https://github.com/nerdvegas/rez/pull/234

Original desc:

some minor fixes for serialize.load_from_file:
added format_ and update_data_callback to the memcache key for _load_from_file (since they affect the results)


src/rez/serialise.py
```diff
@@ -90,9 +90,18 @@ def load_from_file(filepath, format_=FileFormat.py, update_data_callback=None):
...
     st = os.stat(filepath)
-    return str(("package_file", filepath, st.st_ino, st.st_mtime))
+    if update_data_callback is None:
+        callback_key = 'None'
+    else:
+        if hasattr(update_data_callback, 'im_self'):
+            callback_key = '%r - %r' % (update_data_callback.im_self,
+                                        update_data_callback)
+        else:
+            callback_key = repr(update_data_callback)
```
nerdvegas: This is a problem no? repr(some_static_func) gives:
<function blah at 0x4543534546>

So each new proc is going to miss the memcached entry.

PR #263, opened 2015-10-29
Logfile

elrond79:
redo of:

https://github.com/nerdvegas/rez/pull/232

Original desc:

added a flag for easier logging of all output from rez + subprocesses - handy for rez-build/release

### DIFF
--- a/src/rez/utils/logging_.py
+++ b/src/rez/utils/logging_.py
@@ -58,19 +58,69 @@ class _Printer(object):
 #   http://stackoverflow.com/questions/4984428/python-subprocess-get-childrens-output-to-file-and-terminal
 # by user515766 (http://stackoverflow.com/users/515766/user515766)
 
-def tee(infile, *files):
+def tee(infile, *files, **kwargs):
     """Print `infile` to `files` in a separate thread."""
-    from threading  import Thread
+    from threading  import Thread, Condition
+    import os
+    import time
+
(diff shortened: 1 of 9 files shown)


src/rez/utils/logging_.py
```diff
@@ -52,3 +51,93 @@ def __call__(self, msg, *nargs):
...
+        kwargs['stderr'] = PIPE if stderr is not None else None
+    p = Popen(cmd_args, **kwargs)
+    threads = []
+    if stdout is not None:
+        threads.extend(tee(p.stdout, stdout, sys.stdout))
+    if stderr not in (None, STDOUT):
+        threads.extend(tee(p.stderr, stderr, sys.stderr))
+    for t in threads:
+        t.join() # wait for IO completion
+    return p.wait()
```
nerdvegas: Wouldn't it be simpler just to add a file handler to the standard python logger? Is there a reason you did it this way?

elrond79: That would get most of the output by rez itself...  but the aim here was (primarily) to provide a way to log the results of rez-build and rez-release... most of whose output is due to cmake, make, etc... whose output is out of the control of rez itself - and so the only real option is to do a tee.

This evolved because I found that I had created a wrapper shell script that I was always invoking instead of calling rez-build directly, purely to do this logging.

nerdvegas: Hey Paul, thanks for the clarification, that makes sense.

A

PR #265, opened 2015-10-29
Custom package priority

elrond79:
redo of:

https://github.com/nerdvegas/rez/pull/207

Original desc:

This branch provides a way to specify a custom solver prioritization order for the versions of specific packages.

### DIFF
--- a/src/rez/config.py
+++ b/src/rez/config.py
@@ -167,29 +167,6 @@ class OptionalDictOrDictList(Setting):
     _env_var_name = None
 
 
-class VersionPriorityDict(Setting):
-    schema = Or(And(None, Use(lambda x: {})),
-                Schema({
-                    str: And(
-                        [Or(
-                                # Allow False, or "", which is converted to
-                                # False
-                                lambda x: x is False,
-                                And(lambda x: isinstance(x, basestring)
-                                              and not x,
-                                    Use(bool)),
-                                # Or a VersionRange,
-                                VersionRange,
-                                # ...or value which can be converted to a
-                                # VersionRange
-                                And(basestring, Use(VersionRange)),
-                                And(Or(int, float), Use(str),
-                                    Use(VersionRange)))],
-                        # Verify that there's at most one False
-                        lambda x: x.count(False) <= 1)
-                })
-    )
-
 class SuiteVisibility_(Str):
     @cached_class_property
     def schema(cls):
@@ -202,6 +179,43 @@ class VariantSelectMode_(Str):
     def schema(cls):
         from rez.solver import VariantSelectMode
         return Or(*(x.name for x in VariantSelectMode))
+
+
+class VersionPriority_(Setting):
+    @cached_class_property
+    def schema(cls):
+        from rez.solver import VersionPriorityMode
+        return Or(
+            {},
+            # None, in which case it's an empty dict..
+            And(None, Use(lambda x: {})),
+            # Or they give a dict, mapping from package name to setting...
+            Schema({
+                str: Or(
+                    # each package is either "lastest" / "earliest", or...
+                    VersionPriorityMode,
+                    And(Or(*(x.name for x in VersionPriorityMode)),
+                        Use(VersionPriorityMode.__getitem__)),
+
+                    # ...a list, giving an explicit version priority
+                    And(
+                        [Or(
+                            # Allow False, or "", which is converted to
+                            # False
+                            False,
+                            And("", Use(bool)),
+                            # Or a VersionRange,
+                            VersionRange,
+                            # ...or value which can be converted to a
+                            # VersionRange
+                            And(basestring, Use(VersionRange)),
+                            And(Or(int, float), Use(str),
+                                Use(VersionRange)))],
+                        # Verify that there's at most one False
+                        lambda x: x.count(False) <= 1)
+                )
+            }),
+        )
 
 
 class RezToolsVisibility_(Str):
@@ -338,7 +352,7 @@ config_schema = Schema({
     "env_var_separators":                           Dict,
     "variant_select_mode":                          VariantSelectMode_,
     "package_filter":                               OptionalDictOrDictList,
-    "version_priority":                             VersionPriorityDict,
+    "version_priority":                             VersionPriority_,
     "new_session_popen_args":                       OptionalDict,
 
     # GUI settings
--- a/src/rez/config.py
+++ b/src/rez/config.py
@@ -184,38 +184,69 @@ class VariantSelectMode_(Str):
 class VersionPriority_(Setting):
     @cached_class_property
     def schema(cls):
-        from rez.solver import VersionPriorityMode
         return Or(
             {},
             # None, in which case it's an empty dict..
             And(None, Use(lambda x: {})),
             # Or they give a dict, mapping from package name to setting...
             Schema({
-                str: Or(
-                    # each package is either "lastest" / "earliest", or...
-                    VersionPriorityMode,
-                    And(Or(*(x.name for x in VersionPriorityMode)),
-                        Use(VersionPriorityMode.__getitem__)),
-
-                    # ...a list, giving an explicit version priority
-                    And(
-                        [Or(
-                            # Allow False, or "", which is converted to
-                            # False
-                            False,
-                            And("", Use(bool)),
-                            # Or a VersionRange,
-                            VersionRange,
-                            # ...or value which can be converted to a
-                            # VersionRange
-                            And(basestring, Use(VersionRange)),
-                            And(Or(int, float), Use(str),
-                                Use(VersionRange)))],
-                        # Verify that there's at most one False
-                        lambda x: x.count(False) <= 1)
-                )
+                str: cls.version_priority_entry_validate_schema,
             }),
         )
+
+    @classmethod
+    def _make_version_priority_entry_schema(cls, output_python_objs):
+        from rez.solver import VersionPriorityMode
+
+        if output_python_objs:
+            to_ver_prio_mode = Use(VersionPriorityMode.__getitem__)
+            to_ver_range = Use(VersionRange)
+        else:
+            nop = lambda x: x
+            to_ver_prio_mode = nop
+            to_ver_range = nop
+
+        return Or(
+            # each package is either "lastest" / "earliest", or...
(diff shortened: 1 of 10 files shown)


nerdvegas:
Hey Paul,

I'm just about to start work on a rez feature that we have need of at Method. I've realised that it is actually a generalisation of your per-package version priority work. I wanted to describe it to see if you might find it useful also. I'm thinking that it will probably make sense to reimplement this PR in terms of this new feature.

So I wanted to introduce the idea of a "soft timestamp". The idea is that if a soft timestamp is set, rez will prefer to use packages released before this time (as with a usual timestamp), but if necessary, will use packages released after (contrast with a standard timestamp, where newer packages are ignored completely).

The reason we want this is because I'm looking for a good compromise between 'pushing' packages to production (ie, overseeing every package version update before it goes to production) and 'pulling' packages (ie, do resolves at runtime, get whatever latest packages match the request). Both have their downsides - the former requires a lot of maintenance, and there's generally just too much info for a person to take in (they can see that say, 10 packages have newer versions, but it's too much for one person to quantify what they are); the latter isn't stable enough, because newer packages often match a current request, and they get consumed automatically.

Soft timestamps are the compromise. You might, say, set a soft timestamp at the start of each week. That gives stability during the week, but specific requests can be made for newer packages if and when necessary. Certain developers can be chosen to test the latest packages (they would turn the soft timestamp off). A slightly different flavor of soft timestamp - a soft patch timestamp - would let new patches past the timestamp, but not minor or major versions.

In order to implement this, I think we just need to introduce the general idea of a "version priority function". I think you can see that this is only slightly different to what you've done in this PR.

Thx
A

elrond79:
Hmm... yeah, I could totally add the ability to specify a custom function to use in sorting.  Only question would be how to allow specification - via a module and function name, ie `python:myModule.myFunc`, that would then be imported? Or would we want to allow specification of the function body itself somehow, within the config?

Another idea occurs to me, though, which would probably fit your needs and mine, and re-use a paradigm already introduced: what if we just made it take filter "rules"?

Those would allow us to specify both version ranges (like the current implementation uses), and timestamps (like you want).  We'd also get support for globs + regexes for free, which I can see being potentially helpful.  We could do this in addition to custom functions if you still feel we need the full flexibility that would offer.

Lemme know what you think...

nerdvegas:
I thought about combining filters and 'orderers', but I think it complicates things. And it doesn't really match how we think about things. For example, we use a glob package filter to turn 'dev' packages on/off - but we do that entirely independently of changing the package version order. If those were combined it would be harder to manage.

I've had to start on this already because we need this feature asap, but I have your current implementation in mind and I think it'll transfer over just fine.

In terms of configuration, I was going to keep it simple, very similar to how the package_filters works currently - just a simple name lookup for each class, and each config entry is a dict used to initialise the orderer instance(s). Like package filters, I'm not sure that people will ever want to extend this, and we can make it fancier later if need be. 

I also have it so you can set multiple orderers at once, and the orderer class has the option to define a package's versions order or to pass on to potentially the next orderer. I can't think off the top of my head where multiple orderers would be useful but it's essentially free to add this support. If I didn't have that then we'd have to decide which orderer has precendence, and there's not really any right answer to that.

nerdvegas:
Oh I meant to say also, there may also be unforseen consequences in the solver. Meaning, it may be possible to apply filters to packages in the solve earlier than reordering, which could have important implications for resolve speeds. There are still changes to the solver I want to investigate (resolve speed / minimising package loads is extremely important to us) and something tells me that keeping these separate may make that easier also.

elrond79:
I don't mean combining filters and prioritization into one entity - they would be separate things / concepts, with separate configurations. I just meant that, for configuration, they would accept the same sort of things - ie, speak the same language.  In other words, the same logic for processing filter configuration - the "filter rules" system - would be used for processing the prioritization configuration.  ie, so we might do:

version_priority:
    mypackage:
        - glob(*.AwesomeEdition)
        - after(1429830188)
        - range(8+)

...which looks similar to the configuration for filter rules, because it's using the same processing backend.

nerdvegas:
Oh ok I misunderstood.

So you mean, by using the same 'language' in the config as filters, you can
apply reordering to only certain packages? Perhaps we'd have something like:


reorderers:
- myexample:
  class: TimestampPackageOrderFunction
  timestamp: 3012854
  filter:
  - glob(foo_*)

So here we would be applying my 'soft' timestamp reordering only to foo_*
packages.
?

A

elrond79:
What I was originally imagining was something more like this:

```
reorderers:
# items at the top level are explicit package family names
    foo:
        # list memebrs beneath the package are matches against versions
        # within a package family.
        - before(3012854)
```

This would set the "foo" package to use your soft-timestamp behavior, because anything which doesn't match an explicity priority group is automatically added to a priority group at the end - ie, it would be equivalent to:

```
reorderers:
    foo:
        - before(3012854)
        - regex(*)
```

If we wanted to have a "default" reorderer / version_priority for all packages, we could do something like this:

```
reorderers:
    True:
        - before(3012854)
```

Of course, this doesn't support matching against a package name - if we wanted to do that, we could go with something like this:

```
reorderers:
    - family: glob(foo_*)
      priority_groups:
          - before(3012854)
```

If we wanted to expand to allow arbitrary functions, we could do:

```
reorderers:
    - family: glob(foo_*)
      priority_func: mymodule.orderByPopularity
    - family: glob(*)
      priority_groups:
          - before(3012854)
```

I do feel that, even if we allow custom functions, we should have syntax to handle "common" reordering scenariors, such as by version range or timestamp.

nerdvegas:
So you mean that you just want to be able to utilise filters more
generally, and in this case within a reorderer class. I think that's what
you mean. In the soft timestamp case it doesn't really apply - the behavior
is not as simple as taking all packages before a time and prioritising them
(and using any filter other than 'before' wouldn't make sense). So I think
that's why I was confused.

In any case yes I can see how that would be useful in other cases. It's
easy enough to do already: You just need to call
package_filter.Rule.parse_rule(). We'd just have to add that as a validator
in config.py. I'll keep this in mind.

In terms of arbitrary priority functions, there's not much point in this,
unless it's a proper plugin. I don't want to introduce package.py
attributes that the standalone rez installation can't consume. At the same
time, I'm not sure people would be interested enough in extending the
default filters to make it a plugin in its own right - but happy to change
later if I'm wrong.

A

elrond79:
> So you mean that you just want to be able to utilise filters more
generally, and in this case within a reorderer class.

Exactly!

> In the soft timestamp case it doesn't really apply - the behavior
is not as simple as taking all packages before a time and prioritising them

Hmm... the basic scenario seems to fit, but you have a point about your "let patch revisions through, but not minor or major" ability - that would certainly require more complexity. 

> and using any filter other than 'before' wouldn't make sense

I disagree - our original implementation essentially is using the "range" filter... and I can see uses for the text-matching ones, ie:

    `regex((?<!\.beta)$)`

> In terms of arbitrary priority functions, there's not much point in this,
unless it's a proper plugin. I don't want to introduce package.py
attributes that the standalone rez installation can't consume.

Ok, good point.  As long as one of the "standard" reorderers allows reordering based on explicit versions, we're covered.

One last request, though - if we are going with different class based reorderers, I suggest that each reorderer be allowed to determine for itself whether it applies to a given package family - ie, there would be a method like:

```
ReordererClass.applies_to_family(family_name)
```

that would return a True / False value if it should be used for a given package family.  The reason why is because, if not, our use case - of wanting to explicitly specify "soft maximum" versions per package - becomes much more cumbersome.  It would look something like:

```
reorderers:
- class: ExplicitVersionOrderFunction
  filter: [maya]
  version_groups: ['<2016']
- class: ExplicitVersionOrderFunction
  filter: [nuke]
  version_groups: ['<9']
- class: ExplicitVersionOrderFunction
  filter: [other_package]
  version_groups: ['<30']
```

...whereas if each reorderer can have different methods for determining applicability, it could look like:

```
reorderers:
- class: ExplicitVersionOrderFunction
  version_groups:
     maya: ["<2016"]
     nuke: ["<9"]
     other_package: ["<30"]
```

...which is both much tidier, and more efficient (ie, we woldn't have to do a series of checks for applicability, but just a single test for dict membership).

nerdvegas:
Cool, see below.

On Tue, Dec 8, 2015 at 3:33 PM, Paul Molodowitch <notifications@github.com>
wrote:

> So you mean that you just want to be able to utilise filters more
> generally, and in this case within a reorderer class.
>
> Exactly!
>
> In the soft timestamp case it doesn't really apply - the behavior
> is not as simple as taking all packages before a time and prioritising them
>
> Hmm... the basic scenario seems to fit, but you have a point about your
> "let patch revisions through, but not minor or major" ability - that would
> certainly require more complexity.
>
> and using any filter other than 'before' wouldn't make sense
>
> I disagree - our original implementation essentially is using the "range"
> filter... and I can see uses for the text-matching ones, ie:
>
I meant that other filters don't make sense specifically for the soft
timestamp reorderer.


> `regex((?<!\.beta)$)`
>
> In terms of arbitrary priority functions, there's not much point in this,
> unless it's a proper plugin. I don't want to introduce package.py
> attributes that the standalone rez installation can't consume.
>
> Ok, good point. As long as one of the "standard" reorderers allows
> reordering based on explicit versions, we're covered.
>
Yep that was my intention.


> One last request, though - if we are going with different class based
> reorderers, I suggest that each reorderer be allowed to determine for
> itself whether it applies to a given package family - ie, there would be a
> method like:
>
> ReordererClass.applies_to_family(family_name)
>
> Ah yeah they can already do that. Each reorderer has a reorder() function
that can either return a reordered package list, or None. If None, the
package list is passed to the next reorderer, if there is one. As an aside
- there's no such thing as multiple reorders on a package - I think this
would just be confusing and I can't see the use case.

elrond79:
> Ah yeah they can already do that. Each reorderer has a reorder() function
that can either return a reordered package list, or None. If None, the
package list is passed to the next reorderer, if there is one. As an aside
- there's no such thing as multiple reorders on a package - I think this
would just be confusing and I can't see the use case.

Awesome, makes sense.  And yeah, totally agree on multiple reorders - should stop at the first that applies.

nerdvegas:
Closing, since we've talked about implementing this in terms of the new package orderer functionality. Superseded by the issue: https://github.com/nerdvegas/rez/issues/329

PR #270, opened 2015-11-17
Moved the calling of the action callbacks after the main execute.

cwmartin:
Working on some configuration managment that requires that action callbacks be run after the main resolve execution is run. Uncertain if any existing actions callbacks are dependant on the pre-existing call order.


nerdvegas:
Hey Christopher,

I'm a bit hesitent to make this change because it is hard to know how changing the order like this might affect current users. I think it would be good to add an additional 'post_actions_callback' instead - that way current functionality doesn't change, and users can choose whether they want the pre- or post- ordered callback.

Would this give you the functionality you need and if so do you mind making that change?

Thx
A

cwmartin:
Ya, this was this simplest way for my purposes, but I was debating adding a
second actions callback for post, or adding an option to toggle when the
callbacks were called - the default being before.

Both methods would serve my purpose. I'll implement a a seconds
post_actions_callback and send another pull request.

Cheers,
Chris.

cwmartin:
Hi Allan,

I've added the post_actions_callback as discussed.

Cheers,
Chris.

cmartinaf:
Hi Allan, 

Just curious if you've had a chance to look at the change I've made. 

Cheers,
Chris.

cwmartin:
Hi Allan,

Just curious if you've had a chance to look at the change I've made.

Cheers,
Chris.

nerdvegas:
Hey Chris,

I'll be addressing the PR backlog as soon as I can this new year, hoping to
start in the coming weeks. I'll start with the smallest and easiest PRs
first so yours will be on that list.

Thanks,
A


On Wed, Jan 6, 2016 at 8:12 PM, Christopher Martin <notifications@github.com

PR #272, opened 2015-11-28
adding extra args to the pip installer so we can add custom header pa…

larsbijl:
…ths to pip.

(needed for pandas as numpy is a rez dependency)


nerdvegas:
Thanks Lars, I'll merge this soon.

Out of interest, what pip args specifically did you need in this case? I'm working on some further pip integration and I wonder if we can provide a better solution.

larsbijl:
for pandas I need to add the following for numpy's headers to be picked up and build

`
PIPARGS --global-option=build_ext --global-option=-I$ENV{numpy_INCLUDE_DIR}
`

and numpy's variable it set as this in the `package.py`

```
    if building:
        env.numpy_INCLUDE_DIR = "{root}/python/core/include"
```

larsbijl:
also being to passing `-v`  is helpful for debugging.

PR #279, opened 2015-12-09
rex: add an expandvars utility function

elrond79:
adding an expandvars func as discussed in the mailing list.

it's available both in the global namespace, and on env (lemme know if you want only the env.expandvars)

Also, it takes an optional kwarg, which controls whether or not do a format as well (defaults to on).


elrond79:
no longer added to the env namespace (global func only)

PR #306, opened 2016-03-17
Windows Fixes

skral:
Fixes to get rez to work more smoothly on Windows. Some of these changes are based on changes from holofermes (thank you).
- Tilde and path separator on Windows
- Removed "foobar" from cmd prompt on Windows
- Updated version extraction in bind for CMake on Windows
- CMake doesn't like backward slashes in CMAKE_MODULE_PATH so we replace them to be safe
- Sensible defaults for difftool on Windows.
- Not passing -j to NMake on Windows
- Pass native slashes to Pip in CMake template. Otherwise Pip doesn't write the executable files on Windows.
- Default PATHEXT on Windows otherwise which('cmake') can fail if the environment was empty.

### DIFF
--- a/src/rez/utils/platform_.py
+++ b/src/rez/utils/platform_.py
@@ -481,6 +481,10 @@ class WindowsPlatform(Platform):
     def _terminal_emulator_command(self):
         return "CMD.exe /Q /K"
 
+    def _difftool(self):
+        from rez.util import which
+        return which("meld", "fc")
+
     def _physical_cores_from_wmic(self):
         # windows
         import subprocess
--- a/src/rezplugins/shell/cmd.py
+++ b/src/rezplugins/shell/cmd.py
@@ -93,7 +93,7 @@ class CMD(Shell):
     def _bind_interactive_rez(self):
         if config.set_prompt and self.settings.prompt:
             stored_prompt = os.getenv("REZ_STORED_PROMPT")
-            curr_prompt = stored_prompt or os.getenv("PROMPT", "foobar")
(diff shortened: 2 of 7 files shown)


src/rez/utils/platform_.py
```diff
@@ -481,6 +481,10 @@ def symlink(self, source, link_name):
     def _terminal_emulator_command(self):
         return "CMD.exe /Q /K"
 
+    def _difftool(self):
+        from rez.util import which
+        return which("meld", "fc")
```
mstreatfield: This will conflict with a change in pull request #204 - however your change is better.  Hopefully @nerdvegas can fix it up when he does the merge of the pull requests.

src/rezplugins/shell/cmd.py
```diff
@@ -93,7 +93,7 @@ def get_syspaths(cls):
     def _bind_interactive_rez(self):
         if config.set_prompt and self.settings.prompt:
             stored_prompt = os.getenv("REZ_STORED_PROMPT")
-            curr_prompt = stored_prompt or os.getenv("PROMPT", "foobar")
```
mstreatfield: Whoops, sorry, my mistake!

mstreatfield:
There is a change in pull request #204 which might be of use on Windows - in particular 9746cb5

nerdvegas:
Cheers for the head's up Mark,
A

skral:
Thank you guys for looking into it. Anything I should fix before you can merge the pull request? I will create some more PRs in the future, should I try to keep the list of changes smaller, making it easier to merge?

nerdvegas:
Note: Addressed by #319

Issue #309, opened 2016-04-06
UnixPlatform distribution inconsistent


bfloch:
The way that the linux distribution/os is determined in rez is currently
1. parse /etc/lsb-release 
2. try getting the output of the lsb_release program
3. use python's dist detection

The  3. is indeed unreliable and is about to become deprecated and removed in python 3.6
http://bugs.python.org/issue1322 (which we do not care about for the time being at python2.7)

However we have an case where Scientific Linux machines are evaluated either as
"Scientific" based on 2. if a machine has redhat-lsb-core installed, or "Scientific Linux" if not and it falls back to 3. (which ends up using /etc/redhat-release as far as I can tell). The difference lies in the implementation of lsb_release (a monster bash script) vs. platform.py from python.

Linux won't give us a satisfying result however we do care about consistency.
We can hack around it but ultimately I would love to see a solution that ends up in rez and at least gives us the expected results.

Following suggestions:
1. Make Platform_.os() a plugin so the maintainer is in control
2. Allow for os aliases within the rezconfig. Scientific would map to Scientific Linux as long as the versions match.
3. Allow to configure the order of fallbacks

Personally I'd prefer the 2nd solution as it would solve our issue. However it will not solve the issue for someone that gets debian reported from an ubuntu system.

Opinions?

nerdvegas:
Option 2 strikes me as the simplest solution. Option 1 feels like overkill,
and 3 may not fix all cases, but I think there's a good chance 2 will work
for everyone.

A

JeanChristopheMorinPerso:
I'll take a guess and consider this as solved since it's 6 years old and we now use the `distro` library which should be more reliable.

PR #310, opened 2016-04-13
Issue 309 osmap

bfloch:
So here is the suggested dictionary for issue #309 

The way it works is you add something like this to your rezconfig.py:

```
platform_map = {
  "os": {
        r"Scientific Linux-(.*)": r"Scientific-\1",
        r"Ubuntu-14.\d": r"Ubuntu-14",
        r"CentOS Linux-(\d+)\.(\d+)(\.(\d+))?": r"CentOS-\1.\2",
    },
}
```
The example would remap: 
* Scientific Linux-6.4 -> Scientific-6.4
* Ubuntu-14.2 -> Ubuntu-14
* CentOS Linux-7.2.1522 -> CentOS-7.2

This gives much more control over the os string and also can fix mismatches between platform.linux_distribution and lsb_release.

It is a pretty basic decorator.
I applied it to os and arch. I doubt there is a usecase for plaform_.name.

It works quite well.
Currently I see following optional improvements:
- The unit test works but is very hacky, do you know an easier way to mock the config?
- The dictionary has an undefined order. Probably should be a list of tuples in case someone goes crazy but I felt usually you can model everything with a single regexp.

Let me know what you think.

### DIFF
--- a/src/rez/config.py
+++ b/src/rez/config.py
@@ -216,6 +216,7 @@ config_schema = Schema({
     "plugin_path":                                  PathList,
     "bind_module_path":                             PathList,
     "implicit_packages":                            StrList,
+    "platform_map":                                 OptionalDict,
     "parent_variables":                             StrList,
     "resetting_variables":                          StrList,
     "release_hooks":                                StrList,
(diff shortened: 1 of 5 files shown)


nerdvegas:
Hi,

I've merged this PR but dropped the unit test - it was causing an error related to swapping out the config module. It's pretty minor functionality though, not sure the test is worth it. Seems to work fine.

bfloch:
Thanks.

PR #311, opened 2016-04-18
Fix a typo in the rez-pip CLI script.

cb109:
Otherwise calling ``pip_install_package()`` fails in line **31**.


nerdvegas:
Note: Addressed by #319

PR #312, opened 2016-05-10
Fixes Windows issue when running rez-env with quiet flag (-q)

skral:
Removed exit command in order to run "rez-env -q" otherwise rez doesn't stay in the environment but exits right away if run with quiet flag.
Updated context call to exit the shell it is run in. Bringing us straight back to where we left off.
Removed commented info calls.

Steps to reproduce on Windows:
works: rez-env cmake
fails: rez-env -q cmake


nerdvegas:
Note: Addressed by #319

Issue #313, opened 2016-05-12
Home directory doesn't expand on Windows with default config


assumptionsoup:
I'm using git-bash on windows with the master branch.  I suspect this is a problem with cmd.exe too.

```
>~$ rez-bind platform
rez: PackageRepositoryError: Lockfile directory C:\Users\jordan\~\packages does not exist - please create and try again
```

On my system, `~/packages` does exist.  What does not exist is `~/~/packages`.
I've tracked down the error to here https://github.com/nerdvegas/rez/blob/master/src/rez/utils/formatting.py#L460
The relevant part is:

```
        i = path.find(os.path.sep, 1)
        if i < 0:
            i = len(path)
        if i != 1:
            return path
```

The problem is that in git-bash, `os.path.sep` is `\\` (as with all nt systems), but the default config specifies `~/packages`.  Thus `~/packages` is not expanded.  In order for rez to work out of the box, I think this section of code should be changed to obey `os.path.sep` and `/`.  Or perhaps the default config should differ across platforms (such that nt systems default to `~\\packages`).  What are your thoughts?

As a side note, `os.path.expanduser('~/packages')` works correctly in this scenario.  It's just the custom expanduser implementation in rez that does not.

**Update:** `"${HOME}/packages"` works on windows and posix since it uses variable expansion instead of user expansion.  Using an explicit `$HOME` variable in the default rezconfig.py instead of `~` could solve this issue.

mvtk:
:+1:  for more sensible platform specific defaults. Though i am not sure if the implementation could maybe be done a little bit different that makes shure weak references are not destroyed but not drop using expanduser entirely on NT but use it selectively instead.

nerdvegas:
Does anyone see a problem with updating rezconfig.py to do like so:

local_packages_path = os.path.join("~", "packages")

? That should fix the issue I think.
A

JeanChristopheMorinPerso:
Seems to be fixed by https://github.com/AcademySoftwareFoundation/rez/pull/319

Issue #318, opened 2016-05-24
rez assumes an 8 core Windows machine is a 1 core


michaelmovies:
Addressed in PR #319  but adding it here to be tracked: current code cannot handle multi-processor, multi-core machines on Windows and falls back to an incorrect default.  Unlikely to be consequential, but does lead to some spamming of the terminal output.

Issue #317, opened 2016-05-24
filesystem.TempDirs leaving directories behind


michaelmovies:
Currently filesystem.TempDirs has an overloaded __del__ method that is failing on my Windows machine because the os module has already been torn down and deleted from sys.modules before the __del__ method is even called on the TempDirs instance.  This results in an obscure Warning message and the temp directory not being removed from disk.  Addressed in PR #319  but added here in case the backlog doesn't clear.

Issue #316, opened 2016-05-23
formatting.expanduser fails with mixed slash paths


michaelmovies:
Currently expanduser('~/path') and expanduser(r'~\path') will give different return values, which causes a failure in one of the default tests on Windows.  Addressed in #319 .

PR #319, opened 2016-05-24
Minimum Windows Fixes

michaelmovies:
Hi,

The master branch currently fails or warns on its own tests out of the box on Windows due to 4 issues.

1) bug in '~' expansion
2) missing diff tool default - this is addressed in another PR as well, but it's significant because it appears Rez is broken on first blush without it
3) bug in physical core determination on multi-CPU machines
4) risky use of del overload on TmpDirs results in noisy and obscure AttributeError warning, and also leaves the temp directories created on disk after shutdown

This pull request addresses all 4.

### DIFF
--- a/src/rez/backport/shutilwhich.py
+++ b/src/rez/backport/shutilwhich.py
@@ -36,7 +36,9 @@ def which(cmd, mode=os.F_OK | os.X_OK, env=None):
             path.insert(0, os.curdir)
 
         # PATHEXT is necessary to check on Windows.
-        pathext = env.get("PATHEXT", "").split(os.pathsep)
+        default_pathext = \
+            '.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC'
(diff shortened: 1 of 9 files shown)


michaelmovies:
Sorry for the duplicated pull request; meant to send it to you from a branch, not master and I'm more used to the gitlab interface.

michaelmovies:
Hi @nerdvegas, this PR now rolls up several vital Windows PRs that are currently open and which block anyone on Windows from using Rez.  It incorporates all of the changes in #306 #311 and #312 and all of those PR's can be closed when this one is fully merged.  It solves issues #316 #317 and #318 and may fix #313 as well.

nerdvegas:
Hey @michaelmovies, this is now merged. Please note that I changed expanduser behavior so that it *never* does the nix-like "~otheruser" type expansion, since that always may clash with a package weak reference. I didn't want to also assume that usr expansion would only ever be wanted in StrPathList type config settings. I doubt anyone is using other user type expansion so I think this is probably the best fix overall. Let me know if it works on W.

PR #321, opened 2016-06-08
Shotgun Toolkit - Rez App Launch

nvontagen:
Hey Allan,

Finally getting around to creating a PR with the Shotgun Toolkit hook that we're using to launch apps through toolkit in a Rez context. I've added a bit of explanation to the file, but the process does assume basic toolkit knowledge. 

We've done a bit of testing with it and seems solid, but it hasn't been used in production yet. 

Cheers!

### DIFF
--- a/shotgun_toolkit/rez_app_launch.py
+++ b/shotgun_toolkit/rez_app_launch.py
@@ -1,0 +1,154 @@
+# Copyright (c) 2013 Shotgun Software Inc.
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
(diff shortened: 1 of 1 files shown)


nerdvegas:
Hi Nick,

Thanks, merged. Note that I've moved the code to a src/support directory - this is for code provided for rez integration with other apps, but isn't part of the rez installation itself (I think that's the intention here yes?).

Note also that I've updated the copyright notice. So far as I can tell from guidelines I've read, contributions to OS projects become part of the same license, and the contributor is added to the copyright alongside the original author, so I've done that. Hit me up if there are any issues.

Thanks
A

PR #326, opened 2016-07-01
Pull in latest from nerdvegas/master ...

michaelmovies:
Incorporates the minimal_windows_fixes.


michaelmovies:
Woops, sorry. Still discovering the little differences between Gihub and Gitlab.

PR #327, opened 2016-07-01
The formatting.expanduser changes broke Windows again

michaelmovies:
Hey,

The formatting.expanduser changes broke Windows support again, and one of the tests in "rez selftest --config" fails.

Ultimately the reason was the removal of an "os.path.normpath" call on the path BEFORE we begin the ~ search, but I also noticed some "bugs" in your code (the main things being it not capturing all forms of whitespace, but also the inability to handle the mixed os.sep and os.altsep case, which is common as hell on Winblows). So in this PR I've switched to a single regex.sub call that explicitly performs the limited replacement you mention in your comment lines (retaining the oddity that both ";" and os.pathsep (':') are allowed on either side of the ~ on Linux).  Personally I find the regex a bit more formal and a little easier to read, with the added bonus of allowing for explicit representation for checking start and end of the string as well as the expanded whitespace definition and mixed sep handling.


nerdvegas:
Thanks Michael and apologies for the breakage, I can't test the side
effects on windows!
A

michaelmovies:
Might not be trivial to get working, considering the way the installer works, but perhaps [Appveyor](https://www.appveyor.com/)?  They've got a free tier for FOSS projects, and they'll run arbitrary jobs on most versions of Windows.

Issue #334, opened 2016-08-01
missing setup.py


larsbijl:
can't seem to find the `setup.py` which is required in the `install.py`

```
running in /python setup.py install
/software/rez/rez-install/CentOS/6/2.0.0/bin/python: can't open file 'setup.py': [Errno 2] No such file or directory
```

instinct-vfx:
setup.py was dropped in the "cleanup" commit ( ff70672c0faba3ee64cbeb6613a7f2c9f572b71c ). As it's in the commit message it seems to be on purpose. 
Re-adding the file works fine here to deploy and the installed version also seems to work for now.

larsbijl:
thanks that worked.
@nerdvegas what is supposed to replace the setup.py?

nerdvegas:
Whoops, that's a fail on my behalf. I haven't touched the installation in a
long time and I thought setup.py was a hangover from long ago.

I can't easily fix this from work. If one of you could submit a PR I can
merge it online and get it fixed ASAP. Otherwise I'll fix it tonight.

Thx
A

nerdvegas:
Scratch that. Just fixed it.

A

PR #335, opened 2016-08-01
Re-add setup.py

instinct-vfx:
As per the discussion on #334 here is a pull request for re-adding setup.py. Sorry for the different accounts.


instinct-vfx:
Fixed by Allan. So closing

PR #338, opened 2016-08-15
Added Travis CI config

skral:
In the wake of the open source initiative I think a Travis CI config might help.
Initial config which runs all the tests. Unfortunately one test fails right now. Didn't look into why.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: python
+
+python:
+  - "2.7"
+
+install:
+  # Adding tcsh
+  - "sudo apt-get install tcsh"
+  # Installing rez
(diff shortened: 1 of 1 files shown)


skral:
This is just working for Linux right now. We should add OSX as well. Does someone have experience with Travis CI and OS X?

nerdvegas:
FYI: Travis doesn't support python on osx (wth?): https://github.com/travis-ci/travis-ci/issues/2312

Travis seems somewhat limited atm.. you also can't indicate separate build statuses in the badge either, only passed/failed. I'l include it anyway but lets keep an eye out for something better.

PR #341, opened 2016-08-19
Restores correct "rez env" behavior on Windows, and allows curated $PATH

### LINKED ISSUE #336: rez-env does not apply the environment in 2.0.0 on windows
I deployed 2.0.0 from master in exactly the same way i did with the RC before. When using `rez-env package` it seems the shell is never really created/applied. (e.g. there are no REZ_\* variables and commands() has not been called). When using `rez-env package -- command` everything seems to work as expected.

yawpitch:
@nerdvegas 

This fixes #336 by reverting an earlier change made by **skrall** (which replaced /C with /K).  

Additionally it adds a configuration variable and logic to allow an organization that has a curated $PATH (ie they already know the exact list of directories and **do not** wish to gather them from the -- at that point incorrect -- Windows registry or __PATHS_ var) to specify that path list and have it respected.

This is primarily of use to those (like me) who are building a configuration management wrapper for Rez, and who -- for better or worse -- do not have the ability to reliably assume that the system registry $PATH set on each and every Windows workstation is (remotely) the same.

### DIFF
--- a/src/rez/bind/hello_world.py
+++ b/src/rez/bind/hello_world.py
@@ -16,6 +16,7 @@ import os.path
 
 def commands():
     env.PATH.append('{this.root}/bin')
+    env.OH_HAI_WORLD = "hello"
 
 
 def hello_world_source():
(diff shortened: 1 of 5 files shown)


michaelmovies:
Additionally this now fixes #342 as well ... sorry for lumping them all together, but our master branch is also what we're using in production.

michaelmovies:
This now also fixes #343 and #344 as well.

instinct-vfx:
👍

PR #349, opened 2016-09-27
rez-pip fine grained control.

smaragden:
## This pull request adds some fine grained control to rez-pip.

Added flags to rez-pip cli command.
 --variant
Example usage:

```
'rez-pip -r --variant platform,arch,os ...' would result in default behavior:
variants = [['platform-linux', 'arch-x86_64', 'os-CentOS-6.6', 'python-2.5']]

'rez-pip -r --variant platform ...' would result in:
variants = [['platform-linux', 'python-2.5']]

To get even more control we could do something like:
'rez-pip -r --variant platform,os-CentOS-6 ...' would result in:
variants = [['platform-linux', 'os-CentOS-6', 'python-2.5']]
```

---

Added config options:

```
release_python_packages_path = "~/.rez/packages/python" # default location for released python packages
pip_default_variant=["platform","arch","os"] # default variant requirements for rez-pip installed packages
```

Python-MAJOR.MINOR will always be added to the variant.
if release_python_packages_path is not set it will fall back on release_packages_path.

---

Added check to make sure release_python_packages_path is both existing and writable.

### DIFF
--- a/src/rez/cli/pip.py
+++ b/src/rez/cli/pip.py
@@ -58,8 +58,7 @@ def command(opts, parser, extra_arg_groups=None):
                 installed_variants.update(installed)
             if skipped is not None:
                 skipped_variants.update(skipped)
-        except Exception exc:
-            print exc, exc.__name__, type(exc).__name__
+        except:
             failed_variants.add(package)
     # print summary
     #
--- a/src/rez/cli/pip.py
+++ b/src/rez/cli/pip.py
@@ -27,7 +27,7 @@ def setup_parser(parser, completions=False):
         help="Specify the variant requirements for package "
         " default=platform,arch,os")
     parser.add_argument(
-        "PACKAGE",
+        "PACKAGES", nargs='+',
         help="package to install or archive/url to install from")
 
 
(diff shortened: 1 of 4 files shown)


smaragden:
I added --ignore-installed flag to pip command. I think this should be harmless. But it makes sure the package and dependencies gets installed even if the package is pip installed to the python installation that is used by rez-pip.

nerdvegas:
Thanks Fredrik, I will get to this in the coming weeks. My only comment
just now is that the config option 'release_python_packages_path' should
probably be 'release_pip_packages_path' as that's a little clearer -
otherwise there may be the perception that a standard rez package that
happens to be python based, might also be released to this location.

RE manual variant control, good to have. Eventually I wanted to potentially
detect from the package contents what the variant should be, but that is
error prone I think - eg if a .so/.a etc is found on Linux then variant
would include arch/os, and so on. Manual control is a good solution in the
interim.

A

mottosso:
This implements `--variant` (great Swedish minds think alike? :)) and solves the auto-detecting of variants from package contents (i.e. wheel metadata) as well.

- https://github.com/nerdvegas/rez/pull/204

bpabel:
Can we close this? The variant information can be gleaned from package metadata and the current implementation of rez-pip does that.

PR #353, opened 2016-10-13
release_vcs/git: make sure it always uses the repository at vcs_root

elrond79:
This commit ensures that the git commands for the release_vcs plugin always work on the intended git repository.

Unfortunately, I don't remember the specific details to recreate anymore, but at one point I ran into a situation where something had altered the cwd, and our git calls were actually operating on an incorrect repo...


nerdvegas:
Heya, you don't remember any more details? Cause the _cmd() call sets cwd to the package root, so I can't see how this could happen?

PR #357, opened 2016-10-19
add max_package_changelog_revisions config option

elrond79:
This adds an option to limit the maximum number of revisions printed to the package.py

It has a purpose similar to the max_package_changelog_chars, but keeps things tidier when you want to set very low limits (ie, you only want THE current commit)


nerdvegas:
I've merged this and fixed the conflict, not sure why GH hasn't picked it up as merged. Closing...

PR #362, opened 2016-11-12
Ability for a package to declare it shouldn't be cached

elrond79:
This allows a package to declare that it's results shouldn't be cached; this can be handy if you want to declare different values for a package depending on context; for instance, if you have a "gitkraken" package, and you want it's tools to be ["gitkraken"] on Linux, but ["GitKraken"] on OSX, you can do this pretty easily in a package.py by checking the current OS... but if it's memcached, then this result will be "baked" and returned to all machines using that memcache server, regardless of OS.

By allowing the gitkraken's package.py to set DO_NOT_CACHE = True, we can keep the proper branching behavior by OS.

### DIFF
--- a/src/rez/serialise.py
+++ b/src/rez/serialise.py
@@ -120,6 +120,9 @@ def _load_file(filepath, format_, update_data_callback):
 
     if update_data_callback:
         result = update_data_callback(format_, result)
+
+    if format_ == FileFormat.py and result.get("DO_NOT_CACHE", False):
+        return DoNotCache(result)
     return result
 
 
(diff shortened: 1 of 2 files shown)


nerdvegas:
Hey,

You can do this now with 'late binding' package attributes instead. Doing it this way would be better because you can still cache the package and get the intended results:

    @late()
    def tools():
        if system.platform == 'osx':
            return ['GitKraken']
        return ['gitkraken']

Do you still see a need for non-cached packages?
Thx, A

elrond79:
Hmm, good point. I think using late binding functions will probably mean we
don't need this... I'll convert our packages to use this when i get a
chance... In the meantime, I'll close this

PR #364, opened 2016-11-12
rex: don't let packages pollute namespace of future packages

elrond79:
The package executor currently has a bug where the global namespace is shared when multiple packages are executed when setting up a context.  This can create differing behavior depending on the order of execution of packages, which is clearly undesired.

This fix patches the behavior, so that the global namespace in which each package executes is "reset" after each package runs.

### DIFF
--- a/src/rez/rex.py
+++ b/src/rez/rex.py
@@ -1169,9 +1169,20 @@ class RexExecutor(object):
             code (str): Rex code to execute.
             filename (str): Filename to report if there are syntax errors.
         """
-        self.compile_code(code=code,
-                          filename=filename,
-                          exec_namespace=self.globals)
+        # we want to execute the code using self.globals - if for no other
+        # reason that self.formatter is pointing at self.globals, so if we
+        # passed in a copy, we would also need to make self.formatter "look" at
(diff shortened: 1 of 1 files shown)


src/rez/rex.py
```diff
@@ -1169,9 +1169,20 @@ def execute_code(self, code, filename=None):
...
+        # the same executor may be used to run multiple packages. Therefore,
+        # we save a copy of self.globals before execution, and restore it after
+        saved_globals = dict(self.globals)
+        try:
+            self.compile_code(code=code,
+                              filename=filename,
+                              exec_namespace=self.globals)
+        finally:
+            self.globals.clear()
+            self.globals.update(saved_globals)
```
nerdvegas: hey, why not just exec_namespace=self.globals.copy()?

nerdvegas: I'm gonna do this as a separate release anyway, since there's the outside chance that a studio is accidentally relying on this buggy behavior (and it'll be clearer to point that out in a separate release). Keen to get this done so I'll also do the self.globals.copy instead, happy to change back after if there's something I've missed.

elrond79:
Awesome, thanks!

PR #372, opened 2016-12-09
Taggable versions that sort BEFORE non-tagged versions

elrond79:
It's come up a number of times that we need to release a "special" version of some already-released version... however, we it to come before the already released version in priority / version sorting.  

To use a real-life example, suppose we initially have:

boost-1.55.0

...which is statically linked.  We now want to release a dynamically-linked version of that. But if we were to just create:

boost-1.55.0.dynamic

...then this will be detected by boost as a "higher" version, since in rez, "1.foo.anything" is always higher than "1.foo".

However, if we add a "__dynamic" tag (tags are indicated by preceding with a double underscore), then it will sort BEFORE boost-1.55.0.  So, to summarize, the sort order would be:

LOWEST version / priority:
boost-1.55.0__dynamic
boost-1.55.0
boost-1.55.0.dynamic
HIGEST version / priority

This behavior is configurable in the .rezconfig.  As a bonus, since it's actually implemented as a new type of VersionToken, .rezconfig now allows choosing a custom Token class for your version objects to use.

### DIFF
--- a/src/rez/config.py
+++ b/src/rez/config.py
@@ -316,6 +316,7 @@ config_schema = Schema({
     "env_var_separators":                           Dict,
     "variant_select_mode":                          VariantSelectMode_,
     "package_filter":                               OptionalDictOrDictList,
+    "version_token":                                Str,
     "new_session_popen_args":                       OptionalDict,
 
     # GUI settings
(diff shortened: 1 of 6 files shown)


bpabel:
There is a similar ticket to deal with these types of issues. https://github.com/nerdvegas/rez/issues/302

Though, in this case, these actually seem like two separate packages (static vs dynamic).

elrond79:
True, there's definitely some overlap.  This PR was created specifically to deal with a situation where a package had already been released, that we didn't want to tamper with, but we wanted another "similar" version that would sort before.

Making the two completely separate packages is certainly a possibility, but has it's own complications - for instance, you would need to deal with what would happen if both packages ended up in a resolve (would you have completely separate env vars? $BOOST_DYNAMIC_ROOT_DIR and $BOOST_STATIC_ROOT_DIR?)?

However, after thinking about it a bit, I think this may be a niche enough feature that it's not worth including in the main codebase, so I'm going to close this PR.  The code is still there in case anyone ever wants to use it, though...

bpabel:
I admit I haven't had to deal with many issues regarding choosing between static/dynamic libraries, but I think you might be able to use the [conflict operator](https://github.com/nerdvegas/rez/wiki/Basic-Concepts#the-conflict-operator) in your package dependency declarations to prevent both the static and dynamic versions from loading into the same environment.  I'm assuming you would never want the static and dynamic versions loaded at the same time.

PR #373, opened 2016-12-13
add ability to modify lists in .rezconfig, instead of overriding

elrond79:
Like the title says - similar to how dicts in ~/.rezconfig are merged, it's frequently handy to add items to lists instead of overwriting. This PR gives that ability (ie, you can prepend/append to the implicit_packages, etc).

### DIFF
--- a/src/rez/config.py
+++ b/src/rez/config.py
@@ -336,6 +336,50 @@ _plugin_config_dict = {
 
 
 # -----------------------------------------------------------------------------
+# ListModify Prepend / Append
+# -----------------------------------------------------------------------------
+
+class ListModify(object):
+    '''Allows modification of an existing list setting.
+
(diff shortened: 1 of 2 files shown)


nerdvegas:
Hey Paul,

I've discovered that this approach doesn't work for the case that I was most interested in - package-level config overrides. I'm working on something that achieves the same goal as here, but looks like this:

```
setting = ModifyList(append=["extra", "stuff"])
```

This format enables package config overrides like so:

```
with scope("config") as c:
    c.setting = ModifyList(append=["extra", "stuff"])
```

Just giving you a heads up, let me know if you have any concerns.

Thx
A

nerdvegas:
Closing, addressed by: https://github.com/nerdvegas/rez/releases/tag/2.18.0

PR #374, opened 2016-12-14
create a set of standard build env vars

elrond79:
Factors out code for setting build system environment variables, applies it to bez.


elrond79:
Awesome, thanks!

Issue #375, opened 2017-01-12
Using rez env interactively in windows


bpabel:
What is the preferred way of creating interactive rez environments?

I have several rez packages built and `rez env` is resolving them, but it doesn't seem to be applying them to the current shell (windows cmd).  

For example, I'm using `rez env` to build a context with the `requests` python library

```
rez env requests
```

Rez appears to find and resolve the package, but it's not updating the `PYTHONPATH` in my shell.  

If I run the following command, it will launch a child shell and the packages are correctly applied.  Is this how it is supposed to work -- that you have to launch a child process to apply the package environment?

```
rez env requests -c CMD
```

nerdvegas:
Hey,

rez-env requests is supposed to put you into an interactive subshell,
configured for that environment. When you're in the new shell, the prompt
is updated so that something like ']$' becomes '>]$' (note prepended
chevron).

There must be some problem in your scenario - some issue with the shell
startup. I'm not familiar with how this should work on Windows so I'm
hoping others will chime in.

Thx
A

instinct-vfx:
Can you share some more information like the code of the package.py file to investigate?

Cheers,
Thorsten

assumptionsoup:
I took another stab at getting rez to work on windows today.  I have this same issue with the hello_world example.  In windows cmd.exe and in git-bash `rez-env` does not put the environment in a sub-shell, and therefore it does not affect $PATH.

elrond79:
Out of curiosity, what happens if you do "rez-env PACKAGES -- cmd.exe"?

bpabel:
@elrond79 Yes, that will drop you into a shell, but that is different from the behavior in other shells.  In other shells, not specifying a command will give you an interactive shell.  The current behavior (without the PR I submitted) when using rez-env in a windows CMD shell is to spawn a shell with the correct rez enviroment, but then immediately exit.

instinct-vfx:
There are different PRs doing similar things or tackling it slightly different. We definitely need to clean up the windows side of rez and rework the cmd plugin a bit.

bpabel:
Fixed with #377

Issue #348, opened 2016-09-26
Error using rez on windows


bpabel:
I'm trying to build the hello_world example package provided with rez but I'm getting errors just doing the cmake compiler tests that I don't get when just running cmake by itself without rez.

When I run rez-build on the hello_world project, I get the following output:

NOTE: I get this same output even if the `CMakeLists.txt` file only contains the `cmake_minimum_required` command

```
Executing: c:\program files (x86)\cmake\bin\cmake.exe -d C:\Users\babel\workspace\rez\example_packages\hello_world -Wno-dev -DCMAKE_ECLIPSE_GENERATE_S
OURCE_PROJECT=TRUE -D_ECLIPSE_VERSION=4.3 --no-warn-unused-cli -DCMAKE_INSTALL_PREFIX=C:\Users\babel\packages\hello_world\1.0.0 -DCMAKE_MODULE_PATH=%C
MAKE_MODULE_PATH% -DCMAKE_BUILD_TYPE=Release -DREZ_BUILD_TYPE=local -DREZ_BUILD_INSTALL=0 -G NMake Makefiles
Not searching for unused variables given on the command line.
-- The C compiler identification is MSVC 19.0.24210.0
-- The CXX compiler identification is MSVC 19.0.24210.0
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin/cl.exe
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin/cl.exe -- broken
CMake Error at C:/Program Files (x86)/CMake/share/cmake-3.6/Modules/CMakeTestCCompiler.cmake:61 (message):
  The C compiler "C:/Program Files (x86)/Microsoft Visual Studio
  14.0/VC/bin/cl.exe" is not able to compile a simple test program.

  It fails with the following output:

   Change Dir: C:/Users/babel/workspace/rez/example_packages/hello_world/build/CMakeFiles/CMakeTmp



  Run Build Command:"nmake" "/NOLOGO" "cmTC_62b41\fast"

        "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\nmake.exe" -f
  CMakeFiles\cmTC_62b41.dir\build.make /nologo -L
  CMakeFiles\cmTC_62b41.dir\build

  Building C object CMakeFiles/cmTC_62b41.dir/testCCompiler.c.obj

[...]

PR #377, opened 2017-01-26
Better Windows Support

bpabel:
Several fixes and improvements

* Added a patched version of the `pbs` package to vendor libraries and updated the git vcs plugin and release hook so that `rez-release` will work on windows with git.
* Added pycharm project files to .gitignore
* Updated the windows cmd shell plugin to keep the shell open (via the `/K` flag instead of `/C` flag) when run in interactive mode.  Non-interactive mode still uses the `/C` flag so that the shell exits when the non-interactive command exits.
* Updated the way that `detached` is handled in windows shells to use `START` instead of `CMD /Q /K`.  The former will cause windows to open a new shell, which is what I believe is the purpose of `detached`.
* Updated the windows cmd shell plugin to correctly write comments and join multiple commands.

Primarily a fix for #375 but it may solve other tickets as well.


instinct-vfx:
How do these fixes relate to this PR? https://github.com/nerdvegas/rez/pull/341 There seems to be quite some overlap.

In regards to pbs and sh i opted to remove the dependency altogether as it was only used for the export method anyways. Could you share why you went for pbs instead?

bpabel:
> How do these fixes relate to this PR? #341 There seems to be quite some overlap.

I *think* the end result would be about the same, the main difference I can see is how we deal with interactive shells.  I omit adding the `exit` command, whereas #341 keeps the `exit` command for both interactive and non-interactive shells, but launches an additional subshell on line 160 in interactive mode, and that shell gets exited by the exit command rather than the interactive shell.  The results *should* be mostly the same, but with an additional shell being launched in interactive mode, if there are autorun scripts that run on shell launch, they're going to be run twice in the #341 implementation.

There is a comment about a rez code exec operation that passes an empty string command.  I'd have to check on that because I believe that would get treated as a interactive shell in my implementation, which may not be correct.

PR #341 also adds some features that my PR doesn't address

* Character escaping fixes
* Alias fixes.
* Defining default system paths in the config file.

And my PR addresses theses issues that aren't part of #341 

* Detached support for cmd shells.
* Fixes for cmd comments
* Fixed git support for rez-release

I think these changes could be merged with the exception of the interactive shell implementation, I'd have to do more research on that to see which way makes more sense.

> In regards to pbs and sh i opted to remove the dependency altogether as it was only used for the export method anyways. Could you share why you went for pbs instead?

I only chose `pbs` because `sh` isn't supported on windows.  I even had to patch `pbs` to get it to work with the way it was being used (`from pbs import git`).  I'm fine with dropping the dependency altogether.  I just needed the plugin to import to use `rez-release` in windows.

instinct-vfx:
Thanks a lot for the in-depth explanation! I simply removed the import, and the method. The only reason the method can not use the same implementation as the rest of the git commands in the module is, that it is marked to be a classmethod and hence can not use self. I am unsure if this is needed or if it maybe should live elsewhere, or construct it's own command helper.

assumptionsoup:
This fixes interactive shells for me on windows, but not #348  Also, I was a little surprised that running rez-env from a git-bash environment opened up a cmd.exe subshell.  I suppose this is the expected behavior?

bpabel:
Yeah, git bash is a different shell.  It looks like rez is currently hardcoded to only use the `cmd` shell plugin on windows - https://github.com/nerdvegas/rez/blob/master/src/rez/system.py#L78

You could add some logic there to figure out the correct shell is on windows, then add a plugin to work with the git-bash shell (might just be able to repurpose most of the unix bash shell plugin).

bpabel:
What is the status of this?  I pulled down the latest master, which looks like it has #341 merged (except looking at the #341 now, the changes look different than I remember).  Either way, it looks like rez is still not dropping into an interactive shell on windows.

instinct-vfx:
Interested and confused a bit too here. I would love to see the different overlapping PRs for windows fixed once and for all. There is still some other areas to work on, but fixing this issue along with fixing the git plugin ( #460 ) would be the minimal base to start deploying master here.

bpabel:
It looks like you already fixed the git issue, right?  Is #460 ready to be merged?  If you do that, I can pull it down and update this PR to be a cleaner update that just addresses the windows issue.

>  I would love to see the different overlapping PRs for windows fixed once and for all. 

I don't see any other open PR's dealing with Windows support.  The only other one I remember was #341, which has already been merged, but doesn't fix the windows interactive-shell issue.  If you know of other overlapping PR's, I can definitely take a look at them and merge those changes into this PR.

instinct-vfx:
Yes, #460 is ready to be merged. I made the changes according to the discussion with Alan in #459 

I don't have write access though, so i can not merge myself.

src/rez/utils/platform_.py
```diff
@@ -482,7 +482,7 @@ def symlink(self, source, link_name):
                 raise ctypes.WinError()
 
     def _terminal_emulator_command(self):
-        return "CMD.exe /Q /K"
+        return "START"
```
bpabel: "terminal emulator" is probably a bit misleading.  This command is only used when launching a rez environment using `detached=True`.  `CMD.exe` just spawns a subshell within the current shell.  `START` will launch a separate shell.

bpabel:
Includes the changes from #460 / #459 and removes `sh` and `pbs` dependencies.

nerdvegas:
Righto, I'm on this today.

PR #386, opened 2017-02-17
Fix: query the specified python when bind to.

gst:
Hi,

when rez-bind with --exe specified I found that rez actually copied the python files from my system python instead of the one specified.

This fix the problem.


gst:
Closing in favor of #400

PR #400, opened 2017-03-14
Enh/Fix: rez-bind python: do not copy system dirs to the packages fol…

### LINKED ISSUE #399: rez-bind python erroneously set some PYTHONPATH for python/lib and python/extra
Hi,

I discovered, already some days/week ago, that `rez-bind python` actually creates a package.py which contains instructions to add to PYTHONPATH the paths to 2 of the "main" libs directories of the python which was bound to. (it's the directories where the os and setuptools modules are available under the python which is bound to).
It (rez-bind) also has copied everything under these 2 paths into the package installation directory of that python.

We so end with  : 

`../packages/python/x.y.z/platform-linux/arch-x86_64/bin/python`  ->  link to the real python binary which was bound to.

But:

```
../packages/python/x.y.z/platform-linux/arch-x86_64/python/lib
../packages/python/x.y.z/platform-linux/arch-x86_64/python/extra
```

are now also in PYTHONPATH. 

BUT, now anything which use a python under the hood/internally will have to use that PYTHONPATH containing directories to the selected (by rez-env) python own system libs/modules.

We can see that effect like this : 

```
$ python -m site
sys.path = [
    '/home/gregory',
    '/home/gregory/rez/master-2.9.1/packages/python/2.6.6/platform-linux/arch-x86_64/os-CentOS-6.7/python/lib',
    '/home/gregory/rez/master-2.9.1/packages/python/2.6.6/platform-linux/arch-x86_64/os-CentOS-6.7/python/extra',
    '/usr/lib64/python26.zip',
    '/usr/lib64/python2.6',
    '/usr/lib64/python2.6/plat-linux2',
    '/usr/lib64/python2.6/lib-tk',
    '/usr/lib64/python2.6/lib-old',
    '/usr/lib64/python2.6/lib-dynload',
    '/home/gregory/.local/lib/python2.6/site-packages',
    '/usr/lib64/python2.6/site-packages',
    '/usr/lib64/python2.6/site-packages/PIL',
    '/usr/lib64/python2.6/site-packages/gst-0.10',
    '/usr/lib64/python2.6/site-packages/gtk-2.0',
    '/usr/lib64/python2.6/site-packages/webkit-1.0',
    '/usr/lib64/python2.6/site-packages/wx-2.8-gtk2-unicode',
    '/usr/lib/python2.6/site-packages',
]
USER_BASE: '/home/gregory/.local' (exists)
USER_SITE: '/home/gregory/.local/lib/python2.6/site-packages' (exists)
ENABLE_USER_SITE: True
```

As you can see, given they are in PYTHONPATH, they take precedence over my .local  (USER_SITE) directory.. which is a first not good.

But, if I would have rez-env with something using a python under hood, like nuke for instance, then Houston we have a (big) problem : 

because nuke uses its own python, not necessarily compatible with the one/thoses I've bound to.

That is :  nuke starts, it starts its own embedded python, which early import many important modules, like openssl, but then boum because the openssl for one python version isn't compatible with another. and given PYTHONPATH the python under nuke happens so to import the important modules from the python that was resolved in my rez-env call, not from his own python.

It seems to me you're trying, somehow, to do a _kind of_ virtualenv manually (but only a kind of) but the way it is actually triggers the problem(s) I mention / try to explain here. But PYTHONPATH is not done for that. well I mean putting a python system library/directory in PYTHONPATH is calling for problems.. as soon as you have to play with different python versions.

Actually, only doing the link to the python which is bound-to is enough to make it all available.

waiting answer / feedback or question ..

I can propose the PR (which is simply to not do the copy of these dirs and not add them to PYTHONPATH in the package.py which is generated)  if you agree with this ..

gst:
should fix #399

### DIFF
--- a/src/rez/bind/python.py
+++ b/src/rez/bind/python.py
@@ -43,24 +43,6 @@ def bind(path, version_range=None, opts=None, parser=None):
     check_version(version, version_range)
     log("binding python: %s" % exepath)
 
-    # find builtin modules
-    builtin_paths = {}
-    entries = [("lib", "os"),
-               ("extra", "setuptools")]
-
-    for dirname, module_name in entries:
-        success, out, err = run_python_command([
-            "import %s" % module_name,
-            "print %s.__file__" % module_name])
-
-        if success:
-            pypath = os.path.dirname(out)
-            if os.path.basename(pypath) == module_name:
-                pypath = os.path.dirname(pypath)
-
-            if pypath not in builtin_paths.values():
-                builtin_paths[dirname] = pypath
-
     # make the package
     #
 
@@ -69,21 +51,11 @@ def bind(path, version_range=None, opts=None, parser=None):
         link = os.path.join(binpath, "python")
         platform_.symlink(exepath, link)
 
-        if builtin_paths:
-            pypath = make_dirs(root, "python")
-            for dirname, srcpath in builtin_paths.iteritems():
-                destpath = os.path.join(pypath, dirname)
-                log("Copying builtins from %s to %s..." % (srcpath, destpath))
-                shutil.copytree(srcpath, destpath)
-
     with make_package("python", path, make_root=make_root) as pkg:
         pkg.version = version
         pkg.tools = ["python"]
         pkg.commands = commands
         pkg.variants = [system.variant]
-
-        if builtin_paths:
-            pkg.post_commands = post_commands
 
     return pkg.installed_variants
 


gst:
@nerdvegas 
won't you merge this one ?

I would even not wish to my worst enemy to encounter the difficulties I had to undertand the problematic situation I got as described in #399  .. ;)

nerdvegas:
Sorry I didn't reply to this @gst, I missed this comment somewhere along the way.

rez-bind'ed packages aren't really intended for production use, I hope to deprecate rez-bind at some point actually - the newer replacement will be "native" packages (with the addition of package attributes-as-functions it's much easier to write packages that act as interfaces to existing software installed on the system). Will take a look at this now.

PR #402, opened 2017-03-21
Add support for multiple paths for REZ_CONFIG_FILE

instinct-vfx:
Add support for multiple paths in REZ_CONFIG_FILE. Paths are simply appended. Basically the last path wins as far as my testing went. I hope this is the intended behaviour?

### DIFF
--- a/src/rez/config.py
+++ b/src/rez/config.py
@@ -541,7 +541,7 @@ class Config(object):
         filepaths.append(get_module_root_config())
         filepath = os.getenv("REZ_CONFIG_FILE")
         if filepath:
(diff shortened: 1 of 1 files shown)


instinct-vfx:
Updated with current master (which equals 2.14.1)

instinct-vfx:
Is there anything blocking this still open? Using this in production since quite a while and seems fine to me

nerdvegas:
This is actually getting released quite soon, start of next week. I
included another config-related PR into the next release and hit some
issues with that, which I was fixing this week (ability to modify list
settings rather than overwrite).

Cheers
A

PR #408, opened 2017-03-29
Enh: allow to give install and global option to underlying pip of rez…

gst:
This allow to pass global & install options of pip.

### DIFF
--- a/src/rez/cli/pip.py
+++ b/src/rez/cli/pip.py
@@ -23,6 +23,16 @@ def setup_parser(parser, completions=False):
         help="install as released package; if not set, package is installed "
         "locally only")
     parser.add_argument(
+        "--install-option", action='append',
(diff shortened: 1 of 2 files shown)


gst:
@nerdvegas  I wish this would work directly but strangely the argument parser seems to miss the argument I give, like with : 

`rez-pip --install PySide --install-option --qmake=/usr/bin/_qmake`

I get : 

rez pip: error: argument --install-option: expected one argument

seems related to LazyArgumentParser and 'co but I don't find what's the problem..

gst:
Hmm, actually I have started a quite better change/PR than this one:

I add `--pip-before-option` and `--pip-option`, which can be used to specify **any** pip argument(s), and not only the install-option and global-option ones..

will push it ASAP..

PR #424, opened 2017-05-16
Fixes #422 - Infinite solve if variants can not be solved

### LINKED ISSUE #422: Infinite solve if variants can not be solved
Tested with rez 2.13.0. Attached minimal setup with README.
This seems to be the minimal setup:

**Package 1 "python_provider":**

variants = [
  ['maya', '!houdini', '!nuke']
]

**Package 2 "pyqt":**

varriants = [
  ['houdini-15'],
  ['nuke-8'],
]

**Package 3 "maya":**

Regular.

Solve:
rez-env python_provider pyqt maya -vv

Give us a python_provider (maya, but not houdini and not nuke), maya and pyqt
(which does not have a variant that would not break the python_provider variant requirement).

Instead of failing the solver solves forever.

[rez_solver_bug.zip](https://github.com/nerdvegas/rez/files/994202/rez_solver_bug.zip)

bfloch:
This should fix #422.
I've added a regression test. Please take extra care to verify that it will not affect other cases.

### DIFF
--- a/src/rez/solver.py
+++ b/src/rez/solver.py
@@ -833,7 +833,7 @@ class _PackageVariantSlice(_Common):
             if result:
                 entry, next_entry = result
                 entries = self.entries[:i_entry] + [entry]
-                next_entries = [next_entry] + self.entries[i_entry:]
+                next_entries = [next_entry] + self.entries[i_entry + 1:]
             else:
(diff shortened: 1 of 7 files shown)


bfloch:
I just notice the packages for the test will break existing tests. I will move them to a different folder.

Issue #426, opened 2017-05-16
Request: Allow to define multiple local/release folders 


bfloch:
This includes the details for the request.
https://groups.google.com/forum/#!topic/rez-config/nazmdCCg1XM

In a nutshell, in the config:

- local_packages_path (Str) becomes local_packages_path**s** (StrList)
- release_packages_path (Str) becomes release_packages_path**s** (StrList) 
- The default selected path is set via packages_index (Int)

The benefit is that now different paths might act as local/release and will be correctly discovered.

To release/build to a different local/release folder you just need to override the packages_index in the package.py:

~~~python
config = {
    "packages_index":  1,
}
~~~

For the sake of clean implementation it does break config compatibility (to some little degree).
For example:
~~~python
local_packages_path = "~/packages"
release_packages_path = "~/.rez/packages/int"
~~~
Becomes:
~~~python
local_packages_paths = ["~/packages"]
release_packages_paths = ["~/.rez/packages/int"]
packages_index = 0
~~~

I will make a pull request for this feature soon.

bfloch:
Here an example of how this might be used. Given a rezconfig.py
~~~python
local_packages_paths = ["~/packages/inhouse", "~/packages/external",]
release_packages_paths = ["~/.rez/packages/inhouse", "~/.rez/packages/external",]
packages_index = 0
~~~

All inhouse developent packages do not need to do anything special.

In a package.py for a external package you might just:

~~~python
config = {
    "packages_index":  1,
}
~~~

Now rez-build / rez-release will release to /external for this particular package.
Make sure to include all the paths to the package search paths.

nerdvegas:
Hey Blazej,

A couple of things:

1. As mentioned earlier I really need to keep backwards compatibility. I
think the setting name should not change (ie, local_packages_path, not
local_packages_paths), and it should support both string (old format) and
list (new format). I also don't want to support both local_packages_path
and local_packages_paths, that's too confusing. Even though strictly
speaking your new names are more correct, backwards compatibility wins!

2. 'packages_index': If I understand correctly, this controls both the
local and release path that the package will use? If so, that's probably ok
for most cases, but I think there should also be the option to
rez-build/rez-release to different paths as well. The name is also not
quite descriptive enough - how about 'packages_path_index'?


So taking these issues into account, may I suggest something like this:

# in rezconfig
local_packages_path = {  # can also be a string (backwards compat)
  "internal": "~/packages/int",
  "external": "~/packages/ext"
}

[...]

PR #427, opened 2017-05-16
Implements: Allow to define multiple local/release folders

bfloch:
Based on the request at #426.

All tests pass.
For the sake of clean implementation it does break config compatibility (to some little degree).
For example:
~~~python
local_packages_path = "~/packages"
release_packages_path = "~/.rez/packages/int"
~~~
Becomes:
~~~python
local_packages_paths = ["~/packages"]
release_packages_paths = ["~/.rez/packages/int"]
packages_index = 0
~~~


bfloch:
This includes the other pull request. Sorry for that.

PR #428, opened 2017-05-16
Implements #426: Allow to define multiple local/release folders

bfloch:
Based on the request at #426.

All tests pass.
For the sake of clean implementation it does break config compatibility (to some little degree).
For example:
~~~python
local_packages_path = "~/packages"
release_packages_path = "~/.rez/packages/int"
~~~
Becomes:
~~~python
local_packages_paths = ["~/packages"]
release_packages_paths = ["~/.rez/packages/int"]
packages_index = 0
~~~

### DIFF
--- a/src/rez/config.py
+++ b/src/rez/config.py
@@ -244,8 +244,9 @@ config_schema = Schema({
     "implicit_styles":                              OptionalStrList,
     "alias_styles":                                 OptionalStrList,
     "memcached_uri":                                OptionalStrList,
-    "local_packages_path":                          Str,
-    "release_packages_path":                        Str,
+    "local_packages_paths":                         StrList,
+    "release_packages_paths":                       StrList,
+    "packages_index":                               Int,
     "dot_image_format":                             Str,
     "build_directory":                              Str,
     "documentation_url":                            Str,
@@ -475,11 +476,22 @@ class Config(object):
         return d
 
     @property
+    def local_packages_path(self):
+        """Returns the local_package_path based on the index"""
+        return self.local_packages_paths[self.packages_index]
+
+    @property
+    def release_packages_path(self):
+        """Returns the release_package_path based on the index"""
+        return self.release_packages_paths[self.packages_index]
+
+    @property
     def nonlocal_packages_path(self):
         """Returns package search paths with local path removed."""
         paths = self.packages_path[:]
-        if self.local_packages_path in paths:
-            paths.remove(self.local_packages_path)
+        for local_packages_path in self.local_packages_paths:
+            if local_packages_path in paths:
+                paths.remove(self.local_packages_path)
         return paths
 
     def get_completions(self, prefix):
--- a/src/rez/rezconfig.py
+++ b/src/rez/rezconfig.py
@@ -51,11 +51,17 @@ packages_path = [
 ]
 
 # The path that Rez will locally install packages to when rez-build is used
-local_packages_path = "~/packages"
+# This uses the package_index to identify which path to use.
+local_packages_paths = ["~/packages"]
 
 # The path that Rez will deploy packages to when rez-release is used. For
 # production use, you will probably want to change this to a site-wide location.
-release_packages_path = "~/.rez/packages/int"
+# This uses the package_index to identify which path to use.
+release_packages_paths = ["~/.rez/packages/int"]
+
+# Defines which local/release_packages_paths are enabled
+# Feel free to override this within a package to release to different locations.
+packages_index = 0
 
 # Where temporary files go. Defaults to appropriate path depending on your
 # system - for example, *nix distributions will probably set this to "/tmp". It
--- a/src/rez/packages_.py
+++ b/src/rez/packages_.py
@@ -92,9 +92,12 @@ class PackageBaseResourceWrapper(PackageRepositoryResourceWrapper):
     @cached_property
     def is_local(self):
         """Returns True if the package is in the local package repository"""
-        local_repo = package_repository_manager.get_repository(
-            self.config.local_packages_path)
-        return (self.resource._repository.uid == local_repo.uid)
+        for local_packages_path in self.config.local_packages_paths:
+            local_repo = package_repository_manager.get_repository(
+                local_packages_path)
+            if self.resource._repository.uid == local_repo.uid:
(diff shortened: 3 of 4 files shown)


nerdvegas:
Hey Thanks Blazej,

One thing though.. it breaks compatibility, as in local_packages_path etc
must be a list now? I can't introduce breaking changes though, I put a lot
of effort into ensuring that rez stays backwards compatible unless breaking
changes are absolutely necessary. Happy for the fix but it does have to be
backwards compatible.

Cheers
A

bfloch:
Changes based on the comments in https://github.com/nerdvegas/rez/issues/426#issuecomment-302254762
It acts like usual with old settings (strings).

I've added the labels as suggested.
I took the liberty to also add them to non local packages in case there is a package_paths_index. This helps navigation.

If there are no multiple projects then the output is as is.
Else it looks something like:
~~~
requested packages:
qtpy
realm
~platform==linux  (implicit)
~arch==x86_64     (implicit)
~os==CentOS-7.2   (implicit)

resolved packages:
arch-x86_64             ###################################          (internal)
logger-0.2.0            ###################################          (internal)
os-CentOS-7.2           ###################################          (internal)
platform-linux          ###################################          (internal)
qtpy-1.0.0b3.tbe1       ###################################          (local: thirdparty)
tool-0.26               ###################################          (local: internal)
rez-2.0.rc1.39          ###################################          (thirdparty)
config-0.1.0            ###################################          (internal)
~~~

nerdvegas:
Looks great Blazej, cheers

src/rez/config.py
```diff
@@ -505,6 +533,33 @@ def _get_plugin_completions(prefix_):
...
+        if isinstance(self._data[local_mode + '_packages_path'], basestring):
+            return None
+
+        if isinstance(self.packages_path_index, basestring):
+            return self.packages_path_index
+
+        try:
+            result = self.packages_path_index[local_mode]
+        except KeyError as e:
+            raise ConfigurationError('packages_path_index missing %s' % e.message)
```
bfloch: Raises in case non-string based path is detected.

bfloch:
I updated rez-bind so that a manual --target-packages-path can be given.

Btw. in any misconfiguration a ConfigurationError is raised.
For example:
~~~
rez-bind -t doesnotexist --no-deps rez
11:38:53 ERROR    ConfigurationError: Invalid target_packages_path "doesnotexist" for local
~~~

bfloch:
Hold your horses. I found a weird bug (maybe regression on the feature) where a local package is not recognized correctly. Give me a sec.

bfloch:
Ok this includes the bugfix as well as fixes the label to be more accurate even if no config override is given in the package (by comparing the repository paths).

The bug was introduced by the property that I created that overrides the config value. I needed to reference back the original data, else I am already using a single value.

Anyhow this works for me now.

bfloch:
I will do some more testing over the week.

bfloch:
Just a heads up. We have this in production since the original date.
If you ever plan the next minor or major version consider this.
It is backwards-compativle / non-destructive to existing package-definitions.

I would love to have someone external do some testing.

nerdvegas:
Hey Blazej, I think I'll be getting to this in the next week actually, so
keep an eye on the releases.

Cheers
A

src/rez/rezconfig.py
```diff
@@ -45,17 +45,39 @@
...
 ]
 
 # The path that Rez will locally install packages to when rez-build is used
-local_packages_path = "~/packages"
+# The target_packages_path controls which path to use.
+# Note: Can be a string for backwards compatibility.
+local_packages_path = {
+    "internal": "~/packages/int",
+    "external": "~/packages/ext",
+}
```
nerdvegas: Just a head's up - I can't accept this change, because it changes an existing default config value, which is a backwards compatibility issue. I will set these back to previous defaults.

bfloch:
Let me know if you need me to change anything.
I can also write documentation once we have it merged.

adriankrupa:
@nerdvegas @bfloch Hey, any news in that pr?

nerdvegas:
I'm getting to this soon, another project is taking up my time at the
moment but this is top of my list.

Thx
A

bfloch:
I hope that fixed the minor conflict.

mottosso:
Heads-up, this breaks backwards compatibility with the API. :(

- [Example](https://github.com/mottosso/rez-localz/blob/2ece69a354d751ea93581ccbf3b34506e4c40a47/python/localz/__main__.py#L171)

For the usecase of separating releases into `/int` and `/ext`, isn't this what the `package_preprocess_function` is for?

- [Wiki](https://github.com/nerdvegas/rez/wiki/Package-Definition-Guide#overriding-config-settings-in-preprocessing)
- [Example](https://github.com/mottosso/rez-for-projects/blob/37d86b31eb94d4529449efb5571cf9ac5f5ef96f/rezconfig.py#L51)

Makes sense for the `--no-local` argument however.

nerdvegas:
Hey @bfloch,

Apologies I never got back to this PR. I'm making a concerted effort recently to get on top of development, and I'll be revisiting this. I did run into some issues that stopped me from merging it, so in the coming weeks I will get a chance to go over the PR again and make some comments.

FYI, I'll be out of contact over the next 3-4 days as my family and I are moving house. If you can't contact me, that's why!

To expand on what @mottosso has said, yes you can achieve the same goal using a global preprocessor - it would simply read some custom package attribute, and alter release_packages_path appropriately. The same would work for local_packages_path also. However, what that would not do is make rez aware that there are multiple local package paths. So options like --no-local wouldn't respect that, and the output of rez-context would only show `(local)` for the one configured local_package_path.

bfloch:
@mottosso Yeah, internally we went with the config in the packages, but it killed the no-local option which caused some confusion, especially since the print-out becomes misleading.

@nerdvegas Yeah it's been a while. Looking forward to your comments. I'd gladly give it another shot.
I remember vaguely that there is more tools which would need to be aware of the paths, so I'll have to take a look at it myself.

src/rez/rezconfig.py
bfloch: It seems I never addressed this one. Must have slipped my sight.

PR #434, opened 2017-06-01
Ensure that "{install}" placeholder is properly expanded in package "build_commands".

KelSolaar:
References https://github.com/nerdvegas/rez/issues/431#issuecomment-305647516

This ensures that the `{install}` placeholder is formatted correctly instead of using `"install"` string in current implementation.

I have tested and it does fix the issue I had with `rez-build -i` and `rez-release`.

### DIFF
--- a/src/rezplugins/build_system/custom.py
+++ b/src/rezplugins/build_system/custom.py
@@ -112,7 +112,7 @@ class CustomBuildSystem(BuildSystem):
 
         def expand(txt):
             root = self.package.root
-            install_ = "install" if install else ''
+            install_ = install_path if install else ''
             return txt.format(root=root, install=install_).strip()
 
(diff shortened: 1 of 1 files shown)


nerdvegas:
Hi, I'm confused by this PR? {install} is _supposed_ to expand to the "install" string... if you want the actual installation path, that's what {root} is for.

{install} is handy for when your custom build command actually runs make or something:

`build_command = "make {install}"`

Here your Makefile would presumably set the installation path based on REZ_BUILD_INSTALL_PATH.

Thx
A

KelSolaar:
Hey Allan, I think it is me not understanding properly what `{install}` was meant to achieve (No Documentation 101 :) ). As a matter of fact I think I'm not even using that anymore. I will close the PR.

Issue #442, opened 2017-07-09
"rez-depends" and "private_build_requires ".


KelSolaar:
Hi,

I noticed that *rez-depends* does not list the `private_build_requires ` attribute:

``` python
private_build_requires = [
    'git',
    'python',
    'rez_api']
```

Would be something great to have as a command line argument.

Cheers,

Thomas

bfloch:
That was bugging us for a while to. I just implemented this in pull-request #444 
Enjoy.

bfloch:
@nerdvegas This can be closed now after the merge. Thanks!

nerdvegas:
I need to update this actually, for 2 reasons.

1. There's currently a bug, this doesn't work for released packages because `private_build_requires` is stripped at release time (easily fixed)
2. The `-p` arg needs to specify a particular package - it doesn't make sense to expose the private build reqs of all packages, as this does not represent any real env that ever gets resolved. Only the private build reqs of a package being built directly are ever included in a resolve.

nerdvegas:
https://github.com/nerdvegas/rez/releases/tag/2.25.0

PR #444, opened 2017-07-12
Implements #442 - Adds build_requires and private_build_requires to rez-depends

bfloch:
No changes to the existing interface. Tests pass.


KelSolaar:
Awesome!

PR #448, opened 2017-08-01
solver fix for variant splitting

elrond79:
A fix for the core solver that I'm surprised doesn't come up more often. Also added a test to illustrate the problem.


bfloch:
It does :) That's the same fix here: https://github.com/nerdvegas/rez/pull/425

I'm fine with either pull.

nerdvegas:
Thanks guys, good to know the resolve error is one and the same. I'll refer
to both PRs and take the tastiest of both :)

A

PR #454, opened 2017-08-31
Add standard_system_paths setting to set default PATH

skral:
It can be very helpful to configure the paths that should be part of your default PATH when rez runs. Especially under Windows where it will fall back to the registry which is horribly slow.

Hope someone else finds this useful. Let me know if there is anything we should improve before you deem it ready.


bpabel:
It doesn't look like there is a way to specify that the default environment should be initialized with an empty PATH, since the empty list is interpreted to mean "use the system default".

Perhaps the default for this option should be ``None``, which would be interpreted to mean use the system default.

PR #460, opened 2017-10-10
Remove sh dependency in git vcs plugin

### LINKED ISSUE #459: Remove sh dependency
As discussed on the mailing list already we want to remove the .sh dependency. It is not supported on Windows and just a fancy wrapper to replace subprocess and the likes essentially and it seems the only place it is still used is the git vcs plugin anyways.
And even in the git vcsplugin it is only used in a single classmethod.

instinct-vfx:
Fixes #459 

For now it is enough to just remove the import and drop the module from the vendor package

I did a grep and could not find any other uses of sh so it should be safe to drop.

Note: This leaves the export() classmethod defunct. It was before as it used retain_cwd which is not imported to begin with. Will be addressed separately.


bpabel:
The `sh` package is used in [/rez/src/rezplugins/release_hook/command.py](https://github.com/nerdvegas/rez/blob/f8c312e18b78d9f15585c9a1955e97e34d948878/src/rezplugins/release_hook/command.py#L12) as well.

instinct-vfx:
Good spot. Will take a look asap.

nerdvegas:
Addressed by https://github.com/nerdvegas/rez/pull/377

PR #462, opened 2017-10-19
Updated cmd.py subprocess.Popen calls to use "null" handler for stdin

eldritchsong:
There's an issue where subprocess raises "The handle is invalid" if any of the stdio handles are not real io handles (see https://bugs.python.org/issue3905). This becomes an issue for Maya on windows if Maya is launched via cmd because of how Maya attempts to redirect stdio. Making this change allows the Rez python API to be used in Maya launched via cmd.

### DIFF
--- a/src/rezplugins/shell/cmd.py
+++ b/src/rezplugins/shell/cmd.py
@@ -67,9 +67,13 @@ class CMD(Shell):
             cmd = ["REG", "QUERY", "HKLM\\SYSTEM\\CurrentControlSet\\Control\\Session Manager\\Environment", "/v", "PATH"]
             expected = "\r\nHKEY_LOCAL_MACHINE\\\\SYSTEM\\\\CurrentControlSet\\\\Control\\\\Session Manager\\\\Environment\r\n    PATH    REG_(EXPAND_)?SZ    (.*)\r\n\r\n"
 
-            p = subprocess.Popen(cmd, stdout=subprocess.PIPE,
-                                 stderr=subprocess.PIPE, shell=True)
-            out_, _ = p.communicate()
+            # pass null handle to stdin in order to fix Maya handle error on windows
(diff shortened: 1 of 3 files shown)


src/rezplugins/shell/cmd.py
```diff
@@ -67,9 +67,13 @@ def get_syspaths(cls):
             cmd = ["REG", "QUERY", "HKLM\\SYSTEM\\CurrentControlSet\\Control\\Session Manager\\Environment", "/v", "PATH"]
             expected = "\r\nHKEY_LOCAL_MACHINE\\\\SYSTEM\\\\CurrentControlSet\\\\Control\\\\Session Manager\\\\Environment\r\n    PATH    REG_(EXPAND_)?SZ    (.*)\r\n\r\n"
 
-            p = subprocess.Popen(cmd, stdout=subprocess.PIPE,
-                                 stderr=subprocess.PIPE, shell=True)
-            out_, _ = p.communicate()
+            # pass null handle to stdin in order to fix Maya handle error on windows
+            with open(os.devnull, 'w') as stdin_handle:
```
nerdvegas: The downside to this approach is that valid piped input (when not in maya) will also be lost.

I'm not on Windows, but can you confirm that this approach works? - 

```
if sys.stdin.fileno() not in (0, 1, 2):
    stdin = subprocess.PIPE
else:
    stdin = None
```

If this works, we should add it as a util func and use it in the other shells also, since it appears to be a general python bug: https://bugs.python.org/issue3905. We may have to add checks in future too (I'd not be surprised if this is fixed in py3), so a separate util func makes sense.

nerdvegas:
I have addressed this as per suggestion above, and as part of the merge of https://github.com/nerdvegas/rez/pull/377, so closing this for now.

PR #464, opened 2017-10-31
Add rez-bundle command.

Joshuabaker2:
This adds a `rez-bundle` command that resolves the environment for a given package, and then "bundles" that environment by copying all the given packages into a single directory. That directory can then be a new `REZ_PACKAGES_PATH`, allowing us to resolve that new directory with the standard `rez-env` mechanism.

### DIFF
--- a/bin/rez-bundle
+++ b/bin/rez-bundle
@@ -1,0 +1,3 @@
+#!/usr/bin/env python
+from rez.cli._main import run
+run("bundle")
--- a/src/rez/cli/_util.py
+++ b/src/rez/cli/_util.py
@@ -8,6 +8,7 @@ from rez.vendor.argparse import _SubParsersAction, ArgumentParser, SUPPRESS, \
 subcommands = [
     "bind",
     "build",
+    "bundle",
     "config",
     "context",
     "complete",
--- a/src/rez/cli/bundle.py
+++ b/src/rez/cli/bundle.py
@@ -1,0 +1,168 @@
+"""
+Bundle a package and all of it's dependencies into a single directory.
+"""
+
+
+def setup_parser(parser, completions=False):
+    parser.add_argument(
+        "--nl", "--no-local", dest="no_local", action="store_true",
+        help="don't load local packages")
+    parser.add_argument(
+        "--paths", type=str, default=None,
+        help="set package search path")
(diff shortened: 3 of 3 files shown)


nerdvegas:
Hey Josh,

Great addition and I'll definitely pull this soon. Just FYI though - this will only work with "filesystem"-based package repositories. Rez supports different package repo types, and a full rez-bundle solution should work with all of them. This tool will work for many current rez users, but not all.

The proper solution is a fair bit more work though, and what you have here will get 90% of users something useful right away. To give an idea of what we're aiming for in the end is:

    rez-env --output --context-type=embedded ...

This would create an rxt directory rather than a file, and would bundle packages into that directory. However it would work with any package repo type, and would only copy in the variants that the solve uses, rather than the entire package.

I'll add a warning to the tool before merging it in but I expect this will add value to a lot of users immediately.

zachlewis:
(just wanted to say, this is super useful for me. thanks!)

fnaum:
Hi @Joshuabaker2 , have you been using this rez-bundle? 
I'm curious about how this works with 3rd party packages that bake some path into the executables or config at build/release time. 
We do strip any RPATH from c++ packages, but there are some packages like Qt, Maya, that do bake full paths all over the place and then when you move that package somewhere el do not fully work .

Joshuabaker2:
Hi @fnaum. I have used it and it works well, but haven't tested it with an application that uses shared libraries that aren't defined in the rez file. Give it a spin and let us know :)

nerdvegas:
Relevant recent conversation:
https://groups.google.com/forum/#!topic/rez-config/QyJT6p3eXUA

I'm still keen to take Josh's work and use it as the basis for a formal
implementation (by that I mean one that works regardless of package
repository type). We need that at Method as we'd like to have standalone
service installations.

In terms of what to do about rpathed/non-relocatable packages, a new
"relocatable" package attribute will be added, they will simply be left out
of any bundling process (ie still refer to original package).

Cheers
A

fnaum:
Even when you define the PATH/LD_LIBRARY_PATH in the package there are some
tools that bake paths into the executables, or some config files that are
later used at runtime.
But I will give it a try since I'm quite curious

PR #467, opened 2017-12-18
Add rez info command

fnaum:
A simple new command to be able to get information of the current package without the need to be in a rez environment.  ( `rez env rez`)
For example, we will use that from Jenkins pipelines to get information of the current package to build the name of the artifacts.

usage:
```
>rez info name
usdBase

>rez info version
0.8.1

>rez info --separator  ", " requires
stdlib-4.4.6+<5, tinyxml-2.5.2+<2.6, cpp_btree-1.1, libcurl-7

>rez info variants
CentOS-6.2+<7 AL_boost-1.55.0
CentOS-6.2+<7 boost-1.55.0

>rez info --raw  variants
[['CentOS-6.2+<7', 'AL_boost-1.55.0'], ['CentOS-6.2+<7', 'boost-1.55.0']]

### CODE REVIEW

**Comment by fnaum** [comment]
I checked the Travis build error, it does not seem it is coming from the changes in this PR

**Comment by fnaum** [comment]
Someone asked me today what is the difference from this with `rez search`.

In order to use `rez search,` you need to know the **name** of the package. 
The scenario where we use this is on some Jenkins jobs that build muti rez package git repositories. When a commit happens in one of those repos, we only want to build the rez package that changed (and it's downstream dependencies). 
From the commit, we found the information of the paths that were changed and if it under a valid package (contains package.py|yaml + CmakeLists.txt) we build that package, but we need information from the current package being built i.e we need the name and version to create the artifact name or the dynamic name of the stages in the Jenkins scripted pipeline. 

The other thing is that is a bit faster as it just opens the file and gets the info it needs. It is basically a command to avoid `cat package* | grep | sed` lines. And of course, something that works across all platforms.

If you think there is no need for a new command, and there is a lot of overlap with `rez search`, I can think on extending the `rez search` capabilities to include the case where you want info from the current package without needing the name of the package.

**Comment by nerdvegas** [comment]
How is the 'current package' determined?

Thx
A

**Comment by nerdvegas** [comment]
I don't know what's going on with Travis, and it's annoying that the badge on my master now says 'failing' even though it's just the PR. I've been a bit disappointed with Travis tbh. I might turn off auto-building of PRs for the moment.

**Comment by KelSolaar** [comment]
Hey Allan,

You should be able to specify the branch for which the badge is shown as follows:

`https://travis-ci.org/nerdvegas/rez.svg?branch=master`: Rez

or using shields.io:

`https://img.shields.io/travis/nerdvegas/rez/master.svg?style=flat-square`: Rez

Cheers,

Thomas

**Comment by nerdvegas** [comment]
Yeah I already have that URL. I suspect it's matching to any `master`
branch, including PRs (and this one does happen to be from another `master`
- tsk tsk!)

A

ps - build fail is related to Travis updating their build image... I
suspect this is hitting the cmake-3 issue we currently have. I'm going to
look into this next.

**Comment by KelSolaar** [comment]
I have had quite many build fails in recent days, did not look into details beyond restarting the jobs but yeah they seem to have updated things.

**Comment by JeanChristopheMorinRodeoFX** [comment]
I'd be interested in something similar, though I would see it more like `rez-view . --format yaml` maybe? From Jenkins you can easily open a YAML file with [readYaml](https://jenkins.io/doc/pipeline/steps/pipeline-utility-steps/#readyaml-read-yaml-from-files-in-the-workspace-or-text) `def packageContent = readYaml file: <path to file.yaml>`. And I guess we could even add support for a JSON format in case someone would not be able to easily parse YAML from a CI infra.

I personally feel `rez-info` would cause confusion with `rez-view`. But I might be wrong.

What do you guys think?

**Comment by nerdvegas** [comment]
I agree that updating rez-view is probably the best approach here. I would:

* add options to print a specific field;
* add option to view the "current" package (meaning: The unbuilt package in
cwd with package.py present).

This has been raised before, specifically wrt the need to get access to a
package's attributes from a build script during build. Makes sense.

A

**Comment by fnaum** [comment]
I noticed that rez-view already has a -c (--current) flag but is used for the current context.
Also,  you will need the ability to extract just one field (raw and formatted), you can copy the code from this PR.

Our use cases are all in the test, so as long as those test pass I do not mind consolidating it into rez-view

**Comment by JeanChristopheMorinRodeoFX** [comment]
Here is the new PR: https://github.com/nerdvegas/rez/pull/539/files, 2 weeks later that what I initially said. Anyway, the important thing is it's there.

@fnaum You are welcome to test my branch. Let me know if there is a difference with what you had (in theory it's good, but if you could confirm to me, that would be great).

**Comment by bfloch** [comment]
Great work!

Can we add indices to the formatted variants dumps?
I wrote something similar with indices so that I can lookup what to feed `rez-build --variants XXX` with.
I guess at this point we can also consider hashed variants for build packages, which would help https://github.com/nerdvegas/rez/issues/754#issuecomment-538110357

Maybe like:
```
>rez info variants
0: CentOS-6.2+<7 AL_boost-1.55.0 [HASHIFAVAILABLE]
1: CentOS-6.2+<7 boost-1.55.0
```

PR #471, opened 2018-01-17
added warm option to rez-memcache

### LINKED ISSUE #275: add --warm option to rez-memcache
I sometimes find it useful to flush the memcached instance, and warm it afterwards so that all package.py/yaml files are cached, otherwise resolve times take a big hit for a while since loading many small files over NFS is often slow.

You can do the warm like so currently:

```
rez-search -t package -f '{q} {authors}'
```

But this would be more intuitive:

```
rez-memcache --warm
```

adriankrupa:
resolve #275 
caching only non-local packages


nerdvegas:
I am merging this, but please refrain from master-based PRs, thx.

PR #473, opened 2018-01-18
allow to user input and packages in rez-env at the same time

### LINKED ISSUE #294: rez-env doesn't support input context and packages at the same time
rez env: error: Cannot use --input and provide PKG(s) at the same time

adriankrupa:
Using context patching system
Resolves #294

### DIFF
--- a/src/rez/cli/env.py
+++ b/src/rez/cli/env.py
@@ -156,12 +156,14 @@ def command(opts, parser, extra_arg_groups=None):
         pkg_paths = [os.path.expanduser(x) for x in pkg_paths if x]
 
     if opts.input:
-        if opts.PKG:
-            parser.error("Cannot use --input and provide PKG(s) at the same time")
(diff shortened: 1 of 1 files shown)


nerdvegas:
Please note: #294 - I'm going to merge this, but with some modifications.

I don't think it should be possible to simply do the following:

```
]$ rez-env -i foo.rxt pkg1 pkg2
```

IMO it is not clear enough that an existing env (foo.rxt) is being patched here, because the `-p` arg is not explicitly specified. However, I do think that this should work:

```
]$ rez-env -i foo.rxt -p pkg1 pkg2
```

IMO this is a natural extension to current behavior, the only difference being that the 'current' context is the imported one, rather than the current rez-env'd shell.

Also, `--strict` shouldn't be implied - like a standard patch request, it is strict if the user has specified with `--strict`.

PR #490, opened 2018-02-27
Fix alias command in Windows when PATH is modified

### LINKED ISSUE #343: env.PATH manipulations break alias command on Windows
On Windows if a package.py (or **any** previously loaded package.py) does an _env.PATH.prepend_ or _env.PATH.append_ **BEFORE** a call to the _alias_ function, the _alias_ function will fail with the following error printed to the launching terminal:

> 'doskey' is not recognized as an internal or external command

This is because by default (ie without setting rez to pass through the PATH var from the parent), the PATH variable is being cleared at the moment of the first call to _env.PATH.prepend_ / _env.PATH.append_, which then makes it impossible for the shell that's running the resolved context file to find the _doskey.exe_ command. The resolved_context.py file then runs _executor.append_system_paths()_ which by default re-establishes the basic part of the PATH required by Windows to once again find the _doskey.exe_ file, but at that point it's already too late for the _alias_ to have worked.

End result: the alias fails to be created with an unexpected error message, but the executable the error message describes is once more available to the user by the time the shell has resolved.  This was not easy to track down.

This all goes back to what can only be described as a **"**Feature**"** of Windows; CMD.exe as a shell has built-in commands (set, call, etc), and doskey is in theory one of those commands, but unlike most of the others it's a distinct .exe, so you can get yourself into a situation where a shell cannot find one of the built-in commands that is a from-the-manufacturer part of that shell.  Go figure.

Working on a potential fix where we lazy-load the actual path to the doskey executable and use that instead.

### LINKED ISSUE #389: "alias" unexpected behaviour.
Hi,

I was under the *naive* assumption that the `alias` definition would allow me to call the application directly on `rez-env` invokation:

```shell
>>>>>>>> C:\Users\thomas>rez-env nuke-9.0v8 -- nuke
'nuke' is not recognized as an internal or external command,
```

if I launch a shell and then invoke `nuke` it is fine though:

```shell
>>>>>>> C:\Users\thomas>rez-env nuke-9.0v8 -- cmd
Microsoft Windows [Version 10.0.14393]
(c) 2016 Microsoft Corporation. All rights reserved.

>>>>>>>> C:\Users\thomas>nuke
Nuke 9.0v8, 64 bit, built Oct 26 2015.
Copyright (c) 2015 The Foundry Visionmongers Ltd.  All Rights Reserved.
Disk cache C:/Users/thomas/AppData/Local/Temp/nuke\ViewerCache/??: 10065MB (99% of 10240MB) used in 1042 files.
```

This is my `package.py` file by the way:

```python
# -*- coding: utf-8 -*-

name = 'nuke'
version = '9.0v8'
authors = ['The Foundry Visionmongers Ltd.']
description = 'Powerful compositing, editorial and review.'
tools = ['nuke', 'nukex']  # Testing stuff here.
uuid = 'third-party.the-foundry-visionmongers.nuke'


def commands():
    import os
    import platform

    application_name = '{0} {1}-{2}'.format(
        'The Foundry Visionmongers Ltd.', this.name.title(), this.version)
    error_message = '"{0}" is not available on your platform!'.format(
        application_name)

    if platform.system() == 'Darwin':
        raise SystemError(error_message)
    elif platform.system() in ('Windows', 'Microsoft'):
        root_directory = 'C:\\Program Files\\{0}{1}'.format(
        this.name.title(), this.version)
    else:
        raise SystemError(error_message)

    assert os.path.exists(root_directory), (
        '"{0}" is not installed at a standard location on your system!'.format(
        application_name))

    # Setting aliases after *PATH* manipulation will break the aliases command:
    # https://github.com/nerdvegas/rez/issues/343
    alias('nuke', "Nuke9.0")
    alias('nukex', "Nuke9.0 --nukex")

    env.PATH.append(root_directory)
```

Cheers,

Thomas

skral:
Fixes #389, fixes #343

Includes an easy test to see if the `alias` command works after modifying the `PATH` on Windows. This is not meant to be a fix for how to deal with `PATH` and modifications to it. It's just about the `alias` command and not getting anymore errors in Windows when using it.

### DIFF
--- a/src/rez/tests/test_shells.py
+++ b/src/rez/tests/test_shells.py
@@ -273,6 +273,40 @@ class TestShells(TestBase, TempdirMixin):
 
         _execute_code(_rex_appending, expected_output)
 
+    @shell_dependent()
+    def test_rex_code_alias(self):
+        """Ensure PATH changes do not influence the alias command.
+
+        This is important for Windows because the doskey.exe might not be on
+        the PATH anymore at the time it's executed. That's why we use the full
+        path to doskey.exe and figure out the path to it before we modify the
+        PATH.
+
+        """
+        def _execute_code(func):
+            loc = inspect.getsourcelines(func)[0][1:]
+            code = textwrap.dedent('\n'.join(loc))
+            r = self._create_context([])
+            p = r.execute_rex_code(code, stdout=subprocess.PIPE)
+
+            out, _ = p.communicate()
+            self.assertEqual(p.returncode, 0)
+
+        def _alias_after_path_manipulation():
+            # Appending something to the PATH and creating an alias afterwards
+            # did fail before we implemented a doskey specific fix.
+            env.PATH.append("hey")
+            alias('alias_test', '"echo test_echo"')
+
+            # We can not run the command from a batch file because the Windows
+            # doskey doesn't support it. From the docs:
+            # "You cannot run a doskey macro from a batch program."
+            # command('alias_test')
+
+        # We don't expect any output, the shell should just return with exit
+        # code 0.
+        _execute_code(_alias_after_path_manipulation)
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/src/rez/tests/test_shells.py
+++ b/src/rez/tests/test_shells.py
@@ -278,9 +278,9 @@ class TestShells(TestBase, TempdirMixin):
         """Ensure PATH changes do not influence the alias command.
 
         This is important for Windows because the doskey.exe might not be on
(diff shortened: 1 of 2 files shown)


flavienliger:
I don't have the alias error (doskey) anymore thanks of you.

KelSolaar:
Excellent @skral, thank you!

skral:
Perfect. Thanks for confirming. Glad it's working for you too.

asztalosdani:
It would be great, to see this merged.

Issue #469, opened 2018-01-12
working 'getting started' guide for windows


mrpurest:
My first step when assessing rez for use in our (mostly windows) studio was to go through the getting started guide.

However the steps in the guide do not currently work on windows due to issues already reported.

I understand that there are workarounds, but I don't understand what they are at this stage - I think its important that this is addressed in the documentation so that new users can step through the getting started guide without getting stuck.

nerdvegas:
Could you summarize what those issues are? I don't develop on windows and
have a lot of other issues to track so it isn't easy to be on top of this.
I released a new rez version a month or so back which merges several
Windows-related PRs, and so I also don't know to what extent those do or do
not address the problems you're hitting.

Please also include the rez version you're using in case issues have been
fixed since (though I assume you're working from master or latest?).

Thanks,
A

mrpurest:
I guess this is a duplicate of  #431  in a way.

The rez-build does not work because of cmake issues.   ( see #348 )

The way to get around this seems to be to use the rezbuild.py file rather than CMakeLists.txt
The example hello_world does have a file rezbuild.py.example and this does build without error, however it looks like it has not been tested on windows since you cannot run the 'hello' script with the package afterwards.

We are experimenting at the moment, and the system looks very promising, but we will have to make decisions as to how to use it ourselves before even completing the 'getting started'.

Should we make a 'hello_world_windows' in the example_packages directory?

R

nerdvegas:
Hello,

Please include the rez version you are testing with.

See: https://github.com/nerdvegas/rez/releases/tag/2.17.1

Does this version fix the issue?

Thx
A

ISkipy:
Hello,

Here is the issues that we got:
![capture](https://user-images.githubusercontent.com/35566104/35099215-05222740-fc4f-11e7-99ef-48de4a7bc88e.PNG)

and:
![capture2](https://user-images.githubusercontent.com/35566104/35099222-0dc94c2a-fc4f-11e7-9d8d-79f87fc63d37.PNG)


I understand that as we got an error on the last one we are not able to run this:
![capture3](https://user-images.githubusercontent.com/35566104/35099303-6639f936-fc4f-11e7-979c-b0f828cfc025.PNG)

instinct-vfx:
Disclaimer: I have not looked specifically into this case, so take with a pinch of salt.

[...]

PR #491, opened 2018-02-28
Added build system for python setup.py files.

bpabel:
Initial step towards python setup.py build support as discussed in #469 

There would still be some redundant information between the `package.py` and `setup.py`, but this removes the need for a wrapper `bez` or `cmake` just to call `setup.py` and copy the install files.

Also, the install locations for the generated `lib` and `scripts` files are hardcoded to `/python` and `/bin`.  Not sure if this is something that should be configurable (possibly in `package.py`?),  because the `package.py` will need to add those paths to the `PYTHONPATH` and `PATH` in the `commands()` section

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## [2.31.4](https://github.com/nerdvegas/rez/tree/2.31.4) (2019-06-22)
+[Full Changelog](https://github.com/nerdvegas/rez/compare/2.31.3...2.31.4)
+
+**Merged pull requests:**
+
+- Expose Python standard module __file__ and __name__ to rezconfig [\#636](https://github.com/nerdvegas/rez/pull/636) ([mottosso](https://github.com/mottosso))
+
 ## [2.31.3](https://github.com/nerdvegas/rez/tree/2.31.3) (2019-06-22)
 [Full Changelog](https://github.com/nerdvegas/rez/compare/2.31.2...2.31.3)
 
--- a/src/rez/utils/_version.py
+++ b/src/rez/utils/_version.py
@@ -1,7 +1,7 @@
 
 
 # Update this value to version up Rez. Do not place anything else in this file.
-_rez_version = "2.31.3"
+_rez_version = "2.31.4"
 
 try:
     from rez.vendor.version.version import Version


asztalosdani:
Hi,

what's up with this? I would like to use it. ;)

bpabel:
I was hoping to get some feedback on how people would like to see this implemented.

It's probably possible to eliminate the need for the `package.py` file entirely and pull everything from the `setup.py` script and the built distributions, and possibly from build arguments passed to the plugin, for things like overrides to the package name, or which variants to build, which python to use, etc.  Or possibly just include a cmd to generate package files from the setup.py and use that (since they need to be installed with the package anyway).

I know several people already have their own implementaions of `rez-pip`, which this may help address.

One of the things I currently find a bit awkward is contributing to python projects and working with people that don't use `rez`, because rez requires me to have a `package.py` file in my project directory, which I *do* want under version control for my own purposes, but I don't want to share with other people, because they don't need it.

instinct-vfx:
Sorry for the late reply. Things are a bit busy here atm. This is interesting but i need to think some more about it. The way we currently tackle this is by having a custom rez-pip as you mentioned. The main difference for our own version is windows support and some conventions we want to have for python packages. For python projects essentially we have a completely standard python project with setup.py et.al with no rez stuff whatsoever. This is built and uploaded to our internal devpi server. This is in turn used to install the package using our rez-pip installation. As we need the devpi wheels anyways it just felt natural.

So in our setup rez is an afterthought and package.py files for python packages are created fully automatically.

bpabel:
@instinct-vfx That's an interesting workflow.  So you're not using `rez-build` or `rez-release` at all in your development workflow?  How do test packages or work with them locally prior to releasing them?

instinct-vfx:
This is only true for python projects. And they are being developed and tested like most python packages i guess. With support for setup.py building that may change, but i still would not want to maintain a custom package.py as i do not really see any advantage but a lot of potential errors. It would be great though to be able to do a locally installed rez-build from setup.py.

JeanChristopheMorinRodeoFX:
On our side, our implementation of "rez-pip" (in quotes, you will see why) is completely bypassing rez logic. In our package.py, we added support for a `python_requires` attribute that we can simply put real pip packages in. During a resolve, rez passes the packages to the pip command under the hood and installs the pip packages locally on the user machine, in something really similar to a virtualenv (but its not a virtualenv). That local path is added to PYTHONPATH on the fly.

All of this to say that we went away of `rez-pip`. `python_requires` works exactly the same as a normal pip command. We can even pass pip packages to the `rez-env` command like this: `rez-env maya -P Flask` where `-P` is for the pip packages.

The output would look like:
```
(bash)[jcmorin@dev025l~jcmorin]$ rez-env maya -P Flask
09:36:15 INFO     Already installed and not too old. Reusing env. '/softwareLocal/rez_python_envs/shared/2.7/0xd77f4ebcc0e6736d'
09:36:15 INFO     Successfully setup or updated python env with following python distributions: ['Flask']

resolved by jcmorin@dev025l.rodeofx.com, on Thu Jul 05 09:36:15 2018, using Rez v2.14.1

requested packages:
maya                  
~platform==linux      (implicit)
~arch==x86_64         (implicit)
~os==CentOS-7.4.1708  (implicit)

resolved packages:
maya-2018.3                /rdo/software/rez/packages/maya/2018.3/platform-linux   
platform-linux             /rdo/software/rez/packages/platform/linux               
python-2.7.5               /rdo/software/rez/packages/python/2.7.5/platform-linux  
rdo_license_servers-0.0.3  /rdo/software/rez/packages/rdo_license_servers/0.0.3    

python distributions:
Package      Version
------------ -------
click        6.7    
Flask        1.0.2  
itsdangerous 0.24   
Jinja2       2.10   
MarkupSafe   1.0    
pip          9.0.3  
setuptools   39.2.0 
Werkzeug     0.14.1 
wheel        0.31.1 
```

That means a lot. When we rez-build a rez package that is also a pip package at the same time, we simply have a package.py in our repo and a setup.py. The setup.py imports the package.py to get its information. So no data duplication and the developer still have the option to ship different files with the pip package. Once built, we push a wheel into our internal devpi server.

package.py
```
name = 'package_a'

version = '1.0.0'

authors = ['RodeoFX']

description = 'Rez package'

private_build_requires = ['some_package_utils']  # some_package_utils is used for demonstration purposes. On our side, the packge is named "rdo_package_utils"

python_requires = ['Flask']

uuid = '3ba595a2-d69c-4491-9279-9053d1de0a2f'

# The process of building the pip wheel and uploading it is done inside the build.py by using some code
# that comes from the package "some_package_utils"
build_command = 'python {root}/build.py {install}'

def commands():
    import os
    env.PYTHONPATH.append(os.path.join('{root}', 'python'))
```

setup.py
```
import setuptools

import package

if __name__ == '__main__':
    setuptools.setup(
        name=package.name,
        version=package.version,
        description=package.description,
        install_requires=package.python_requires,
        author='RodeoFX',
        author_email='dev@rodeofx.com',
        url='https://bitbucket.org/<team>/{0}'.format(package.name),
        packages=setuptools.find_packages('python'),
        package_dir={'': 'python'}
    )
```

Having such a setup with pip inside rez allows us to be able to develop a non-rez repository and contribute to it without adding rez logic in it since we can later on use `python_requires` to get our custom pip package that we contributed to.

I hope what I said is clear and might help to spec out this PR. And if you guys would like to see our work for pip support in Rez, let me know.

src/rezplugins/build_system/python.py
```diff
@@ -0,0 +1,325 @@
...
+    def name(cls):
+        return "python"
+
+    @classmethod
+    def is_valid_root(cls, path):
+        return (os.path.isfile(os.path.join(path, "setup.py"))
+                and not os.path.isfile(os.path.join(path, "rezbuild.py")))
+
+    def __init__(self, working_dir, opts=None, package=None, write_build_scripts=False,
+                 verbose=False, build_args=[], child_build_args=[]):
```
ylathouris: Is the use of mutable defaults here intentional? I'm referring to `build_args=[]` and `child_build_args=[]`. From my understanding, this is an anti-pattern that can lead to hard-to-find-bugs:

- https://docs.python-guide.org/writing/gotchas/
- https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

Could we achieve the same thing using something like:

```python
 def __init__(self, working_dir, opts=None, package=None, write_build_scripts=False,
                 verbose=False, build_args=None, child_build_args=None):
        super(PythonBuildSystem, self).__init__(
            working_dir,
            opts=opts,
            package=package,
            write_build_scripts=write_build_scripts,
            verbose=verbose,
            build_args=build_args or [],
            child_build_args=child_build_args or [])
```

src/rezplugins/build_system/python.py
```diff
@@ -0,0 +1,325 @@
...
+
+        def _pip_build():
+            #TODO: pip 10 will install wheel builds by default, making this intermediate wheel build step unnecessary
+            # Generate wheel
+            print('Building wheel...')
+            pip = 'pip'
+            # These args speed up pip builds
+            pip_args = ['--disable-pip-version-check', '--no-deps', '--no-index']
+            cmds = [pip, 'wheel', '.', '-w', dist_root] + pip_args
+            # subprocess.call(cmds, cwd=source_path)
```
ylathouris: Is this for reference or is it just what was used before `_run_context_shell(cmds, cwd=source_path)`?

ylathouris:
We are also very interested in this feature! From our point of view, we are creating python packages that live outside the rez ecosystem but would like to share them with people who are working with rez. If we could get away with NOT maintaining a package.py we would. 

We generally run unit tests with `tox` which allows us to define our pip-style dependencies. For functional/E2E tests we use `Docker` and `pip` for setting up a suitable environment.

src/rezplugins/build_system/python.py
bpabel: It's an undeleted reference comment.  This code was originally a utility library for the bez build system using rezbuild.py.  I basically broke apart that utility script and inserted it into a rez build plugin skeleton, making some changes to get it to work correctly.  This is one of the parts that need to be changed to work with the build plugin and the original code from the utility script just got left in by accident.

bpabel: Hopefully I'll get some time this month to revisit this branch and clean some stuff like this up.  If you seen any other areas that can be improved, please let me know, thanks.

ylathouris: Cool. Yeah, even though Git has marked it all as new, I figured that wasn't completely true. I get that you're just making it play nice with setup.py so please take my comments with a grain of salt. If you get time to clean up some of the commented code then great, otherwise no worries at all.

bpabel:
Ok, this should be ready.

PR #494, opened 2018-03-03
Support keyword arguments in rez python

### LINKED ISSUE #492: Support for keyword arguments in `rez-python`
Looking at how rez-python is implemented, this looks like a difficult request, but I'll ask it anyway.

Right now `rez-python` only supports positional arguments- no flags or kwargs.

So, `rez-python ./scriptname /path/to/file/1 /path/to/file2` works, but `rez-python ./scriptname -i --sequence baz --shot 100` will not, because rez tries to eat the flags and kwargs, then errors out when it doesn't understand them.

I feel like it'd be a neat addition, because it would allow people to place `#!/usr/bin/env rez-python` as a script shebang line, and have it behave exactly as if you called the script with `rez-env rez -c "{sys.argv}"`, which is how I initially expected rez-python to behave.

skral:
Fixes #492.

@bpabel You were right there wasn't a lot missing to get this to work.

I added some more docstrings and tests as well.

### DIFF
--- a/src/rez/cli/python.py
+++ b/src/rez/cli/python.py
@@ -1,41 +1,68 @@
-"""
-Start a python interpreter or execute a python script within Rez's own execution context.
+"""Start a Python interpreter or execute a Python script.
+
+This will happen within Rez's own execution context. You can pass keyword
+arguments or flags by using a double dash: rez python FILE ARG -- --flag
 """
 
 
 def setup_parser(parser, completions=False):
+    """Set up command specific arguments.
+
+    Args:
+        parser (ArgumentParser): A preconfigured base parser we can add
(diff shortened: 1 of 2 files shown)


nerdvegas:
I am closing this as I've addressed it another way in https://github.com/nerdvegas/rez/pull/523

I've updated subcommands so they can parse cli args differently where necessary. This comes with some advantages:

* commands like rez-python now support passing args directly through to underlying proc (python in this case);
* other commands (eg rez-config) no longer accept rez-env-style arg groupings, ala "rez-env pkg1 -- command -- other stuff". They only did before as a matter of convenience, and these extra args were ignored.

PR #495, opened 2018-03-04
Update print statements to be Python 3 compatible

skral:
I updated all print statements in the entire code base to use the Python 3 syntax `print(value)` instead of the deprecated `print value`. It's a first step towards being Python 3 compatible.

Please take a look and let me know if you see any issues with merging these changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ Resolved environments can also be created via the API:
     >>> p = r.execute_shell(command='which hescape', stdout=subprocess.PIPE)
     >>> out, err = p.communicate()
     >>>
-    >>> print out
+    >>> print(out)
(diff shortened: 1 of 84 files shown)


nerdvegas:
Sebastian is ON FIRE.

Cheers for this, py3 compatibility is nearing necessity, especially with
the first moves in VFX towards adopting it officially.

FYI, py3 support in rez will need to happen as 2 distinct phases:

1. Ensure rez's own code (and thus the code in the rez 'rez' package
itself) is py3-compatible;
2. Port rez itself to py3.

In the interim, py2-based rez should be able to manage py3-based
environments just fine, and I've seen that working in practice already.

Thx
A

Issue #431, opened 2017-05-30
How are people building and releasing packages on Windows?


KelSolaar:
Hi,

Given #348, and other related issues, I wanted to know if there is a canonical way to build and release packages on Windows. I have been creating a lot of them manually but I would like to proof the system and reach a point where I can build and release them using *rez-build* and *rez-release*.

Cheers,

Thomas

instinct-vfx:
My current setup for things require builds is using bez and calling cmake from there. This also gives me a bit more flexibility for things that need to go into the call of cmake itself (e.g. platform target and alike).

I always wanted to go in there and look into doing proper cmake support on windows including required support for platform target overrides and other things, just never really found the time. Willing to join forces though if other people are interested.

KelSolaar:
@instinct-vfx : Thanks, that is quite useful, I have just given a spin at `bez` with the *hello_world* example. I have it semi working, I will need to massage the `rezbuild.py` file. Which VCS are you using, *git*?

instinct-vfx:
Yes we are using git. This is on of our (slightly stripped)  rezbuild.py files:

```
def build(source_path, build_path, install_path, targets):
    # Generate visual studio project files
    subprocess.call(["cmake", "-G", "Visual Studio 14 2015 Win64", "-T", "v100", ".."], cwd=os.path.join(source_path, "build_cmake"))

    ## Build solution and install into source_dir   
    subprocess.call(["cmake", "--build", ".", "--config", "Release"],
        cwd=os.path.join(source_path, "build_cmake"))    

    if "install" not in (targets or []):
        return

    # Copy files to repository when installing
    shutil.copytree(os.path.join(source_path, "bootstrap"), os.path.join(install_path, "bootstrap"))
    shutil.copytree(os.path.join(source_path, "plugins", "Release"), os.path.join(install_path, "plugins"))
```

KelSolaar:
Thanks!

How do you pick which compiler to use (x86, x64, etc...)? I made *cmake* and *vs_buildtools* packages and I picked the compiler doing something like that: `rez-env cmake vs_buildtools -- vcvars64.bat`

KelSolaar:
I will double check tomorrow but it seems like `cmake` actually manages to find the compiler this way.

instinct-vfx:
[...]

PR #501, opened 2018-04-04
Added mingw as a rez build_system for cmake.

willjp:
Barely a pull request, but I wanted the option to use mingw in addition to nmake on windows.

``` powershell
rez-build                      # still defaults to nmake
rez-build --bs mingw   # uses mingw32-make

# ~/.rezconfig can set either as cmake's build_system.
```

Thanks very much for rez!

### DIFF
--- a/src/rezplugins/build_system/cmake.py
+++ b/src/rezplugins/build_system/cmake.py
@@ -41,6 +41,7 @@ class CMakeBuildSystem(BuildSystem):
                      'codeblocks':  "CodeBlocks - Unix Makefiles",
                      'make':        "Unix Makefiles",
                      'nmake':       "NMake Makefiles",
+                     'mingw':       "MinGW Makefiles",
                      'xcode':       "Xcode"}
 
     build_targets = ["Debug", "Release", "RelWithDebInfo"]
@@ -175,14 +176,20 @@ class CMakeBuildSystem(BuildSystem):
             return ret
 
         # assemble make command
+        make_binary = self.settings.make_binary
         if self.settings.make_binary:
             cmd = [self.settings.make_binary]
+        elif self.cmake_build_system == 'mingw':
+            cmd = ["mingw32-make"]
+        elif self.cmake_build_system == 'nmake':
+            make_binary = 'nmake'
+            cmd = ["nmake"]
         else:
             cmd = ["make"]
         cmd += (self.child_build_args or [])
 
         # nmake has no -j
-        if self.settings.make_binary != 'nmake':
+        if make_binary != 'nmake':
             if not any(x.startswith("-j") for x in (self.child_build_args or [])):
                 n = variant.config.build_thread_count
                 cmd.append("-j%d" % n)
--- a/src/rezplugins/build_system/rezconfig.py
+++ b/src/rezplugins/build_system/rezconfig.py
@@ -30,7 +30,6 @@ cmake = {
 
 if os.name != "posix":
     cmake["build_system"] = "nmake"
-    cmake["make_binary"] = "nmake"
 
 
 # Copyright 2013-2016 Allan Johns.


willjp:
# tl;dr

If all you're trying to do is get rez working with cmake on windows, I wrote out instructions on how I was able to get it to work at the _bottom of this issue_: #348  

If you can get away with it though, I found mingw to be faster, since it doesn't need to run `vcvarsall.bat` 3x separate times for each build/install. If you're using all python source though, I'd probably just write something for bez. You'll have a faster turnaround. Alternatively, I noticed another pull request where someone was adding support for `setup.py` , which will likely turn out to be faster on windows.

``` bash
# test with cmakefile that copied a single, emptyfile.txt
# not compiling anything using:    ``rez-build --install``


# =============    ================  ======================
# Build System:    First Rez-Build   Subsequent Rez-Builds
# =============    ================  ======================
NMake                 38s                36s    ## most time spent running vcvarsall.bar
MinGW                 28s                5s     ## first-build, cmake spends time testing that installed/works
Bez                   2s                 2s     ## most time spent setting up rez-envs

```


# answering your question:

Full disclaimer, I'm entirely new to compiling software on windows, and I'm pretty new to Cpp/compilers in general. 

If i'm understanding it correctly though, ``NMake`` is still using visual-studio to do the compile.  It gets installed with the microsoft-build-tools under ``'C:\\Program Files (x86)\\Microsoft Visual Studio\\<year>\\BuildTools\\VC\\<version>\...``. At a glance, the ``Makefile`` that it spits out looks very similar to a gnu-make ``Makefile``.

CMake can also generate Visual Studio projects, targeting specific years, or versions. It creates what I'm guessing is a Visual Studio Project that can be opened in the Visual Studio IDE. These can still be compiled on the commandline using ``MSBuild`` . 
https://cmake.org/cmake/help/v3.0/generator/Visual%20Studio%2010%202010.html?highlight=visual%20studio



**non-rez NMake Workflow**
``` powershell
cd $your_project/build
 
cmd.exe                        # open cmd.exe within your powershell (reqd for vcvarsall.bat)
vcvarsall.bat x64              # set environment variables for compiling for x64 windows
 
cmake -G "NMake Makefiles" ..  # generate makefile
nmake                          # compile project
nmake install                  # install compiled project
 
 
exit                           # return to powershell
```


**non-rez visual studio cli Workflow**

``` powershell
cd $your_project/build
 
cmd.exe                     # open cmd.exe within your powershell (reqd for vcvarsall.bat)
vcvarsall.bat x64           # set environment variables for compiling for x64 windows
 
cmake ..                    # generate makefile
MSBuild BUILD_ALL.vcxproj   # compile project
MSBuild INSTALL.vcxproj     # install project
 
exit                        # return to powershell
```

# Why rez cmake builds don't work out-of-the-gate on windows

The reason that the default of 'NMake Makefiles' is not working out of the gate on windows could be because of one of the following:

* Visual Studio needs to set up a bunch of environment variables before compiling. ( https://msdn.microsoft.com/en-us/library/f2ccy3wt.aspx )
* According to my reading of other issues on here (I think one of your threads, actually) rez-envs do not inherit the environment of the shell it is invoked in. So even if you run vcvarsall.bat to set your environment, they are not set in the rez-env used for the build.
* A Makefile provides instructions to build a project. CMake is a scripting language to generate makefiles that work on a variety of platforms, and for a variety of compilers. Cmake itself doesn't actually compile or install your files, that is done by gnu-make, Nmake, or another build-system. Even if your project is entirely python, and requires no compilation at all CMake still needs to be pointed at something it can generate a Makefile for, that will be made responsible for copying your python source code from A to B.

nerdvegas:
Just a clarification on:


*rez-envs do not inherit the environment of the shell it is invoked in. So
even if you run vcvarsall.bat to set your environment, they are not set in
the rez-env used for the build.*
Rez envs DO inherit the original env, however any env-var appends/prepends
done in any of the resolved packages will OVERWRITE that env-var. Case in
point is PYTHONPATH - if it remained intact and was just added to, you
could easily end up with a PYTHONPATH that references existing packages
that might be in conflict with the newly resolved env.

Now having said that, something that's been on the cards for a long time is
more fine-grained control of this behaviour. I'll cite a problem - PATH.
There's something of a hack in place at the moment, which treats PATH as a
special case - otherwise, any append to PATH overwrites it completely, and
suddenly you lose your standard tools like ls, cd etc (which is... a little
problematic!).

I would prefer to have default behaviour which:

* Remembers the pre-rev-env value of a var;
* Applies currently behaviour (overwrite on first append/prepend);
* However, also appends the original pre-rez value back again at the end
(perhaps avoid dups at the same time).

There are some more complexities to this that I won't go into here, but
this is the general idea.

Thx
A

instinct-vfx:
For reference i posted an example of the setup i currently have in the issue below. I need to be able to properly drive cmake msvc builds with support for switching both versions and platform targets.

https://github.com/nerdvegas/rez/issues/431

assumptionsoup:
Thanks for your help @willjp and @instinct-vfx I'm going to go read through #348 , but before that, I wanted to make one small correction to your suggested cmake workflow on windows.  It's true that cmake creates a visual studio project, and you *can* call `MSBuild` to compile that project.  However, cmake understands most (all?) generators that it creates. So the more generic workflow is this:

```
cmake .. -G "Visual Studio 14 2015 Win64"
cmake.exe --build . --config Release
```
I'm not sure about the install step, but it probably exists as well.

willjp:
Thanks again @nerdvegas and @instinct-vfx ! I can see how changing the PATH/PYTHONPATH can very quickly get complicated. You have me curious now, time for me to do some more code-spelunking :) .

@assumptionsoup - oh! not necessarily faster, only faster with the way that I've been using it. I created a rez package for visual-studio that ran `vsvarsall.bat` in package.py's `commands()` . So it was running `vcvarsall.bat` 3x times during a `rez-build --install` .  I'm looking forwards to trying out something more similar to instinct-vfx's setup in #431 , in addition to seeing if I can rig something up so that it only needs to run once.  I'm just starting out, trying to decide on a workflow.

instinct-vfx:
Newer plugin setups are based on  build_command actually. And i also added a custom argument "-g", "--generate" which only generates but does not trigger a build. This is what i use when i start work on something: I clone the repo, and don't need a build of the last release, i just want to start looking into code. So i run with -g and build from VS and only build through rez-build/rez-release for testing and releasing.

For simplicity i am still using rezbuild.py though but call it through build_command like this:
`build_command = "python {root}/rezbuild.py {install}"`

in rez_build i first run cmake to generate:
` subprocess.call(["cmake", "-G", vs_version, source_path], cwd=build_path)`
and then check if --generate was used, and if it was used i simply return_

```
    if int(os.environ["__PARSE_ARG_GENERATE"]):
        # Return if --generate was given as we only want to generate solution and project files.
        log.info(
            "Solution and project files have been generated in the build folder ("
            + build_path + ")")
        return
```

If not, i actually build using something like this:
```
    # Build binaries
    subprocess.call(
        ["cmake", "--build", ".", "--config", "Release"], cwd=build_path)
```

I am still copying my artifacts manually if release is used as it feels a tad more flexible than having it in cmake thought that could be done easily.

My `parse_build_args.py` looks as follows for reference:
`parser.add_argument("-g", "--generate", action="store_true", help="Generate solution files only (=do not build or install)")`

Cheers,
Thorsten

nerdvegas:
Just reading up on this thread folks. Would it help to have rez-build run `vcvarsall.bat x64` when appropriate and before cmake is invoked?

Issue #517, opened 2018-07-12
Linux distribution not detected on Alpine Docker image


Aphosis:
It seems none of the methods used to find the Linux OS/distribution works in an Alpine Docker image.
This issue prevents Rez installation (or at least the `rez-bind --quickstart` command tu run).

I was trying to add a `rez-test` call in a GitLab pipeline job, and using a container seemed like a good starting point.

After some research I found that OS specific data was registered in a **/etc/os-release** file:

https://www.freedesktop.org/software/systemd/man/os-release.html

I managed to make it work by parsing this file in case the `lsb_release` command fails, but I was wondering if adding another resolution method was the way to go.

skral:
I wonder if we should replace finding the distribution code on Linux with [distro](https://github.com/nir0s/distro) instead of trying to find it ourselves. It handles a lot of distributions, it's pretty much self contained and we don't need to maintain that code ourselves.

nerdvegas:
Great find, that looks promising. It'd be ideal to offload this code from
rez.

A

PR #518, opened 2018-07-12
Added /etc/os-release Linux OS detection

Aphosis:
This pull request addresses #517.

It adds a new way to find the linux OS, by parsing /etc/os-release, which is described here:

https://www.freedesktop.org/software/systemd/man/os-release.html


nerdvegas:
Thanks Alexy,

Detecting linux distro seems to be a mixed bag of whatever works so I'm
happy to take this fix, cheers.
A

PR #496, opened 2018-03-04
Add all shells to Travis CI configuration

skral:
This adds all shells in Mac OS and Linux to the Travis CI configuration. Running `selftest` in every shell will increase our confidence in merging new code changes.

While setting up the configuration I noticed tests are breaking for `sh` on Linux and `tcsh` on Mac OS. I will try to take a look at why they are failing so we don't end up with a "build:error" badge right after the merge.

@nerdvegas I know you haven't been a big fan of Travis CI so far but I am sure we can figure out why the badge hasn't been all green all the time.

Let me know what you think of this PR.

PR #521, opened 2018-08-20
Shell based completion path selection

j0yu:
Basic selection logic to output the relevant shell completion command.

@skral, would this be of interest to #496 ?

### DIFF
--- a/install.py
+++ b/install.py
@@ -166,6 +166,15 @@ if __name__ == "__main__":
     print dest_bin_dir
 
     if completion_path:
-        print "You may also want to source the relevant completion script from:"
+        print
+        print "You may also want to source the",
+
+        shell = os.getenv('SHELL')
+        if shell:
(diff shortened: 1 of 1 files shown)


skral:
@j0yu I don't see a relation between these PRs but I might be wrong. Do you see one?

j0yu:
Hello @skral,

Sorry, I'm still fairly new to setting up and tinkering with the backend of `rez`.

I was thinking of utilising the additional tests of various shells in #496 as this will output different texts to `stdout` based on different shells.

Then again in regards to tests, the changes I introduced would probably come under a test specifically for installation rather than the `selftest`?

skral:
@j0yu You're right there is no installation specific test yet. #496 makes sure all shells rez supports are available in the CI environment on Linux and OSX, installs rez and runs `selftest`. I think it would be great to have some installation specific tests to check the integrity of rez after the installation, I don't remember seeing any PR including that yet though.

PR #522, opened 2018-09-05
-don't use memcache when reading developer package

nerdvegas:
resolves https://github.com/nerdvegas/rez/issues/514


JeanChristopheMorinRodeoFX:
Thanks a lot for that fix!

PR #539, opened 2018-11-08
Add a developer mode to rez-view

JeanChristopheMorinRodeoFX:
`--developer` is used to query a package in a given path. It's useful to get the package information from a CI infra (Jenkins, Travis, etc)

This PR is a retake from https://github.com/nerdvegas/rez/pull/467. It does exactly the same thing, but with all the goods of `rez-view`. It also unifies the viewing functionalities (see discussions in the linked PR).

`rez-view . --developer` will give us the package in the current directory in YAML format.
`rez-view /path/to/package/dir --developer` does the same thing that `rez-view . --developer`

New arguments added to `rez-view` (except --developer):
* `--fields`: Allows to only print the given fields
* `--separator`: If used, separator to be used when printing lists. defaults to empty space.
* `--pretty`: Prints the fields in a pretty manner, by default they are printed in raw format.

For examples, please see the tests included in this PR.

### DIFF
--- a/src/rez/cli/build.py
+++ b/src/rez/cli/build.py
@@ -61,9 +61,6 @@ def setup_parser(parser, completions=False):
         help="create build scripts rather than performing the full build. "
         "Running these scripts will place you into a build environment, where "
         "you can invoke the build system directly.")
-    parser.add_argument(
-        "--view-pre", action="store_true",
-        help="just view the preprocessed package definition, and exit.")
     setup_parser_common(parser)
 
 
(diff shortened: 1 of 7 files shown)


fnaum:
Hi @JeanChristopheMorinRodeoFX , I will check it against our rez info tomorrow. 
At first glance, I can see I will need to type a bit more when I need something from a developer package.
For example, something we use pretty often is 
`rez info version`
now (if I'm not wrong) it becomes
`rez view . --developer --fields version`
Not a huge deal because we use this mostly on scripts, but I will have to change all the instances.

I will do a review and think a bit more when I'm at work tomorrow

JeanChristopheMorinRodeoFX:
Indeed it's a little longer to type. But I had to do concessions since I had to combine two different tools in one. But if you want you can use `-d` instead of `--developer`, it's shorter.

If you or anybody  have any suggestions for the interface of the command line for `rez-view`, I'm all open.

nerdvegas:
I haven't gotten to this PR yet, but want to soon.

Off the top of my head, I think I'd make cwd ('.') implicit if not
provided, so then will become:

    rez-view -d -f version

I would also avoid 'fields' as an arg, and use 'format' instead. This is
more flexible and matches other cli tools (such as rez-search). Eg:

    rez-view -d -f 'this package is called {name}'

Cheers
A

JeanChristopheMorinRodeoFX:
Ok, but `--format` is already used in the `rez-view` command to specify in which format you want to see the result (py, yaml and text).

Also if you want the same kind of format than in `rez-search`, that will require some more work. It would also make `rez-view` a quite complicated command to use (a lot of different ways to use it) IMO.

I personally feel like it's getting convoluted with the package serialization + the search formatting that now seems to almost do the same thing... What are your thoughts on this?

Issue #549, opened 2018-12-13
--build-system not always available


brentVilla:
In rez-build's finding on what build-system plugin to use, it only defines the --build-system flag if two or more build systems are detected.  This is problematic because there is no way to force the build system if you have build infrastructure that is building codebases that may find a single build system and others find multiple.  You can't configure the build infrastructure to always pass a --build-system flag because that will fail for codebases where only a single build system is defined.  And you can't leave out that flag because then you'll get errors for multi-build system codebases.

Some options:

1. Always define the --build-system flag, even for single build system codebases.  Raise an error if the value of --build-system doesn't match the build system found.
2. Have a default build system configuration value that rez-build will use in codebases with when multiple build systems are found and a --build-system flag is not provided.
3. Be able to specify the default build system in the package.py file.

nerdvegas:
I'm going to include (1) and (3) in a fix shortly. I'm also going to remove the predetermined 'choices' list for `--build-system`... this isn't really necessary, and it causes an extra package.py evaluation. Typically this doesn't matter much, but I'm starting to think about packages with more expensive evaluations (ie early-bound functions). For eg, consider a smart installer package that retrieves info from a website.

nerdvegas:
So I've stumbled on an unfortunate problem. The cmake build system has its own `--build-system` option. This has never conflicted before, because the cmake args are only added to the argparser if there's one valid build system... in which case, the standard `--build-system` option isn't added.

There is no fix for this other than removing or changing the cmake `--build-system` arg, so it's a rare case where I have to break backwards compatibility. I will change it to `--cmake-build-system`... hopefully, if anyone is using this, they're using the `--bs` short form.

nerdvegas:
https://github.com/nerdvegas/rez/releases/tag/2.26.0

PR #550, opened 2018-12-14
Issue 549

nerdvegas:
#### Addressed Issues

* [#549](https://github.com/nerdvegas/rez/issues/549) '--build-system' rez-build option not always
  available

#### Notes

To fix this issue:
* The '--build-system' rez-build option is now always present.
* To provide further control over the build system type, the package itself can now specify its build
  system - see https://github.com/nerdvegas/rez/wiki/Package-Definition-Guide#build_system

#### COMPATIBILITY ISSUE!

Unfortunately, the 'cmake' build system had its own '--build-system' commandline option also. This
was possible because previous rez versions suppressed the standard '--build-system' option if only
one valid build system was present for a given package working directory. **This option has been
changed to '--cmake-build-system'**.

### DIFF
--- a/src/rezplugins/build_system/cmake.py
+++ b/src/rezplugins/build_system/cmake.py
@@ -38,22 +38,22 @@ class CMakeBuildSystem(BuildSystem):
     """
 
     build_systems = {
-        'eclipse':     "Eclipse CDT4 - Unix Makefiles",
-        'codeblocks':  "CodeBlocks - Unix Makefiles",
-        'make':        "Unix Makefiles",
-        'nmake':       "NMake Makefiles",
-        'mingw':       "MinGW Makefiles",
-        'xcode':       "Xcode"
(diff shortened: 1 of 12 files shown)


src/rezplugins/build_system/cmake.py
```diff
@@ -75,8 +75,8 @@ def bind_cli(cls, parser):
                             type=str, choices=cls.build_targets,
                             default=settings.build_target,
                             help="set the build target (default: %(default)s).")
-        parser.add_argument("--bs", "--build-system", dest="build_system",
-                            type=str, choices=cls.build_systems.keys(),
+        parser.add_argument("--bs", "--cmake-build-system",
```
fnaum: this needs 
`dest="cmake_build_system"`,

If not when running with --bs it won't be picked in line 96

nerdvegas:
Fix included in https://github.com/nerdvegas/rez/releases/tag/2.29.1

Cheers
A

PR #572, opened 2019-02-05
package_test.py: fix rez-test header command with %

e2lka:
We had a test with a command that contains a % that was raising because of a bad use of print_header in package_test.py.

```
print_header("Running test commange: my command 3%")
FAIL

print_header("Running test commange",  "my command 3%")
OK
```

### DIFF
--- a/src/rez/package_test.py
+++ b/src/rez/package_test.py
@@ -203,7 +203,7 @@ class PackageTestRunner(object):
                 else:
                     cmd_str = ' '.join(map(quote, command))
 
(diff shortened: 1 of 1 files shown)


JeanChristopheMorinRodeoFX:
@nerdvegas Another easy one. In the same line as https://github.com/nerdvegas/rez/pull/540. Though this current PR is not just cosmetic, it is really a bug that prevents us from running `rez-test` in some cases.

We targetted the `develop` branch and hope it's the right branch to target... Let us know if it's not.

Thanks!

PR #586, opened 2019-03-22
Issue 583 hashed variants

### LINKED ISSUE #583: hash-based variant subpaths
Currently, variant subpaths are based on that variant's requirements. This is problematic for variants with conflict/range requirements.

Add a mode (enabled via a new 'hashed_variants' pkg attrib) which places variants into subpaths which are a sha256(?) hash of the variant (ie, of its requirements).

This will create unwieldy install paths, however I have thoughts on a separate feature to address this.

nerdvegas:
Closes #583

* Adds new "hashed variants" feature.
* Also adds "variant shortlinks" for hashed variant packages. This creates symlinks so that long variant root paths are avoided.


src/rez/package_resources_.py
```diff
@@ -429,9 +437,26 @@ def _uri(self):
     def _subpath(self):
         if self.index is None:
             return None
+
+        if self.parent.hashed_variants:
+            h = sha1(str(map(str, self.variant_requires)))
```
bpabel: One issue to think about, especially if you're planning on eventually supporting python 3, is that with the str/unicode/bytes change in python 3, the arguments to the hashlib hashes are a bit different, and they no longer accept unicode strings, only encoded byte strings.  And ideally, you'd want the hash to be the same whether it was run from a python 2 or python 3 rez.

The following should produce the same hash in py2/3:

```
import six
variant = ['one', 'two']
h = sha256()
h.update(u','.join(map(six.text_type, variant)).encode('utf8'))
h.digest()
```

nerdvegas:
Cheers I'll try this out.

Will be declining this PR as I've found issues related to package copy that
weren't taken into account, however that should be sorted pretty soon.

A

PR #588, opened 2019-03-29
Fix path list with whitespace

asztalosdani:
The packages path is split badly if a path contains whitespace.
If I set `REZ_PACKAGES_PATH `to `c:/foo bar/baz`, rez will interpret it as `["c:/foo", "bar/baz"]`.

This PR fixes this issue. I also made a test to prove this.

### DIFF
--- a/src/rez/tests/test_config.py
+++ b/src/rez/tests/test_config.py
@@ -234,6 +234,9 @@ class TestConfig(TestBase):
         packages_path = [
             "c:/foo bar/baz",
             "c:/foo 2 bar/baz",
+            r"c:\foo bar\baz",
+            "/home/foo bar/baz",
+            r"/home/foo\ bar/baz"
         ]
         os.environ["REZ_PACKAGES_PATH"] = os.pathsep.join(packages_path)
 
--- a/src/rez/tests/test_config.py
+++ b/src/rez/tests/test_config.py
@@ -226,6 +226,19 @@ class TestConfig(TestBase):
         finally:
             os.environ = old_environ
 
+    def test_7(self):
+        """Test path list environment variable with whitespace."""
+        c = Config([self.root_config_file], locked=False)
+
+        # test basic env-var override
+        packages_path = [
+            "c:/foo bar/baz",
+            "c:/foo 2 bar/baz",
+        ]
+        os.environ["REZ_PACKAGES_PATH"] = os.pathsep.join(packages_path)
+
+        self.assertEqual(c.packages_path, packages_path)
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/src/rez/config.py
+++ b/src/rez/config.py
@@ -126,6 +126,10 @@ class OptionalStrList(StrList):
 class PathList(StrList):
     sep = os.pathsep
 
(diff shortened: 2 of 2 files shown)


oktomus:
You haven't committed with your github account. You can fix this by doing the following
``` 
git config user.email "github email"
git reset [hash of previous commit]
git add 
git commit "Fix path list with whitespace"
git push -f asztalosdani:master
```

asztalosdani:
Thank you, I fixed it.

src/rez/tests/test_config.py
```diff
@@ -226,6 +226,19 @@ def _data(self):
...
 
+    def test_7(self):
+        """Test path list environment variable with whitespace."""
+        c = Config([self.root_config_file], locked=False)
+
+        # test basic env-var override
+        packages_path = [
+            "c:/foo bar/baz",
+            "c:/foo 2 bar/baz",
+        ]
```
oktomus: You could add more samples to your test. For instance:
- /home/foo bar/baz
- /home/foo\ bar/baz
- c:\foo bar\baz

asztalosdani: done

src/rez/tests/test_config.py
```diff
@@ -226,6 +226,22 @@ def _data(self):
         finally:
             os.environ = old_environ
 
+    def test_7(self):
```
nerdvegas: fails on linux (pathsep is `:`)

src/rez/tests/test_config.py
```diff
@@ -226,6 +226,22 @@ def _data(self):
...
+
+        # test basic env-var override
+        packages_path = [
+            "c:/foo bar/baz",
+            "c:/foo 2 bar/baz",
+            r"c:\foo bar\baz",
+            "/home/foo bar/baz",
+            r"/home/foo\ bar/baz"
+        ]
+        os.environ["REZ_PACKAGES_PATH"] = os.pathsep.join(packages_path)
```
fnaum: I think it is not a good idea to change an environment variable in the UT as it affects all other tests that are run as part of the same process (i.e when running rez selftests)
This line breaks a test on other unrelated area.
If you really need to, then put the original value back...something along the lines

```
original_packages_path = os.environ["REZ_PACKAGES_PATH"]
os.environ["REZ_PACKAGES_PATH"] = os.pathsep.join(packages_path)
self.assertEqual(c.packages_path, packages_path)
os.environ["REZ_PACKAGES_PATH"] = original_packages_path
```
or if you are going to use it several times create a `@contextmanager` and with `with ...`

Also, I do not understand the original use case of this change.
Why not set the path like `c:/foo;c:/bar/baz` or "/foo:baz/bar` why do you need the space?

mottosso: Agreed on not changing variables that persist across tests, however:

> Why not set the path like c:/foo;c:/bar/baz or "/foo:baz/bar` why do you need the space?

The folder `c:\foo bar\` is different from `c:\foo` and `c:\bar` I think.

fnaum: Would quotes work in this case? ie. "c:\foo bar\"?

PR #589, opened 2019-03-30
Python3 compatibility proposal

jware-vc:
- I have a simple python3 compatible implementation of rez. 
- I added a rez python3 setup file so you can bind python3 in your current environment into a rez package. 
- The hello_world has been updated to use build_requires instead of requires. It is possible to deploy a python2 and python3 variant of the same version of hello_world. If we do not specify the requires python, the version pulled in will be determined by which python3 package you specify at resolve time.

The resolves seem to work as expected. Thoughts?

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,1 +1,0 @@
-theme: jekyll-theme-midnight
--- a/example_packages/hello_world/package.py
+++ b/example_packages/hello_world/package.py
@@ -15,10 +15,12 @@ tools = [
     "hello"
 ]
 
-requires = [
-    "python"
+build_requires = [
+    "python3",
 ]
-
+variants = [
+    ['python3-3'],
+]
 uuid = "examples.hello_world_py"
 
 def commands():
--- a/example_packages/hello_world/python/hello_world.py
+++ b/example_packages/hello_world/python/hello_world.py
@@ -1,4 +1,4 @@
 
 
 def hello():
-    print "Hello world!"
+    print("Hello world!")
--- a/src/rez/bind/python3.py
+++ b/src/rez/bind/python3.py
@@ -1,0 +1,67 @@
+"""
+Binds a python executable as a rez package.
+"""
+from __future__ import absolute_import
+from rez.bind._utils import check_version, find_exe, extract_version, \
+    make_dirs, log, run_python_command
+from rez.package_maker__ import make_package
+from rez.system import system
+from rez.utils.lint_helper import env
+from rez.utils.platform_ import platform_
+from rez.vendor.version.version import Version
+import shutil
+import os.path
+
+
+def setup_parser(parser):
+    parser.add_argument("--exe", type=str, metavar="PATH",
+                        help="bind an interpreter other than the current "
+                        "python interpreter")
+
+
+def commands():
+    env.PATH.append('{this.root}/bin')
+
+
+def bind(path, version_range=None, opts=None, parser=None):
+    # find executable, determine version
+    exepath = find_exe("python3", opts.exe)
+    code = "import sys; print('.'.join(str(x) for x in sys.version_info))"
+    version = extract_version(exepath, ["-c", code])
+
+    check_version(version, version_range)
+    log("binding python: %s" % exepath)
+
+
+    # make the package
+    #
+
+    def make_root(variant, root):
+        binpath = make_dirs(root, "bin")
+        link = os.path.join(binpath, "python3")
+        platform_.symlink(exepath, link)
+
+
+    with make_package("python3", path, make_root=make_root) as pkg:
+        pkg.version = version
+        pkg.tools = ["python3"]
+        pkg.commands = commands
+        pkg.variants = [system.variant]
+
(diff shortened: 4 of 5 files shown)


bpabel:
Just to be clear, rez isn't python 3 compatible and won't run under python 3 yet.  But a python 2 rez is perfectly capable of launching python 3 environments like you've done here.

That being said, you should probably install python 3 to the same package family as python 2, not create a new one called `python3`, since that would allow *both* packages to be loaded in the same environment unless you made sure to specify conflicts between them, and packages that are compatible with both wouldn't be able to just list `python` as a dependency and still run with `python-3`

Also, check out [this discussion](https://groups.google.com/forum/#!topic/rez-config/R1iJ9iE8Txg) on the rez-config google group.  It discusses how `rez-bind` is likely to be deprecated in the near future in favor of using some cool new features in rez package definitions that achieve essentially the same thing.

jware-vc:
Sorry, thank you for correcting me. I had poorly worded my comment. 

I was liking python3 being its own package family, and the py version conflict is much more manageable with rez than without. To be honest, I was having difficulties coming up with a workflow to manage python 2 and 3 as versions in the same package family. How do you handle the situation where variation a requires python-2, and variation b requires python-3?

bpabel:
Is there a reason you can't just do 

```
variations = [
    ['python-2'],
    ['python-3'],
]
```

jware-vc:
Holy crap. I had been using this wrong. I didn't realize rez could build multiple variations at once. I need to look into this more. Thanks for the info!

jasoncscott:
Jon,

Remember that each variant is treated like its own resolve, so a build is
repeated for each variant as well.


Jason Scott

jware-vc:
Howdy,

So I had been using a py3 feature branch in git to manage two separate versions of my software. Suppose that hello_world package's python folder has completely different file or folder structure for python 2 and 3. I have been needing to do a bit of jiggery pokery to make transition my larger projects. Do you think it would be possible to link variants with feature branches in git?

jasoncscott:
If you're asking about the future, I guess anything is possible?  If you're
asking about an existing feature to utilize git branches, I don't think
that would work, since rez wants to tag a commit with the release version.
You could probably bypass that, but it sort of feels wrong to do that.

Having different file or folder structures is absolutely do-able, so that's
not an issue--it's just the git branching that's probably the problem right
now.


Jason

jware-vc:
Can you explain more about the issue with tagging? I don't think I understand the problem. Wouldn't it just be tagging the release on the py3 branch? and tag the master for py2?

I was thinking something along the lines of:
`
variants = {
    'git:master': ['python-2'],
    'git:branchName2': ['python-3'],
}
`

bpabel:
The way you could make this work now is to switch branches manually and specify which variant it should be building under when you build/release it.

```
$ rez-build -i --variants 0
```

Then switch to your py3 branch and build the python-3 variant

```
$ rez-build -i --variants 1
```

The variant numbers are zero-indexed, so this assumes your variants look like
```
variants = [
    ['python-2'],
    ['python-3'],
]

jware-vc:
Oh, I started playing around with rez-suites. This is pretty amazing. That may be a good solution for one of my needs. You are all amazing people. Thank you for your help.

Has there been any concern with the length of PYTHONPATH environment variable as the amount of resolved packages increases? I have been playing around with an python import hook to import from rez packages without needing PYTOHNPATH to modified. Could that be useful?

bpabel:
> Has there been any concern with the length of PYTHONPATH environment variable as the amount of resolved packages increases?

It's been discussed, but mostly with regards to the added latency of reading many paths off of network shares.

> I have been playing around with an python import hook to import from rez packages without needing PYTOHNPATH to modified. Could that be useful?

Sounds interesting.  You should bring it up on the google group, as you'll likely get more visibility and input there since it's better designed for long-form discussion compared to github isssues/PR's.

nerdvegas:
Has there been any concern with the length of PYTHONPATH environment
variable as the amount of resolved packages increases?

Yes. Two points to note:

* I want to add a 'package localization' feature soon, which will auto
cache package payloads to local disk. Even though this will not shorten
PYTHONPATH, it does have the capacity to dramatically decrease the cost of
all those file stats that would typically be happening over shared storage
at a given facility.

* There's a feature I have in mind called "context filesystems". At resolve
time, packages would have the opportunity to remap their payload contents
into a shared directory. These directories would be temporary, but I would
probably add a caching system for them at some point. You would also be
able to specify whether a package's contents should be symlinked or copied.
If your packages are aware of this feature and express how they should copy
their payloads, then long PYTHONPATH should be a thing of the past.

Hth
A

nerdvegas:
Hi @jware-vc, what would you like to do with this PR? I want to deprecate rez-bind, but adding a binding for py3 is reasonable. My only issue with the PR currently is that the package is `python3` rather than `python-3.*`, as mentioned in these comments.

jware-vc:
Thank you for your help. This PR is not necessary.

PR #597, opened 2019-04-18
Implement preprocess support for rezconfig.py

mottosso:
Here's a feature proposal, to help simplify life for beginners, without breaking advanced use.

**rezconfig.py**

```python
# Option 1, original
package_preprocess_function = "my_module.function"

# Option 2, this proposal
def package_preprocess_function(this, data):
    # do fancy things
```

I found the original a little heavy on the up-front investment to give the feature a try. Needing to:

1. Write and maintain a separate module
2. Add and maintain another environment variable, i.e. `PYTHONPATH`, outside of the otherwise excellent Rez environment management.

But I can see how the feature circumvents some of the conceptual benefits of the original; e.g. if you've already got an established PYTHONPATH of global e.g. `rezutils.py`, then the overhead isn't as noticeable. However I can see this help beginners getting into Rez, and does not exclude the original use.

### Usecase

My usecase was to prefix packages with a "category".

**package.py**

```python
name = "core_pipeline"
category = "int"
```

**rezconfig.py**

```python
packages_path = [
    "~/packages/int",
    "~/packages/ext",
    "~/.rez/packages/int",
    "~/.rez/packages/ext"
]

def package_preprocess_function(this, data):
    if not data.get("category"):
        return

    try:
        data["config"]["local_packages_path"]
        data["config"]["release_packages_path"]

    except KeyError:
        pass

    else:
        # Already explicitly specified by package
        return

    category = data["category"]
    config = data.get("config", {})

    config["release_packages_path"] = os.path.join(_release_path, category)
    config["local_packages_path"] = os.path.join(_local_path, category)

    data["config"] = config
```

### DIFF
--- a/src/rez/config.py
+++ b/src/rez/config.py
@@ -109,6 +109,10 @@ class OptionalStr(Str):
     schema = Or(None, basestring)
 
 
+class OptionalObject(Str):
+    schema = Or(None, object)
+
+
 class StrList(Setting):
     schema = Schema([basestring])
     sep = ','
@@ -312,7 +316,7 @@ config_schema = Schema({
     "implicit_back":                                OptionalStr,
     "alias_fore":                                   OptionalStr,
     "alias_back":                                   OptionalStr,
-    "package_preprocess_function":                  OptionalStr,
+    "package_preprocess_function":                  OptionalObject,
     "context_tracking_host":                        OptionalStr,
     "variant_shortlinks_dirname":                   OptionalStr,
     "build_thread_count":                           BuildThreadCount_,
--- a/src/rez/developer_package.py
+++ b/src/rez/developer_package.py
@@ -171,6 +171,7 @@ class DeveloperPackage(Package):
 
             if preprocess_func:
                 print_info("Applying preprocess from package.py")
+
             else:
                 # load globally configured preprocess function
                 dotted = self.config.package_preprocess_function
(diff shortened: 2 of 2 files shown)


JeanChristopheMorinRodeoFX:
This PR makes sense. Though one thing that would make it easier to use is if both the `package_preprocess_function` in the `rezconfig.py` and the `preprocess` functon of a rez package would be additive. We use the `preprocess` function in `package.py` in what @nerdvegas calls the "native package". These packages uses a common library defined inside `package_definition_build_python_paths` that is used to compute the path to where our DCCs are installed. That way the logic is centralized and using them in the preprocess allows us to bake the information in the released packages. See https://github.com/nerdvegas/rez/issues/514#issuecomment-396608518.

I'm not sure if making both additive wuold make sense in this PR or not. Also, note that the tests will need to be changed in order to test the new behavior.

mottosso:
Sounds better suited for a separate PR; something that could get merged with or without this one.

nerdvegas:
Agreed RE separate PR for Jean's suggestion. This should be driven by a new
build-time-only package attribute called 'preprocess_mode', which would be
one of ["before", "after"]. If present, the global preprocessor would also
be called, "before" or "after" the package's. If not specified, only the
package's preprocess would be applied.

RE this PR, I have no problems with the feature, but is it WIP as specified
in the description, or is it ready?

Thx
A

mottosso:
>but is it WIP as specified in the description, or is it ready?

Was in the process of testing it as I was submitting, but so far there hasn't been any issues on either Windows nor Linux. Happy to merge!

JeanChristopheMorinPerso:
I opened an issue to track what I was talking about, see https://github.com/nerdvegas/rez/issues/609

nerdvegas:
Requires tests before I can merge. Should be easy, see existing tests/test_config.py

src/rez/config.py
```diff
@@ -109,6 +109,10 @@ class OptionalStr(Str):
     schema = Or(None, basestring)
 
 
+class OptionalObject(Str):
+    schema = Or(None, object)
```
nerdvegas: This schema is a bit too loose; also the subclassing of `Str`, while it will work currently, is a bit shaky (Str._parse_env_var gives expected behaviour for OptionalObject, but that's not reason enough for the subclassing). It's also somewhat misleading for an `OptionalObject` to also support env-var parsing as a string, but not anything else. Suggest:

```
class OptionalStrOrFunction(Setting):
    schema = Or(None, str, callable)
    def _parse_env_var(self, value):
        # note: env-var override only supports string, eg 'mymodule.preprocess_func'
        return value
```

nerdvegas:
*ping*!

PR needs to address points made above before I can merge.

JeanChristopheMorinPerso:
@nerdvegas I spoke with @mottosso and I'll take care of addressing your comments and adding the appropriate tests. I'll then push a PR for the additive preprocess (#609).

mottosso:
More than happy with that, in fact all of my PRs and issues are open for anyone to contribute to.

JeanChristopheMorinPerso:
@nerdvegas I had to create a new PR in order to be able to add commits to this. We can close this current PR and continue in #650 .

nerdvegas:
Closed in lieu of https://github.com/nerdvegas/rez/pull/650

PR #599, opened 2019-04-18
WIP Feature/pip wheels windows

mottosso:
Based on https://github.com/lambdaclan/rez/commits/feature/pip-install-modern.

Specific to Windows, but also affects Linux, see below.

### Problem

Some packages - like `lxml`, `numpy` and `PySide` - have wheels that avoid a need to compile, solving primarily for Windows users - whom may not have the necessary compilers installed for building those.

The cost however is that this approach couldn't separate the source, header and lib files, like a vanilla `rez-pip` would.

### Solution

```bash
$ rez pip --install lxml --use-wheel
```

Add `rez pip --use-wheel` to recommend `pip` to use wheels when possible, and fall back to building otherwise (default pip behavior). Backwards compatibility is maintained by having the argument be optional.

### Test

Confirmed to work with these, both wheel and non-wheels alike.

```bash
$ rez pip --install --release --use-wheel ^
    Cython ^
    Jinja2 ^
    MarkupSafe ^
    Pillow ^
    PySide ^
    Qt.py ^
    blockdiag ^
    certifi ^
    excel ^
    funcparserlib ^
    lockfile ^
    lxml ^
    ordereddict ^
    pyblish-base ^
    pyblish-lite ^
    pyblish-maya ^
    pythonnet ^
    pywin32 ^
    setuptools ^
    urllib3 ^
    webcolors ^
    xlrd ^
    six
```

### Moving forward

`--use-wheel` doesn't actually enforce the use of wheel; it merely calls `pip install` without additional arguments, which just so happens to default to using wheels. The question is whether `--use-wheel` should force a wheel, and fail if none is found. Or if `--use-wheel` is the best name for this argument.

I also removed the ability to pick a `pip` version; not because it isn't important, but because it simplified the implementation and I honestly couldn't see a need to separate between both Python and Pip versions; whichever `pip` is available as a package during `rez pip` would get picked up by this approach - which in a nutshell is `python -m pip`. I'm probably missing something though, so something to look out for.


mottosso:
Superseded by https://github.com/nerdvegas/rez/pull/614

Issue #590, opened 2019-04-03
"parse_build_args.py" file parser arguments are not accessible anymore in "os.environ".


KelSolaar:
Hi,

I'm trying to update Rez from 2.16.0 to 2.28.0 and one of the first thing I noticed is my builds failing :)

We create `parse_build_args.py` files to support some custom build arguments with content for example as follows:

```python
# -*- coding: utf-8 -*-
parser.add_argument(
    '--skip-code-formatting',
    action='store_true',
    help='Skips code formatting.')
```

and then in the build file (we use Fabric, thus `fabfile.py`), we do things like that:

```python
    if not int(os.environ['__PARSE_ARG_SKIP_CODE_FORMATTING']):
        code_formatting()
```

Unfortunately the `__PARSE_ARG_SKIP_CODE_FORMATTING` is nowhere to be seen now.

Cheers,

Thomas

instinct-vfx:
Hey Thomas,

We ran into this too. Here is my report: https://github.com/nerdvegas/rez/issues/587

An here is a commit fixing this. PR is coming soon (Thanks Iraklis!) https://github.com/lambdaclan/rez/commit/af81b2492d297ad3350d1988bca47de0b4dcae7c

KelSolaar:
@instinct-vfx : Ah good to know, will follow closely the future PR!

KelSolaar:
I made a custom build with two cherry-picked commits from @lambdaclan and it fixed the issue for me.

Would be fantastic to have a PR for that so that we don't have to deploy a custom version, nothing really bad about it but it is extra maintenance steps.

Cheers,

Thomas

nerdvegas:
I'm keeping an eye out for this PR also. There will be another round of
merges coming soon and this one's on the list.
A

lambdaclan:
@KelSolaar  @nerdvegas I have just added a PR for the issue +1 more incoming!

nerdvegas:
I believe this is now fixed in 2.30.0, pls reopen if incorrect

Issue #587, opened 2019-03-26
Custom build arguments broken since 2.26.0


instinct-vfx:
This occured and is reproducible for me on windows 7.

In 2.26.0 cli/build.py was changed from:

```
    if clss:
        if len(clss) == 1:
            cls = clss[0]
            cls.bind_cli(parser)
        else:
            types = [x.name() for x in clss]
            parser.add_argument(
                "-b", "--build-system", dest="buildsys", type=str, choices=types,
                help="the build system to use.")
``` 

to:

```
    if clss:
        if len(clss) == 1:
            cls_ = clss[0]
            title = "%s build system arguments" % cls_.name()
            group = parser.add_argument_group(title)
            cls_.bind_cli(group)

        types = [x.name() for x in clss]
    else:
        types = None
```

This change leads to custom build arguments (defined in parse_build_args.py) to be visible when calling `rez-build --help` as an additional group at the end of the argument help (which i assume is the intention here) but the corresponding environment variables are not created in the build environment.

This seems to be due to `cls_.bind_cli(group)` only passing on the group rather than the parser which breaks the argument parsing in custom.py

lambdaclan:
Hello Thorsten,

This indeed seems to be a bug, it can easily be remedied by using the `parser._action_groups` private method to access the parser's groups. 

Simply passing the parser object (LazyArgumentParser - subclass of normal ArgumentParser) instead of the group object (AgumentParserGroup) would also work but the extra arguments will be in the main help menu (group item will be ignored).

![Selection_001](https://user-images.githubusercontent.com/47409392/55057063-72f80f80-50ab-11e9-83ff-00fe5c6f0df3.png)

This means that we should pass the parser object to `cls_bind_cli` like we used to before 2.26 but then get the argument group object to insert the extra arguments defined in the  parse_build_args.py file on that instance (enabling the custom group when using --help) and finally set the attribute to the parser object instead of the argument group object. 

![Selection_002](https://user-images.githubusercontent.com/47409392/55057452-9b343e00-50ac-11e9-8d11-9815896e7e17.png)

I have already got it working locally so we can open a PR. 👍

nerdvegas:
PR please, if it isn't there already :)

A

lambdaclan:
@instinct-vfx @nerdvegas I have just added a PR for the issue :+1:  more incoming!

nerdvegas:
fixed in https://github.com/nerdvegas/rez/releases/tag/2.30.0

PR #601, opened 2019-04-19
Bugfix/custom build arguments

lambdaclan:
# Description

Prior to rez version 2.26, the function call to `bind_cli` within `src/rez/build.py` was passing the parser object as an argument which was then used to store the extra args via `setattr` defined within `parse_build_args.py` onto so that we can get to it in self.build.  

From version 2.26 onwards the passed arguments changed from `ArgumentParser` to an **argument group object** in order to display the extra arguments in a separate group for help messages. 

The above change created a regression because although the custom build arguments (defined in parse_build_args.py) are visible when calling rez-build --help as an additional group the corresponding environment variables are not created in the build environment because the `setattr` function is using the wrong object.

This pull request fixes the problem by passing both the group and parser objects to correctly generate the help menu but also ensure the environment variables are set.

Fixes # (issue)
#590, #587  

## Type of change

## Screenshots

Please see #587

# How Has This Been Tested?

I have manually tested the changes myself multiple times with various arguments by proving
a series of `parse_build_args.py` files. Furthermore other members of the community have
tried my changes are have reported them to be working.

> I made a custom build with two cherry-picked commits from @lambdaclan and it fixed the issue for me.
> 
> Would be fantastic to have a PR for that so that we don't have to deploy a custom version, nothing really bad about it but it is extra maintenance steps.
> 
> Cheers,
> 
> Thomas

## Todos
- [ ] Tests
- [ ] Documentation

@nerdvegas Please advice :pray: 

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/rez/cli/build.py
+++ b/src/rez/cli/build.py
@@ -46,7 +46,7 @@ def setup_parser_common(parser):
             cls_ = clss[0]
             title = "%s build system arguments" % cls_.name()
             group = parser.add_argument_group(title)
-            cls_.bind_cli(group)
+            cls_.bind_cli(parser, group)
 
         types = [x.name() for x in clss]
     else:
--- a/src/rezplugins/build_system/custom.py
+++ b/src/rezplugins/build_system/custom.py
@@ -60,7 +60,7 @@ class CustomBuildSystem(BuildSystem):
             child_build_args=child_build_args)
 
     @classmethod
-    def bind_cli(cls, parser):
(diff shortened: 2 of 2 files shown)


lambdaclan:
🎆 First merge,  hopefully more to come :)

Issue #503, opened 2018-04-06
rez-pip NO packages were installed


asztalosdani:
Hi,

when installing Qt.py package with rez-pip, it fails.
```
C:\Users\daniel.asztalos>rez-pip --install Qt.py
15:57:58 INFO     Using pip-1.5.6 (C:\Users\daniel.asztalos\packages\pip\1.5.6\package.py[0])
Downloading/unpacking Qt.py
  Downloading Qt.py-1.1.0-py2.py3-none-any.whl
Installing collected packages: Qt.py
Successfully installed Qt.py
Cleaning up...

NO packages were installed.```

**Comment by assumptionsoup** [comment]
I've had this happen on and off too.  On my system I have anaconda installed, and I have noticed that rez-pip completes an install of Qt.py into anaconda instead of a rez package.  I wonder if this is an issue with anaconda, windows, or both.

**Comment by assumptionsoup** [comment]
Running the command twice, results in rez telling me that the requirement is satisfied:
```
>>>rez pip --install qt.py
15:12:33 INFO     Using pip-1.5.6 (F:\packages\int\pip\1.5.6\package.py[0])
Requirement already satisfied (use --upgrade to upgrade): qt.py in c:\users\jordan\appdata\local\continuum\anaconda2\lib\site-packages
Cleaning up...

NO packages were installed.
```

Which isn't really helpful in any scenario.  The fact that I have a package on my system already,shouldn't affect my ability to create a rez pacakge IMO.  Just because it's on my system doesn't mean it's on everyone's system.  That's rez's job.

**Comment by willjp** [comment]
I've also had this happen on windows - I tried a few of my deps without success, then went searching on the google-group and tried installing the package that was being used as an example when rez-pip was launched (flask).

https://groups.google.com/forum/#!searchin/rez-config/rez-pip%7Csort:date/rez-config/Pro1FGuRzIY/wjSXFydfDQAJ

That did mostly work, enough that I was able to use what worked as a base to create other packages. It's a bit of a chore, but it works well enough, and similar to the amount of effort I was putting in before rez.

``` 
.../packages/mock/2.0.0/
    platform-windows/arch-AMD64/os-windows-10.0.16299/
        maya-2017/
        maya-2018/
        python2.7.14/
        python3.6.4/
            python/
                <site-packages contents...>
    package.py
```


``` python
# package.py
# -*- coding: utf-8

name = 'mock
version = '2.0.0'
requires = [
    'six',
    'funcsigs',
[...]

Issue #475, opened 2018-01-29
Rez pip doesn't handle optional requirements


adriankrupa:
Rez 2.17.1 Windows

For example `rez pip --install pytest` installs on windows succesfully with `colorama`, but it's not added to `requires`. In addition right now rez installs PyPi packages with egg-info and not with dist-info and `distlib` doesn't parse optional requirements. 
Maybe `requires` shoud be changed to late binding function to support this optional requirements.

instinct-vfx:
We're not using the official rez-pip here on windows for various reasons. Most notably the way it is implemented to install into a temp directory (using pip's --install-options) enforces use of source packages rather than binary packages if available. This also causes installing as egg rather than with dist-info by the way. On windows this is quite a hassle as building binary extensions has a lot of pre-requisites on the machine used to build and is often non-trivial and also often does not work out of the box.

In addition we want some custom handling to support different python builds for the same version (e.g. the official python is built with VS2008 which makes official binary packages incompatible and we need to be able to deploy binary packages twice to support python hosts that were built using different toolchains. We have a dedicated wrapper package for MSVC versions that allows for that. We detect binary extensions (by checking all installed files for binary extensions like dll / pyd etc. because metadata is often unreliable) and if an extension is binary we add the toolchain used.

We detect optional requirements, but currently set them as non-optional obviously. Well, we handle some of them. They can be pretty diverse and supporting all possible cases seems kind of daunting.

We also have a fair bit of special handling in there (e.g. generated exe files like with bumpversion, we have a custom python version with patched setuptools/distlib wrappers to support non-absolute paths in generated exe wrappers) to support all kinds of specialities and metadata version quirks and all the other things bringing tears to my eyes.

Anyways, i would welcome some more discussion in that area with other windows based people / shops!

Cheers,
Thorsten

JeanChristopheMorinPerso:
It looks like it was fixed by #602

PR #596, opened 2019-04-17
WIP Implement support for multiple packages with rez-pip install

mottosso:
Hi all,

This PR adds support for passing multiple packages to `rez pip install`, to better align with vanilla `pip`, and simplify the release of lots of packages at once.

```bash
$ rez pip install packageA packageB packageC --release
```

PR #602, opened 2019-04-19
[WIP] Feature/pip install modern

lambdaclan:
# Description

[In collaboration with #599]

The Python community is steadily moving away from eggs and transitioning into wheels. 

As per the Python [wheels website](https://pythonwheels.com/)
> Wheels offer faster installation for pure Python and native C extension packages, avoid arbitrary code execution for installation. (avoids setup.py) and **offer more consistent installs across platforms and machines**.

One of the most important differences is the cross platform compatibility. Furthermore the current
implementation of the pip install command is using install options to split and point the installation to a custom directory (libs, scripts, includes etc) is enforcing the use of source packages and disables the
modern pip default of using a wheel if available. This means that everything that is not a pure python package needs a proper, working and matching compiler setup making it a real pain for Windows users whom may need multiple visual studios versions etc...

The difference with #599 is that it enforces the use of wheels via [PEP 517](https://www.python.org/dev/peps/pep-0517/#source-distributions) which seems to be the way forward. If a package does not have a wheel then one is being built prior to installation ensuring
consistency.

There is still ongoing discussion [here](https://discuss.python.org/t/pep-517-and-projects-that-cant-install-via-wheels/791/3) from the pip maintainers but a few things that got my attention are the following statements:

>we intend (at some point) to remove the legacy "install via setup.py" code path from pip.

> We have that legacy mode at the moment - setup.py installs are (at least in my mind) legacy since PEP 517 support was released. The problem is that we can’t support that (or any other) legacy mode indefinitely, so the question is when (and how) we remove it.

Another interesting discussion to follow is the one [here](https://github.com/pypa/pip/issues/2677) about handling install options with wheel packages.

Relates # (issue)
rez #475, #503 maybe more...

pip https://github.com/pypa/pip/issues/4611 https://github.com/pypa/pip/issues/4501

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Remarks

* Requires pip >=19.0
* Requires  distlib >= 0.28

# How Has This Been Tested?

I have installed a plethora of packages from known to obscure with good success rate.
Some issues still occur (unable to build a wheel - very rare) depending on the platform but the same problems would also occur with the current implementation.

I have also successfully installed the majority of the packages listed on the Python [wheels website](https://pythonwheels.com/) which do not yet provide a wheel without issues.

**Test Configuration**:
* Python version: 2.7.15
* OS: Linux, Windows 10
* Toolchain: pip 19.0.3, distlib 0.2.8

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Further investigation about dependency gathering (run-time and optional)
- [ ] Metadata and distlib API research notably [DependencyFinder](https://distlib.readthedocs.io/en/latest/reference.html#distlib.locators.DependencyFinder) and [Metadata 2.1](https://packaging.python.org/specifications/core-metadata/#requires-dist-multiple-use) along with [PEP 508](https://www.python.org/dev/peps/pep-0508/)

# Checklist:

- [ ] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -244,16 +244,26 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
     # Collect resulting python packages using distlib
     distribution_path = DistributionPath([destpath])
     distributions = [d for d in distribution_path.get_distributions()]
-    metadata = next(iter([dist.metadata for dist in distributions if dist.key == source_name]), None)
 
-    # TODO deal with the scenario when a wheel fails to build, ie no metadata
-    if not metadata:
-        raise ValueError('Wheel metadata for {} was not found.'.format(source_name))
+    for distribution in distribution_path.get_distributions():
+        requirements = []
+        if distribution.metadata.run_requires:
+            # Handle requirements. Currently handles conditional environment based
+            # requirements and normal requirements
+            # TODO: Handle optional requirements?
+            for requirement in distribution.metadata.run_requires:
+                if "environment" in requirement:
+                    if interpret(requirement["environment"]):
+                        requirements.extend(_get_dependencies(requirement, distributions))
+                elif "extra" in requirement:
+                    # Currently ignoring optional requirements
+                    pass
+                else:
+                    requirements.extend(_get_dependencies(requirement, distributions))
 
-    tools = []
-    src_dst_lut = {}
+        tools = []
+        src_dst_lut = {}
 
-    for distribution in distributions:
         for installed_file in distribution.list_installed_files():
             if 'bin' in installed_file[0]:
                 installed = os.path.relpath(installed_file[0], os.path.join('..','..'))
@@ -261,6 +271,7 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
                 installed = installed_file[0]
 
             source_file = os.path.normpath(os.path.join(destpath, installed))
+
             if os.path.exists(source_file):
                 destination_file = source_file.split(stagingsep)[1]
                 exe = False
@@ -277,58 +288,60 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
             else:
                 _log("Source file does not exist: " + source_file + "!")
 
-    def make_root(variant, path):
-        """Using distlib to iterate over all installed files of the current
-        distribution to copy files to the target directory of the rez package
-        variant
-        """
-        for source_file, data in src_dst_lut.items():
-            destination_file, exe = data
-            destination_file = os.path.normpath(os.path.join(path, destination_file))
+        def make_root(variant, path):
+            """Using distlib to iterate over all installed files of the current
+            distribution to copy files to the target directory of the rez package
+            variant
+            """
+            for source_file, data in src_dst_lut.items():
+                destination_file, exe = data
+                destination_file = os.path.normpath(os.path.join(path, destination_file))
 
-            if not os.path.exists(os.path.dirname(destination_file)):
-                os.makedirs(os.path.dirname(destination_file))
+                if not os.path.exists(os.path.dirname(destination_file)):
+                    os.makedirs(os.path.dirname(destination_file))
 
-            shutil.copyfile(source_file, destination_file)
-            if exe:
-                shutil.copystat(source_file, destination_file)
+                shutil.copyfile(source_file, destination_file)
+                if exe:
+                    shutil.copystat(source_file, destination_file)
 
-    # determine variant requirements
-    # TODO detect if platform/arch/os necessary, no if pure python
-    variant_reqs = []
-    variant_reqs.append("platform-%s" % _system.platform)
-    variant_reqs.append("arch-%s" % _system.arch)
-    variant_reqs.append("os-%s" % _system.os)
+        # determine variant requirements
+        # TODO detect if platform/arch/os necessary, no if pure python
+        variant_reqs = []
+        variant_reqs.append("platform-%s" % _system.platform)
+        variant_reqs.append("arch-%s" % _system.arch)
+        variant_reqs.append("os-%s" % _system.os)
 
-    if context is None:
-        # since we had to use system pip, we have to assume system python version
-        py_ver = '.'.join(map(str, sys.version_info[:2]))
-    else:
-        python_variant = context.get_resolved_package("python")
-        py_ver = python_variant.version.trim(2)
+        if context is None:
+            # since we had to use system pip, we have to assume system python version
+            py_ver = '.'.join(map(str, sys.version_info[:2]))
+        else:
+            python_variant = context.get_resolved_package("python")
+            py_ver = python_variant.version.trim(2)
 
-    variant_reqs.append("python-%s" % py_ver)
+        variant_reqs.append("python-%s" % py_ver)
 
-    name = metadata.name
+        name = distribution.metadata.name
 
-    with make_package(name, packages_path, make_root=make_root) as pkg:
-        pkg.version = metadata.version
-        if metadata.summary:
-            pkg.description = metadata.summary
+        with make_package(name, packages_path, make_root=make_root) as pkg:
+            pkg.version = distribution.metadata.version
+            if distribution.metadata.summary:
+                pkg.description = distribution.metadata.summary
 
-        pkg.variants = [variant_reqs]
+            pkg.variants = [variant_reqs]
+            if requirements:
+                pkg.requires = requirements
 
-        commands = []
-        commands.append("env.PYTHONPATH.append('{root}/python')")
+            commands = []
+            commands.append("env.PYTHONPATH.append('{root}/python')")
 
-        if tools:
-            pkg.tools = tools
-            commands.append("env.PATH.append('{root}/bin')")
+            if tools:
+                pkg.tools = tools
+                commands.append("env.PATH.append('{root}/bin')")
 
-        pkg.commands = '\n'.join(commands)
+            pkg.commands = '\n'.join(commands)
 
-    installed_variants.extend(pkg.installed_variants or [])
-    skipped_variants.extend(pkg.skipped_variants or [])
+        installed_variants.extend(pkg.installed_variants or [])
+        skipped_variants.extend(pkg.skipped_variants or [])
 
     # cleanup
     shutil.rmtree(tmpdir)
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -231,10 +231,8 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
 
     # Build pip commandline
     cmd = [pip_exe, "install",
-           "--install-option=--install-lib=%s" % destpath,
-           "--install-option=--install-scripts=%s" % binpath,
-           "--install-option=--install-headers=%s" % incpath,
-           "--install-option=--install-data=%s" % datapath]
+           "--use-pep517",
+           "--target=%s" % destpath]
 
     if mode == InstallMode.no_deps:
         cmd.append("--no-deps")
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -303,13 +303,12 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
 
         variant_reqs.append("python-%s" % py_ver)
 
-        name, _ = parse_name_and_version(distribution.name_and_version)
-        name = distribution.name[0:len(name)].replace("-", "_")
+    name = metadata.name
 
         with make_package(name, packages_path, make_root=make_root) as pkg:
-            pkg.version = distribution.version
-            if distribution.metadata.summary:
-                pkg.description = distribution.metadata.summary
+        pkg.version = metadata.version
+        if metadata.summary:
+            pkg.description = metadata.summary
 
             pkg.variants = [variant_reqs]
             if requirements:
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -220,7 +220,7 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
     stagingsep = "".join([os.path.sep, "rez_staging", os.path.sep])
 
     destpath = os.path.join(stagingdir, "python")
-    # TODO use binpath once https://github.com/pypa/pip/pull/5983 is approved
+    # TODO use binpath once https://github.com/pypa/pip/pull/3934 is approved
     binpath = os.path.join(stagingdir, "bin")
 
     if context and config.debug("package_release"):
@@ -245,6 +245,10 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
     distribution_path = DistributionPath([destpath])
     distributions = [d for d in distribution_path.get_distributions()]
 
+    folders = [folder for folder in os.listdir(destpath) if os.path.isdir(os.path.join(destpath, folder))]
+    if "bin" in folders:
+        shutil.move(os.path.join(destpath, "bin"), binpath)
+
     for distribution in distribution_path.get_distributions():
         requirements = []
         if distribution.metadata.run_requires:
@@ -265,23 +269,31 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
         src_dst_lut = {}
 
         for installed_file in distribution.list_installed_files():
-            if 'bin' in installed_file[0]:
-                installed = os.path.relpath(installed_file[0], os.path.join('..','..'))
-            else:
-                installed = installed_file[0]
+            # a relative directory that needs to be appended without the
+            # os.sep (via os.path.join) in order for normpath to return the
+            # correct path, this is in order to match the wheel RECORD file
 
-            source_file = os.path.normpath(os.path.join(destpath, installed))
+            # Example
+
+            # destpath = /tmp/pip-HBkRGa-rez/re]z_staging/python
+            # installed_file[0] = ../../bin/pydocstyle
+            # normpath of /tmp/pip-HBkRGa-rez/rez_staging/python../../bin/pydocstyle =  /tmp/pip-HBkRGa-rez/rez_staging/bin/pydocstyle OK
+            # normapth of /tmp/pip-HBkRGa-rez/rez_staging/python/../../bin/pydocstyle = /tmp/pip-HBkRGa-rez/bin/pydocstyle WRONG
+            if installed_file[0].startswith("."):
+                installed = destpath + installed_file[0]
+            else:
+                installed = os.path.join(destpath, installed_file[0])
+
+            source_file = os.path.normpath(installed)
 
             if os.path.exists(source_file):
                 destination_file = source_file.split(stagingsep)[1]
                 exe = False
 
                 if is_exe(source_file):
-                    if destination_file.startswith("%s%s" % (os.path.join("python", "bin"), os.path.sep)):
-                        destination_file = os.path.join("bin", os.path.basename(source_file))
-                        _, _file = os.path.split(destination_file)
-                        tools.append(_file)
-                        exe = True
+                    _file = os.path.basename(destination_file)
+                    tools.append(_file)
+                    exe = True
 
                 data = [destination_file, exe]
                 src_dst_lut[source_file] = data
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -255,8 +255,9 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
     distribution_path = DistributionPath([destpath])
     distributions = [d for d in distribution_path.get_distributions()]
 
-    folders = [folder for folder in os.listdir(destpath) if os.path.isdir(os.path.join(destpath, folder))]
-    if "bin" in folders:
+    # moving bin folder to expected relative location as per wheel RECORD files
+    staged_binpath = os.path.join(destpath, "bin")
+    if os.path.isdir(staged_binpath):
         shutil.move(os.path.join(destpath, "bin"), binpath)
 
     for distribution in distribution_path.get_distributions():
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -284,10 +284,13 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
             # when in fact ../bin seems to be the resulting path after the
             # installation as such we need to point the bin files to the
             # expected location to match wheel RECORD files
-            if installed_file[0].startswith("."):
-                installed = destpath + installed_file[0]
+            installed_filepath = installed_file[0]
+            bin_prefix = os.path.join('..', '..', 'bin') + os.sep
+            if installed_filepath.startswith(bin_prefix):
+                # account for extra parentdir as explained above
+                installed = os.path.join(destpath, '_', installed_filepath)
             else:
-                installed = os.path.join(destpath, installed_file[0])
+                installed = os.path.join(destpath, installed_filepath)
 
             source_file = os.path.normpath(installed)
 
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -295,7 +295,7 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
             source_file = os.path.normpath(installed)
 
             if os.path.exists(source_file):
-                destination_file = source_file.split(stagingsep)[1]
+                destination_file = os.path.relpath(source_file, stagingdir)
                 exe = False
 
                 if is_exe(source_file):
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -25,6 +25,41 @@ import sys
 import os
 import re
 
+VERSION_PATTERN = r"""
+    v?
+    (?:
+        (?:(?P<epoch>[0-9]+)!)?                           # epoch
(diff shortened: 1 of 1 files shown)


mottosso:
Differences between this and https://github.com/nerdvegas/rez/pull/599

This PR..

1. Enforces use of wheel
2. Backwards breaking
3. No update to distlib

(1) seems like a good idea, though not 100% sure we'd want to require anything this close to bleeding-edge `pip`.

For (2), this doesn't have to be a breaking change I think. What do you think about the proposal in #599, of maintaining current behavior and adding a `--use-wheel` argument? It could speed up merging and provide a transition path for anyone interested in using this new feature. Without it, it's a little risky to update Rez once this has been merged, especially for those not keeping up to date with PRs.

For (3), I thought we needed to update it? You found a way of using the still-old release? :O

JeanChristopheMorinRodeoFX:
> though not 100% sure we'd want to require anything this close to bleeding-edge pip

Wheels are far from being bleeding edge. They've been used for a couple of years now. Wheels are the standard since a year or 2.

> Without it, it's a little risky to update Rez once this has been merged, especially for those not keeping up to date with PRs.

If there are some people that don't look at release notes or even testing before updating, they live an extremely extreme life. Though, I know Allan likes to not break compatibility AFAIK. On my side, if we can avoid the breakage, why not. But if it's necessary, why not too (take it with a grain of salt, coming from me that doesn't use rez-pip and use my own implementation that has nothing to do with the current implementation).

mottosso:
> Wheels are far from being bleeding edge. 

I was actually referring to pip-19.

JeanChristopheMorinRodeoFX:
I see, thanks for the clarification. I don't see any problem with using pip >= 19. 19.0 was released in January. 19.1 just got released (like 11 hours ago). So >19.0,<19.1 isn't considered bleeding edge while 19.1+ could be considered bleeding edge, but normally pip is quite stable. In 4-5 years, I've rarely hit pip issues. If there is something to worry about, it's [PEP-518](https://www.python.org/dev/peps/pep-0518/), not `pip>=19.0` nor PEP-517` IMO.

nerdvegas:
I agree with assessment of current rez-pip limitations. Wheels seem the way
forward.

Backwards compatibility in rez is paramount. For this reason I am in favour
of supporting this new code path either via a `--use-wheel` option as
Marcus suggests; or, a separate cli altogether (rez-pip2 or rez-wheel
perhaps - leaning towards rez-pip2).

The reason I suggest a separate cli is because I'd be concerned about
ending up with a confusing mix of wheel-compatible and non-wheel-compatible
options available in the same tool (ie some options that make sense in
combo with --use-wheel and some that do not).

Ira if you're getting good results then I'm keen to see more.

Thx
A

nerdvegas:
Hi @lambdaclan , @motto,

How does this and https://github.com/nerdvegas/rez/pull/599 relate? Are they competing implementations that we need to choose from? Would like to talk more about these PRs in reference to one another before moving forward.

Thx
A

lambdaclan:
Hello Allan, thank you very much for taking the time to look at the pull request.

I have been using my changes above for quite some time without any issues and I am sure Marcus can attest to this since he has been doing the same. 

Although the PR from Marcus is based on this one we are doing things a bit differently such as me enforcing wheel installs. I have also updated distlib but I did not push those changes so as not to make the PR harder to read. I will do so before the final merge. The PR clearly states that a new version of distlib is required for wheel installs to work.

Maybe I was overly ambitious in thinking that we can get rid of the old style installs since they are considered legacy now but since backwards compatibility is important then I can keep the original install method in place.

As such,  I am happy to comply with the suggested solution of adding a new separate cli that will only work with wheels  instead of using a flag such as --use-wheels to avoid the issue of wheel compatible incompatible packages as you mention. Using a specific command for wheel installs will be immediately apparent that packages are installed using wheel packages .

In doing so, It will also be easier to deprecate the old code base when the time comes rather than trying to keep the code from a flag option.

@nerdvegas @mottosso I am willing to do the changes outlined above if it is OK with you so please let me know so that I can proceed. 

As I mentioned in another post, I do see some duplication of work happening and it is a shame. We need a way to track who is doing what to encourage people to work on other features/issues or promote collaboration between community members. This is really important in my opinion, what do you think would be a suitable solution? Maybe having a rez kanban board?

I am happy to work with Marcus on this if he wants otherwise I will just carry on.

Looking forward to your input. Thank you.

Ira

mottosso:
> How does this and #599 relate?

In short, #599 and #596 is consolidated into #614 which is akin to #628, but different.

I would recommend consolidating this into #628 as well, so we can have a look at both and determine what we like and dislike about each and ultimately consolidate each into one for final merge. Smorgasbord!

> We need a way to track who is doing what to encourage people to work on other features/issues or promote collaboration between community members.

IMO this is a little over the top. A simple "Hey, I'm working on X" in the chat should suffice. The reason for these duplicate PRs isn't a lack of kanban, but rather:

1. I needed pip
2. You had a branch I wasn't able to merge (squashed)
3. I re-implemented the feature on my end and made #599 
4. You refactored your branch and made #602
5. I continue the work in #614, where things start to diverge
6. You continue yours in #628, divergence continues

And as a result, we've got some lovely exploration and features to choose between. :)

lambdaclan:
> I would recommend consolidating this into #628 as well, so we can have a look at both and determine what we like and dislike about each and ultimately consolidate each into one for final merge. Smorgasbord!

I do not mind either way so I will leave it up to Allan to decide. We will still need to extract the wheel installation method to a new separate cli tool as outlined above and add some tests. I am happy to help
if Marcus is on it then I will do something else.

> > We need a way to track who is doing what to encourage people to work on other features/issues or promote collaboration between community members.
> 
> IMO this is a little over the top. A simple "Hey, I'm working on X" in the chat should suffice. The reason for these duplicate PRs isn't a lack of kanban, but rather:
> 
>     1. I needed pip
> 
>     2. You had a branch I wasn't able to merge (squashed)
> 
>     3. I re-implemented the feature on my end and made #599
> 
>     4. You refactored your branch and made #602
> 
>     5. I continue the work in #614, where things start to diverge
> 
>     6. You continue yours in #628, divergence continues
> 
> 
> And as a result, we've got some lovely exploration and features to choose between. :)

That would be nice but unfortunately I do not think that everyone publicly announces what they are working on furthermore a message on chat can easily missed by other members of the community and also not seen at all by people not part of the chatting group.

Having a kanban is more official and lets everyone know what active contributors are working on. It will also allow us to work on features/issues that matter based on a roadmap.

I am not in control so there is not much I can do but these are my thoughts. I dont want to know how many people have branches doing the same work as others because their studio needs it...

I want to make clear that I do not mind collaborating on the same issue with you or anyone else but we do need a way to track everything.

mottosso:
> I am happy to help if Marcus is on it then I will do something else.

Sure, I'm happy to handle pip. Each of yours and Allans requirements (both here and #628) are included in #614 so we could continue from there.

src/rez/pip.py
```diff
@@ -265,18 +262,24 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
         tools = []
         src_dst_lut = {}
 
-        for installed_file in distribution.list_installed_files(allow_fail=True):
-            source_file = os.path.normpath(os.path.join(destpath, installed_file[0]))
+        for installed_file in distribution.list_installed_files():
+            if 'bin' in installed_file[0]:
```
nerdvegas: It's not obvious what's going on here, a comment would help. Also, I take it that this covers a case where `installed_file[0]` is present within a 'bin' dir; in this case, a substring search isn't robust enough, I would search for 'bin' in the list generated from `str.split(os.sep)`. I gather that placement of said bin dir is important also, I'd check for that as well.

src/rez/pip.py
```diff
@@ -265,18 +262,24 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
...
 
-                if is_exe(source_file) and \
-                        destination_file.startswith("%s%s" % ("bin", os.path.sep)):
-                    _, _file = os.path.split(destination_file)
-                    tools.append(_file)
-                    exe = True
+                if is_exe(source_file):
+                    if destination_file.startswith("%s%s" % (os.path.join("python", "bin"), os.path.sep)):
+                        destination_file = os.path.join("bin", os.path.basename(source_file))
+                        _, _file = os.path.split(destination_file)
```
nerdvegas: _file = os.path.basename(destination_file)

nerdvegas:
Hey @lambdaclan,

I'm pretty happy to merge this very soon, I just had one suggestion regarding backwards compatibility. Since support for pep517 was added in pip-19, I think it'd be sufficient to check the pip version and use this new code in that case, rather than introducing a --use-wheel as suggested before. It feels like that may be a cleaner way to go about it, and will avoid the confusion of extra options. It seems to me that wheels are prevalent enough that we can go that route regardless.

What do you think? I'll be back on this tomorrow, I'm happy to make those changes if we want to do that and if you haven't already. I know this is holding up #628 so I'm keen to get it merged.

Cheers
A

src/rez/pip.py
JeanChristopheMorinPerso: I seems like I also commented this line in https://github.com/nerdvegas/rez/pull/628/files#r296459995. I'm always confused when the same change appears in two different PRs.

Here is the specification of the database of installed python distributions: https://www.python.org/dev/peps/pep-0376/#record

src/rez/pip.py
```diff
@@ -315,11 +318,10 @@ def make_root(variant, path):
 
         variant_reqs.append("python-%s" % py_ver)
 
-        name, _ = parse_name_and_version(distribution.name_and_version)
-        name = distribution.name[0:len(name)].replace("-", "_")
+        name = distribution.metadata.name
 
         with make_package(name, packages_path, make_root=make_root) as pkg:
-            pkg.version = distribution.version
+            pkg.version = distribution.metadata.version
```
JeanChristopheMorinPerso: Note that we need to be careful with this. The python version schema specification isn't 100% compatible with rez.

lambdaclan:
Hello @nerdvegas and @JeanChristopheMorinRodeoFX 

Apologies for the delayed reply. With the latest progress taking place in the repository it is becoming diffucult to keep track of all the incoming emails.

I was waiting to see what will happen between this PR and Marcuse's one from #614 

> > I am happy to help if Marcus is on it then I will do something else.
> 
> Sure, I'm happy to handle pip. Each of yours and Allans requirements (both here and #628) are included in #614 so we could continue from there.

Since you have decided to move forwards with this PR and #628 then I will go ahead and the do the changes requested.

If anything else pops up please comment on either PR and I will look into it. I will keep an eye on both to make sure I do not miss anything.

nerdvegas:
Thanks Ira! Both rez-pip and the installation process are things I want to
get sorted asap, so I'll be as responsive as possible on this.

Just a note - my family and I are moving Friday this week so I'll more than
likely go dark for 3-4 days, so if I don't reply to emails around then,
that's why.

Cheers
A

lambdaclan:
Hello Allan, no problem at all, thanks for letting us know. 

I hope the move goes smoothly!

I am going through your and JC comments now and should have something ready for review fairly soon.

Ira

JeanChristopheMorinPerso:
Thanks for letting us know @nerdvegas . @lambdaclan I'll keep an eye open and will do another pass of review when you will push. You might have to look at https://github.com/nerdvegas/rez/pull/628 because I might have do some comments there that are applicable here (sorry about that).

Just so we are all on the same page, this PR enforces the usage of wheels only and I think we will need a PR to update distlib right?

lambdaclan:
Hello JC, thank you very much for your time reviewing my PRs.

Feedback from you and Allan is highly appreciated since I am fairly new to the project and want to make sure that everything works as intended without comprising existing functionality.

Indeed, this PR enforces the usage of wheels, if a packages does not have a wheel one is automatically built during install and then process continues as expected. As pip updates keep coming in we will possibly need to make adjustments to rez-pip as well. For example enforcing pep-517 might become the default in the near future.

In terms of distlib, it is a pre-requisite for this feature (and #628) to work so I can either make another PR or add it to this one. I purposely kept it out to make it easier to read.

PR #628 depends on this one since wheel installation method is a per-requisite. Let us get this one sorted out first and then I will move onto that one.

Hope this clears things a bit.

Ira

nerdvegas:
Another PR for the distlib update would be great. I'll be able to merge
that tonight, and if #602 is ready then that also.

Thx
A

src/rez/pip.py
lambdaclan: I have been looking into ways to improve the readability of this. Since this is installed as a wheel distribution, all the installed files are listed in a CSV file called RECORD including the bin scripts as such there is no worry of missing any files from the list.

I am aware of the specification for the database generated pointed out by JC but there seems to be some inconsistencies when combining the wheel install with the target flag to set a directory. Specifically, the bin script paths in the RECORD file
are wrong.

Essentially it points the bin scripts to be two directories up instead of being in the base location.

Have a look at an example install of pydocstyle library:

image

The bin folder is right there along with the rest of source files but a quick look into the RECORD file, revelas that the bin location should be somewhere else...

image

This is why you see all that weird parsing and path combining taking place. I am trying to see if I am missing something but the rest of the path are correct starting right from the destpath location as base.

Using --prefix instead of --target will put the files in a site-packages directory and the scripts in a bin folder

image

image

This time the RECORD file points the bin scripts to be 3 directories up from base.

image

which is strangely correct but doing so will require to parse the source packages out of lib/python/site-packages..

I will keep looking into for a bit longer and see if I find anything.

lambdaclan: Relevant issue: https://github.com/pypa/pip/issues/3934

lambdaclan: I am now moving the bin folder to where the RECORD file expects it to be ../../bin and everything is working fine. The checks have become very simple.

src/rez/pip.py
lambdaclan: I have been using metadata in a couple of places for summary and name, those shouldn't be an issue I guess. I never had an issue with version but I can revert back to the original if necessary.

This is an issue I have been getting but I fixed it on the other PR

https://github.com/nerdvegas/rez/pull/628/commits/98be1156cc61d54d846304f13028d6bcfeb0e369

lambdaclan:
Excluding tests, I believe this is now good for a second review. I will be keeping an eye on it so that we can sort it out as soon as possible and move onto the next one (#628). Preparing a simple PR for distutils now.

bin_handling

nerdvegas:
Hey @lambdaclan,

It seems like retaining backwards compatibility for pip-<19 may be more effort than it's worth. I think it'd be fine to move ahead without out it, so long as there's a check for this which prints an obvious error message. Let's just go with that?

nerdvegas:
Um, that close was an accident!!

src/rez/pip.py
nerdvegas: Yeah I see what you mean, it's odd!

src/rez/pip.py
```diff
@@ -245,9 +242,13 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
     _system = System()
 
     # Collect resulting python packages using distlib
-    distribution_path = DistributionPath([destpath], include_egg=True)
+    distribution_path = DistributionPath([destpath])
     distributions = [d for d in distribution_path.get_distributions()]
 
+    folders = [folder for folder in os.listdir(destpath) if os.path.isdir(os.path.join(destpath, folder))]
```
nerdvegas: I would add a comment along the lines of, "moving bin folder to expected relative location, rather than under python/" etc. Also, simpler code would suffice:

```
staged_binpath = os.path.join(destpath, "bin")
if os.path.isdir(staged_binpath):
    shutil.move(...)
```

src/rez/pip.py
```diff
@@ -267,16 +268,30 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
...
+            # a relative directory that needs to be appended without the
+            # os.sep (via os.path.join) in order for normpath to return the
+            # correct path, this is in order to match the wheel RECORD file
+
+            # Example
+
+            # destpath = /tmp/pip-HBkRGa-rez/re]z_staging/python
+            # installed_file[0] = ../../bin/pydocstyle
+            # normpath of /tmp/pip-HBkRGa-rez/rez_staging/python../../bin/pydocstyle =  /tmp/pip-HBkRGa-rez/rez_staging/bin/pydocstyle OK
+            # normapth of /tmp/pip-HBkRGa-rez/rez_staging/python/../../bin/pydocstyle = /tmp/pip-HBkRGa-rez/bin/pydocstyle WRONG
```
nerdvegas: The comment and implementation threw me for a little bit here.

The comment should just explain the odd case: that ../../bin/etc is provided by distlib when in fact ../bin/etc seems to be the correct path instead. I was a little confused because the comment here delves into the implementation more than it does explain the situation.

src/rez/pip.py
```diff
@@ -267,16 +268,30 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
...
+            if installed_file[0].startswith("."):
```
nerdvegas: I would make this clearer, perhaps like so:

```
installed_filepath = installed_file[0]
bin_prefix = os.sep.join('..', '..', 'bin') + os.sep
...

if installed_filepath.startswith(bin_prefix):
    # account for extra parentdir as explained above
    installed = os.path.join(destpath, '_', installed_filepath)
else:
    ...
```

src/rez/pip.py
```diff
@@ -267,16 +268,30 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
...
+                installed = destpath + installed_file[0]
+            else:
+                installed = os.path.join(destpath, installed_file[0])
+
+            source_file = os.path.normpath(installed)
 
             if os.path.exists(source_file):
-                destination_file = installed_file[0].split(stagingsep)[1]
+                destination_file = source_file.split(stagingsep)[1]
```
nerdvegas: I apologise for my own (old) rubbish code! This split is odd, simple use of os.path.relpath would be clearer

```
destination_file = os.path.relpath(source_file, stagingdir)
```

src/rez/pip.py
```diff
@@ -317,11 +332,10 @@ def make_root(variant, path):
 
         variant_reqs.append("python-%s" % py_ver)
 
-        name, _ = parse_name_and_version(distribution.name_and_version)
-        name = distribution.name[0:len(name)].replace("-", "_")
+        name = distribution.metadata.name
```
nerdvegas: The rez package name can't be simply set to the dist name, because some pip packages have hyphen in the name. In rez this is not a valid package name (it would be interpreted as the start of the version, eg my-pkg-1.2 is 'my', version 'pkg-1.2').

src/rez/pip.py
nerdvegas: I'd second JC here, even if initially there's no remapping at all, there should be a separate `pip_to_rez_version` function.

Taking a look at https://www.python.org/dev/peps/pep-0440/#version-scheme though, there are some things to take into account. I can see two potential issues
1: version epochs (they make no sense to rez, so they'd just get stripped of the leading `N!`;
2: python versions are case insensitive, so they should probably be lowercased when converted to a rez version.

nerdvegas:
Thanks @lambdaclan, please see comments. I think we're close.

I'll merge the distlib version update PR along with this once it's ready.

FYI tomorrow night is my last chance to do the merge before I move house! My pc will be packed up after that :)

Cheers all
A

JeanChristopheMorinPerso:
@lambdaclan Have you tested with both a package that is only available as a source distribution and also tested with a wheel and an egg?

src/rez/pip.py
JeanChristopheMorinPerso: Local versions are also not compatible with rez (https://www.python.org/dev/peps/pep-0440/#local-version-segments) BTW, local is something we use at RodeoFX for our pip packages when we fork a repo for example. Though, we don't use `rez-pip` since we have our own implementation. But some other studios could use the local versions. At Rodeo, we basically convert `package-1.0.0+<something>` to `package-1.0.0-<something>`.

Most possibilities are described in https://www.python.org/dev/peps/pep-0440/#examples-of-compliant-version-schemes.

lambdaclan:
Hello @nerdvegas 

> It seems like retaining backwards compatibility for pip-<19 may be more effort than it's worth. I think it'd be fine to move ahead without out it, so long as there's a check for this which prints an obvious error message. Let's just go with that?

Sounds good to me, definitely keeping the old codebase in for installations from source would take some time, with wheels available for almost every package a pip version check as you said, should suffice. Furthermore, when a wheel is not available one is being built automatically.

@JeanChristopheMorinRodeoFX 

>  Have you tested with both a package that is only available as a source distribution and also tested with a wheel and an egg (Maybe we can skip eggs, as this is quite an old format that I've not seen in a while in PyPI)?

Thinking quickly, PyYAML for Linux doesn't have a wheel in PyPI and same for psutil, PyPI only contain sources for these packages on Linux.

Hello JC, I have been using this for quite sometime and did not have any issues, the issue I has I mentioned before was a Schema error -->

In the PR post I mentioned the following:

>  I have installed a plethora of packages from known to obscure with good success rate.
Some issues still occur (unable to build a wheel - very rare) depending on the platform but the same problems would also occur with the current implementation.

> I have also successfully installed the majority of the packages listed on the Python wheels website which do not yet provide a wheel without issues.

Saying that, I am happy to try more packages that might give issues such as the ones you mention so I will try them both (PyYAML and psutil) and let you know how it goes.

> I'll merge the distlib version update PR along with this once it's ready.
FYI tomorrow night is my last chance to do the merge before I move house! My pc will be packed up after that :)

That is great Allan, lets try to get this merged today if everything goes smoothly. I will start working on the changes necessary as per your comments right away. In terms of the metadata version name, summary etc, I revert back to the orginal function for getting those now that I am aware of the issues.

Ira

src/rez/pip.py
lambdaclan: Never knew that normpath will normalize a pathname by collapsing **all** redundant separators and up-level references. I always thought it was limited to dots and slashes, awesome. :+1:

src/rez/pip.py
lambdaclan: Agreed, I was overly verbose aplogies. I will remove the implementation details and reword the comment a bit.

src/rez/pip.py
lambdaclan: Yes indeed, since we only care about that path no reason to loop. It always seems to be under the same location so no need to recurse the directory for other bin folders. That is what I had in mind initially.

lambdaclan:
Tests:

## pip < 19

* ignore the info showing valid pip version, I have harded the check to 20 to make the error happen for the test

pip


## PyYAML

pyyaml

## psutil

psutil

Review round 3. Only thing left is the version conversion I believe. Can you provide an example of a version that will not work via distribution.version? I can seem to find any libraries with exotic versions to try...

Something like 1!1.0 (epoch plus case insensitive?) and what it should turn into.

I am asure we can use this to extract and convert it to a rez compatible version --> https://www.python.org/dev/peps/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions

lambdaclan:
Using the REGEX mentioned above, we can easilly retrieve all the parts of a pip version and modify it accordingly.

`rez-pip -i pydocstyle`

.....

within the pip_to_rez_version function

parse segments and construct rez-compatible version
....

Sample output:

```
configparser
version_segements:  {'pre': None, 'dev_l': None, 'dev_n': None, 'post_l': None, 'pre_n': None, 'pre_l': None, 'dev': None, 'epoch': None, 'post_n1': None, 'post_n2': None, 'release': u'3.7.4', 'post': None, 'local': None}
pydocstyle
version_segements:  {'pre': None, 'dev_l': None, 'dev_n': None, 'post_l': None, 'pre_n': None, 'pre_l': None, 'dev': None, 'epoch': None, 'post_n1': None, 'post_n2': None, 'release': u'3.0.0', 'post': None, 'local': None}
six
version_segements:  {'pre': None, 'dev_l': None, 'dev_n': None, 'post_l': None, 'pre_n': None, 'pre_l': None, 'dev': None, 'epoch': None, 'post_n1': None, 'post_n2': None, 'release': u'1.12.0', 'post': None, 'local': None}
snowballstemmer
version_segements:  {'pre': None, 'dev_l': None, 'dev_n': None, 'post_l': None, 'pre_n': None, 'pre_l': None, 'dev': None, 'epoch': None, 'post_n1': None, 'post_n2': None, 'release': u'1.2.1', 'post': None, 'local': None}
```

What I need to know is some sort of a ruleset, for example if only release key is set then keep as it is if it has epoch etc then remove it etc.

nerdvegas:
Looks good, only some minor changes. My testing worked, let's deal with any further issues in separate PR. Thanks for the hard work!

lambdaclan:
Wow, that was fast! Sounds good Allan. Please let me know of anything that needs to be changed and I will take care of it. I will focus on #628 next.

PR #608, opened 2019-04-22
Support iterating over `request` in package.py

mottosso:
This enables the `request` variable of `package.py` to be iterated over.

```python
def commands():
  for req in request:
    print("%s was requested!" % req)
```

### Problem

With support for `if "maya" in request`, I expected the object to be iterable, but was instead greeted by a cryptic traceback.

```bash
AttributeError: request does not exist: '0'
```

Especially considering errors during `commands()` and anything with `@early` or `@late` in it doesn't provide accurate line numbers (as they are dynamically concatenated and `exec`'ed), there was little way of telling what actually caused this to happen.

In practice, I found myself doing e.g.:

```python
environ = {
  "any": {
    "PYTHONPATH": ["{root}/python"],
  },
  "maya": {
    "MAYA_PLUG_IN_PATH": ["{root}/maya/plugins"],
  }
}

def commands():
  environ = this.environ["any"].copy()

  for req in request:
    items.update(this.environ.get(req, {}))

  for key, values in items.items():
      for value in values:
        env[key].prepend(value)
```

An alternative (working) solution was this.

```python
for key, values in this.environ.items():
  if key not in request:
    continue

  result.update(values)
```

Which works, but isn't as nice to look at as the above.

For symmetry, I also added iteration support to the `resolve` object.

### DIFF
--- a/src/rez/rex_bindings.py
+++ b/src/rez/rex_bindings.py
@@ -140,6 +140,11 @@ class VariantsBinding(Binding):
     def __contains__(self, name):
         return (name in self.__variants)
 
+    def __iter__(self):
+        """Support for `for res in resolve`"""
+        for req in self.__variants:
+            yield req
+
 
 class RequirementsBinding(Binding):
     """Binds a list of version.Requirement objects."""
--- a/src/rez/rex_bindings.py
+++ b/src/rez/rex_bindings.py
@@ -162,7 +162,13 @@ class RequirementsBinding(Binding):
             self._attr_error(name)
 
     def __contains__(self, name):
+        """Support for `if req in request`"""
         return (name in self.__requirements)
+
+    def __iter__(self):
+        """Support for `for req in request`"""
+        for req in self.__requirements:
+            yield req
 
 
 # Copyright 2013-2016 Allan Johns.


JeanChristopheMorinPerso:
Hey @mottosso , that's a really great adition. I always wanted to add it be never had time.

Would you mind adding tests for that please? See in https://github.com/nerdvegas/rez/blob/master/src/rez/tests/test_rex.py. This is as per https://github.com/nerdvegas/rez/blob/master/CONTRIBUTING.md#contributing-to-rez.

The information on how to run the tests, see https://github.com/nerdvegas/rez/blob/master/src/rez/tests/README.

nerdvegas:
Yes test please, especially for any additions to the rex side of things.

Thx
A

mottosso:
Having a look at this, but am having trouble figuring out how to gain access to the `request` global.

Here's what I've tried adding to `test_rex.py`

```python

    def test_iterate_over_request(self):
        """Iterating over request global"""

        def _rex():
            for req in request:
                print(req)

        self._test(func=_rex,
                   env={},
                   expected_actions=[],
                   expected_output={})
```

But am getting:

```bash
marcus@toy:$ rez selftest --rex
ERROR: test_iterate_over_request (rez.tests.test_rex.TestRex)
Iterating over request global
  File "/home/marcus/rez/local/lib/python2.7/site-packages/rez/tests/test_rex.py", line 413, in test_iterate_over_request
    request
NameError: global name 'request' is not defined

Ran 12 tests in 0.045s

FAILED (errors=1)
```

nerdvegas:
I'll take a look at this. But something that is apparent right away is that
request, as defined in the docs, is *dict-like*. Therefore I would expect
this instead:

    for pkg_name, req in request.iteritems():
        print "Pkg %s is requested as %s" % (pkg_name, req)

I'm also not sure of a case where you'd want to do this (generally a
package just sets the evars appropriate to itself - eg maya would set
`env.MAYA_PLUGIN_IN_PATH=...`). But that doesn't invalidate this PR either,
so I'll take in on board.

Thx
A

nerdvegas:
*ping*!

I'm aware of this PR still, but just pointing out that it needs work so it'll take me longer to get to it. Specifically it needs a test, and it needs to be dict-like in behaviour as per previous comment. The test will have to be updated so that a `request` object is made available.

Issue #612, opened 2019-04-24
Make rez pip useful


mottosso:
### Goal

Make pip great again. Or rather, make pip actually useful.

### Motivation

`rez pip --install` currently creates a Rez package from a Python package on PyPI, which is great. However it also embeds overly specific requirements for the package.

Consider the following example.

```bash
$ rez pip --install Qt.py
```

The resulting package is..

```
~/packages/Qt.py/1.1.0/platform-windows/arch-AMD64/os-windows-10.0.17763/python-3.6/
```

Meaning that in order to use this package, you'll need:

- A Windows OS
- Of the exact version `10.0.17763`
- Python 3.6
- A 64-bit machine

Even though Qt.py works on..

- Any OS
- Any version of Windows (and Linux, and MacOS)
- Any arch
- Any Python

The same is true for any package, including compiled ones. A compiled PyQt5 would require an exact arch, an exact platform and Python distribution, but not the exact version of OS.

Some packages require Python 2 or 3, but not necessarily the minor version; thus `python-3` is sufficient to support all 3.x versions.

### Proposal

Ditch `rez pip` as it exists today, and make this the new default.

```bash
$ rez pip --install Qt.py
# Installed using wheels, as per #602
```

Resulting in:

```
~/packages/Qt.py/1.1.0/package.py
```

With options for:

```bash
$ rez pip --install Qt.py --variant platform-windows --variant python-3.6
```

For wheels specific to a particular OS or interpreter, like PySide, it could then look towards a provided `--variant python-X` to determine which Python to build/download for.

That's assuming you, the one installing a package from pip, knows full well what a package requires and aren't dependent on the information being inferred from its metadata.

However, for when you don't and the information is there:

```bash
$ rez pip --install UnfamiliarPackage --auto-variants
```

Which could look at whatever sources is available to a package on pip to figure out which variants to package it under.

Thoughts?

j0yu:
We haven't addressed them yet due to our current size and complexity, but those are great points and issues that I definitely support.

There was a point where I noticed it could be an issue e.g. upgrading from even just `centos-7.4` to `centos-7.6` (testing in docker container first) , none of the pypi packages we have will show up. I have thought about it then that thought went down the rabbit hole of potentially even having (amongst other things):
- auto `requirements.txt` or `pipenv` scanning to determine any requirements or variants
- custom package (re)name that's compatible with rez studio's naming conventions e.g. `Qt.py` released as `qt_py`

[...]

Issue #610, opened 2019-04-23
Overly specific implicit Windows `os` package


mottosso:
### Problem

Windows gone did an update overnight, and suddenly all of my packages stopped resolving, saying:

```bash
$ rez env python
The context failed to resolve:
The following package conflicts occurred: (os-windows-10.0.17134 <--!--> ~os==windows-10.0.17763)
```

And after staring myself blind at this seemingly contradictory line, I spotted the last 3 digits have increased since yesterday.

Windows isn't good for much, but it's really good at backwards compatibility. It would be safe to strip those last 3 digits I think. Heck, it'd be safe to strip almost every digit, as anything working on Windows XP is likely still compatible with Windows 10.

On Linux, the implicit packages are specific enough I think.

```bash
~platform==linux  (implicit)
~arch==x86_64     (implicit)
~os==Debian-9.7   (implicit)
```

Here's what they look like on Windows.

```bash
requested packages:
~platform==windows       (implicit)
~arch==AMD64             (implicit)
~os==windows-10.0.17763  (implicit)
```

### Solution

Would recommend making this level of specificity opt-in, to avoid newcomers from falling into this trap. How about a default of 2 digits, e.g. `windows-10.0`?

instinct-vfx:
First: I do agree that the os is overly specific. That being said i usually work around this by not requiring os and not putting it into the variants. I am usually sticking to platform and arch only. The implicit packages are weak references so they are only matched IF the package is required by your package. So that conflict should not arise unless your python package has a very specific os requirement, no? Is this python package created using rez-bind?

Depending on your requirements you could also set a platform map in your config: https://github.com/nerdvegas/rez/wiki/Configuring-Rez#platform_map

There are quite some differences for Win7->Win10 btw. Nothing major but i have not really gotten the hang of their new "versioning" :(

smaragden:
We are using a platform_map in rezconfig t osolve this exact issue.
This is for CentOS but you can do the same for windows
```python
platform_map = {
    "os":{
        r'CentOS-(\d+)\.(\d+)(\.(\d+))?': r'CentOS-\1',
    }
}
```
This will change **CentOS-7.6.1810** to **CentOS-7**

mottosso:
> That being said i usually work around this by not requiring os and not putting it into the variants

[...]

PR #598, opened 2019-04-18
WIP Fix file permissions of package.py on Windows

mottosso:
This hasn't been confirmed yet; on some machines, this hasn't been an issue, but on mine (Windows 10, non-privileged user, local disk) it was a showstopper.

The permission bits here [aren't supported](https://docs.python.org/2/library/os.html#os.chmod) on Windows which shouldn't be a problem, as they should fall back to the only two bits that are, `IWRITE` and `IREAD`. Here however, they result in read-only file permissions that cannot be overwritten and breaks building of variants amongst others

```bash
$ rez build --install
== Anima Rez Config ==
16:50:44 INFO     Applying preprocess function package_preprocess_function
16:50:44 INFO     Package attributes were changed in preprocessing:
Added attributes: ['config']
16:50:44 INFO     Applying preprocess function package_preprocess_function
16:50:44 INFO     Package attributes were changed in preprocessing:
Added attributes: ['config']

--------------------------------------------------------------------------------
Building ATC-0.3.15...
--------------------------------------------------------------------------------

--------------------------------------------------------------------------------
Building variant 0 (1/2)...
--------------------------------------------------------------------------------
16:50:44 INFO     Applying preprocess function package_preprocess_function
16:50:44 INFO     Package attributes were changed in preprocessing:
Added attributes: ['config']
Resolving build environment: globals-0.0.5+ core_pipeline-1.2+ maya-2018
resolved by manima@toy.fritz.box, on Thu Apr 18 16:50:44 2019, using Rez v2.29.0

requested packages:
globals-0.0.5+
core_pipeline-1.2+
maya-2018
~platform==windows       (implicit)
~arch==AMD64             (implicit)
~os==windows-10.0.17134  (implicit)

resolved packages:
Qt.py-1.1.0            C:\Users\manima\Dropbox\dev\anima\.cache\packages\ext\Qt.py\1.1.0\platform-windows\arch-AMD64\os-windows-10.0.17134\python-2.7
arch-AMD64             C:\Users\manima\packages\arch\AMD64                                                                                             (local)
core_pipeline-1.2.0    C:\Users\manima\Dropbox\dev\anima\.cache\packages\int\core_pipeline\1.2.0
globals-0.0.6          C:\Users\manima\packages\int\globals\0.0.6
maya-2018.0.0          C:\Users\manima\packages\ext\maya\2018.0.0
maya_base-1.0.7        C:\Users\manima\packages\ext\maya_base\1.0.7
[...]

PR #614, opened 2019-04-24
WIP Implements #612, Useful pip

mottosso:
A first draft of #612.

```bash
$ rez-pip --install six  # no variant
$ rez-pip --install urllib3 --variant python-2.7
$ rez-pip --install PySide --variant platform-windows --variant arch-AMD64 --variant python-2.7
```

Also works in conjunction with `--python-version` in case you've got multiple.

```bash
$ rez-pip --install urllib3 --variant python-2.7 --python-version 2.7
```

Also works (better) with #599 or #602. This makes no assumptions about the variant being installed. 

Testing this out, not yet ready to merge.


mottosso:
That last commit is due to [this line](https://github.com/nerdvegas/rez/blob/3d251f5af6307a9f63dc1ac5d7bc4da9a18419e7/src/rez/cli/pip.py#L58) which in the case of not passing `--variant` throws this exception, rather than evaluating to `False`.

mottosso:
Discovered an issue with this approach, and the `rez pip` approach in general, which has to do with dependencies of the `pip` package.

E.g. `packageA` is pure-Python, but depends on `PyQt5`. So when calling:

```bash
$ rez pip --install packageA
```

You'll end up with both `packageA` and `PyQt5` in a variant-less directory. Likewise, if you call:

```bash
$ rez pip --install packageA --variant platform-windows --python-3.6
```

You'll end up with PyQt5 in the right place, but `packageA` overly specified.

### Solution

For that, I would probably do:

1. `rez pip --install somePackage`
2. If `somePackage` have no dependencies, or all dependencies are already installed, proceed
3. Else error out, with a message detailing which dependencies need installing first

That way, installing `packageA` becomes a two-step process.

```bash
$ rez pip --install packageA
# ...
# ERROR: Must install PyQt5 first
$ rez pip --install PyQt5 --variant platform-windows --variant python-3.6
$ rez pip --install packageA
# Success
```

More tedious, but unavoidable I think, without a lot of assumptions.

Thoughts?

mottosso:
Another issue with `rez pip` in general that I'll attempt to solve in this PR.

- If the package already exists relative the Python `site-package/`, then `rez pip` won't make a package out of it.

For example, if `python` isn't a package when calling `rez pip`, then Rez will pick up the system-wide Python and carry on. If that system Python then already has e.g. `mkdocs` in its `site-packages/`, then a helpful (but hidden) message will kindly tell you how it's skipped.

```bash
rez pip --install mkdocs==1.0.4
13:44:49 INFO     Using python-3.6 (C:\Users\manima\Dropbox\dev\anima\github\mottosso\rez-for-projects\local_packages_path\python\3.6\package.py[])
c:\python36\lib\site-packages\pip\_internal\commands\install.py:244: UserWarning: Disabling all use of wheels due to the use of --build-options / --global-options / --install-options.
  cmdoptions.check_install_build_global(options)
Requirement already satisfied: mkdocs==1.0.4 in c:\python36\lib\site-packages (1.0.4)
Requirement already satisfied: PyYAML>=3.10 in c:\python36\lib\site-packages (from mkdocs==1.0.4) (3.13)
Requirement already satisfied: tornado>=5.0 in c:\python36\lib\site-packages (from mkdocs==1.0.4) (5.1.1)
```

As a result, the command will finish successfully, but not have made the packages you expected it to.

mottosso:
I've included #599 into this PR as it didn't make much sense without it.

I've now also addressed the above issues.

1. **Auto-variants** Dependencies having variants of their own, via variant auto-detection. The `classifiers` section of each PyPI package is parsed for clues on whether a distribution requires Python 2 or 3. I'd also like for it to figure out whether the platform is important, i.e. whether a library is binary or not
2. **Ignore Existing** Package already installed to `site-package`, now those are simply ignored.

The solution for (1) was not sufficient however; as some packages may support both Python 2 and 3, *but* have separate distributions for each. As such, the `classifiers` wouldn't be enough. What should happen is that it should look towards the "python_tag" of the bdist_wheel for an exact match.

For example `PyYAML-5.1-cp27-cp27m-win32.whl` means `python-2`.

- https://pypi.org/project/PyYAML/#files
- https://www.python.org/dev/peps/pep-0427/#file-name-convention

mottosso:
Magic, the wheel contains information on whether a package is binary and needs the `platform-` variant, *and* the Python version for the `python-` variant.

All is looking well, looking out for trouble now..

mottosso:
The automatic detection of Python version for a given package works great, except it currently only includes the major version, e.g. `python-2` or `python-3`, whereas some packages depend on the minor version as well, such as PySide which requires `python-2.6` or `python-2.7`.

Probably safe to always using the minor version too, which is already available [here](https://github.com/nerdvegas/rez/pull/614/commits/f3974a252a7eed80610571ed9cad34dca8997c90#diff-9d98373f96aec3a65234c137885661ccR346) but simply not used.

mottosso:
With a few weeks of use, requirements are clearing up.

| Package                      | Properties | Description | Example
|:-------------------------|:------------|:----------|:------
| Qt.py | pure python, any version | Should install without variant | `~/packages/Qt.py/1.0.0`
| PyYAML | pure python | Should install with major version variant | `~/packages/PyYAML/1.0.0/python-2`
| PySide | compiled | Should install with platform and major.minor version variant | `~/packages/PySide/1.0.0/platform-windows/python-2.7`

Need to figure out whether and what to include for `os`, as packages built for Ubuntu may not work for OpenBSD, but anything built for ubuntu.18.0.13151 will run on ubuntu.18.0.13151; so a balance of sorts.

I'd also like for the `--platform` option to pip to be made available, so that you can install Linux packages from Windows and vice versa; something only possible for wheels, which isn't everything but every little helps.

```bash
$ rez pip --platform linux --install PySide2
$ rez pip --platform windows --install PySide2
```

Also with this PR you can pass multiple packages to `--install`, but you can't continue passing arguments which is unexpected.

```bash
$ rez pip --install rez --python-version 3
# "--python-version" and "3" are considered pip packages
```

What's missing at the moment is those two, and the `minor` version for compiled packages, resulting in PySide being installed for `python-2`, despite working for `2.7` specifically.

mottosso:
rez-wheel

### Changes

- `pip wheel --install`
- Collaborative `rez wheel`
- Installation preview
- CLI and API separation
- Automatically detect variants per default
- Python and pip version preview
- Lowercase package names
- Multiple versions, and subsequent arguments
- User friendly tracebacks
- Minor version included for compiled packages
- Major/Minor `os` variant
- Tests

<br>

### pip wheel

Following [this conversation](https://github.com/nerdvegas/rez/pull/602#issuecomment-495959765), to preserve backwards compatiblity with `rez pip` I've moved all functionality to `pip wheel`.

```bash
$ rez wheel --install six mkdocs requests[security]
```

I figured this made more sense than `rez pip2` considering `wheel` is the flavour of the month at the moment, preceded by "egg", and would avoid resorting to `rez pip`, `rez pip2`, `rez pip3` and so forth once PyPI announces the next great packaging method of the future.

<br>

### Collaborative `rez wheel`

Now `rez wheel` (optionally) uses `rez env` to determine the version of Python and pip.

```bash
$ rez env python2.7 pip-19 -- rez wheel install six
```

Without it, it refers to whichever Python and pip is available (if any) on the users PATH/PYTHONPATH/site-packages.

```bash
$ rez wheel --install six
```

<br>

### Installation Preview

Now packages are previewed prior to being installed, akin to `apt-get install`, including:

- Total size, e.g. `13.2 mb`
- Destination repository, e.g. `~/packages`

```bash
$ rez wheel --install mkdocs
Using Python 2.7.13
Using pip 19.1 from /home/marcus/rez/local/lib/python2.7/site-packages/pip (python 2.7)
Reading package lists... ok - 14.51s
Discovering existing packages... ok - 0.27s
The following NEW packages will be installed:
  markupsafe        1.1.1    platform-linux/os-Debian-9.7/python-2.7
  backports_abc     0.5
  livereload        2.6.1
  pyyaml            5.1      platform-linux/os-Debian-9.7/python-2.7
  futures           3.2.0    python-2
  setuptools        41.0.1
  mkdocs            1.0.4
  singledispatch    3.4.0.3
  six               1.12.0
  tornado           5.1.1    platform-linux/os-Debian-9.7/python-2.7
  click             7.0
  jinja2            2.10.1
  markdown          3.1.1
Packages will be installed to /home/marcus/packages
After this operation, 12.46 mb will be used.
Do you want to continue? [Y/n]
(1/13) Installing markupsafe-1.1.1... ok - 0.05s
(2/13) Installing backports_abc-0.5... ok - 0.04s
(3/13) Installing livereload-2.6.1... ok - 0.03s
(4/13) Installing pyyaml-5.1... ok - 0.04s
(5/13) Installing futures-3.2.0... ok - 0.03s
(6/13) Installing setuptools-41.0.1... ok - 0.04s
(7/13) Installing mkdocs-1.0.4... ok - 0.03s
(8/13) Installing singledispatch-3.4.0.3... ok - 0.03s
(9/13) Installing six-1.12.0... ok - 0.03s
(10/13) Installing tornado-5.1.1... ok - 0.03s
(11/13) Installing click-7.0... ok - 0.03s
(12/13) Installing jinja2-2.10.1... ok - 0.03s
(13/13) Installing markdown-3.1.1... ok - 0.03s
13 installed, 0 skipped
Cleaning up temporary files.. ok - 0.91s
Success
```

This also handles and previews existing packages.

```bash
$ rez wheel --install mkdocs
Using Python 2.7.13
Using pip 19.1 from /home/marcus/rez/local/lib/python2.7/site-packages/pip (python 2.7)
Reading package lists... ok - 13.26s
Discovering existing packages... ok - 0.29s
The following NEW packages will be installed:
  mkdocs            1.0.4
The following packages will be SKIPPED:
  markupsafe        1.1.1
  backports_abc     0.5
  livereload        2.6.1
  pyyaml            5.1
  futures           3.2.0
  setuptools        41.0.1
  singledispatch    3.4.0.3
  six               1.12.0
  tornado           5.1.1
  click             7.0
  jinja2            2.10.1
  markdown          3.1.1
Packages will be installed to /home/marcus/packages
After this operation, 12.46 mb will be used.
Do you want to continue? [Y/n]
```

Since each "stage", such as querying PyPI, can take between 1-20 seconds, timing is included to give the user some idea of how long future invocations are likely to take.

If no packages were installed, no input is required.

```bash
$ rez wheel --install six
Using Python 2.7.13
Using pip 19.1 from /home/marcus/rez/local/lib/python2.7/site-packages/pip (python 2.7)
Reading package lists... ok - 1.90s
Discovering existing packages... ok - 0.09s
six-1.12.0 was already installed
No new packages were installed
Cleaning up temporary files.. ok - 0.02s
Success
```

<br>

### Python and pip version preview

Before, it wasn't clear which version of Python and pip was being used to install packages, resulting in the user having to remove packages afterwards and retry.

Now the versions are previewed before installation begins, via `python --version` and `pip --version`.

```bash
$ rez wheel --install mkdocs
Using Python 2.7.13
Using pip 19.1 from /home/marcus/rez/local/lib/python2.7/site-packages/pip (python 2.7)
```

<br>

### CLI and API separation

Due to the preview-ability, a few more features were added to `pip.py`, and as a result I've separated those features into something usable from Python, and wrapped those into a CLI with proper messages, handling and user-input.

```bash
$ rez wheel --install mkdocs
```

```python
from rez import pip
pip.install("mkdocs")
```

Or the more fine-grained.

```python
from rez import pip
distributions = pip.download(["mkdocs"]) # includes lots of dependencies

for dist in distributions:
    package = pip.convert(dist)

    if not pip.exists(package):
        pip.deploy(dist, package)
```

<br>

### Major/Minor `os` variant

Variants use a more forgiving version of `os`.

- `os-windows-10.0.18362` ---> `os-windows-10.0`
- `os-Debian-9.7.5` ---> `os-Debian-9.7`

This is the maximum amount of granularity provided by a pip WHEEL and should help counter #610 too.

<br>

### Lowercase package names

Previously, the original PyPI package name was used as a Rez package name, e.g. `PySide`. For consistency, and since Rez is case-insensitive, names are converted to lowercase, e.g. `pyside`.

- PySide -> pyside
- Qt.py -> qt.py

<br>

### User friendly traceback

Before, all `pip` output spammed the terminal. Now those are hidden unless an exception occur in which case the exception value is printed.

```bash
$ rez wheel --install pip==19.0.2 pip==19.0.3
Using Python 2.7.13
Using pip 19.1 from /home/marcus/rez/local/lib/python2.7/site-packages/pip (python 2.7)
Reading package lists...
ERROR ----------------------------------------------------------------

ERROR: Double requirement given: pip==19.0.3 (already in pip==19.0.2, name='pip')

ok - 0.75s
Discovering existing packages... ok - 0.00s
No new packages were installed
Cleaning up temporary files.. ok - 0.00s
Success
```

Where the part in-between `---` is the `traceback.value`.

<br>

### Automatically detect variants per default

Before, variants were automatically detected using the `--auto-variants` flag. Since it's worked so well, it's now default and can be overridden via the `--variant` argument. See `--help` for details.

<br>

### Minor version and compiled packages

Before, compiled packages were given a variant of e.g. `python-2` but actually depend on the minor version as well.

<br>

### Multiple versions, and subsequent arguments

Before, you could pass multiple packages to `pip`, but couldn't follow up with additional arguments as one would expect, now you can.

```bash
$ rez wheel --install six mkdocs --variant python-3.6 --no-deps
```

<br>

### Tests

This PR now includes tests specific to pip.

Since installing anything via pip is a lengthy process (1-10 seconds), the total time of running tests have increased by 90 seconds (!).

- See [`test_wheel.py`](https://github.com/mottosso/bleeding-rez/blob/feature/useful-pip/src/rez/tests/test_wheel.py)

<br>

### Next Steps

This is using pip via the command-line (like before), rather than via the `pip.py` "API" which could provide more fine-grained control, and improve performance by not invoking shells. But I think we need to stick to using the shell, so as to enable Rez to provide an environment for it with a Python that in turn is used to call on `pip.

`--platform` isn't implemented yet, I'm not sure how to approach this, since we're now relying on pip to handle what platform and OS a wheel is coming from, and compare this against the `rez.utils.platform_.platform_` data for a Rez-compatible name. If we override `--platform` for pip, it would no longer match the current system and hence require explicit handling of the platform tags from  a WHEEL. The problem with that is that there isn't a standard or limit to the number of tags available; it could be just about anything.

src/rez/tests/test_wheel.py
```diff
@@ -0,0 +1,258 @@
...
+    def test_compiled(self):
+        """Install a compiled Python package"""
+        self._test_install("pyyaml", "5.1")
+
+    # def test_nowheel(self):
+    #     self._install("PySide==1.2.4")
+
+    def test_dependencies(self):
+        """Install mkdocs, which carries lots of dependencies"""
+        installed = self._install("mkdocs==1.0.4")
```
JeanChristopheMorinPerso: You should be careful with this. `mkdocs` sub-depedencies (transitive) could change and the tests will break. Like here: https://github.com/mkdocs/mkdocs/blob/master/setup.py#L59. A new version of click would make the tests fail.

Also, I don't think we should fetch the internet in tests. It's normally error prone and it brings instability to the tests. One could want to develop rez withuot internet. Developers in big studios don't necessarily have access to internet from a shell, or someone could want to write/run some tests while having no internet.

src/rez/wheel.py
```diff
@@ -0,0 +1,527 @@
...
+
+        # Handle case where the Python distribution used alongside
+        # pip already has a package installed in its `site-packages/` dir.
+        "--ignore-installed",
+
+        # rez pip users don't have to see this
+        "--disable-pip-version-check",
+
+        # Prevent user-settings from interfering with install
+        "--isolated",
```
JeanChristopheMorinPerso: Some studios don't have internet and uses a global config on machines or environments to points to a local PyPI index. This option will break this. What was the reasoing being using `--isolated`?

src/rez/wheel.py
```diff
@@ -0,0 +1,527 @@
...
+
+
+def _get_dependencies(requirement):
+
+    requirements = ([requirement] if isinstance(requirement, basestring)
+                    else requirement["requires"])
+
+    result = []
+    for package in requirements:
+        if "(" in package:
```
JeanChristopheMorinPerso: The `packaging` package should help you here I think.

src/rez/tests/test_wheel.py
mottosso: Hey jc, thanks for having a look!

> mkdocs sub-depedencies (transitive) could change and the tests will break

Ah, yes that's true. I had assumed that by picking a fixed version of mkdocs, we'd also be getting fixed versions of its dependencies, but that's not the case..

> Also, I don't think we should fetch the internet in tests.

This could probably be managed via the `rez selftest --skip wheel` argument. Alternatively, when `cmake` isn't available on a system, the test is skipped. Maybe the same could apply if the user doesn't have internet.

src/rez/wheel.py
mottosso: Oh. The reason was to reduce the room for error, and the number of tests we'd need to write. I figure it wouldn't hurt, but I did not consider this usecase. Will remove it.

src/rez/wheel.py
```diff
@@ -0,0 +1,527 @@
...
+
+    tempdir = tempdir or os.getcwd()
+
+    # Build pip commandline
+    cmd = [
+        "python", "-m", "pip", "install",
+        "--target", tempdir,
+
+        # Only ever consider wheels, anything else is ancient
+        "--use-pep517",
```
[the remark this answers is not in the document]
mottosso: Yes, it's from pip-19 and above. Should make this a minimum required version for using `rez wheel`.

src/rez/wheel.py
```diff
@@ -0,0 +1,527 @@
...
+
+    popen = subprocess.Popen(cmd,
+                             stdout=subprocess.PIPE,
+                             stderr=subprocess.STDOUT,
+                             universal_newlines=True)
+
+    output = []
+    for line in iter(popen.stdout.readline, ""):
+
+        if line.startswith("DEPRECATION"):
```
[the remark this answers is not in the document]
mottosso: In the context of `rez wheel`, that's a `pip` implementation detail. But more importantly, there isn't anything the `rez wheel` user can do about it. :( At the moment, all it does is distract from the actual problem the package author or user needs to see, such as whether a pip version exists or whether there isn't any internet.

src/rez/wheel.py
```diff
@@ -0,0 +1,527 @@
...
+            # Mute warnings about Python 2 being deprecated.
+            # It's out-of-band for the casual Rez user.
+            continue
+
+        output.append(line.rstrip())
+
+    popen.wait()
```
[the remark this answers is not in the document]
mottosso: Hm, that isn't a recommendation though, but a warning. We're emptying the buffer in the block above, so that could never happen. We're in the clear.

src/rez/wheel.py
```diff
@@ -0,0 +1,527 @@
...
+
+    Arguments:
+        distribution (distlib.database.InstalledDistribution): Source
+        variants (list, optional): Explicitly provide variants, defaults
+            to automatically detecting the correct variants using the
+            WHEEL metadata of `distribution`.
+
+    """
+
+    name, _ = parse_name_and_version(distribution.name_and_version)
```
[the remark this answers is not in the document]
mottosso: This is a remnants of the current `pip.py`, and I wasn't sure enough about what it did to remove or reimplement it. Good to know there's an official solution for it.

src/rez/wheel.py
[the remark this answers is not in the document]
JeanChristopheMorinPerso: Still, we should avoid filtering out deprecation warnings. Because one could be not related to python 2.7 end of support.

>In the context of rez wheel, that's a pip implementation detail.

I don't really agree. It's far from an implementation detail for a developer. A developer should be able to live with deprecation warnings and other warnings in general. There are there for a reason. Yes they are annoying, but they are important. And I also expect someone using `rez-pip` to know how to use pip and so know what kind of message he is suppose to see (since he will get the warnings anyway without rez).

Also, I think the PyPA org was planning to let PyPI return warnings so that pip, pipenv, and others would receive some warnings to pass to the user. That way, the python community is able to pass their messages without having people update their tools. These messages are important.

mottosso:
That last test involved writing to an existing `package.py`, which doesn't work on Windows without feature/windows-fileperm #598 so I included that here.

mottosso:
I've added a `--index-url` option to account for studios using a custom index, and are also checking for a minimum required `pip` version, to avoid unexpected error.

```bash
$ rez env python-2 pip-18 -- rez wheel --install six
ERROR: Requires pip>=19
```

mottosso:
Restored upper/lowercase of package names, so now PySide remains PySide etc.

1. The next thing I'd like to address is the `_get_dependencies()` function which is a big mess.
2. I'm also not too fond of how the `list_installed_files` is able to list files not actually installed; such as those in `bin/` which aren't included with when pip is used with `--target`. But also, it seems based on what the author has put in his `setup.py` which can differ from what is actually installed. In the case of Qt.py, I found that a `LICENCE` file was included there, but not actually on disk. We could list strictly installed files ourselves, since they're right there on disk. That would be safest perhaps, but may include *more* than one asked for. Having said that, if more files are included in a wheel than what is actually documented, it's not our problem and can be resolved upstream.

mottosso:
Spotted two edgecases.

### 1. Python version mismatch

```bash
rez env python-3 pip-19 -- rez wheel --install PyQt5==5.7.1                                                                         
Using python-3.6                                                                                                                      
Using pip-19.1.1                                                                                                                      
Reading package lists... ok - 29.67s                                                                                                  
Discovering existing packages... ok - 0.02s                                                                                           
The following NEW packages will be installed:                                                                                         
  PyQt5    5.7.1   platform-windows/os-windows-10.0/python-3.4                                                                        
  sip      4.19.8  platform-windows/os-windows-10.0/python-3.6                                                                        
Packages will be installed to C:\Users\manima\packages                                                                                
After this operation, 182.53 mb will be used.   
```

Here, PyQt5 is classified as being supported for 3.4, whereas the Python I'm using is 3.6. This would result in a package that, as far as Rez is concerned, isn't compatible with the Python it's being installed for.

This should ideally create a package compatible with `>=3.4,<=3.6` but we can't create a variant like this. We also can't make the variant `python-3`, because the next version of PyQt5 may well come classified as supporting 3.7, and we wouldn't want 3.3 to pick this up.

So instead I think it'll need to discard the version coming from the WHEEL, and instead use the version of Python used to install it with.

### 2. Hacked package

```bash
rez env python-3 pip-19 -- rez wheel --install PyQt5                                                                                
Using python-3.6                                                                                                                      
Using pip-19.1.1                                                                                                                      
Reading package lists... ok - 30.20s                                                                                                  
Discovering existing packages... ok - 0.02s                                                                                           
The following NEW packages will be installed:                                                                                         
  PyQt5        5.12.2   platform-windows/os-windows-10.0/python-3.5                                                                   
  PyQt5_sip    4.19.17  platform-windows/os-windows-10.0/python-3.6                                                                   
Packages will be installed to C:\Users\manima\packages                                                                                
After this operation, 125.47 mb will be used.                                                                                         
Do you want to continue? [Y/n] 
```

PyQt5 has started distributing `sip` not as a separate Python module or package, but as something that installs itself into an existing PyQt5 install. This doesn't work well with Rez and requires some manual tweaking (countering the manual tweaking done to create this distribution in the first place).

Once installation is complete, PyQt5 expects to find `PyQt5.sip`, but Rez has instead gone ahead and made:

```
~/packages/PyQt5/5.12.2
~/packages/PyQt5_sip/4.9.17
```

This could *potentially* be resolved by addressing point (2) in the previous post; of gathering files straight from disk rather than reading from the distribution. The distribution is lying in this case, as files will have been moved around on disk post-install to accommodate for this esoteric package behaviour.

JeanChristopheMorinPerso:
After thinking about this, I think `rez-wheel` is not a good name. The wheel format should be an implementation detail, not a first class citizen. Pip is the first class citizen that we need to expose, not wheel.

>Having said that, if more files are included in a wheel than what is actually documented, it's not our problem and can be resolved upstream.

This is the way to go. We should never try to be smarter than the package maintainer. If you are unhappy with which files appears in wheel, then you need to contribute to the project and propose a change for removing those said files from the wheel.

src/rez/wheel.py
```diff
@@ -0,0 +1,570 @@
...
+        # requirements of its own, and variants to go with it.
+        cmd += ["--no-deps"]
+
+    cmd += names
+
+    popen = subprocess.Popen(cmd,
+                             stdout=subprocess.PIPE,
+                             stderr=subprocess.STDOUT,
+                             universal_newlines=True,
+                             shell=True)
```
JeanChristopheMorinPerso: Any reason for using `shell=True` here?

src/rez/wheel.py
```diff
@@ -0,0 +1,570 @@
...
+    # Try system Python
+    popen = subprocess.Popen(
+        """\
+        python -c "import sys;print('.'.join(map(str, sys.version_info[:2])))"
+        """,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        universal_newlines=True,
+        bufsize=10 ** 4,  # Enough to capture the version
+        shell=True,
```
JeanChristopheMorinPerso: Same here, any reason to use `shell=True`?

src/rez/wheel.py
```diff
@@ -0,0 +1,570 @@
...
+def platform_name():
+    return platform_.name
+
+
+@lru_cache()
+def python_version():
+    """Return major.minor version of Python, prefer current context"""
+
+    import subprocess
+    from rez.status import status
```
JeanChristopheMorinPerso: subprocess is already imported at the top of the file. Also, could `rez.status` be moved at the top of the file too?

src/rez/wheel.py
mottosso: Yes, in both cases, `shell=True` enables `rez wheel` to be prefixed with `rez env -- rez wheel`. Without it, it wouldn't be able to see the Python exposed by `rez env`.

mottosso:
Should add `python` into the `requires` of each pip package, such that you can ask for just e.g. six and get Python too.

**six/package.py**

```python
name = "six"
requires = ["python"]
"""

Question about that, how does this work with variants? I.e. some packages have Python as a variant, do we still add `python` to `requires`? For example:

- PyQt5 is installed, with `python-3` as a variant, then `python-3` doesn't also have to be a requirement?
- PyQt5 is installed again, this time for Python 2. If we made `python-3` a requirement, do we then change the requirement?

Is it safe to simply add `python` to every pip package, and trust variants to safeguard the version?

**Comment by bpabel** [comment]
> Should add python into the requires of each pip package

While it's probably safe to do this, it's not always needed if you're using the package in a DCC with embedded python like Maya or nuke.

**Comment by mottosso** [comment]
Ah, yes that's true.. Did not think of that. If it's added, it could even *prevent* use in e.g. Nuke which provides its own binary called `python.exe` that may in turn conflict with the implicitly added `python` package.

**Comment by mottosso** [comment]
> After thinking about this, I think rez-wheel is not a good name.

Having tried this in practice for a few days, it doesn't sit right with me either.

Let's have a look at our options.

<table>
	<tr>
		<th width=230>Option</th>
		<th>Why</th>
		<th>Why not</th>
	</tr>
	<tr>
		<td>

`rez pip --install`

</td>
		<td>Simple, expected</td>
		<td>Breaks backwards compatibility</td>
	</tr>
	<tr>
		<td>

`rez pip --some-flag --install`

</td>
		<td>Backwards compatible, predictable</td>
		<td>

Tedious, lends itself to forgetting or not know about `--some-flag`
</td>
	</tr>
	<tr>
		<td>

`rez pip2 --install`

</td>
		<td>Simple</td>
		<td>

Conflicts with `pip2` and `pip3` which are typically reserved for installing using pip for Python 2 and 3.</td>

</tr>
	<tr>
		<td>

`rez wheel --install`

</td>
		<td>Simple</td>
		<td>

Unexpected and isn't clear what the relationship is with `rez pip`

</td>
	</tr>
</table>

Normally, I'm on board with the idea of maintaining backwards compatibility, but considering we'd be maintaining compatibility with an unusable version, I would personally make an exception and let the old pip die.

**Comment by mottosso** [comment]
One of the issues at the moment with installing using `--target` is that we don't get any of the associated scripts. I noticed these are actually preserved when using `--prefix` instead.

```bash
$ pip install black --prefix .
$ tree
Lib/
  site-packages/
    black/
Scripts/
  black.exe
```

> And `/bin/black` on Linux

We could probably leverage this for the corresponding Rez package, adding those to a `black/1.2.3/bin` subdirectory, and `env.PATH.prepend("{root}/bin")` alongside `env.PYTHONPATH.prepend("{root}/python")`

The only crux is that the `DistributionPath` and `InstalledDistribution` classes don't seem to be aware of their corresponding scripts. :( Therefore, if you install two or more packages that both provide executables, there would be no way of distinguishing which script came from which package.

We could probably work around it by pip installing each package separately, into it's own `--prefix` directory. But I'm hoping there's something I'm missing, any ideas? On doing that however, we do leverage the fact that Python creates this folder-structure to (presumably) account for things other than `site-packages` being installed, but I haven't encountered anything like that as of yet. If we keep it, we could simply `env.PYTHONPATH.prepend("{root}/Lib/site-packages")` and not have to worry about copying files individually from each installed package like we are now.

instinct-vfx:
I am not up to date with a bunch of the more recent changes in pip and setuptools. But in the past you could redirect individual components of a package using install-settings options. BUT the big show-stopper here was that this always enforced using the source package which is a no-go on windows.

bpabel:
Yeah, I had my own pip installer tool that had to do two install steps, once to get a list of all the downloaded dependencies, and then it would create an install package for each one separately using `--no-deps`

mottosso:
> We could probably leverage this for the corresponding Rez package

I gave this a try, but quickly realised this won't work.

No matter how the binaries are made, they'll be hard-linked to whichever Python executable was used to generate them, which isn't OK. It could be a `python-3` package, or a `python-2` package, or any other arbitrary version, even system-python. Whichever binary gets linked, that's the one the generated executable would use once called, regardless of which Python was requested after having been generated.

```bash
$ rez env python-3.7 rez wheel --install black
$ rez env python-2 black
> $ black
# Screw you, I'll use 3.7
```

The "binaries" we make must instead be e.g. shell scripts, that reference just `python` rather than a full path.

> Yeah, I had my own pip installer tool that had to do two install steps, once to get a list of all the downloaded dependencies, and then it would create an install package for each one separately using --no-deps

Did you experience the above problem? :O

instinct-vfx:
In order to party get around this problem we have a patched python distribution that includes custom versions of the executable wrappers that don't include the full path to python (so it takes the first python on PATH essentially). That comes with its own can of worms of course.

I wonder if it is possible to rely on python launcher these days? Disclaimer: Have not looked into it at all. And if not: is it possible to override the executable generation? Could we just generate .bat files on windows instead?

mottosso:
Moved to here: https://github.com/mottosso/rez-pipz

PR #618, opened 2019-04-26
Fix/add support for reversed version range

instinct-vfx:
This is a new pull request replacing https://github.com/nerdvegas/rez/pull/470 because i managed to completely fubar the original one. This is based off current master (which is 2.29.1)

Thanks a lot to @skral for fixing some issues in my original PR and for adding tests!

Sorry for the hassle, i shall take more care in the future.


nerdvegas:
Merging this now. Please note that the docstring here should have been updated (but no worries I am doing that!): https://github.com/nerdvegas/rez/blob/master/src/rez/vendor/version/version.py#L723

I also noticed that while there are docs on version format in the wiki, there isn't for version range syntaxes. I will add that also, in a separate PR.

Cheers
A

PR #623, opened 2019-04-29
Automatically create missing package repository dir

mottosso:
```bash
$ rez build --install
09:34:22 ERROR    PackageRepositoryError: Package repository path does not exist: C:\Users\marcus\packages
$ rez pip --install six
09:34:22 ERROR    PackageRepositoryError: Package repository path does not exist: C:\Users\marcus\packages
```

As choosing a destination directory is either default to users home directory, or explicitly chosen as a config option, installing to a non-existing directory shouldn't throw an error.

This should also help users new to Rez with the `rez bind --quickstart` option which currently throws an exception at the user right after install.

### DIFF
--- a/src/rezplugins/package_repository/filesystem.py
+++ b/src/rezplugins/package_repository/filesystem.py
@@ -5,6 +5,7 @@ from contextlib import contextmanager
 import os.path
 import os
 import stat
+import errno
 import time
 
 from rez.package_repository import PackageRepository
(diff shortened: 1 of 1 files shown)


mottosso:
Could we merge this? This is a critical feature for any new user, who will not have this directory present on his machine. An error is currently the first encounter with anyone looking at Rez, and that's not good.

nerdvegas:
I'm considering it, but consider:

* the existing error message is clear, and
* attempting to create the dir could fail anyway (it may not be a local packages dir), and
* if it isn't a local packages dir, then creating it is making an assumption about what the owner and permissions of the created directory should be. This isn't great, as someone may create a shared package repo dir, that another user may then not have write access to. Rez doesn't really have enough info to know what attributes the package repo dir should have.

If this functionality existed only for local package dirs, I think it would be more acceptable. What are your thoughts?

mottosso:
If I got to pick, I would let Rez try, and give the user an error only as a last-resort.

instinct-vfx:
For me i'd say restrict to local folders, then error on fail.

nerdvegas:
I'll merge as-is, I don't feel strongly enough about it and I don't think it's worth the extra complication of checking for local dir. Mind you, the error needs changing (it swallows the existing OSError, which could give useful info). I'll adjust and release.

Issue #616, opened 2019-04-26
Shell history not working in cmd.exe or PowerShell


mottosso:
Confirmed by @instinct-vfx on [Slack](https://rez-talk.slack.com/archives/CHELFCTFB/p1556261890473900) and this old [Google Groups thread](https://groups.google.com/forum/#!searchin/rez-config/history%7Csort:date/rez-config/erxLThK7Yrk/amhtxvCVAAAJ), using cmd.exe or PowerShell with Rez doesn't let you step through history via the traditional up-arrow key.

**To reproduce**

```bash
$ rez env
> $ echo Hello World!
> $ # hit up-arrow
> $ # no response.. :(
```

It does however (for some reason) work when using Rez with ConEmu, Cmder and Alacritty (and probably others too); each of which call on the same `rez_shell.bat` file. Odd!

nerdvegas:
Closed by https://github.com/nerdvegas/rez/releases/tag/2.35.0

PR #625, opened 2019-04-30
Rez on PyPI

mottosso:
[](https://ci.appveyor.com/project/mottosso/bleeding-rez/feature/rez-on-pypi-part3)

Hi all,

Here's a version of Rez compatible with PyPI.

rezonpypi

<br>

| Questions | Answers
|:----------|:------------
| What's changed? | Installation and upgrades has been made easier
| At what cost? | No embedded virtualenv
| What can we do? | Create your own virtualenv
| How can I try it? | `pip install bleeding-rez`

<br>

**Try latest branch**

```bash
$ pip install git+https://github.com/mottosso/bleeding-rez.git@feature/rez-on-pypi-part3 -U
```

All tests pass on both Windows and Linux (the PR includes an AppVeyor Windows environment for tests). See below for current minor obstacles.

<br>

### Overview

I've approached this in stages, whereby you are welcome to merge either one or all.

- [x] [**Part 1 - Scripts**](https://github.com/mottosso/bleeding-rez/tree/feature/rez-on-pypi-part1) Convert the `scripts/` directory to ["entry points"](), so as to get rid of the manual creation of binaries on all platforms
- [x] [**Part 2 - ~~install.py~~ is no more**](https://github.com/mottosso/bleeding-rez/tree/feature/rez-on-pypi-part2) I got rid of `install.py` and all that goes with it (i.e. `build_utils/`)
- [ ] [**Part 3 - Clean up**](https://github.com/mottosso/bleeding-rez/tree/feature/rez-on-pypi-part3) Address below issues and ensure all is working.

<br>

### Highlights

Ticked boxes mean I'm happy with it, but let me know if you aren't. Unticked means it's something we need to address.

- [x] **PyPI** Rez is now compatible with PyPI, and as proof and testing ground I've uploaded my fork under the name `bleeding-rez`, which is what you're seeing above. There's not much to say about it other than it works as you would expect. No tricks.
- [x] **PyPI Auto-deploy** Rez is automatically submitted to PyPI whenever a GitHub release (or tag) is made. Keeping PyPI in sync with GitHub
- [x] **Windows tests** I've replaced Travis with AppVeyor; you shouldn't notice any difference, apart from tests now also running on Windows. Travis is still there for the time being, to make you comfortable with the switch.
- [x] **Windows tests** On-top of that, I noticed some tests hadn't been run on Windows before and fixed those. I suspect it was related to how some tests would silently fail (a.k.a. get skipped) if certain binaries wasn't found on the PATH. The problem was, the way binaries was discovered on Windows made it impossible to find `cmake`, and thus `cmake` tests were never run. I did however run into the following issue..
- [ ] **Visual Studio on Windows** As a side-effect of addressing tests on Windows, I found that `vcvarsall.bat` doesn't work with the way the `cmd` shell is implemented; `cmd` is inheriting `PATH` from the system registry, whereas Visual Studio assumes `PATH` is appended interactively. As a consequence of this, it does mean Rez now inherits `PATH` more generally, which may or may not be what we want. See [1](#PATH)
- [x] **Code Coverage** In addition to Windows tests, I've also added code coverage, to get some sense of (1) what is left untested and (2) what is left actually unused and can be stripped.
- [ ] **~~system.is_production_install~~ is no more** This problem may be better addressed in another way, see [2](#is-production-install)
- [ ] **sup_world test disabled** I couldn't figure out how to address this; somehow, the `CMAKE_MODULE_PATH` got appended with a Windows-backslash which threw cmake. See [commit message](https://github.com/mottosso/bleeding-rez/pull/10/commits/65cda97611a8b3bd9b92458115f8607328e16d60) for details.
- [x] **`--target`** Custom install directory with `--target` doesn't work with how Rez manages your `PYTHONPATH`. Because if you do use `--target` you would also need to manually add Rez to your PYTHONPATH but you can't, since `rez env` will take control of that PYTHONPATH and remove it once you've entered into an environment. Without `--target`, rez will still be accessible as `site-packages` is accessible despite not being on your `PYTHONPATH`. Workaround:
   1. Use a virtualenv
   2. Add destination folder to your system PYTHONPATH

<br>

### No Virtual Envrionment

From what I can tell, the embedded virtualenv was used to let packages gain absolute control over `PYTHONPATH`, whilst still being able to expose the `rez` Python package to `rez env`.

- [Reference 1](https://github.com/nerdvegas/rez/commit/56aa8b4a4a7a11dd61d5e18b6fe4efc44bd9ed6c)
- [Reference 2](https://github.com/nerdvegas/rez/commit/bd9fac0739716494b32cb4bd8310cc16cb3f9bc0)

Now that there is no virtual environment, `rez` is made available via the `site-packages` of the target Python install.

```bash
$ c:\python27\lib\site-packages\rez
```

But that means you can't install Rez elsewhere anymore, which can be a problem if you host Rez itself as a package.

To solve that, you can either can install any PyPI package in a virtualenv, including this Rez from this PR.

<br>

### Is Production Install

Previously, a file called `.is_production_install` was embedded in the `Scripts\rez` directory alongside the Rez binaries. I was placed there to help identify the install having been made via `install.py`, which in turn was designed to solve the aforementioned issue of [`PATH`](#path).

With this PR, that safety guard is off per default, and replaced with using your own virtual environment.

```bash
$ python -m virtualenv my_rez_env
$ my_rez_env\scripts\activate
(my_rez_env) $ pip install bleeding-rez
```

The advantage is two-fold.

1. Rez on Pip, improve adoption and reduce learning curve
2. Simplified getting-started process for new users

I'd imagine the README would then be split between a "I just want to take Rez for a spin" and "I like Rez, now I want to use it in my large corp", whereby the latter involves a virtual env.

<br>

### Post-merge

In order to test deployment, I've deployed to my account under a different name and version.

Once merged, you'll need to:

1. Create an account on AppVeyor
1. Replace my username and password in `appveyor.yml`
1. Create an account on CodeCov
1. Add your `CODECOV_TOKEN` to AppVeyor's Environment

Let me know your thoughts!

Edit: This also closes https://github.com/nerdvegas/rez/pull/339


mottosso:
### Help!

I'm having trouble solving the unticked boxes above, primarily the one involving CMake on Windows finding the Visual Studio install.

- [Full log](https://ci.appveyor.com/project/mottosso/bleeding-rez/builds/24196381/job/2r8512py5ia8awaj?fullLog=true#L42) (It's heavy)

```bash
rez selftest -s cmd
Not searching for unused variables given on the command line.
-- The C compiler identification is unknown
-- The CXX compiler identification is unknown
CMake Error in CMakeLists.txt:
  The CMAKE_C_COMPILER:
    cl
  is not a full path and was not found in the PATH.
  To use the NMake generator with Visual C++, cmake must be run from a shell
  that can use the compiler cl from the command line.  This environment is
  unable to invoke the cl compiler.  To fix this problem, run cmake from the
  Visual Studio Command Prompt (vcvarsall.bat).
  Tell CMake where to find the compiler by setting either the environment
  variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
  the compiler, or to the compiler name if it is in the PATH.
CMake Error in CMakeLists.txt:
```

Any ideas? I suspect I need to add some CMake modules to enable it to find VS on the local disk, but I don't know how. :S

bpabel:
Having a working pip install would be great.  Ditto for automated uploads.

I think we should still have an install script to create a virtualenv or standalone python installation.  The `poetry` project has a good system that we may want to emulate.  It's installable via pip, but they also provide a downloadable `getpoetry.py` script that installs poetry and it's dependencies into a virtualenv and updates the system path to make it runnable.

Also, Rez currently patches the generated console scripts to run with whatever python is in the environment (by default, setuptools bakes in the python executable path when building the script), not sure if we still want to do that or not.

mottosso:
> I think we should still have an install script to create a virtualenv or standalone python installation.

There could still be, but frankly, I think that's better suited for the README.

---

**Installation**

Choose your installation method.

<details>
  <summary>Simple</summary>
<table>
<tr><td>
<br>

**Simple** The simple approach is well suited for beginners and those looking to learn more about Rez and whether it is suited to their problem and environment.

</td></tr>
<tr><td>

```bash
$ pip install rez
```
</td></tr>
<tr><td>

<br>

**Advantages**

- **User-friendly and familiar installation method** Everybody loves pip

</td></tr>
</table>

</details>
</details><details>
  <summary>Recommended</summary>

<table>
<tr><td>
<br>

**Recommended** The recommended approach is better suited for live production, where you've committed to Rez and want precision and control.

</td></tr>
<tr><td>

```bash
$ python -m virtualenv rez
$ rez\Scripts\activate
(rez) $ pip install rez
```
</td></tr>
<tr><td>
<br>

**Advantages**

- **Isolated PATH** No interference from external binaries in your `Scripts/` Python directory.
- **Isolated site-packages** No interference from system-wide Python packages

</td></tr>
</table>
</details>

src/rezplugins/shell/sh.py
```diff
@@ -58,6 +59,13 @@ def get_syspaths(cls):
             if path not in paths:
                 paths.append(path)
         cls.syspaths = [x for x in paths if x]
+
+        # add Rez binaries
+        exe = which("rez")
```
JeanChristopheMorinRodeoFX: This will fail under OSX. OSX has already something called `rez`, see https://www.unix.com/man-page/osx/1/Rez/. I've never been able to use the rez-commands without `-` because of that thing.

JeanChristopheMorinRodeoFX:
We definitely need to keep the `install.py` for now (and same for the patched scripts), until we revisit the entire install process IMO.

src/rezplugins/shell/sh.py
mottosso: Thanks @JeanChristopheMorinPerso, fixed.

mottosso:
> We definitely need to keep the install.py for now (and same for the patched scripts), until we revisit the entire install process IMO.

Thanks @JeanChristopheMorinPerso, what advantages do you see in this approach?

README.md
```diff
@@ -121,17 +121,100 @@ Resolved environments can also be created via the API:
...
+<tr><td>
+<br>
+
+**Simple** The simple approach is well suited for beginners and those looking to learn more about Rez and whether it is suited to their problem and environment.
+
+</td></tr>
+<tr><td>
+
+```bash
+$ pip install rez
```
JeanChristopheMorinRodeoFX: We shouldn't advertise this method first. We should advertise to install in a virtualenv or install with `pip install rez --user`, but not install on the system. Installing on the system is never recommended and the python community puts a lot of effort to emphasis on the fact that we should never alter the system's python and always install with `--user` or use virtualenvs.

Why should we never alter the system's python site-packages? Because for platforms that comes with python already installed and that the OS is using python to work properly (Linux and OSX), it's almost impossible (if not completely impossible) to revert to the initial state if something goes wrong without re-installing the OS.

README.md
```diff
@@ -121,17 +121,100 @@ Resolved environments can also be created via the API:
...
+
+</td></tr>
+<tr><td>
+
+```bash
+$ python -m virtualenv rez-dev
+$ rez-dev\Scripts\activate
+(rez) $ git clone https://github.com/nerdvegas/rez.git
+(rez) $ cd rez
+(rez) $ pip install . -U
```
JeanChristopheMorinRodeoFX: The real development mode for a pip package is to use `pip install . -e`
```
  -e, --editable <path/url>   Install a project in editable mode (i.e. setuptools "develop mode") from a local project path or a VCS url.
```
That symlinks your code in the site-packages directory. That way you do it only once and your code is kept up-to-date in your site-packages.

JeanChristopheMorinRodeoFX:
@mottosso Sure, let me explain a little bit more in details my thoughts.

First of all, a pip package is great if you need to use rez the API. Your contribution here is great as far as the pip package goes. It effectively makes the life of more advanced users easier.

When sourcing a virtualenv, you get the `bin` of the virtualenv added into `$PATH`. That means it can mess up really bad with your rez environment because the `python` executable inside the `bin` becomes available in your `$PATH`. This is exactly what the patched scripts were preventing by copying the scripts into a seperate `rez` folder inside the `bin` folder that we can just add into our `$PATH` and don't need to activate the virtualenv.

Once patched, the scripts looks something like:
```python
#!/rdo/rodeo/setup/rez/linux/2.23.1.rdo5/bin/python -E

from rez.cli._main import run
run()
```
so they use the virtualenv's python only to execute the rez commands. That allows you to have python as a rez package and don't have to worry about which python will be picked up if you `rez-env python -- python` (that is if you didn't use rez-bind to create your python rez package. So if your python rez package isn't using your system's python).

One could argue that the shebang in the console scripts of a pip package points to the virtualenv's python and that it wouldn't be a problem. That is true. Although, I personally think that it could cause confusions to new comers that would want to activate the virtualenv to start to work with it (which is what we are all used to, at least most of the time when you create a virtualenv you will end up activating it).

Also, as I previously said (in slack and in this PR), I would really like if we do the exercise of reviewing all our options for the new installer before going forward with removing the `install.py` file. All package managers I've used so far have their own bootstrap process to install themselves. This is to make the life of new comers easy and help them avoid doing mistakes (like with python, virtualenvs, site-packages, user's site-packages and all that stuff that people don't necessarily need to know to use rez or the other package managers, no matter what the language is).

Let's get some examples:

* Pip: https://pip.pypa.io/en/stable/installing/ (although one could argue that we can't install pip without pip, though it's not true since it's what the installer does, it installs pip with pip).
* Pipenv: https://github.com/pypa/pipenv#installation
* https://github.com/sdispater/poetry#installation
* saltstack (not so a package manager, but a configuration/deployment system entirely written in pure python, and so available as a pip package): https://repo.saltstack.com/#bootstrap (see the pip package at https://pypi.org/project/salt/). Salt is actually a really good example, as it's an "app" that has a python API that you can use (like rez). They hide all the complexity under a bootstrap process (and yes there is a lot of complexities in salt). If one wants to dive into the complex stuff, he is allowed and can use the pip package.

I won't list all of them, but it gives a pretty good idea that everybody gives a bootstrap process.

I hope this clarifies a little bit more my thoughts behind my last post (that I confess was too short and was written too fast, my bad).

Let me know what you think.

README.md
```diff
@@ -121,17 +121,113 @@ Resolved environments can also be created via the API:
...
+**Optional**
+
+In case you also want to make `python` into a Rez package, you may also want to remove any reference to `python` from your virtual environment.
+
+```bash
+# Windows
+(rez) $ del rez\Scripts\python.exe
+
+# Linux
+(rez) $ rm rez\bin\python
```
JeanChristopheMorinRodeoFX: This won't work. And actually, it would break the entire thing. The scripts installed by your pip package directly link to this specific python.

Like here is the `pylint` binary created when installing pylint inside a virtualenv:

```python
#!/mnt/users/jcmorin/.local/share/virtualenvs/tractor-to-elasticsearch-T8zSJag0/bin/python2

# -*- coding: utf-8 -*-
import re
import sys

from pylint import run_pylint

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(run_pylint())
```
```
lrwxrwxrwx 1 jcmorin artists     7 Apr 25 17:29 python -> python2
-rwxrwxr-x 1 jcmorin artists  7216 Apr 25 17:29 python2
lrwxrwxrwx 1 jcmorin artists     7 Apr 25 17:29 python2.7 -> python2
```
Well, it will "work" but it's a really really dirty hack. Because the recommended way for launching python is to use `python2` or `python3` (so we are sure which version is really used).

So it doesn't cover all cases.

mottosso:
> the python executable inside the bin becomes available in your $PATH

I had thought about this, and figured one could simply remove the `python` binary from the `/bin` directory. But alas, the entry points are hard-wired to use that binary. :(

However, giving this a try, it doesn't look like that `python` actually interferes with a packaged Python, if one wanted to use that.

```bash
$ rez env
> $ where python
/path/to/virtualenv/bin/python
> $ exit
$ rez env python-3
> $ where python
/path/to/python3/python
```

Was there another issue having the virtualenv `python` on your `PATH`?

> This is exactly what the patched scripts were preventing by copying the scripts into a seperate rez folder inside the bin folder that we can just add into our $PATH and don't need to activate the virtualenv.

It doesn't actually look like you need to activate this environment. Like before, you can still add it to your PATH explicitly.

**Before**

```bash
$ python install.py
$ export PATH=/path/to/rez-install/Scripts/rez
```

**After**

```bash
$ pip install rez
$ export PATH=/path/to/your/virtualenv/Scripts
```

mottosso:
I dove into setuptools to find whether we could add a post-install script to avoid a separate `install.py`, but alas didn't manage. I found a method of adding a script to when `setup.py` has finished, either by (1) adding an `atexit` handler to trigger whenever Python is done with the file and closes and (2) a `cmdclass` argument (aptly named) to `setup()` which does trigger on completion, but doesn't affect the resulting wheel which is ultimately what is called on a users machine.

Attaching script below for completeness in case someone else ventures down this (darkly) road.

<details>
<summary>Post-install script for <code>setup.py</code></summary>

```python
from __future__ import with_statement
import fnmatch
import shutil
import os
import errno
import sys


try:
    from setuptools import setup, find_packages
    from setuptools.command.install import install
except ImportError:
    sys.stderr.write("install failed - requires setuptools\n")
    sys.exit(1)


if sys.version_info < (2, 6):
    sys.stderr.write("install failed - requires python v2.6 or greater\n")
    sys.exit(1)


def find_files(pattern, path=None, root="rez"):
    paths = []
    basepath = os.path.realpath(os.path.join("src", root))
    path_ = basepath
    if path:
        path_ = os.path.join(path_, path)

    for root, _, files in os.walk(path_):
        files = [x for x in files if fnmatch.fnmatch(x, pattern)]
        files = [os.path.join(root, x) for x in files]
        paths += [x[len(basepath):].lstrip(os.path.sep) for x in files]

    return paths


# get version from source
fname = os.path.join("src", "rez", "utils", "_version.py")
with open(fname) as f:
    code = f.read().strip()
_rez_version = None  # just to keep linting happy
exec(code)  # inits _rez_version
version = _rez_version


scripts = [
    "rez-config",
    "rez-build",
    "rez-release",
    "rez-env",
    "rez-context",
    "rez-suite",
    "rez-interpret",
    "rez-python",
    "rez-selftest",
    "rez-bind",
    "rez-search",
    "rez-view",
    "rez-status",
    "rez-help",
    "rez-depends",
    "rez-memcache",
    "rez-yaml2py",
    "_rez_fwd",  # TODO rename this _rez-forward for consistency
    "_rez-complete",
    "rez-gui"
]


class SecurePathPostInstall(install):
    """Generate a rez/ subdirectory for optional addition to PATH

    These help protect a given `rez env` from the contained `python`
    executable of a system or virtual environment, and may optionally
    be added to your PATH in place of the root Scripts/ folder on
    Windows or bin/ folder on Unix

    """

    def run(self):
        install.run(self)

        base = self.install_base
        bin_dir = os.path.join(base, "Scripts")
        rez_bin_dir = os.path.join(bin_dir, "_rez")

        try:
            os.makedirs(rez_bin_dir)
        except OSError as e:
            if e.errno == errno.EEXIST:
                # This is ok, probably from a prior install
                pass
            else:
                return sys.stderr.write(
                    "Couldn't generate optional, isolated PATH for Rez"
                )

        # NOTE: Print statements aren't visible when installing via
        # pip install, unless -vvv is passed for increased verbosity
        print("Generating isolated Rez binary directory..")
        for _, _, files in os.walk(bin_dir):
            for fname in files:
                if not fname.startswith("rez"):
                    continue

                src = os.path.join(bin_dir, fname)
                dst = os.path.join(rez_bin_dir, fname)
                print("  copying %s -> %s" % (src, dst))
                shutil.copyfile(src, dst)
            break

        print("Success")


setup(
    name=os.getenv("REZ_PYPI_NAME", "rez"),  # for development on fork
    version=version,
    description=("A cross-platform packaging system that can build and "
                 "install multiple version of packages, and dynamically "
                 "configure resolved environments at runtime."),
    keywords="package resolve version build install software management",
    long_description=None,
    url="https://github.com/nerdvegas/rez",
    author="Allan Johns",
    author_email="nerdvegas@gmail.com",
    license="LGPL",
    entry_points={
        "console_scripts": [
            "rez = rez.cli._main:run",
            "bez = rez.cli._bez:run",

            # Alias
            "rezolve = rez.cli._main:run",
        ] + [
            "{cmd} = rez.cli._main:{func}".format(
                cmd=script,
                func=script.replace("-", "_")
            )
            for script in scripts
        ]
    },
    cmdclass={
        "install": SecurePathPostInstall,
    },
    include_package_data=True,
    zip_safe=False,
    package_dir={'': 'src'},
    packages=find_packages('src', exclude=["tests"]),
    package_data={
        'rez':
            ['rezconfig', 'utils/logging.conf'] +
            ['README*'] +
            find_files('*', 'completion') +
            find_files('*', 'tests/data'),
        'rezplugins':
            find_files('rezconfig', root='rezplugins') +
            find_files('*.cmake', 'build_system', root='rezplugins') +
            find_files('*', 'build_system/template_files', root='rezplugins'),
        'rezgui':
            find_files('rezguiconfig', root='rezgui') +
            find_files('*', 'icons', root='rezgui')
    },
    classifiers=[
        "Development Status :: 4 - Beta",
        "License :: OSI Approved :: GNU Lesser General Public License v3 (LGPLv3)",
        "Intended Audience :: Developers",
        "Operating System :: OS Independent",
        "Programming Language :: Python",
        "Programming Language :: Python :: 2",
        "Programming Language :: Python :: 2.6",
        "Programming Language :: Python :: 2.7",
        "Topic :: Software Development",
        "Topic :: System :: Software Distribution"
    ]
)

```

</details>

On the upside, it looks like the solution is not far away!

---

**Optional**

Virtualenv exposes a `python` and `pip` binary on your `PATH` which make them available from within a Rez context despite no `python` or `pip` package having been included. For additional protection, you can create a separate directory with only Rez binaries and expose this on your `PATH` instead.

```bash
(rez) $ mkdir rez\Scripts\rez
(rez) $ cp rez\Scripts\rez*.exe rez\Scripts\rez\
(rez) $ deactivate
$ set PATH=%cd%\rez\Scripts\rez
$ rez env
> $ python
'python' is not recognized as an internal or external command,
```

---

In other news, @instinct-vfx do you know how to get CMake to find Visual Studio on Windows? Is there maybe a module missing or the like?

bpabel:
I also just recently learned of a tool called `pipx` that works similar to `pip` but will install the library and its dependencies into a virtualenv and put the generated scripts into a globally accessible location.  Basically, the same thing the install script would do, except it has options for upgrading and it just works as long as your project works with pip.  It only copies the scripts for that library, so you don't have to worry about scripts from dependent packages or `python.exe`.

Unfortunately, it only officially supports python>3.6.  But the nice part is that if rez works with pip (and eventually becomes python3 compatible), then everything just works with `pipx` without any extra work.

mottosso:
> I also just recently learned of a tool called pipx

Yes! With Rez as a regular pip package, those are the kind of options you'd have as a user. I also have my own system for managing environments and so to me virtualenv is just another layer of unnecessary abstraction.

mottosso:
Help! Does anyone use CMake on Windows? I haven't been able to figure out how to get `rez selftest` to work when `cmake` is available on `PATH`, I suspect CMake isn't working that well with Visual Studio and Rez. Any ideas?

bpabel:
@motosso My studio has it working, but it's pretty hacky.  It's a separate package that sets some of the required envvars for vcvarsall, and then runs vcvarsall and parses the output for env vars that get set and then sets them in the current rez environment.  I imagine it's not a true msvc environment that could actually work, but it's enough to get cmake to work.

nerdvegas:
I used to have rez installable via pip and hosted on pypi ages ago. That was abandoned due to issues that it looks like you've been facing here. Notably, there were repeated cases where rez itself would cease to function once the user is resolved into an environment containing a different (and incompatible) version of python. This approach works fine if all you want is the rez API, but isn't useful if you need rez itself (ie its tools).

As @JeanChristopheMorinPerso says above, we need to properly review options to replace the current installer. The rez installation must be able to operate in a way that's independent of the environment that it has constructed for you in any given resolved env. Lots of work did go into ensuring that this is currently the case.

My current thinking on this is to aim for a pip-compatible installation, that also installs a set of platform-specific wrappers (rez-env etc). These wrappers would only become exposed to PATH when within a resolved env. When invoked, they would set the environ back to whatever it was pre-rez-env, and would then invoke the analogous native tool (rez-env wrapper would call rez-env, for eg).

mottosso:
> Notably, there were repeated cases where rez itself would cease to function once the user is resolved into an environment containing a different (and incompatible) version of python. This approach works fine if all you want is the rez API, but isn't useful if you need rez itself (ie its tools).

Any idea how I can reproduce this? I'm not sure what you're referring to, this is the version I've been using for the past few weeks in a number of different circumstances.

> That was abandoned due to issues that it looks like you've been facing here.

What other issues do you have in mind?

> As @JeanChristopheMorinPerso says above, we need to properly review options to replace the current installer.

He did say that, but I wasn't sure of what to make of it. Isn't this an option we're reviewing? And isn't it proper?

> My current thinking on this is to aim for a pip-compatible installation, that also installs a set of platform-specific wrappers (rez-env etc). These wrappers would only become exposed to PATH when within a resolved env. When invoked, they would set the environ back to whatever it was pre-rez-env, and would then invoke the analogous native tool (rez-env wrapper would call rez-env, for eg).

It sounds a little out of scope for this particular PR, but if you'd like I could amend it.

Which wrappers would you like exposed from within an environment? Most of them are either useful or necessary from outside of a context I think:

- [ ] status
- [x] help
- [x] depends
- [x] plugins
- [x] pip
- [x] diff
- [x] cp
- [x] build
- [x] env
- [x] suite
- [x] config
- [x] python
- [x] selftest
- [x] test
- [x] memcache
- [x] yaml2py
- [x] interpret
- [x] search
- [x] bind
- [x] gui
- [ ] context
- [x] release
- [x] view

nerdvegas:
Frustratingly, this is a PR I cannot merge. It removes the existing installation script, and I'm hard pressed to think of anything that could break backwards compatibility more. I also think it's too large in scope - it is described as "Rez on PyPi", not as "deleting/replacing the existing installer." To make clearer the importance of backwards compatibility, I've updated the contributing guidelines: https://github.com/nerdvegas/rez/blob/master/CONTRIBUTING.md.

Not only would this break studios' existing installation processes for rez, but I also think it would introduce regressions from years ago, and I'll explain below. There's no way you can assume from your own testing over a few weeks, that a new installation process is going to work everywhere - there are too many variables to account for - different python versions, different platforms, compiled vs non-compiled packages and so on. Anything like this needs to be released alongside the existing installer, so we can get a good amount of real-world testing in place before making a transition. This has to be advertised to studios via the release notes, so they have a chance to test out the new and experimental installation process.

Some background info is hopefully going to help.

Years ago in rez, we had a PyPi registration for it, and it was pip installable (in fact the PyPi entry was only removed a few months ago I believe). This worked fine, with one big caveat - because we couldn't guarantee stability of tools such as rez-env once within a resolved env, the pip-installed version deliberately disabled these once within a rez-env. The intent of the pip package was primarily for when a developer wanted to use rez as an API and as a native python package, rather than a toolset.

From there, we determined that a virtualenv-based install should work. It almost did, bar a couple of issues: -

The first, someone mentioned earlier - the binaries within the virtualenv end up exposed in the rez-env, so `python` for eg would unexpectedly run the virtualenv copy, and lots of things would break. This is why the installer places the rez binaries only into their own `rez` subdirectory.

The second we didn't know about for a while, until we started hitting issues a few months in. The problem had to do with code on PYTHONPATH still being visible when within a virtualenv. I wish I could recall further details, but this is years ago and I cannot. But the problem had to be solved - in some circumstances, rez would crash. If you look at any of the rez binaries within the virtualenv, you'll see that they shebang with `python -E`. Passing -E is not natively supported in virtualenv, and a bunch of stuff (venv, distlib) had to be patched to get it to work.

So, the installer is responsible for taking care of these details. Deleting it and build_utils/ is just deleting these fixes that were put in place years ago. Nobody wanted to patch virtualenv/distlib - we simply found that it had to be done.

I would like to see this PR broken down, and changes to/removal of the existing installation process removed.

First, we could have a PR which actually just does what this title suggests - ie, adds to PyPi. A pip rez installation should disable tools when within a rez-env as mentioned earlier, because they may not be stable.

Second - and I've mentioned this several times - I think it's worth investigating an approach where setup.py also installs sets of platform-specific wrapper scripts that, when invoked, revert the environ (or part of it) to the pre-rez-env environ, and then exec the equivalent entry point. If this works, it would mean the ability to simply `pip install rez`, _and_ not lose the cli tools within a rez-env. We may find that this is enough functionality for everyone, and eventually could do away with virtualenv-based installations completely.

I would like nothing more than to see the removal of build_utils/, and a pip-based install. But that has to be done in parallel with the existing installer, until such time that we can be confident enough of its stability.

mottosso:
> Frustratingly, this is a PR I cannot merge.

I'm happy to address your concerns, but there's no need for frustration. I didn't expect this to get merged before a few rounds of tennis. :) I appreciate you chiming in.

> It removes the existing installation script, and I'm hard pressed to think of anything that could break backwards compatibility more.

There's a (good?) reason for that, which is that if Rez is on PyPI, the installation script is obsolete. I hadn't added it back in because so far all concerns had been accounted for. Backwards compatibility however was not one of them, and is a good reason to keep it in.

> and I've mentioned this several times

Just to make sure I'm reading the tone of this right, is there something that has been said you feel I haven't listened to? JC also used this phrasing [earlier](https://github.com/nerdvegas/rez/pull/625#issuecomment-488293714) and I just want to make sure there's no frustration going on and that we're all happy. :) Sometimes things can get lost in text.

> I would like to see this PR broken down, and changes to/removal of the existing installation process removed. First, we could have a PR which actually just does what this title suggests - ie, adds to PyPi. 

Sounds good.

> A pip rez installation should disable tools when within a rez-env as mentioned earlier, because they may not be stable.

This I don't follow. Do you mean this?

**A**

```bash
$ pip install rez
$ rez
Command not found
```

Or maybe this?

**B**

```bash
$ pip install rez
$ rez env
> $ python
Command not found
```

Or even this?

**C**

```bash
$ pip install rez
$ rez env
> $ rez env
Command not found
```

nerdvegas:
Hi Marcus,

I mean frustrating in the sense that there's stuff here I would like to
merge, but cannot because the install.py removal isn't backwards compatible.

RE mentioning several times, I'm referring specifically to the idea of a
pip install that also installs platform-specific wrappers that get exposed
to PATH on entry to a rez-env, and which revert environ and run the
original entry point. I think this should give us functioning rez tools
within a rez-env, out of the box (no need for production-specific
virtualenv install etc), but I don't believe there's been much feedback on
that proposal.

*"'A pip rez installation should disable tools when within a rez-env as
mentioned earlier, because they may not be stable.""*

I mean:

    ]$ pip install rez
    ]$ rez-env floob  # sets PATH, floob is non-python
    ]$ rez-context
    /usr/bin/env: 'python': No such file or directory

If you could provide a PR which contains just the following parts of this
PR, that would be ideal and can I hope be a step towards removing the
current installer:

* entry points and related setup.py changes;
* Auto deploy to Pypi (nice).

In general it's good to aim for smaller PRs that specifically address one
thing. There are too many things in this PR, for example:

* AppVeyor (I'm not adverse to ditching Travis at all btw, but there should
be a dedicated PR for this. I recall hitting issues with Travis wrt support
for multiple build envs);
* Windows test fixes
* Some py3 compatibility stuff (switching print >> sys.stderr to
sys.stderr.write).

I myself am guilty of throwing unrelated things into PRs at times, and I'm
trying not to do that. In general (both in and outside of rez dev) I've
found that smaller, more targeted PRs get more feedback and get merged more
quickly.

Cheers
A

nerdvegas:
Note: I see there is a separate PR for Appveyor at https://github.com/nerdvegas/rez/pull/339 and that you mention that above. I'll look into sorting this out soon unless someone else gets to it first.

On the (hopefully rare) occasion where a PR includes merges from other PRs, please make that explicit and state it at the top of the PR description. I'd like to avoid unintentionally merging multiple PRs (just a mention can easily be mistaken for an issue instead).

Cheers
A

mottosso:
> I mean:
>    ]$ pip install rez
>    ]$ rez-env floob  # sets PATH, floob is non-python
>    ]$ rez-context
>    /usr/bin/env: 'python': No such file or directory

I'd be happy to get started, but have just a few questionmarks here:

1. Why would you want Rez to stop functioning within its own context? `rez-context` especially is only (?) useful from within a context, should we remove this command altogether?
2. What's the relevance of `rez-env floob` being non-Python, would it not stop functioning if it wasn't?
1. What constitutes a "non-Python" environment? E.g. a dependency on a package called "python" or a "python" executable being on PATH?
1. Is this really what `install.py` does? Should a pip install change the behavior of Rez?

nerdvegas:
On Tue, Jun 4, 2019 at 4:01 PM Marcus Ottosson <notifications@github.com>
wrote:

> I mean:
> ]$ pip install rez
> ]$ rez-env floob # sets PATH, floob is non-python
> ]$ rez-context
> /usr/bin/env: 'python': No such file or directory
>
> I'd be happy to get started, but have just a few questionmarks here:
>
>    1. Why would you want Rez to stop functioning within its own context?
>    rez-context especially is only (?) useful from within a context,
>    should we remove this command altogether?
>
> Not want, it's just reality. If the target env doesn't contain python,
rez-* tools don't work.

>
>    1. What's the relevance of rez-env floob being non-Python, would it
>    not stop functioning if it wasn't?
>
> If it's a compiled tool that doesn't use python, and sets PATH to expose
its tools, then python won't be available in the target env. Rez doesn't
and shouldn't treat python as a special case.

>
>    1. What constitutes a "non-Python" environment? E.g. a dependency on a
>    package called "python" or a "python" executable being on PATH?
>    2. Is this really what install.py does? Should a pip install change
>    the behavior of Rez?
>
> It would be great if a pip install didn't change the behaviour of rez,
which is why I've proposed the wrapper solution previously.

mottosso:
> Not want, it's just reality. If the target env doesn't contain python, rez-* tools don't work.

This doesn't make sense, and I think you're confusing Python with Rez.

`rez-context` uses Python internally, yes, but that's just implementation detail. It may as well have been written in Go, and if so should it stop working unless a Go-package is included?

Now, this is one thing:

```bash
$ rez-env floob
> $ python
Command not found
```

Because there is no reason for `python` to exist here, unless `floob` somehow referenced it, and we can solve that, like `install.py` has done, by putting Rez-executables in its own sub-directory.

But I'm not sure I see the advantage (or even desire?) to let a package affect how Rez works internally.

Furthermore, what you're suggesting would make `rez-context` use a *different* Python for itself, depending on which package exposed which `python` to the environment. What if it's an ARM Python, or Micro-Python, or the Python from The Foundry Nuke? There's little guarantee `rez-context` is even compatible with these, and nor should it be? The way it works right now is that `rez-context` references its *original* Python interpretr, which is the one it was installed for, and is compatible with. But again, those are implementation details.

bpabel:
> Not want, it's just reality. If the target env doesn't contain python, rez-* tools don't work.

It will if you're using setuptools-generated console_scripts entry points.  setuptools will embed the path to the python interpreter the generated executables were created with, so it doesn't matter if python is on the PATH or not. As long as the directory the rez-* tools reside in is on the PATH, they'll work.

> Rez doesn't and shouldn't treat python as a special case.

Does it need to treat it as a special case, though?  Surely the rez-* tools don't *need* to (and probably shouldn't) run with the version of python configured in the current environment.  As an analogy, rez already "special-cases" all the vendored packages it uses, it's not switching them out based on what version of those same packages are available in the configured rez environment.

nerdvegas:
On Tue, Jun 4, 2019 at 4:22 PM Marcus Ottosson <notifications@github.com>
wrote:

> Not want, it's just reality. If the target env doesn't contain python,
> rez-* tools don't work.
>
> This doesn't make sense, and I think you're confusing Python with Rez.
>
Ok, yeah I'm pretty sure I'm not confusing python with rez.

> rez-context uses Python internally, yes, but that's just implementation
> detail. It may as well have been written in Go, and if so should it stop
> working unless a Go-package is included?
>
I think we simply are not on the same page. I am saying that if rez is
installed as a standard python package, and then by running rez-env, you
create an env that is not configured to run python anymore, then rez won't
work within that env. To make it work, regardless of how that env is
configured, you need to do other things, like what install.py is already
doing.

Step 1: Have rez installable as a standard package via pip, using entry
points, and automatically deploying to PyPi on pushes to master.
Step 2: See if it's possible to get this standard package also working
within an env, using the wrapper approach I have suggested earlier.
Step 3: If (2) doesn't work, then see if it's possible to simplify/remove
install.py to provide the standalone behaviour that rez currently has (ie
it does not break within a rez-env).

Now, this is one thing:

nerdvegas:
On Tue, Jun 4, 2019 at 4:29 PM Brendan Abel <notifications@github.com>
wrote:

> Not want, it's just reality. If the target env doesn't contain python,
> rez-* tools don't work.
>
> It will if you're using setuptools-generated console_scripts entry points.
> setuptools will embed the path to the python interpreter the generated
> executables were created with, so it doesn't matter if python is on the
> PATH or not. As long as the directory the rez-* tools reside in is on the
> PATH, they'll work.
>
No it won't. Other packages in the resolved env can do other things, like
set LD_LIBRARY_PATH / LD_PRELOAD and all sorts of things. If the situation
has changed since we first discovered this years ago, then awesome. Let's
find out.

> Rez doesn't and shouldn't treat python as a special case.
>
> Does it need to treat it as a special case, though? Surely the rez-* tools
> don't *need* to (and probably shouldn't) run with the version of python
> configured in the current environment. As an analogy, rez already
> "special-cases" all the vendored packages it uses, it's not switching them
> out based on what version of those same packages are available in the
> configured rez environment.
>
No it doesn't, and no it shouldn't, so yes it is as you describe here.

bpabel:
> If the situation has changed since we first discovered this years ago, then awesome. Let's find out.

Yeah, I think that's the misunderstanding.  Entry points have probably changed to be less fragile since the install script was first created.  

For example, I just ran this in python 2.7 with setuptools 28.8.0

### setup.py

```python
from setuptools import find_packages, setup
setup(
    name="fakerez",
    version="0.0.1",
    packages=find_packages(exclude=('tests',)),
    entry_points={
        'console_scripts': ['fakerez-env=fakerez:env'],
    },
)
```

### Running on cmd line without python
```
C:\Users\007br\workspace\fakerez>python setup.py install
running install
running bdist_egg
running egg_info
creating fakerez.egg-info
writing fakerez.egg-info\PKG-INFO
writing top-level names to fakerez.egg-info\top_level.txt
writing dependency_links to fakerez.egg-info\dependency_links.txt
writing entry points to fakerez.egg-info\entry_points.txt
writing manifest file 'fakerez.egg-info\SOURCES.txt'
reading manifest file 'fakerez.egg-info\SOURCES.txt'
writing manifest file 'fakerez.egg-info\SOURCES.txt'
installing library code to build\bdist.win-amd64\egg
running install_lib
running build_py
creating build
creating build\lib
creating build\lib\fakerez
copying fakerez\__init__.py -> build\lib\fakerez
creating build\bdist.win-amd64
creating build\bdist.win-amd64\egg
creating build\bdist.win-amd64\egg\fakerez
copying build\lib\fakerez\__init__.py -> build\bdist.win-amd64\egg\fakerez
byte-compiling build\bdist.win-amd64\egg\fakerez\__init__.py to __init__.pyc
creating build\bdist.win-amd64\egg\EGG-INFO
copying fakerez.egg-info\PKG-INFO -> build\bdist.win-amd64\egg\EGG-INFO
copying fakerez.egg-info\SOURCES.txt -> build\bdist.win-amd64\egg\EGG-INFO
copying fakerez.egg-info\dependency_links.txt -> build\bdist.win-amd64\egg\EGG-INFO
copying fakerez.egg-info\entry_points.txt -> build\bdist.win-amd64\egg\EGG-INFO
copying fakerez.egg-info\top_level.txt -> build\bdist.win-amd64\egg\EGG-INFO
zip_safe flag not set; analyzing archive contents...
creating dist
creating 'dist\fakerez-0.0.1-py2.7.egg' and adding 'build\bdist.win-amd64\egg' to it
removing 'build\bdist.win-amd64\egg' (and everything under it)
Processing fakerez-0.0.1-py2.7.egg
Copying fakerez-0.0.1-py2.7.egg to c:\python27\lib\site-packages
Adding fakerez 0.0.1 to easy-install.pth file
Installing fakerez-env-script.py script to C:\Python27\Scripts
Installing fakerez-env.exe script to C:\Python27\Scripts

Installed c:\python27\lib\site-packages\fakerez-0.0.1-py2.7.egg
Processing dependencies for fakerez==0.0.1
Finished processing dependencies for fakerez==0.0.1

C:\Users\007br\workspace\fakerez>set path=C:\Python27\Scripts

C:\Users\007br\workspace\fakerez>fakerez-env
Doing env

C:\Users\007br\workspace\fakerez>set path
Path=C:\Python27\Scripts
PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC
```

mottosso:
> It will if you're using setuptools-generated console_scripts entry points. setuptools will embed the path to the python interpreter the generated executables were created with

This has been my understanding as well, but having had a look just now, it doesn't appear to be the case, at least not on Windows.

**rez.exe/__main__.py**

```python
# -*- coding: utf-8 -*-
import re
import sys

from rez.cli._main import rez

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(rez())
```

But having done exactly this and seen it work in practice, I'm a little confused. I suspect this is what's going on, in your case too @bpabel:

1. The `rez.exe` binary is associated with a `py.exe`, the [Python Launcher](https://www.python.org/dev/peps/pep-0397/)
2. `py.exe` is in turn on PATH and references its corresponding `python.exe`
3. Hence, calling `rez.exe`, works despite `python.exe` not being on PATH

But that's an admittedly narrow requirement. I've got many versions of Python installed on my system, and I'm unsure of which `py.exe` is referencing. I think it must be the last one installed, as the installer gives you the option to install this launcher.

On Linux, does it not include a shebang the Python interpreter? I think I'm starting to see now why these executables were created by hand using `install.py`

instinct-vfx:
On windows with python 2.7 setuptools uses pre-compiled wrapper executables that are than patched to call the corresponding scripts using a hardcoded python interpreter. We actually have a patched python build that replaces theses pre-compiled wrappers with ones that only call python.exe instead (so it only has to be somewhere on PATH). But as you can imagine, that is its own can of worms.

bpabel:
@mottosso 

I'm using python 2.7 without a `py.exe`.  I can also copy the generated entry point script and exe into any directory and add it to the path and it will still work.

```
c:\temp>dir /B
fakerez-env-script.py
fakerez-env.exe

c:\temp>set PATH=C:\temp

c:\temp>set path
Path=C:\temp
PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC

c:\temp>fakerez-env
Doing env

```

mottosso:
> I'm using python 2.7 without a py.exe. I can also copy the generated entry point script and exe into any directory and add it to the path and it will still work.

Yes, I am too, this is why I'm confused. If you look inside of `fakerez-env.exe` (it's a zip file), you'll find no mention of any hardcoded path. :S

bpabel:
Well, it's a zip file, but all the embedded files are binary, so the hardcoded path is probably in there somewhere, I just can't read it with a text editor.

mottosso:
Yes, you are right. I just had a closer look. At the very top there, a hardcoded reference. Win!


```hex
00018ff0:  0000 0000 0000 0000 0000 0000 0000 0000 2321 633a 5c75 7365  :................#!c:\use
00019008:  7273 5c6d 616e 696d 615c 6472 6f70 626f 785c 6465 765c 616e  :rs\manima\dropbox\dev\an
00019020:  696d 615c 7265 7a5c 7363 7269 7074 735c 7079 7468 6f6e 2e65  :ima\rez\scripts\python.e
00019038:  7865 0a0d 0a50 4b03 0414 0000 0000 008a 819e 4ea8 bfd5 caca  :xe...PK...........N.....
00019050:  0000 00ca 0000 000b 0000 005f 5f6d 6169 6e5f 5f2e 7079 2320  :...........__main__.py# 
00019068:  2d2a 2d20 636f 6469 6e67 3a20 7574 662d 3820 2d2a 2d0a 696d  :-*- coding: utf-8 -*-.im
00019080:  706f 7274 2072 650a 696d 706f 7274 2073 7973 0a0a 6672 6f6d  :port re.import sys..from
00019098:  2072 657a 2e63 6c69 2e5f 6d61 696e 2069 6d70 6f72 7420 7265  : rez.cli._main import re
000190b0:  7a5f 656e 760a 0a69 6620 5f5f 6e61 6d65 5f5f 203d 3d20 275f  :z_env..if __name__ == '_
000190c8:  5f6d 6169 6e5f 5f27 3a0a 2020 2020 7379 732e 6172 6776 5b30  :_main__':.    sys.argv[0
000190e0:  5d20 3d20 7265 2e73 7562 2872 2728 2d73 6372 6970 745c 2e70  :] = re.sub(r'(-script\.p
000190f8:  7977 3f7c 5c2e 6578 6529 3f24 272c 2027 272c 2073 7973 2e61  :yw?|\.exe)?$', '', sys.a
00019110:  7267 765b 305d 290a 2020 2020 7379 732e 6578 6974 2872 657a  :rgv[0]).    sys.exit(rez
00019128:  5f65 6e76 2829 290a 504b 0102 1400 1400 0000 0000 8a81 9e4e  :_env()).PK.............N
00019140:  a8bf d5ca ca00 0000 ca00 0000 0b00 0000 0000 0000 0000 0000  :........................
00019158:  8001 0000 0000 5f5f 6d61 696e 5f5f 2e70 7950 4b05 0600 0000  :......__main__.pyPK.....
00019170:  0001 0001 0039 0000 00f3 0000 0000 00                        :.....9.........
```

instinct-vfx:
If you rename to zip and open it you only see a python script. But that is not the whole story of course (as you can not execute a zip file that in turn executes python obviously). If you open the exe directly in a text editor and search for "python" you will find it.

mottosso:
Does this also apply on Linux? :O

instinct-vfx:
The template wrappers that setuptools uses are here:

Lib\site-packages\setuptools

called cli.exe and gui.exe + various platform variants.

nerdvegas:
I'd like to test this on Linux soon, will get back with results. If it does
work and results in stable tools unaffected by things as I've mentioned
earlier, then this is a big win.

I still need to stress that the existing installer has to stay in place
though, until we know from real prod testing at multiple studios that this
works.

A

nerdvegas:
Something to note that fell to the wayside in all this. Many studios
install rez on shared storage (I would say most in fact). That means they
won't be able to use a pip-based install (mind you, I am still all for
having one that works).

So, install.py will have to stay, though I hope it can become a standard
virtualenv-based installation and we can do away with the patching. But to
support this alongside the current installer, we may have to add an
--experimental arg.

If and when this does happen, I can roll it out at my studio and it will be
very heavily tested - we run more than a million resolves a day.

A

bpabel:
> Something to note that fell to the wayside in all this. Many studios
install rez on shared storage (I would say most in fact). That means they
won't be able to use a pip-based install (mind you, I am still all for
having one that works).

Pip supports installing to non-standard (network) paths

```
pip install PACKAGE --root /shared/path
pip install PACKAGE --prefix /shared/path
```

mottosso:
> That means they won't be able to use a pip-based install

As Brendan mentions, you can install packages using pip anywhere, they just happen to default into the target Python `site-packages` directory (or somewhere in ~/ if passing `--user`).

But furthermore, if what you're looking for (the user, not you specifically) is a virtualenv, then it would get installed wherever the virtualenv was residing.

```bash
$ python -m virtualenv /some/location
$ . /some/location/bin/activate
(some_location) $ pip install rez
```

nerdvegas:
Yeah, that may be good enough. Let's see how it pans out. I won't get a
chance to test entry point behaviour on linux for a couple of days, but
it's coming.

A

instinct-vfx:
I am not sure how relevant this is here, but keep in mind that using --target does not install everything a normal pip install installs. E.g. executable wrappers generated by setuptools are omitted and most likely everything that is installed outside of site-packages.

JeanChristopheMorinRodeoFX:
Hey guys, I haven't replied since a couple of days on this, but I'm more than happy if we can get rid of the hacks we have in place and simplify the installer. But I think we still need an installer. Yes, pip is cool, but having a simple way to install something is always nice. I think someone shouldn't have to know pip to work with rez. And even with good documentation, the overhead is a little higher when you need to create a virtualenv by hand and then install rez from within the virtualenv if you don't know much about virtualenv and pip.

And I'd still like to see what we can do with the vendored dependencies. I don't know what you guys think but I'm not a big fan of vendored dependencies...

mottosso:
> And I'd still like to see what we can do with the vendored dependencies. I don't know what you guys think but I'm not a big fan of vendored dependencies...

Would recommend opening up a separate issue about this.

JeanChristopheMorinRodeoFX:
Yeah, I wasn't implying that it should be included in this PR here. I was more mentioning that because it will probably affect the installer in some ways if we review the vendored dependencies. I'll keep watching the progress of your PR and will see later for creating an issue.

nerdvegas:
I agree that it'd be good to get rid of the embedded dependencies, and that
that should be dealt with separately. I think we can assess this if and
when the new installation is in place. I also think that this should be
done carefully - first step would be removing the most straightforward of
these dependencies, in lieu of its equivalent native package installed into
rez's site (virtualenv or whatever the case may be).

RE an install.py, agreed that this should still exist, even if it becomes
not much more than a utility for performing the relevant virtualenv install
(for eg). I don't want people to waste their time thinking about what the
right install configuration should be for production use - better that this
is codified in the form of an installation script. Bear in mind that rez is
not a project specifically for managing python packages, so it should not
be a requirement to be familiar with pip - it's just as valid for a
C++-orientated developer to want to install it, for eg.

A

nerdvegas:
So I've just run a really simple test, and have verified that the issue we hit several years ago is still there. That is, PYTHONPATH is still used in a virtualenv, and takes precedence over any packages installed into the venv itself. It's very easy to break a binary within a virtualenv like this.

I used this sample project: https://github.com/pypa/sampleproject. Then:

1. create a virtualenv
2. use that venv's python to `python setup.py bdist_wheel` to create the sampleproject whl
3. install the whl with the venv's pip
4. test it by running the 'sample' binary in the venv (it works)
5. create a temp dir, with 'sample' subdir and empty __init__.py within
6. export PYTHONPATH=(this temp dir)
7. Run 4 again. It explodes. 

I installed via wheel because that removes the setuptools dependency in the 'sample' binary (one less thing to worry about).

So in any case, this doesn't work. I do have an idea now though - it might be enough to have the rez entry points do this:

1. if _REZ_PATCHED_ENV envvar is defined, run as normal. ELSE:
2. exec a proc (replace existing) with same sys args, but in environ with PYTHONPATH removed, and _REZ_PATCHED_ENV set.

Will try this out soon.
A

bpabel:
Using those steps, can confirm the same error happens on windows and mac.

Another possible idea.  You can run the python executable with the `-E` flag to ignore the PYTHONPATH and other env var settings.  Maybe it's possible to patch the entry points executable to include that flag?

JeanChristopheMorinRodeoFX:
The installer was also adding `-E` to the script's shebang (see https://github.com/nerdvegas/rez/blob/master/install.py#L80).

I don't know if it's possible to customize the entry point scripts without adding custom code as we do in the installer right now... But here is the doc (https://distlib.readthedocs.io/en/latest/tutorial.html#specifying-scripts-to-install) in case it can help.

bpabel:
> The installer was also adding -E to the script's shebang

Doh! Of course, should have checked there first.

It looks like setuptools supports adding [post-install scripts](https://stackoverflow.com/questions/20288711/post-install-script-with-python-setuptools)

Perhaps just moving the patching code from `install.py` into `setup.py` as a post-install routine?  I haven't used that feature before though.  Will have to check how it works when generating wheels.

nerdvegas:
Yes, getting the -E in there is the reason we have patched distlib/virtualenv in the first place.

Correct me if I'm wrong but I believe that patching the executables in a post-install routine would be problematic on windows, since these are .exes at that point (rather than a simple shebanged script as per linux).

I was thinking about this some more, and I'd like to try out the re-exec idea I mentioned earlier, but only where PYTHONPATH is non-empty. That may do the trick, not sure.

On a related note, I recall why we have those embedded dependencies now (vendor/ dir) - or at least, a major reason why. It's because it is a common need to install rez as a rez package - but if rez had external dependencies, then those would also have to be installed as rez packages. To do this would mean either rez itself being able to install packages and their dependencies (which it cannot do yet, for a multitude of reasons, which I want to go over in the roadmap); or, we would have to ensure that rez-pip worked to the extent needed to do this - and that's another complex piece of code that then becomes a potential barrier to installing the rez rez package. Also, at the time, rez-pip did not exist.

JeanChristopheMorinRodeoFX:
>Correct me if I'm wrong but I believe that patching the executables in a post-install routine would be problematic on windows, since these are .exes at that point (rather than a simple shebanged script as per linux).

Technically, if I understand things correctly, distlib supports Windows for this. At least there is no warning about Windows in there doc for the shebang. I would guess they do some dark magic (I don't know much about Windows, so everything is dark magic to me :stuck_out_tongue: ) ... To be verified I guess.

>I was thinking about this some more, and I'd like to try out the re-exec idea I mentioned earlier, but only where PYTHONPATH is non-empty. That may do the trick, not sure.

From what I know, `exec` caused us a lot of headaches in the past and we've put some rules where nobody can use them. If I remember correctly, processes that were spawned with `exec` where losing their parent or something, and then if the process was killed, the exec'ed process would become detached and the process was never really killed. Or something like that. So for RodeoFX, if we can avoid `exec`, that would be great. No idea if you guys had problems in the past with this sort of things.

As for the vendored dependencies, I was looking at still having them, but not as we have right now. The archive that we provide (wheel or whatever the format is) would contain the vendored dependencies, but they wouldn't be in git. I'll open an issue so we can discuss future avenues? What do you think?

nerdvegas:
JC, yeah I see what you mean RE vendored deps. I do wonder if this should
be left til after any potential changes to installation though, as one may
affect the other.

Thx for the info re exec, will have to look into that. I suppose it
wouldn't have to be an exec - it could just be a subprocess if that avoids
the issues you mention.

A

nerdvegas:
So I've done some testing using execve + unsetting PYTHONPATH from environ, and it seems to work so far. JC, I know you said you've had issues with exec, any chance you could find out more? Otherwise I'll go ahead with it anyway and if there are issues we'll find out. At that point it shouldn't be hard to switch to a subprocess.

Right now I'm thinking it shouldn't be too hard to put together another PR drawing from this one, but implementing this exec fix and keeping compatibility with the current installer. Then we can put it through some real studio testing and see what breaks.

Ps - other evars that can affect rez/python need to be unset also - eg LD_LIBRARY_PATH and LD_PRELOAD. What this won't fix is if someone provides a shared lib on LD_LIBRARY_PATH that overrides one of python's own. I don't know if anyone out there has hit this situation. The libs in question are (Ub 16.04):

linux-vdso.so.1
libpthread.so.0
libc.so.6
libdl.so.2
libutil.so.1
libz.so.1
libm.so.6
/lib64/ld-linux-x86-64.so.2

mottosso:
Nice find, guys!

> Right now I'm thinking it shouldn't be too hard to put together another PR drawing from this one

I'd be happy to implement this here? You can also push directly to this PR, as I've enabled maintainer push privileges.

nerdvegas:
Hey Marcus, no worries if you'd like to tackle it in this PR.

To expand a little on what I've been thinking:

* install.py has a new --experimental arg
* existing setup.py is moved to setup_old.py; new setup.py for pip goodness is added
* install.py switches between the two depending on --experimental
* the check venvs + execve is done in one place that every entry point hits, from memory I think this is rez/cli/_main.py:run?
* we still need to think about what it means if everything in the venv/bin dir is visible on PATH. This might be ok if we ensure that its at the end of PATH in a rez-resolved env. Right now it's probably at the _start_, because the current bin/rez folder only contains the rez binaries. Bear in mind though that we could still hit roadblocks related to this, and it may be necessary to wrangle the venv a bit and create that rez subdir still.

mottosso:
Cool, having a look at this, and a few more ideas.

1. Revert to using the `scripts=` argument, and provide `.sh` and `.bat` files, primed with `python -E`
2. Since (1) isn't nice on Windows (i.e. #626), have a look at [Shimmy](https://github.com/pfmoore/shimmy) for making `.exe`
3. Since (2) is based on Pip's shims which may then also include #616, consider https://github.com/71/scoop-better-shimexe that addresses this problem

Was there any issue with installing `.sh` scripts on Linux?

nerdvegas:
I feel like we should see how the current approach goes before looking at
alternatives. The shimming that looks to be necessary on Windows could end
up a complicated problem that the current approach may avoid completely.

(1) might work, but the scripts would have to be able to detect if they're
in a venv, and then use the appropriate interpreter. But yeah, linux only.

Thinking about it more and looking back at current installer.py, I think we
still need to perform the step where rez binaries are copied into their own
dir also. There's just too much possibility that a resolved env would
accidentally end up using the venv's python/pip. Consider especially a
resolved env that doesn't contain any python packages. In that case,
running 'python' would definitely use the venv's python, which would be bad
and unexpected (rez shouldn't change anything unrelated to the packages
asked for).

Cheers
A

mottosso:
Having a look at this, but not sure I'm doing it right.

> I'd like to try out the re-exec idea I mentioned earlier, but only where PYTHONPATH is non-empty. That may do the trick, not sure.

Is this what you mean?

**rez.exe/sh/bat**

```python
import os
import subprocess

# Clear PYTHONPATH
environ = os.environ.copy()
environ.pop("PYTHONPATH")

# Re-launch Rez with clean env
subprocess.Popen(["rez"], env=environ)
```

If so, then I'm not sure this solves the problem, because in order to get to this point in the Python script, you would first need to find it on the PYTHONPATH/site-packages, which is what the original problem was? E.g. your step (6) and (7) from the sampleproject example.

E.g.

```bash
$ mkdir rez  # fake rez
$ touch rez/__init__.py
$ export PYTHONPATH=$(pwd)
$ python -m rez  # Wrong Rez picked up
```

And if the wrong Rez is picked up, we couldn't run our code to re-exec a clean Rez. Let me know if I got this wrong somehow. You could also have meant to do this in a `.sh` or `.bat` script, in which case it might work, but would eliminate use of `entry_points` from `setup()` (which isn't necessarily a bad thing).

nerdvegas:
In essence yes this is the approach, although I'd try an os.execve instead
first. The proc tree in a rez-env is already a bit deep. JC did mention
that they had run into issues exec'ing, but it'd be good to find out for
sure what was involved there, we can always switch to subproc easily enough
if we hit issues.

This approach works if both the rez installation, and the rez rez package,
are in on the game, and both run the same exec'ing code. Of course it is
possible for someone to put a fake rez on PYTHONPATH and break it, but this
is such a narrow case that I'm not concerned by it - if someone wants to go
out of their way to break the system by faking a rez python module, then
whatever! The bigger concern is that current version rez rez package won't
work in combination with this new version. I think we can live with this
caveat because it's a one-off, so long as this incompatibility is made
clear.

mottosso:
> Of course it is possible for someone to put a fake rez on PYTHONPATH and break it

Hold up, I thought this was the issue you were demonstrating with sampleproject? Of calling `sample.exe` when another `sample/__init__.py` was on the PYTHONPATH?

If you're OK with Rez clashing, then perhaps there's an even simpler method that doesn't involve more processes.

```python
import sys
sys.path[:] = []
```

That way, regardless of what was on PYTHONPATH at boot is swept away. We could do this at the start of e.g. `rez/cli/_main.py` such that any subsequent import would see nothing but Python's `lib/` directory (not even `site-packages/`).

While we're at it, we could clean up the environment too.

```python
import os
os.environ.pop("PYTHONPATH", None)
os.environ["PATH"] = which("bash")  # on Linux
os.environ["PATH"] = which("cmd")  # on Windows
```

> The bigger concern is that current version rez rez package won't work in combination with this new version.

You mean as it exists at the moment in this PR? Because this is how I'm using it here and haven't noticed any issues.

nerdvegas:
On Fri, Jun 14, 2019 at 4:31 PM Marcus Ottosson <notifications@github.com>
wrote:

> Of course it is possible for someone to put a fake rez on PYTHONPATH and
> break it
>
> Hold up, I thought this was the issue you were demonstrating with
> sampleproject? Of calling sample.exe when another sample/__init__.py was
> on the PYTHONPATH?
>
I just happened to use sample itself to demonstrate the issue. I want to
get to a point where all those embedded vendor deps can just be standard
packages in the rez venv instead, and in that scenario it's more likely
that one of _those_ packages could be interfered with in this way. But that
aside, I know that PYTHONPATH was causing issues in the past that I'm
pretty sure weren't as simple as rez itself being involved (I wish I could
recall one of the cases we hit).

> If you're OK with Rez clashing, then perhaps there's an even simpler
> method that doesn't involve more processes.
>
> import sys
> sys.path[:] = []
>
> That way, regardless of what was on PYTHONPATH at boot is swept away. We
> could do this at the start of e.g. rez/cli/_main.py such that any
> subsequent import would see nothing but Python's lib/ directory (not even
> site-packages/).
>
I'm not sure that's going to cover other cases such as LD_LIBRARY_PATH on
linux. Even if it works presently, it could cause problems later if and
when rez uses any compiled packages (to put that in context, a C++-based
implementation of the version submodule could speed up the solver
substantially). I also doubt that altering os.environ would do anything for
those cases. We'd have to test to be 100% sure of that though, but library
linking is something that happens outside of the interpreter, I would think
that os.environ affects the interpreter only.

Also, os.execve won't introduce more processes, it will just replace the
current one. If doing this doesn't cause other issues then it'll probably
be the most robust way of removing env vars that can alter rez's behaviour.

While we're at it, we could clean up the environment too.
>
> import os
> os.environ.pop("PYTHONPATH", None)
> os.environ["PATH"] = which("bash")  # on Linux
> os.environ["PATH"] = which("cmd")  # on Windows
>
> I don't think you'd need to touch PATH, and in fact this is too
simplistic, on linux at least... it's common to have multiple standard
systems paths on PATH, not just the one.

> The bigger concern is that current version rez rez package won't work in
> combination with this new version.
>
> You mean as it exists at the moment in this PR? Because this is how I'm
> using it here and haven't noticed any issues.
>
Sorry by 'current' I meant any existing rez versions that studios have
installed as a rez package. Both the installed version and the rez rez
package need to have this same bootstrapping code in order for this to work.

mottosso:
Ok, I see what you mean.

For completeness, these are the unknowns that we're working with at the moment.

- [ ] PYTHONPATH was causing issues in the past that I'm pretty sure weren't as simple as rez itself being involved
- [ ] I'm not sure that's going to cover other cases such as LD_LIBRARY_PATH on linux. 
- [ ] I also doubt that altering os.environ would do anything for those cases. 
- [ ] JC had issues with exec (`os.execve` or `subprocess.call("exec")`?)

For pure-Python packages, editing `environ` and `sys.path` would have an effect on imported modules, but I think you're right that `os.environ` may not be what compiled modules are actually looking at, and that Python has an internal `os.environ` which we aren't able to edit as easily.

In either case, better safe than sorry; once we've have a working version covering the unknowns, we can start to peel away at them one at a time.

nerdvegas:
Yup, those were my thoughts also.

Cheers
A

mottosso:
Having a look at this with `execve` but it isn't looking good on Windows, which doesn't appear to actually replace the process, resulting in the calling shell acting as though the child Python has exited and starts accepting input at the same time.

- https://stackoverflow.com/questions/7004687/os-exec-on-windows

Instead, I used `subprocess.Popen` despite then creating one more level deep in the process hierarchy.

To test out before/after, I also added `REZ_UNSAFEMODE`, where "unsafe" means allowing for PYTHONPATH to follow along into the context. Without it, this extra level of processes is created.

```bash
$ rez env  # pid 1000
> $ # pid 1001
> $ exit  # pid 1001 dies
# pid 1000 dies
$
```

```bash
$ export REZ_UNSAFEMODE=1
$ rez env  #pid 1000
> $ # still pid 1000
```

At least two things to improve here.

1. Is there a way to avoid another level in the hierarchy? I attempted a "fire-and-forget" for `Popen`, whereby I just close the parent Python process, but like with `execve` it causes the parent shell to think it finished and start accepting input again which isn't great.
2. Is there another way of getting the new Python to launch `rez`? At the moment, since PYTHONPATH is removed, I launched the subprocess in the parent directory of the rez Python package, leveraging Python's ability to pick up modules from the current working directory. It's relatively safe, so long as Rez doesn't reside amongst other packages (like it would if installing in a non-virtual environment), but means Rez loses hold of what it's actual current working directory is. Perhaps a better alternative is importing the module by absolute path, via e.g. `importlib.find_module`?

nerdvegas:
RE execve and REZ_UNSAFEMODE:

I fully expect there may be differences between platforms. For example, Linux may be fine with execve, and if so we should use it. I also think we may find ourselves in need of multiple settings that relate to this bootstrap-level stuff.

I have a suggestion: we should have one _REZ_BOOTSTRAP_CONFIG evar, which is a JSON-encoded dict. Also, for each platform, we have a hardcoded dict which contains the defaults for these settings. In _main, this dict is updated with anything from _REZ_BOOTSTRAP_CONFIG.

So for eg:

```
if platform.system.lower() == "linux":
    bootstrap_conf = {
        "use_execve": True,
        "inherit_pythonpath": False
    }
elif platform.system.lower() == "windows":
    bootstrap_conf = {
        "use_execve": False,
        "inherit_pythonpath": False
    }
```

This will avoid a bunch of extra evars if we find we need further control; it also gives us a nice analogy to the main config (which we can't use in this case, of course). It's a little technical to have to provide JSON, but in this case I think it's good - anyone changing this stuff should be fairly technical.

In terms of whether or not you can avoid the extra subproc on Windows, I have no idea :)

"""Is there another way of getting the new Python to launch rez"""

I would have thought you'd just be using the exact same sys.argv again? Presumably this would then just invoke the very same entry point again. Is there a reason you aren't doing this?

mottosso:
> For example, Linux may be fine with execve, and if so we should use it

Oh, I would have thought you'd prefer consistency, despite platform. If subprocess works on Linux too, then why implement it twice? They're bound to have issues of their own. Are you sure?

> I have a suggestion: we should have one _REZ_BOOTSTRAP_CONFIG evar

That's a good idea, once related variables start growing that's one way of handling it.

> I would have thought you'd just be using the exact same sys.argv again? Presumably this would then just invoke the very same entry point again. Is there a reason you aren't doing this?

It does forward the arguments, but it needs to take into account the calling executable which isn't Python itself, but rather the compiled executable.

```bash
$ rez env  # print(sys.argv)
['C:\\Users\\marcus\\rez-virtualenv\\Scripts\\_rez\\rez.exe', 'env']
```

In the current commit, it's replacing `rez.exe` with `python -m rez` and forwards the rest.

nerdvegas:
On Tue, Jun 18, 2019 at 10:47 PM Marcus Ottosson <notifications@github.com>
wrote:

> For example, Linux may be fine with execve, and if so we should use it
>
> Oh, I would have thought you'd prefer consistency, despite platform. If
> subprocess works on Linux too, then why implement it twice? They're bound
> to have issues of their own. Are you sure?
>
If we have to use an extra subprocess on Windows then that's just what we
have to live with. I suspect that on linux, using execve instead will work
and will avoid the extra subprocess, and I think this is worth the cost in
consistency. At the end of the day though, what matters is that the
behaviour is as expected. Bear in mind that this only affects invocation of
the rez cli tools, certainly if the effect was more widespread I would have
more reservations about differing implementation across compilers.

> I have a suggestion: we should have one _REZ_BOOTSTRAP_CONFIG evar
>
> That's a good idea, once related variables start growing that's one way of
> handling it.
>
> I would have thought you'd just be using the exact same sys.argv again?
> Presumably this would then just invoke the very same entry point again. Is
> there a reason you aren't doing this?
>
> It does forward the arguments, but it needs to take into account the
> calling executable which isn't Python itself, but rather the compiled
> executable.
>
> $ rez env  # print(sys.argv)
> ['C:\\Users\\marcus\\rez-virtualenv\\Scripts\\_rez\\rez.exe', 'env']
>
> In the current commit, it's replacing rez.exe with python -m rez and
> forwards the rest.
>
I understand that that's what you're doing, I just am not clear as to why.
If you clear PYTHONPATH and then start a subproc with the exact same argv,
does that not work? Perhaps I'm missing something.

nerdvegas:
Closing as we're already publishing to pypi via gh workflow now

Issue #575, opened 2019-02-13
Incompatible windows-os package between local job and cloud render


wizofe:
When I generate a job from [deadline](https://deadline.thinkboxsoftware.com/) that renders on the cloud, I get the `os-windows-xxx package` is incompatible. There's a different version of Windows running in the Cloud and a different one locally (Server vs 7).

The version of `rez` currently running is the 2.0 (I know it isn't the latest but there are a lot of systems to upgrade). The weird thing is that the job is running when executed locally but not when sent to the cloud. Could that be an `os` variant definition in some `package.py`? Or the issue that is fixed by version [2.26.0](https://github.com/nerdvegas/rez/releases/tag/2.26.0)?

Thanks!

instinct-vfx:
You can use the platform-map configuration option to map versions to be less specific:
https://github.com/nerdvegas/rez/wiki/Configuring-Rez#platform_map

Does that maybe help?

wizofe:
Didn't really help. Getting still ... intersected with range ... resulted in no packages.

nerdvegas:
Hi,

There isn't enough info to go on really. What is the error you get
specifically? By 'incompatible' do you mean that you get a resolve conflict?

If so, you could try simply manually setting REZ_IMPLICIT_PACKAGES to
include the windows OS variant present on cloud, and do a resolve locally.
If that fails then there must be some package with a requirement on
windows-7.

I'm not sure what potentially-fixed issue you're referring to?

Thx
A

wizofe:
Ok. Here's a log [LOG](https://paste.ofcode.org/373tZVZEmHFPi66HGXQN7ZX) if I try to create a `rez env` on the cloud.

The issue is that when I submit a job from the local deadline manager to the cloud I can't resolve the environment and the job fails. Note: the packages are read from a central server location if that matters.

Here's [LOG1](https://paste.ofcode.org/KfPPut5i6jdrVExY2CMT2Q) and I am going to attach the LOG2 when I have it (that shows the message of intersected packages instead).

PR #628, opened 2019-05-17
[Feature] Pure python package detection

lambdaclan:
# Description

[Depends on PR #602, #654 #656]

## Update this PR has been split into 3 different PRs

**Since all 3 PRs are based on #656, it might make sense to resolve that one first.**

1. This PR #628 for pure python package detection

2. #657 for the experimental pip QOL commands

3. #658 for the schema and package request error fixes

### Pure python package detection

As we embrace modern Python and try to improve the way rez interfaces with pip by enforcing
Wheel package installation rather than the aging and legacy source and Egg distributions we can further improve rez-pip's functionality by taking advantage of the rich metadata that comes with Wheel distributions.

This PR introduces a purity detector for Python packages. As defined on the [Wheel Binary Package Format PEP491](https://www.python.org/dev/peps/pep-0491)

> Wheel preserves the "purelib" vs. "platlib" distinction, which is significant on some platforms. For example, Fedora installs pure Python packages to '/usr/lib/pythonX.Y/site-packages' and platform dependent packages to '/usr/lib64/pythonX.Y/site-packages'.

> A wheel with "Root-Is-Purelib: false" with all its files in {name}-{version}.data/purelib is equivalent to a wheel with "Root-Is-Purelib: true" with those same files in the root, and it is legal to have files in both the "purelib" and "platlib" categories.

> **In practice a wheel should have only one of "purelib" or "platlib" depending on whether it is pure Python or not and those files should be at the root with the appropriate setting given for "Root-is-purelib".**

This addition enables us to streamline  variant requirements:
 > TODO detect if platform/arch/os necessary, no if pure python

To further ensure compatibility with current platform this PR also adds some *experimental* features
such as package version listing, querying and Python version constraint based search. These features
use the most "reliable" source of package data being the PyPi via the new [JSON API](https://warehouse.readthedocs.io/api-reference/json/). 

The reason these features are marked experimental is because currently there is no standard of how the metadata should be defined and sometimes it varies widely between packages. According to my tests it seems to be working for the most part but still a lot of things to consider.

Relates # (issue)
rez #610, #575  maybe more...

## Breakdown

* pure/impure package detection

* pure packages now only use major Python version (for example python-2)

* drop os version dependency all together platform / architecture / (os) / python for all packages (pure and impure)

## Process

* Python package purity detector
  * this is using the Wheel metadata file named WHEEL. WHEEL is the wheel metadata specific to a build of the package. The Wheel metadata contents are defined [here](https://www.python.org/dev/peps/pep-0491/#file-contents)

## Type of change

## Remarks

* Requires pip >=19.0
* Requires  distlib >= 0.28

# How Has This Been Tested?

I have installed a plethora of packages from known to obscure  in order to check the
purity detector with good success rate.
Some issues still occur (unable to build a wheel - very rare) depending on the platform but the same problems would also occur with the current implementation.

**Test Configuration**:
* OS: Linux, Windows 10

## Todos
- [ ] Tests

## Demo

* Pure / Impure package detector

purity

### DIFF
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -19,7 +19,9 @@ from rez.system import System
 from tempfile import mkdtemp
 from StringIO import StringIO
 from pipes import quote
+from email.parser import Parser
 import subprocess
+import pkg_resources
 import os.path
 import shutil
 import sys
@@ -198,7 +200,7 @@ def pip_to_rez_package_name(distribution):
     return name
 
 
-def run_pip_command(command_args, pip_version=None, python_version=None):
+def run_pip_command(command_args, process_output=False, pip_version=None, python_version=None):
     """Run a pip command.
 
     Args:
@@ -210,7 +212,17 @@ def run_pip_command(command_args, pip_version=None, python_version=None):
     pip_exe, context = find_pip(pip_version, python_version)
     command = [pip_exe] + list(command_args)
 
-    if context is None:
+    if process_output:
+        try:
+            subprocess.check_output(command, stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            regex = r"(?<=\(from versions:).*?(?=\))"
+            pattern = re.compile(regex)
+            match = pattern.search(e.output)
+            output = match.group(0).split(",")
+            return output
+
+    elif context is None:
         return popen(command)
     else:
         return context.execute_shell(command=command, block=False)
@@ -335,6 +347,7 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
     """
     installed_variants = []
     skipped_variants = []
+    variant_types = []
 
     pip_exe, context = find_pip(pip_version, python_version)
 
@@ -370,6 +383,20 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
 
     _cmd(context=context, command=cmd)
     _system = System()
+
+    def pure_python_package(installed_dist):
+
+        true_table = {
+            "true": True,
+            "false": False
+        }
+
+        packages = pkg_resources.find_distributions(destpath)
+        dist = next((package for package in packages if package.key == installed_dist.key), None)
+        wheel_data = dist.get_metadata('WHEEL')
+        wheel_data = Parser().parsestr(wheel_data)
+
+        return true_table[wheel_data["Root-Is-Purelib"]]
 
     # Collect resulting python packages using distlib
     distribution_path = DistributionPath([destpath])
@@ -446,18 +473,29 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
                     shutil.copystat(source_file, destination_file)
 
         # determine variant requirements
-        # TODO detect if platform/arch/os necessary, no if pure python
         variant_reqs = []
-        variant_reqs.append("platform-%s" % _system.platform)
-        variant_reqs.append("arch-%s" % _system.arch)
-        variant_reqs.append("os-%s" % _system.os)
+
+        pure = pure_python_package(distribution)
+        if not pure:
+            variant_types.append("non-pure")
+            variant_reqs.append("platform-%s" % _system.platform)
+            variant_reqs.append("arch-%s" % _system.arch)
+        else:
+            variant_types.append("pure")
 
         if context is None:
             # since we had to use system pip, we have to assume system python version
-            py_ver = '.'.join(map(str, sys.version_info[:2]))
+            if pure:
+                py_ver = '.'.join(map(str, sys.version_info[:1]))
+            else:
+                py_ver = '.'.join(map(str, sys.version_info[:2]))
         else:
             python_variant = context.get_resolved_package("python")
-            py_ver = python_variant.version.trim(2)
+
+            if pure:
+                py_ver = python_variant.version.trim(1)
+            else:
+                py_ver = python_variant.version.trim(2)
 
         variant_reqs.append("python-%s" % py_ver)
 
@@ -487,7 +525,7 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
     # cleanup
     shutil.rmtree(tmpdir)
 
-    return installed_variants, skipped_variants
+    return installed_variants, skipped_variants, variant_types
 
 
 def _cmd(context, command):
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -200,29 +200,17 @@ def pip_to_rez_package_name(distribution):
     return name
 
 
-def run_pip_command(command_args, process_output=False, pip_version=None, python_version=None):
+def run_pip_command(command_args, pip_version=None, python_version=None):
     """Run a pip command.
-
     Args:
         command_args (list of str): Args to pip.
-
     Returns:
         `subprocess.Popen`: Pip process.
     """
     pip_exe, context = find_pip(pip_version, python_version)
     command = [pip_exe] + list(command_args)
 
-    if process_output:
-        try:
-            subprocess.check_output(command, stderr=subprocess.STDOUT)
-        except subprocess.CalledProcessError as e:
-            regex = r"(?<=\(from versions:).*?(?=\))"
-            pattern = re.compile(regex)
-            match = pattern.search(e.output)
-            output = match.group(0).split(",")
-            return output
-
-    elif context is None:
+    if context is None:
         return popen(command)
     else:
         return context.execute_shell(command=command, block=False)
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -335,7 +335,6 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
     """
     installed_variants = []
     skipped_variants = []
-    variant_types = []
 
     pip_exe, context = find_pip(pip_version, python_version)
 
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -382,8 +382,10 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
         packages = pkg_resources.find_distributions(destpath)
         dist = next((package for package in packages if package.key == installed_dist.key), None)
         wheel_data = dist.get_metadata('WHEEL')
+        # see https://www.python.org/dev/peps/pep-0566/#json-compatible-metadata
         wheel_data = Parser().parsestr(wheel_data)
 
+        # see https://www.python.org/dev/peps/pep-0427/#what-s-the-deal-with-purelib-vs-platlib
         return true_table[wheel_data["Root-Is-Purelib"]]
 
     # Collect resulting python packages using distlib
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -465,11 +465,8 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
 
         pure = pure_python_package(distribution)
         if not pure:
-            variant_types.append("non-pure")
             variant_reqs.append("platform-%s" % _system.platform)
             variant_reqs.append("arch-%s" % _system.arch)
-        else:
-            variant_types.append("pure")
 
         if context is None:
             # since we had to use system pip, we have to assume system python version
@@ -513,7 +510,7 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
     # cleanup
     shutil.rmtree(tmpdir)
 
-    return installed_variants, skipped_variants, variant_types
+    return installed_variants, skipped_variants
 
 
 def _cmd(context, command):


mottosso:
Nice one, how about we merge our efforts on pip into one great PR? There's some slight duplication of effort at the moment, and with the number of PRs growing and nothing seeming to get merged, at least anyone interested in improving their pip experience would know which branch to merge locally.

- https://github.com/nerdvegas/rez/pull/595
- https://github.com/nerdvegas/rez/pull/596
- https://github.com/nerdvegas/rez/pull/614

How about I merge the above into this one, and then we continue from there?

lambdaclan:
Helle Marcus, sure thing this is the beauty of working on open source projects, the ability to collaborate. What is important is to get the improvements merged into master so that everyone can benefit from them. Saying that I am sure there will be some conflicts if you try to merge your changes with mine because of work duplication as you mentioned.

We will need to get Allan involved to get feedback on the best approach to follow and then we can work from there to get the changes merged.

In terms of work duplication, this is something I was discussing with Thorsten, maybe we need some sort of task board where everyone can specify what they are working on to avoid doing the same things twice. This will also promote collaborations between contributors which can only be a good thing.

I would love to hear the community's opinion on this.

src/rez/pip.py
```diff
@@ -97,7 +100,17 @@ def run_pip_command(command_args, pip_version=None, python_version=None):
     pip_exe, context = find_pip(pip_version, python_version)
     command = [pip_exe] + list(command_args)
 
-    if context is None:
+    if process_output:
```
mottosso: What is this meant for?

mottosso:
I've been contemplating how to make the implementation of `pip.py` a little simpler.

- Currently, `rez pip` creates a context internally (or uses system-python)

But what if it didn't need to do that? What if instead could call on whichever Python and `pip` was available at the time of being called? If it did, we could strip those parts where it's establishing a context and checking for a system Python, and get called like this.

```bash
# Use system Python, if any
$ rez pip --install six
ERROR: Could not find Python or pip
             Usage: rez env python-2 pip-19 -- pip install six

# Use explicit Python, and pip
$ rez env python-2.7 pip-19 -- rez pip --install six
```

Now `rez pip` is able to collaborate with `rez env` as expected, and as an added bonus makes the syntax explicit, and able to handle requests like normal. Then, for users who prefer `pip` be installed as part of `python` (e.g. me), they could simply omit `pip-19` altogether.

```bash
$ rez env python-3.7 -- rez pip --install six
```

Clear, familiar and consistent with a potential integration of `rez conda`, `rez choco` and `rez vcpkg` etc.

Would you be interested in this for this PR, @lambdaclan?

JeanChristopheMorinPerso:
>Clear, familiar and consistent with a potential integration of rez conda, rez choco and rez vcpkg etc.

I'm not sure we would really want to go the route of implementing conda, choco and others package managers IMHO. If you ever have the chance to talk to a conda developer, there is a high number of chances he will tell you it's not _that_ fun to re-package everything.

>drop os version dependency all together platform / architecture / (os) / python for all packages (pure and impure)

@lambdaclan What's the reasoning behind this? And could you elaborate more on this?

>ability to list all available versions of a package based on PyPi

Does this do anything different from `pip search`?

I'll come back to do a real code review later for sure.

bpabel:
> What's the reasoning behind this? And could you elaborate more on this?

Because pure python packages are cross platform.  There's no need to variant them by os/platform/arch.  And many python packages are py2/3 compliant, so you don't have to variant them by python version either.

JeanChristopheMorinPerso:
But why drop the OS/platform/arch for non pure packages? That was more my question (I should have been more specific in the first place).

mottosso:
> for all packages (pure and impure)

Oh, also noticed this. Impure packages, like PySide, would only run on a given platform, arch and OS so those probably need to remain.

src/rez/package_maker__.py
```diff
@@ -128,6 +128,10 @@ def get_package(self):
     def _get_data(self):
         data = self._data.copy()
 
+        if "requires" in data:
+            if None in data["requires"]:
```
JeanChristopheMorinPerso: You can do:
```python
data["requires"] = list(filter(None, data["requires"]))
```
Which is slightly more pythonic. Basically it will do one for loop instead of two.

src/rez/pip.py
JeanChristopheMorinPerso: If it fails because the version doesn't exist, it will return the available versions. Not entirely sure if it's really needed here.

src/rez/pip.py
```diff
@@ -242,8 +254,22 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
...
+
+        true_table = {
+            "true": True,
+            "false": False
+        }
+
+        packages = pkg_resources.find_distributions(destpath)
+        dist = next((package for package in packages if package.key == installed_dist.key), None)
+        wheel_data = dist.get_metadata('WHEEL')
+        wheel_data = Parser().parsestr(wheel_data)
```
JeanChristopheMorinPerso: A little comment here that points to the documentation about that would be great (like https://www.python.org/dev/peps/pep-0566/#json-compatible-metadata).

src/rez/pip.py
```diff
@@ -242,8 +254,22 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
...
+
+        return true_table[wheel_data["Root-Is-Purelib"]]
```
JeanChristopheMorinPerso: Maybe add a little link to https://www.python.org/dev/peps/pep-0427/#what-s-the-deal-with-purelib-vs-platlib or something else that explains a little bit more what's the deal with `Root-Is-Purelib`?

lambdaclan:
Hello Jc, Marcus and Brendan thank you all very much for the feedback regarding the PR. @JeanChristopheMorinPerso  thank for the review! I will be answering your questions so that we can decide the best way to proceed.

src/rez/pip.py
lambdaclan: This has to do with the ability to list all available versions of a package based on PyPi for the current Python interpreter. Although a bit of a hack as outlined [here](https://stackoverflow.com/questions/4888027/python-and-pip-list-all-versions-of-a-package-thats-available) there is currently no way of retrieving this list. Causing the pip install to fail is intentional in a way to retrieve the available packages.

lambdaclan:
> > drop os version dependency all together platform / architecture / (os) / python for all packages (pure and impure)
> 
> @lambdaclan What's the reasoning behind this? And could you elaborate more on this?

Hello Jc, apologies for my formatting which seems to have made things a bit confusing. Only pure packages drop OS, platform and arch. Not pure packages only drop os version dependency the one
in the bracket.

[platform / architecture / (os) / python - major.minor] -- impure
[python major] -- pure

> 
> > ability to list all available versions of a package based on PyPi
> 
> Does this do anything different from `pip search`?

Yes, [pip search]()https://pip.pypa.io/en/stable/reference/pip_search/ only returns the latest version of the package queried along with any other package that might share the same word. The current implementation only returns the versions of the queried package that are available for install. 

I added this feature in case you want to install an older version and work well with the rest of the experimental features added in the PR.

lambdaclan:
> I've been contemplating how to make the implementation of `pip.py` a little simpler.
> 
>     * Currently, `rez pip` creates a context internally (or uses system-python)
> 
> 
> But what if it didn't need to do that? What if instead could call on whichever Python and `pip` was available at the time of being called? If it did, we could strip those parts where it's establishing a context and checking for a system Python, and get called like this.
> 
> ```shell
> # Use system Python, if any
> $ rez pip --install six
> ERROR: Could not find Python or pip
>        Usage: rez env python-2 pip-19 -- pip install six
> 
> # Use explicit Python, and pip
> $ rez env python-2.7 pip-19 -- rez pip --install six
> ```
> 
> Now `rez pip` is able to collaborate with `rez env` as expected, and as an added bonus makes the syntax explicit, and able to handle requests like normal. Then, for users who prefer `pip` be installed as part of `python` (e.g. me), they could simply omit `pip-19` altogether.
> 
> ```shell
> $ rez env python-3.7 -- rez pip --install six
> ```
> 
> Clear, familiar and consistent with a potential integration of `rez conda`, `rez choco` and `rez vcpkg` etc.
> 
> Would you be interested in this for this PR, @lambdaclan?

Hello Marcus, I completely agree that rez-pip is due for an overall revamp and I am in favour of simplifying the process. 

I like the idea of having rez env collaborate with rez-pip but we will also need to decide how to deal with the standalone rez-pip command (ie not using env). If we decide to get rid of context all together and use system python pip then this might work. I know that on Windows I cannot use the pip package because of some symlink issue so I always delete the package after rez-bind in order to fallback to system version. 

Maybe first of all we need to decide how rez will be installed in the future through pip or through a python install? I can see things getting overly confusing with which pip is going to be used when having so many options. Is the same pip that installed rez? Is it the pip that is bundled with the python that installed rez? Is it something else etc...

I am in favour of setting the system pip as the default and allowing rez-env to interface with rez-pip ( and happy to collaborate to make that happen 👍 ) if the user chooses to do so but I have to agree with Jc about being sceptical with regards to adding integration with other packagers such as conda, choco vcpkg etc. This will add unnecessary  complexity for some specialized scenarios. conda is used for scientific python, choco and vcpkg are Windows specific and not widely used so it is worth it overall?

I like having options but too many options can make things intimidating for users so maybe a whole new tool would make more sense rez-conda, rez-choco etc. Rez plugins here we go...

src/rez/cli/pip.py
```diff
@@ -29,17 +38,129 @@ def setup_parser(parser, completions=False):
...
-    installed_variants, skipped_variants = pip_install_package(
+    if opts.list:
+        opts.PACKAGE += "=="
+        output = run_pip_command(["install", opts.PACKAGE], process_output=True)
+        for package_version in output:
+            print(package_version)
+        return
+
+    if opts.query:
+        URL = "https://pypi.python.org/pypi/{0}/{1}/json"
```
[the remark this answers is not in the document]
lambdaclan: Hello Jc good point. I was thinking that anyone using rez-pip would have an active internet connection but did not consider internal PyPI indexes. Although the query option help clearly states that is for use with PyPi.

If we add the URL as an option how will the query work? I am assuming we will need to add implementations for each method? A bit unfamiliar with internal index since I never used one.

src/rez/cli/pip.py
```diff
@@ -29,17 +38,129 @@ def setup_parser(parser, completions=False):
...
+            data = json.load(urllib2.urlopen(URL.format(opts.PACKAGE, opts.query)))
+
+            classifiers = data["info"]["classifiers"]
+            python_requires = [classifier for classifier in classifiers if "Programming Language :: Python ::" in classifier]
+            python_requires = ''.join(python_requires)
+            python_requires = python_requires.split("Programming Language :: Python :: ")
+            python_requires = filter(None, python_requires)
+
+            if not python_requires:
+                print("Package version is valid but PyPi does not contain a specific Python version. Too old?")
```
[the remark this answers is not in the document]
lambdaclan: Agreed, the reason I said too old is because it normally happens with old packages newer ones seem to have that metadata always defined. This was mostly for debugging purposes so will be happy to have more descriptive messages.

src/rez/cli/pip.py
```diff
@@ -29,17 +38,129 @@ def setup_parser(parser, completions=False):
...
+        for package_version in output:
+            print(package_version)
+        return
+
+    if opts.query:
+        URL = "https://pypi.python.org/pypi/{0}/{1}/json"
+        try:
+            data = json.load(urllib2.urlopen(URL.format(opts.PACKAGE, opts.query)))
+
+            classifiers = data["info"]["classifiers"]
```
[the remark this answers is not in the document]
lambdaclan: I thought about using the requires_python metadata as well but did not want it to conflict with the classifier as sometimes they do and then will be hard to decide which to pick. Also requires python only states a single version whereas classifiers are more descriptive...

Maybe having a check to see if they at least match or something. 

The query option is using classifiers whereas the constrain one is using python_version and requires_python. There is definitely some overlap here that can be improved like sticking to one validation method for example always use requires_python or come up with a an order of importance check ala:

1. requires_python
2. python_version
3. classifier

The problem of course lies with the the non standardization of the whole metadata so it is hard to decide which one to use and why.. 

I will take it into consideration thanks for the feedback.

src/rez/package_maker__.py
lambdaclan: >Explicit is better than implicit.
Simple is better than complex.

Just joking, as a Haskeller I definitely approve but sometimes I find it hard to decide which way to go :)

src/rez/pip.py
lambdaclan: Makes sense I will add it.

src/rez/pip.py
lambdaclan: Makes sense I will add it.

lambdaclan:
> Also, it would be great to add tests for this change.

Once the wheel based installation PR is merged I will add tests and adjust the code as requested. I am a bit hesitant in doing so right away because if the wheel install PR changes or is rejected for whatever reason I will need to re-work a lot of this anyway.

nerdvegas:
Hi @lambdaclan, I'm ready to merge #602, just letting you know that that will happen tomorrow. I'll think about backwards compatibility and may make minimal changes to keep that in place, I'll let you know.

Thx
A

nerdvegas:
Just want to mention about this approach:

```
rez env python-3.7 -- rez pip --install six
```

Ie to use the current rez-env as a basis for determining which python interpreter to use for the installation. We should not take this approach, because in all other cases, the rez cli tools are intended to work independently of the current environment, rez-based or not. It also closely couples rez-pip with rez's own installation procedure.

Options to control the python/pip packages used for installation should just be arguments to the rez-pip command itself; falling back to the system installation of anything should be the last choice (or enabled explicitly via another option).

nerdvegas:
Also letting you know - there's another developer here at Method who deals with all our software installation into rez packages. We have our own toolset here for doing that. We would really rather not be maintaining yet another tool for this, so we're keenly interested in this development.

src/rez/package_maker__.py
bpabel: It's worth noting that `filter(None, ...)` will filter out anything that evaluates to `False`, not just `None` values.

src/rez/pip.py
```diff
@@ -267,18 +293,24 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
         tools = []
         src_dst_lut = {}
 
-        for installed_file in distribution.list_installed_files(allow_fail=True):
-            source_file = os.path.normpath(os.path.join(destpath, installed_file[0]))
+        for installed_file in distribution.list_installed_files():
+            if 'bin' + os.sep in installed_file[0]:
```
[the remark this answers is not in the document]
JeanChristopheMorinPerso: Or maybe even just `if os.path.isabs(installed_file[0])`.

lambdaclan:
> Also requires python only states a single version whereas classifiers are more descriptive...

`requires_python` can describe quite a lot actually: `requires_python=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"` would be a valid value. (See https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires)

The thing is that I don't know if the classifiers are automatically added when `requires_python` is used and the classifiers were not manually specified. Maybe you know (i basically didn't have time to try).

>There is definitely some overlap here that can be improved like sticking to one validation method for example always use requires_python or come up with a an order of importance check ala:
>
> 1. requires_python
> 2. python_version
> 3. classifier

That would probably do it yes, but would only be necessary if the classifiers are not automatically added by `requires_python`

_Originally posted by @JeanChristopheMorinPerso in https://github.com/nerdvegas/rez/pull/628_


After thinking a bit more about this, it might make sense to restructure the commands in order to make their intent more apparent, easier to understand.

As discussed above we have the following fields that we can use to obtain some sort of an idea
about Python requirements of a library:

1. requires_python
2. python_version
3. classifiers

Now it goes without saying that since nothing is formalized, which of the values above each package actually defines can vary widely. According to my initial impressions after playing with this for a while, I realized that normally the definition occurrence is as follows: 

classifiers > python_version > requires_python

overview --> 2 vs 3 --> specific requirements

funnily the one of most value is the least defined one...

If we put that aside for now, what I have in mind is to have two commands.

1. A general query command that will give an overview of all the versions of Python compatible
with queried package.

```
rez-pip -q (--query) package_name
```
There will be no version argument anymore. This is just to check if a package (of any version) is compatible with my current Python. For this I am thinking of using the classifiers. I could use requires_python as well but although like you mention above, can be used to describe a lot of info it hardly ever does...

Example django 2.1.1

classifiers:

6 | "Programming Language :: Python"
7 | "Programming Language :: Python :: 3"
8 | "Programming Language :: Python :: 3.5"
9 | "Programming Language :: Python :: 3.6"
10 | "Programming Language :: Python :: 3.7"
11 | "Programming Language :: Python :: 3 :: Only"


requires_python | ">=3.5"

We can see that they dont match...

2. A specific version query so obtain the package version/s compatible with desired Python version range

```
rez-pip -c (--constrain) package_name python_version_range
```

Here the classifiers are pretty much useless and we really need requires_python as python_version really only specifies py2 or py3 but no other details. The problem is that
requires_python is hardly ever correctly defined....

I am starting to feel that this might be to immature to be implemented as a helper tool. I have good intentions but unfortunately, non formalization and uniformity the data is not there to backup the claims that might be made of such tool.

A sceptical on how to continue (hence experimental in title). It might be worth to extract the purity detection part only as that immediately adds value and work on the experimental stuff on a different PR.

Thoughts?

lambdaclan:
Hello @nerdvegas

Thank you for merging #656! I have now merged the changes from master so only the relevant diffs are now visible, hope it is better now.

src/rez/pip.py
```diff
@@ -335,6 +335,7 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
     """
     installed_variants = []
     skipped_variants = []
+    variant_types = []
```
JeanChristopheMorinPerso: I can't find any references to this variable in the function. I think it's no more needed.

JeanChristopheMorinPerso:
This PR looks good to me.

src/rez/pip.py
```diff
@@ -446,18 +462,26 @@ def make_root(variant, path):
...
+            if pure:
+                py_ver = '.'.join(map(str, sys.version_info[:1]))
+            else:
+                py_ver = '.'.join(map(str, sys.version_info[:2]))
         else:
             python_variant = context.get_resolved_package("python")
-            py_ver = python_variant.version.trim(2)
+
+            if pure:
+                py_ver = python_variant.version.trim(1)
```
nerdvegas: I'm thinking that we shouldn't do this; or, that we do it but only if we know it's safe to do so.

"Safe" depends on whether there are markers in the requirements. For eg, if the package requires "foo-1.0 ; python-3.1.*", then the only way we can properly express the package's requirements in rez is if we variant on pthon-MAJOR.MINOR; otherwise we cannot specify foo as a requirement for just that case.

I also suspect that the complexity in determining this from the markers isn't worth it, and that we should probably just stick to varianting on python-MAJOR.MINOR.

lambdaclan: Hello Allan, after discussing the issue with Thorsten we have reached an agreement that we can live without this particular change, trimming the minor version when dealing with pure packages. What is most important for us is the removal of the os as that adds a significant number of rebuilds when dealing with Windows versions (python x plat  x arch x os) removing just one of these segments will make our work much much easier.

Also. as you mention above it makes sense to keep the minor version since it might be a hard requirement for some packages that will cause other issues. I would like to explore the markers solution down the line so I will take a look and see  if there is a way to achieve it, getting rid of the minor version will still be a big improvement if it is reliable.

For now I will remove the minor version trimming so that we can continue with the review before the merge.

nerdvegas:
That sounds good Ira cheers. Yeah I think it's a reasonable compromise that
avoids what may be a lot of complication for not that much payout. We can
certainly revisit at a later date.

Thx
A

nerdvegas:
I'm going to merge this now, been reviewing for a while and it sounds good. Will make that py ver trim change if it hasn't been done already. I also have a PR coming on the tails of this one. Cheers all!

PR #639, opened 2019-05-29
Ordered iter_packages

mottosso:
Previously, `rez.packages_.iter_packages` yields results in a somewhat randomised order, based on how items are returned from the `set()` they are currently stored in once retrieved from the otherwise alphabetical `os.listdir`.

Now packages are yielded in reverse-alphabetical order, which means you're able to fetch the latest version like this.

**Before**

```python
it = iter_packages("maya")
latest = next(it)
for package in it:
  if package.version > latest.version:
    latest = package
```

**After**

```python
latest = next(iter_packages("maya"))
```

Which is not only easier on the eyes, it's also significantly cheaper, as it doesn't incur a filesystem call per available version of package.

### Why

The benefits are quite obvious I think, but for completeness; this is an optimisation in both network traffic and time spent for the user.

### Caveat

It's a backwards compatible change, but comes with a few gotchas.

1. The alphabetical order isn't always accurately representing the latest version, e.g. `1.80` is newer than `1.9`.
2. Versions may come from multiple directories (!)

Number 2 is the big one, consider for example.

```bash
~/
  packages/
    maya/
      2018/

/
  server/
    packages/
      maya/
        2019/
```

In this case, "latest" is relative the order these appear on your `REZ_PACKAGES_PATH`.

```bash
$ export REZ_PACKAGES_PATH=~/:/server/packages  # next(iter_packages) == 2018
$ export REZ_PACKAGES_PATH=/server/packages:~/  # next(iter_packages) == 2019
```

So the prerequisites for this to work, in its current form is to have each unique package come from a single repository. Note that you could still have multiple repositories, just not versions of the same package coming from two or more repositories.

### Discussion

I think we're close to making this work without caveats. The call to `os.listdir` is cheap compared to parsing the `package.py` which currently happens in tandem with directories returned.

(1) is easily solved by gathering versions from each directory (cheap) and comparing them via `Version(text)`.

(2) is a little more tricky, but perhaps not too far off. If we `os.listdir` all paths first, sort them, and then run them through the process of parsing the corresponding `package.py`, we should be able to get an ordered result, without sacrificing any functionality or performance.

What do you think?

### DIFF
--- a/src/rezplugins/package_repository/filesystem.py
+++ b/src/rezplugins/package_repository/filesystem.py
@@ -718,11 +718,11 @@ class FileSystemPackageRepository(PackageRepository):
         # tested regardless. Failed releases may cause 'building files' to be
         # left behind, so we need to clear these out also
         #
-        dirs = set()
+        dirs = list()
         building_dirs = set()
 
         # find dirs and dirs marked as 'building'
-        for name in os.listdir(root):
+        for name in sorted(os.listdir(root), reverse=True):
             if name.startswith('.'):
                 if not name.startswith(self.building_prefix):
                     continue
@@ -732,8 +732,8 @@ class FileSystemPackageRepository(PackageRepository):
 
             path = os.path.join(root, name)
 
-            if os.path.isdir(path) and not ignore_dir(name):
-                dirs.add(name)
+            if name not in dirs and os.path.isdir(path) and not ignore_dir(name):
+                dirs.append(name)
 
         # check 'building' dirs for validity
         for name in building_dirs:
@@ -745,7 +745,7 @@ class FileSystemPackageRepository(PackageRepository):
                 # package probably still being built
                 dirs.remove(name)
 
-        return list(dirs)
+        return dirs
 
     # True if `path` contains package.py or similar
     def _is_valid_package_directory(self, path):


bpabel:
Would it be that more expensive to parse the output of os.listdir into actual Version objects so you can sort them by their actual Version instead of alphabetically?

> the prerequisites for this to work, in its current form is to have each unique package come from a single repository

This isn't really practical.  Even in the simplest use case where all packages are released and stored in the same repository, you will still need a local path for testing.  Also, every studio Ive worked at that used Rez had some type of local caching in addition to their Network repositories, both within the packages path

JeanChristopheMorinPerso:
> Would it be that more expensive to parse the output of os.listdir into actual Version objects so you can sort them by their actual Version instead of alphabetically?

I don't think it would be more expensive. And it would allow us to have a consistent behavior. That is, the order will be the same as everywhere else in rez. This is what the `rez-search` command does (see https://github.com/nerdvegas/rez/blob/master/src/rez/package_search.py#L256)

As for point 2, the problem might potentially be solvable if the sort was happening outside the repository plugins (Like in `rez.packages_`). But that would maybe mean to get rid of the iters... Which is not so great. Or maybe with another clever technique that I' not aware of (which is highly possible).

mottosso:
At the end of the day, parsing with `Version()` is cheap, and calling `os.listdir` on 1-10 directories is cheap. The expensive bit is reading and parsing the corresponding `package.py`'s.

With that in mind, I'd imagine we could turn this:

```python
for r in repositories:  # e.g. ["~/packages", "/server/packages"]
  for p in r.packages():  # e.g. [maya, nuke, houdini]
    for v in p.versions():  # os.listdir is called
      yield parse(v)
```

Into this:

```python
versions = list()
for r in repositories:  # e.g. ["~/packages", "/server/packages"]
  for p in r.packages():  # e.g. [maya, nuke, houdini]
    for v in p.versions():  # os.listdir is called
      versions += [v]

for v in sorted(versions, key=lambda dirname: Version(dirname)):
      yield parse(v)
```

The trick being bringing all three loops into the same scope, as they are now happening across rez and rezplugins. To do that, we'd need to introduce a new method to repository plug-ins, to yield just the absolute path to version directories, so we can gather them all up in rez, and pass that back to the plug-in parsing.

nerdvegas:
I'm sorry but I don't understand the reason for this PR. The iter_packages function returns in non-specific order because this is the cheapest possible implementation.

"""Which is not only easier on the eyes, it's also significantly cheaper, as it doesn't incur a filesystem call per available version of package."""

This demonstrates a lack of understanding of how rez works, as it is not true. Great lengths have been taken to put off loading of package definitions until necessary - and even then, validation of its attributes are then performed only if and when they are accessed.

It is easy to verify what I've just mentioned. Simply set the `debug_file_loads` setting to True, then iterate over packages. You won't see any debug printout at all, because no package definitions have been loaded - a package's version and name is determined from the filesystem (the family dir and version dir), not from the package.py files.

If you want iter_packages in some order, you just sort it yourself. I can understand the reason for the PR if you had thought package definitions were actually getting loaded though, but that is not the case.

nerdvegas:
I'd also like to clarify:

"""In this case, "latest" is relative the order these appear on your REZ_PACKAGES_PATH."""

This is not the case and is not intended behaviour. As the docs specifically state:

If the same package appears in two or more repositories on the search path, the earlier package is used in preference. This happens at the version level - an earlier package "foo-1.0.0" will hide a later package "foo-1.0.0", **but not "foo-1.2.0"**.

mottosso:
> I don't understand the reason for this PR

Which part do you feel could benefit from some clarification?

> You won't see any debug printout at all, because no package definitions have been loaded - a package's version and name is determined from the filesystem (the family dir and version dir), not from the package.py files.

That is great! Then here's what I would approach next.

1. Replace `iter_packages` with a non-iterator `packages`, because why iterate over anything if the results are random?
2. With a non-iterator `packages()`, we could from there sort the results per default, because why not?

> This is not the case and is not intended behaviour. As the docs specifically state:

Oops I think you may have skimmed this part a little too quick; it's about a caveat of the implementation in this PR, rather than in Rez.

nerdvegas:
1. Replacing means breaking backwards compatibility, so no.

Let me rephrase then. What is this PR bringing to the project - what is it making possible that isn't possible currently? If you want packages in a specific order, you just `sort(iter_packages(...))`.

mottosso:
> What is this PR bringing to the project

It's meant as an optimisation for anyone using `iter_packages` to find a given version, in particular any of the latest ones.

```bash
$ python -m timeit -n 1 -s "from rez.packages_ import iter_packages as ip" "next(ip('core_pipeline'))"
1 loops, best of 3: 36.3 usec per loop
$ python -m timeit -n 1 -s "from rez.packages_ import iter_packages as ip" "list(ip('core_pipeline'))"
1 loops, best of 3: 50.4 usec per loop
```

Notice the call to `next()` versus `list()`, that's a 30% performance increase for finding the latest version. Not entirely representative of course, because:

1. The PR proposes changing what `next()` does, which would affect timings
2. Values are cached, such that the second iteration within the same Python session would take less time

On top of that, based on what you're saying, the proposed implementation is already in place; there is no `open()` and versions are determined by their folder name.

In this case, I would shift the issue onto this:

> Replacing means breaking backwards compatibility, so no.

I didn't mean we break backwards compatibility. In fact we'd be safe to assume nothing of what I propose is ever intended to break backwards compatibility. By replace I mean replace it as anything anyone uses for anything, and leave it accessible via Python but not advertise it.

Maybe I'm missing something, but you didn't answer the question before; how come `iter_packages` is an iterator, if it must always be completed?

nerdvegas:
Who said it must always be completed? If the developer using it doesn't care about order, then it doesn't need to be completed. I'm not making assumptions.

What is the goal here? Is it to achieve a fast sorted iterator? If so, why? Generally a PR should fix a bug, or add a feature, or optimise performance - but optimise it in a way that speeds up an existing use case. Perhaps you know the saying - "early optimisation is the root of all evil." I'd rather not see people putting effort into optimisations that don't achieve anything practical.

In this case, I'm faced with a PR that slows the existing iterator, and for some reason, iterates in alphabetical order (which bears no relationship to version order).

mottosso:
> Who said it must always be completed?

Why else would anyone use it, if not to find a particular version of a given package? :O Anything other than a completed iterator would be unreliable; what's *your* usecase?

> and for some reason, iterates in alphabetical order (which bears no relationship to version order).

I think you may have missed the conversation about about using `Version()` to sort, rather than alphabetical, and that the current implementation was a first step on the way.

> In this case, I'm faced with a PR that slows the existing iterator

I think we may be talking past each other; the above example I gave showed a 30% performance *increase*.

> Perhaps you know the saying - "early optimisation is the root of all evil."

I think (seeing a pattern here! :)) we have a different understanding of what the saying means. Whomever is using `iter_packages`, including myself, would benefit from a performance boost, especially one as low-hanging as this. I honestly did not expect much resistance to something this trivial, and I get the feeling we won't be making much progress here.

At the very least, we've now got documentation on why and how `iter_packages` works for future users to stumble upon in the future, small high-five.

PR #644, opened 2019-06-11
Implement PowerShell

mottosso:
This was fun. :)

powerrez

It's almost all there, except the prompt showing you how many levels deep you are. It's using the same functionality from `cmd` to fetch `PATH` from the environment rather than the current shell so that shouldn't cause any issues. It's generating the familiar `rez-context.ps1` script.

### Why?

We had a chat on Slack about that, where the primarily motivation was to avoid hitting the `cmd.exe`-specific [length-limit](https://groups.google.com/forum/#!msg/rez-config/d8M7ZfD2aQk/jNDyKpixBQAJ;context-place=forum/rez-config) on environment variables, particularly PATH and PYTHONPATH. Then there are other advantages too, like tab-completion support and (proper) terminal colors (none of which are implemented in this PR, but are possible with Powershell).

Powershell to the people!

### DIFF
--- a/src/rezplugins/shell/powershell.py
+++ b/src/rezplugins/shell/powershell.py
@@ -1,6 +1,5 @@
-"""
-Windows Command Prompt (DOS) shell.
-"""
+"""Windows PowerShell 5"""
+
 from rez.config import config
 from rez.rex import RexExecutor, OutputStyle, EscapedString
 from rez.shells import Shell
@@ -10,7 +9,6 @@ from rez.backport.shutilwhich import which
 from functools import partial
 import os
 import re
-import subprocess
 
 try:
     basestring
@@ -174,7 +172,8 @@ class PowerShell(Shell):
             if bind_rez:
                 ex.interpreter._bind_interactive_rez()
             if print_msg and not quiet:
-                ex.command("rez context")
+                # Rez may not be available
+                ex.command("Try { rez context } Catch { }")
 
         executor = RexExecutor(interpreter=self.new_shell(),
                                parent_environ={},
@@ -184,10 +183,15 @@ class PowerShell(Shell):
             _record_shell(executor, files=startup_sequence["files"])
             shell_command = startup_sequence["command"]
         else:
-            _record_shell(executor, files=startup_sequence["files"], print_msg=(not quiet))
+            _record_shell(executor,
+                          files=startup_sequence["files"],
+                          print_msg=(not quiet))
 
         if shell_command:
             executor.command(shell_command)
+
+        # Forward exit call to parent PowerShell process
+        executor.command("exit $LastExitCode")
 
         code = executor.get_output()
         target_file = os.path.join(
@@ -207,12 +211,11 @@ class PowerShell(Shell):
         cmd += [self.executable]
         cmd += ['. "{}"'.format(target_file)]
 
-        if not shell_command:
+        if shell_command is None:
             cmd.insert(1, "-noexit")
 
         p = popen(cmd,
                   env=env,
-                  shell=True,
                   universal_newlines=True,
                   **Popen_args)
         return p
@@ -263,7 +266,9 @@ class PowerShell(Shell):
         self._addline(self.setenv(key, value))
 
     def alias(self, key, value):
-        self._addline("Set-Alias -Name %s -Value \"%s\"" % (key, value))
+        value = EscapedString.disallow(value)
+        cmd = "function {key}() {{ {value} $args }}"
+        self._addline(cmd.format(key=key, value=value))
 
     def comment(self, value):
         for line in value.split('\n'):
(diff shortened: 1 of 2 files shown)


instinct-vfx:
I don't think the new syspaths implementation works. While it works for pure Rez stuff i guess, it will break all kinds of locally installed packages. The current registry implementation seems complex but it has the purpose of decoupling syspaths from the shell rez is being called from (because it may as well be a Rez shell already and that introduces all sorts of issues). It also makes sure that everything that works without Rez also works with Rez.
Your commit message is correct in that we do not know what PATH is, but that's precisely why we need to forward it: It may be required for other applications that Rez does not know about.

mottosso:
Yes, this is probably not the right PR for this kind of change. It's applicable both here, in CMD and for Linux too. I'll revert that to better focus the attention.

mottosso:
To properly test this out, I went cold-turkey to Powershell since implementing this and to my surprise everything works; the sub-shell, the `ResolvedContext.execute_shell`, the works. As an added benefit, history works and commands have tab-completion like you would expect from normal Powershell (not Rez commands though, yet).

@instinct-vfx Would be interesting to see a quick 5 minute test in your environment, setting `REZ_DEFAULT_SHELL=powershell` and see what breaks/doesn't break. I expect it should also work in ConEmu and various other alternative terminals. Common commands like `cd` and `dir` work the same, and `set VARIABLE=Value` is `$env:VARIABLE=Value`.

I also had a look at Powershell 6, which is cross-platform (!) which means we could potentially have (finally?) a single unified interface for Rez on every platform, and focus our efforts on interface, tab completion and color across the board.

- https://github.com/PowerShell/PowerShell

mottosso:
```bash
PS> $fx = get-process -name firefox
PS> $fx.path
# C:\Program Files\Mozilla Firefox\firefox.exe
PS> $fx.workingset / 1mb
# 414.23828125
PS> $fx.kill()
```

That's pretty neat. :)

**Powershell Resources**

- [Scoop Package Manager](https://www.youtube.com/watch?v=a85QLUJ0Wbs)
- [Getting Started](https://www.youtube.com/watch?v=IHrGresKu2w)

That last one is plain, and I especially like this statement:

> There's probably some reason in the world not to run PowerShell as administrator, I haven't found it yet so I pretty much always run it as administrator.

Hehehe, but he does point out a few interesting things and there doesn't really seem to be that much PS material out there; beggars can't be choosers. If you find any more, do share!

**Aliases**

```bash
PS> ls
# 
#     Directory: C:\Users\marcus\.ssh
# 
# Mode                LastWriteTime         Length Name
# ----                -------------         ------ ----
# -a----       11/02/2018     15:52           1766 id_rsa
# -a----       18/02/2018     11:30           1486 id_rsa.ppk
# -a----       11/02/2018     15:52            392 id_rsa.pub
# -a----       18/03/2019     08:05           3514 known_hosts

PS> get-alias clear
# CommandType     Name                                               Version    Source
# -----------     ----                                               -------    ------
# Alias           cls -> Clear-Host
```

**Working with processes**

```bash
PS> $notepad = start-process notepad -passthru
PS> $notepad | get-member
#    TypeName: System.Diagnostics.Process
# 
# Name                       MemberType     Definition
# ----                       ----------     ----------
# ...
# Name                       AliasProperty  Name = ProcessName
# Company                    ScriptProperty System.Object Company # {get=$this.Mainmodule.FileVersionInfo.CompanyName;}
# ...
# Path                       ScriptProperty System.Object Path {get=$this.Mainmodule.FileName;}
# WorkingSet                 Property       int WorkingSet {get;}
# Kill                       Method         void Kill()
# Refresh                    Method         void Refresh()
# Start                      Method         bool Start()
# ToString                   Method         string ToString()
# ...
PS> $notepad.company
# Microsoft Corporation
```

instinct-vfx:
I still have not found the time to dive in, hope to do so over the next few days. Just one quick heads up:

> There's probably some reason in the world not to run PowerShell as administrator, I haven't found it yet so I pretty much always run it as administrator.

Well, no-one has admin privileges here and i'd guess you should not run anything as admin unless you need to ;)

mottosso:
> Well, no-one has admin privileges here and i'd guess you should not run anything as admin unless you need to ;)

Hah, yes. I highlighted it for its absurdity.

mottosso:
Implemented at feature parity with `cmd` on Windows in [bleeding-rez](https://github.com/mottosso/bleeding-rez). Welcome to re-open new PR with source from there.

```bash
$ pip install bleeding-rez
$ rez env --shell powershell
> $
# Or
$ set REZ_DEFAULT_SHELL=powershell
$ rez env
> $
```

- [rezplugins/shell/powershell.py](https://github.com/mottosso/bleeding-rez/blob/master/src/rezplugins/shell/powershell.py)

instinct-vfx:
Re-opening to review.

Issue #609, opened 2019-04-22
Support "additive" preprocess functions


JeanChristopheMorinPerso:
Opening an issue regarding https://github.com/nerdvegas/rez/pull/597#issuecomment-484536831 and a comment that followed (https://github.com/nerdvegas/rez/pull/597#issuecomment-485314968).

Right now, as things are, it is only possible to use one single preprocess function. Either in the `rezconfig.py` file or in a `package.py` file. It's one or the other. This is pretty limiting. Local preprocessor is super useful (by local I mean in `package.py`) and global (in `rezconfig.py`) is also super useful.

I propose to add the ability to have what I call additive preprocessors. That means adding the ability to define both a global and a local preprocessor and that both are executed.

@nerdvegas goes one step further and propose to add a new `preprocess_mode` package attribute. Possible values are `before` and `after`. If `preprocess_mode` is defined, both the global and local will be sexcuted. If not, the default (current) behavior is used (so if a global is set but no local, the global will be used. If a local is set (with or withut global), the local wins).

I would push the idea even further and add a way to force both (global and locall) to always be executed, and have the ability to disable the global per package by setting `preprocess_mode` to `None`, a little bit like `build_command`. So `preprocess_mode` could be defined in `rezconfig.py` and have a default value set by a studio. The idea is that there will probably be some studios that will want to enfore a preprocess on 95% of their packages, and so to remove some technical details, having the mode defined at a facility level will make things simpler and better for some teams. And for the 5% package left, the package author would be able to set `preprocess_mode` to `None` so that the global is not executed.

Let me know what you think. If everybody agrees, I'll work on that.

nerdvegas:
I don't think it's very intuitive to disable global preprocessing by
setting `preprocess_mode = None`. In general it's good to avoid having to
set anything to None to select behaviour; for build_command, it's a rare
case that does make sense however.

I see another issue too. Neither approach has taken into consideration that
a package may wish to do local preprocessing before AND after the global
preprocessor. I can imagine legitimate cases for doing this.

So then, I'm thinking that exposing the global preprocessor as a function
that you can call in your own preprocessor makes the most sense. It's
backwards compatible, and it gives the most flexibility:

    # in package.py
    def preprocess(package, data):
        do_some_before_stuff()
        global_preprocess(package, data)
        do_some_after_stuff()

Also, the problem with a global setting to control preprocess mode, is that
an author of a package may have assumed some default order (let's say local
[...]

PR #651, opened 2019-06-23
Add a package_preprocess_mode (#609)

JeanChristopheMorinPerso:
This PR implements #609 , which is make it possible to have both a global and local preprocess function run when building a package.

That means it's now possible to have a global preprocess that applies to all packages, and also have specialized preprocesses per packages if need be. This is made possible by a new configuration called `package_preprocess_mode`. It accepts 3 values: `after`, `before` and `override`.

The default value is `override`, which mean we have the exact same behavior as before: If you defined `package_preprocess_function` and you also have a `preprocess` function in a package, only the package's `preprocess` function is called.

`after` and `before` are what I call the additive modes. They allow you to choose to execute a package's `preprocess` function before or after the global preprocess.

* I updated the docs;
* Wrote tests.

### DIFF
--- a/src/rez/developer_package.py
+++ b/src/rez/developer_package.py
@@ -178,7 +178,6 @@ class DeveloperPackage(Package):
         package_preprocess_mode = self.config.package_preprocess_mode
 
         def _get_package_level():
-            print_info("Applying preprocess from package.py")
             return getattr(self, "preprocess", None)
 
         def _get_global_level():
@@ -233,7 +232,6 @@ class DeveloperPackage(Package):
                 print_error("Function '%s' not found" % package_preprocess_function)
                 return None
 
-            print_info("Applying preprocess function %s" % package_preprocess_function)
             return preprocess_func
 
         with add_sys_paths(config.package_definition_build_python_paths):
@@ -254,6 +252,9 @@ class DeveloperPackage(Package):
             for preprocessor in preprocessors:
                 if not preprocessor:
                     continue
+
+                level = "global" if preprocessor == global_preprocess else "local"
+                print_info("Applying {0} preprocess function".format(level))
 
                 # apply preprocessing
                 try:
--- a/src/rez/developer_package.py
+++ b/src/rez/developer_package.py
@@ -6,9 +6,18 @@ from rez.exceptions import PackageMetadataError, InvalidPackageError
 from rez.utils.system import add_sys_paths
 from rez.utils.sourcecode import SourceCode
 from rez.utils.logging_ import print_info, print_error
+from rez.vendor.enum import Enum
 from inspect import isfunction
 import os.path
 import stat
+
+
+class PreprocessMode(Enum):
+    """Defines when a package preprocess will be executed.
+    """
+    before = 0    # Package's preprocess function is executed before the global preprocess
+    after = 1     # Package's preprocess function is executed after the global preprocess
+    override = 2  # Package's preprocess function completely overrides the global preprocess.
 
 
 class DeveloperPackage(Package):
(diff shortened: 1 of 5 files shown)


JeanChristopheMorinPerso:
I'd like to mention that I feel like `package_preprocess_function` might not be a proper name now that we have additive behavior... I thought about adding a new `global_preprocess_function`/`global_preprocess` or something along these lines.  It could live along side of `package_preprocess_function` and both would do the same thing. `package_preprocess_function` could raise a warning. But there is no warnings in rez as of now, and I don't know if I'm alone to think `package_preprocess_function` is no more a proper name.

Also, please don't merge this for now. I've written tests but I'd like to see if it fits at a bigger scale and see what I can do with it at RodeoFX to confirm the idea and behavior.

src/rez/developer_package.py
```diff
@@ -159,28 +168,36 @@ def _validate_includes(self):
...
         from copy import deepcopy
 
-        with add_sys_paths(config.package_definition_build_python_paths):
-            preprocess_func = getattr(self, "preprocess", None)
+        package_preprocess_mode = self.config.package_preprocess_mode
 
-            if preprocess_func:
-                print_info("Applying preprocess from package.py")
+        def _get_package_level():
+            print_info("Applying preprocess from package.py")
```
nerdvegas: Message printed too early? I would also say '...from package definition' rather than package.py (you can configure rez to use a different filename).

src/rez/developer_package.py
```diff
@@ -212,23 +229,41 @@ def _get_preprocessed(self, data):
                     )
                     return None
 
-                print_info("Applying preprocess function %s" % preprocess_func)
-
             if not preprocess_func or not isfunction(preprocess_func):
                 print_error("Function '%s' not found" % package_preprocess_function)
                 return None
 
+            print_info("Applying preprocess function %s" % package_preprocess_function)
```
nerdvegas: Message printed too early? Would also mention 'applying global preprocess function'

src/rez/tests/data/packages/developer_dynamic_local_preprocess_additive/package.py
```diff
@@ -0,0 +1,23 @@
...
+@early()
+def description():
+    return "This."
+
+requires = [
+    "versioned-*"
+]
+
+def preprocess(this, data):
+    print 'Local preprocess'
```
nerdvegas: use py3 syntax

nerdvegas:
Cheers! Only some minor issues that I can see.

RE the setting name - I wouldn't worry about this much, I think `package_preprocess_function` is fine. It would be odd with both settings as suggested, because one would make sense in the global config and one would not - which would be at odds with existing settings.

I'll wait til you're confident after more testing, just let me know.

Thx
A

src/rez/developer_package.py
```diff
@@ -212,23 +228,43 @@ def _get_preprocessed(self, data):
...
-            except Exception as e:
-                print_error("Failed to apply preprocess: %s: %s"
-                            % (e.__class__.__name__, str(e)))
-                return None
+            for preprocessor in preprocessors:
+                if not preprocessor:
+                    continue
+
+                level = "global" if preprocessor == global_preprocess else "local"
+                print_info("Applying {0} preprocess function".format(level))
```
JeanChristopheMorinPerso: @nerdvegas I changed the message a little bit while fixing the messages being printed too early. I tried to make the message consistent across the local and global process. If you don't want to introduce the "global" term here, let me know and I'll try to rephrase.

nerdvegas:
Looks good JC. Just clarifying that this is ready to merge now, you said you were doing more testing?
Cheers
A

JeanChristopheMorinPerso:
Hey @nerdvegas sorry for taking so long on this. I'm currently in a big rush at work and it takes me quite a lot of my time. I've booked myself some time tomorrow to test this out a little bit more.

JeanChristopheMorinPerso:
@nerdvegas I tested more in depth and it's all good to go. You can merge this PR! Really sorry for the delay on this one.

PR #654, opened 2019-06-25
Update distlib vendor library

lambdaclan:
# Description

Update distlib library from 0.2.4.dev0 to 0.2.9.post0. This
is necessary in order to be able to enforce Wheel distribution
based installations.

Relates # (PRs)
rez #602 , #628

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Remarks

* This should be merged once #602 is fully merged. Both #602 and #628 require a newer version of distlib libary in order to work. This distlib version provides a different API as such merging this first before Wheel distribution installation is enforced will cause errors. For example the `allow_fail` in         

```
for installed_file in distribution.list_installed_files(allow_fail=True)
```
is no longer available.


JeanChristopheMorinRodeoFX:
@lambdaclan could you update the vendor README please? Apart from that, I don't have much to say. :+1:

PR #657, opened 2019-07-08
[Feature] Pip QOL commands -- EXPERIMENTAL

lambdaclan:
# Description

[Depends on PR #602 #654 #656]

### pip QOL improvements

To further ensure compatibility with current platform this PR adds some *experimental* features
such as package version listing, querying and Python version constraint based search. These features
use the most "reliable" source of package data being the PyPi via the new [JSON API](https://warehouse.readthedocs.io/api-reference/json/). 

The reason these features are marked experimental is because currently there is no standard of how the metadata should be defined and sometimes it varies widely between packages. According to my tests it seems to be working for the most part but still a lot of things to consider.

## Breakdown

* ability to list all available versions of a package based on PyPi

* ability to query Python version prerequisites of a specific version of a package based on PyPi

* ability to retrieve and install a specific version of a Python package based on a Python version range (for example 3.6+)

## Process

* Python package listing uses pip install {package_name} ==
  * not [specifying a version](https://stackoverflow.com/a/26664162) returns a list of available versions for the current python interpreter
* Python package querying is using the Python programming language classifier metadata
  * what are [classifiers](https://pypi.org/classifiers/)?
  * what is the [Python programming language classifier](https://stackoverflow.com/questions/55739269/whats-the-purpose-of-version-in-programming-language-classifier-of-setuptools)?
* Python constraint install is using the requires_python and python_version metadata available via
the PyPi JSON API
  * note: this metadata is not always complete, sometimes only python version is defined which means several adjustments have to be made. Results might vary depending how old the version of a package is or how the maintainer defines the package metadata
  image
* Python package purity detector
  * this is using the Wheel metadata file named WHEEL. WHEEL is the wheel metadata specific to a build of the package. The Wheel metadata contents are defined [here](https://www.python.org/dev/peps/pep-0491/#file-contents)

## Type of change

## Remarks

* Requires pip >=19.0
* Requires  distlib >= 0.28

# How Has This Been Tested?

In terms of the experimental features, although it does work fairly well it is a matter of how much trust
we can put into non-standardized metadata. For example a package might state Python version
requirement 3.5+ but might also work perfectly well with Python 2.7 or lower than Python 3.5...

**Test Configuration**:
* OS: Linux, Windows 10

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Further investigation about PyPi metadata

## Demos

* Package version search

list

* Package version query

query

* Python version constraint

limit

### DIFF
--- a/src/rez/cli/pip.py
+++ b/src/rez/cli/pip.py
@@ -179,7 +179,7 @@ def command(opts, parser, extra_arg_groups=None):
 
     print()
     if installed_variants:
-        print "%d packages were installed:" % len(installed_variants)
+        print("%d packages were installed:" % len(installed_variants))
         for variant, purity in zip(installed_variants, variant_types):
             print("[{}]".format(purity))
             print_variant(variant)
--- a/src/rez/cli/pip.py
+++ b/src/rez/cli/pip.py
@@ -37,6 +37,20 @@ def setup_parser(parser, completions=False):
         help="package to install or archive/url to install from")
 
 
+def get_pypi_api_url():
+    from rez.pip import pip_install_package, run_pip_command
+
+    default_url = "https://pypi.python.org/pypi"
+    output = run_pip_command(["config", "get", "rez.pypi-api-endpoint"], process_output=True)
+    if "ERROR" in output:
+        print(("WARNING: pip config or rez.pypi-api-endpoint key not found. "
+              "Falling to default: ({}) API endpoint.").format(default_url))
+        return default_url
+    else:
+        # take second line of output since the first is the Python 2 depreciation warning
+        return output.split('\n')[1]
+
+
 def command(opts, parser, extra_arg_groups=None):
     from rez.pip import pip_install_package, run_pip_command
     from rez.vendor.version.version import Version, VersionRange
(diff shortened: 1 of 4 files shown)


nerdvegas:
Hi @lambdaclan,

I'm a little hesitant about where this is headed. Rez isn't supposed to be a pip wrapper - there are other projects aiming to do that, such as https://github.com/myint/yolk. Specifically, it appears you can get a listing of PyPI versions of a package using `yolk -V matplotlib`, for example. 

Are we in danger of replicating the efforts of other packages, and do you think this is within the scope of rez? Bear in mind that rez was never intended to be a package manager specific to python (in fact it largely managed C++ packages when it was first written). Adding integration with pip did make sense however, since python packages are so widely used in VFX, and the large majority are already made available via pip.

lambdaclan:
Hello Allan, yes I can see what you are saying we definitely do not want to be adding things to rez that are out of scope. The reason I labelled this PR experimental is because I am trying to see if such features would be beneficial in the greater scheme of things since most revolve around version checking, validations etc that can be useful when authoring rez packages with strict requirements.

 yolk does look like a nice alternative but basically I wanted to avoid external dependencies.

I personally do not have any issues with not including such functionality intro rez-pip if it does not make sense. I will consult with @instinct-vfx  once more and see if we have any particular needs for these changes and will get back to you. We can always extract it as an external standalone tool or as you mention above use a different package so there are many options.

I will keep the PR open for now. What is important for us is #628 and #659 so I will focus on those for now.

nerdvegas:
No worries. To be clear, I understand the need for the functionality that
some of these cli options represent - for example, determining the python
version(s) compatible with a given package. But I think it would be better
to concentrate on that functionality first and foremost, then consider
exposing it in the cli tool if that proves useful. The immediate aim of
rez-pip is to successfully convert a pip package and its requirements into
rez packages, with correct analogous rez dependencies.

Thx
A

instinct-vfx:
Sorry for the silence here. I am fine with splitting this off and focus on functionality required for rez-pip itself only first. rez.pip does have a high priority for us so i am all for getting this on track asap.

PR #658, opened 2019-07-08
[Fix] Package schema and request errors

lambdaclan:
# Description

[Depends on PR #602 #654 #656]

## Breakdown

* fix schema issue with packages including dashes for example importlib_metadata->=0.5

## Type of change

**Test Configuration**:
* OS: Linux, Windows 10

## Todos
- [ ] Tests

### DIFF
--- a/src/rez/package_maker__.py
+++ b/src/rez/package_maker__.py
@@ -128,6 +128,10 @@ class PackageMaker(AttrDictWrapper):
     def _get_data(self):
         data = self._data.copy()
 
+        if "requires" in data:
+            if None in data["requires"]:
+                data["requires"] = [x for x in data["requires"] if x is not None]
+
         data.pop("installed_variants", None)
         data.pop("skipped_variants", None)
         data.pop("package_cls", None)
--- a/src/rez/utils/formatting.py
+++ b/src/rez/utils/formatting.py
@@ -14,7 +14,7 @@ import re
 import time
 
 
-PACKAGE_NAME_REGSTR = "[a-zA-Z_0-9](\.?[a-zA-Z0-9_]+)*"
+PACKAGE_NAME_REGSTR = "[a-zA-Z_0-9](\.?[a-zA-Z0-9_-]+)*"
 PACKAGE_NAME_REGEX = re.compile(r"^%s\Z" % PACKAGE_NAME_REGSTR)
 
 ENV_VAR_REGSTR = r'\$(\w+|\{[^}]*\})'
--- a/src/rez/utils/formatting.py
+++ b/src/rez/utils/formatting.py
@@ -14,7 +14,7 @@ import re
 import time
 
 
-PACKAGE_NAME_REGSTR = "[a-zA-Z_0-9](\.?[a-zA-Z0-9_-]+)*"
+PACKAGE_NAME_REGSTR = "[a-zA-Z_0-9](\.?[a-zA-Z0-9_]+)*"
 PACKAGE_NAME_REGEX = re.compile(r"^%s\Z" % PACKAGE_NAME_REGSTR)
 
 ENV_VAR_REGSTR = r'\$(\w+|\{[^}]*\})'


src/rez/package_maker__.py
```diff
@@ -128,6 +128,10 @@ def get_package(self):
     def _get_data(self):
         data = self._data.copy()
 
+        if "requires" in data:
```
nerdvegas: Sorry it's a little hard to track down the relevance of this change. What's going on here, and why would Nones be provided in the 'requires' key given to the package maker in the first place? Surely they should be stripped out beforehand?

src/rez/utils/formatting.py
```diff
@@ -14,7 +14,7 @@
 import time
 
 
-PACKAGE_NAME_REGSTR = "[a-zA-Z_0-9](\.?[a-zA-Z0-9_]+)*"
+PACKAGE_NAME_REGSTR = "[a-zA-Z_0-9](\.?[a-zA-Z0-9_-]+)*"
```
nerdvegas: Again I'm not quite clear on what's going on here.

The problem with this change is that in rez, `-` _always_ separates the package name from its version. If a package name can contain hyphen, then which part is the name and which is the version becomes ambiguous.

nerdvegas:
I'm a little lost on this one - I've gone over the conversation history on https://github.com/nerdvegas/rez/pull/628 but I can't see how these changes relate.

Also might be good to re-merge with master now that #602 is merged, there are loads of diffs here that no longer apply and it's quite difficult to see which changes are specific to this PR.

Thx!
A

lambdaclan:
Re-merged as required!

src/rez/utils/formatting.py
lambdaclan: Hello Allan, yes this is true. This commit was from some time ago before I realized that rez uses dash to separate the package name from its version. After some tests it does not seem to be relevant to the SchemaError that occurs with some packages. I will revert this change and provide an example to the SchemaError.

src/rez/package_maker__.py
lambdaclan: Hello Allan, this seems to be a really particular issue. I have only found one package that triggers it being importlib-metadata. It was a dependency of a package I was testing out (do not remember which one) but installing it directly still causes the issue.

```
  File "/home/iakritas/.rez/bin/rez/rez-pip", line 4, in <module>
    run('pip')
  File "/home/iakritas/.rez/lib/python2.7/site-packages/rez-2.35.0-py2.7.egg/rez/cli/_main.py", line 148, in run
    returncode = run_cmd()
  File "/home/iakritas/.rez/lib/python2.7/site-packages/rez-2.35.0-py2.7.egg/rez/cli/_main.py", line 140, in run_cmd
  File "/home/iakritas/.rez/lib/python2.7/site-packages/rez-2.35.0-py2.7.egg/rez/cli/pip.py", line 47, in command
    release=opts.release)
  File "/home/iakritas/.rez/lib/python2.7/site-packages/rez-2.35.0-py2.7.egg/rez/pip.py", line 482, in pip_install_package
    pkg.commands = '\n'.join(commands)
  File "/home/iakritas/.pyenv/versions/2.7.16/lib/python2.7/contextlib.py", line 24, in __exit__
    self.gen.next()
  File "/home/iakritas/.rez/lib/python2.7/site-packages/rez-2.35.0-py2.7.egg/rez/package_maker__.py", line 187, in make_package
    package = maker.get_package()
  File "/home/iakritas/.rez/lib/python2.7/site-packages/rez-2.35.0-py2.7.egg/rez/package_maker__.py", line 101, in get_package
    package_data = package_schema.validate(package_data)
  File "/home/iakritas/.rez/lib/python2.7/site-packages/rez-2.35.0-py2.7.egg/rez/vendor/schema/schema.py", line 195, in validate
    nvalue = Schema(svalue, error=e).validate(value)
  File "/home/iakritas/.rez/lib/python2.7/site-packages/rez-2.35.0-py2.7.egg/rez/vendor/schema/schema.py", line 235, in validate
rez.vendor.schema.schema.SchemaError: Or(<class 'rez.utils.sourcecode.SourceCode'>, [Or(And(<type 'basestring'>, Use(<function expand_requirement at 0x7fb80733d320>)), And(<class 'rez.utils.formatting.PackageRequest'>, Use(<type 'str'>)))]) did not validate [u'zipp->=0.5', None, u'configparser->=3.5', None]
Or(Or(And(<type 'basestring'>, Use(<function expand_requirement at 0x7fb80733d320>)), And(<class 'rez.utils.formatting.PackageRequest'>, Use(<type 'str'>)))) did not validate None
Or(And(<type 'basestring'>, Use(<function expand_requirement at 0x7fb80733d320>)), And(<class 'rez.utils.formatting.PackageRequest'>, Use(<type 'str'>))) did not validate None
None should be instance of <class 'rez.utils.formatting.PackageRequest'>
```

Here are the contents of the data dictionary (notice the *requires* list):

```
{'commands': "env.PYTHONPATH.append('{root}/python')", 'description': u'Read metadata from Python packages', 'installed_variants': [], 'version': '0.18', 'skipped_variants': [], 'package_cls': <class 'rez.packages_.Package'>, 'requires': [u'zipp->=0.5', None, u'configparser->=3.5', None], 'variants': [['platform-linux', 'arch-x86_64', 'os-Arch-rolling', 'python-2.7']], 'name': u'importlib_metadata'}
```

That is why I have added the change, getting rid of None resolves the issue. I will dig a bit further and see if I can find anything else or any other packages that trigger the error.

lambdaclan: Looking at the importlib-metadata [setup.cfg](https://gitlab.com/python-devs/importlib_metadata/blob/master/setup.cfg) file I can see the install requires being as follows:

```
install_requires =
    zipp>=0.5
    pathlib2; python_version=='3.4.*' or python_version < '3'
    contextlib2; python_version < '3'
    configparser>=3.5; python_version < '3'

```
Now looking at the requires above:

```
'requires': [u'zipp->=0.5', None, u'configparser->=3.5', None]
```
It seems to be missing two of its dependencies? pathlib2 and contextlib2 should have been included.

nerdvegas:
Worth looking into I think.

Related to this - if we are to obey pip package requirements to the letter,
that means taking markers into account (such as in this example). However,
that also implies that the rez package _must_ require/be varianted on at
least major.minor of the python version. It's possible to have a
requirement like "configparser>=3.5; python_version < '3'" in pip, because
the current python version is known ahead of time. Such requirement logic
isn't possible in rez however, because python is just another package. So
if a package has markers, the only way to satisfy the related requirements
is to have those requirements specified alongside the python version range
that requires them.

Here's where it gets tricky: what if you had a marker like so:

    pathlib2; python_version=='3.4.1'

Now we're in trouble. The only way to express this requirement in rez would
be to have a variant on python-3.4.1, and that's getting a bit odd.
Furthermore, you might also have several markers, whose overlapped python
version requirements would cause even stranger varianting on python.

I'm not sure what the best solution is. It seems to me atm that the
considerable extra complexity needed to exactly define markered
requirements (ie, varianting the package on python in such a way that the
requirements are correct, but python version is never overspecified) is
more trouble than it's worth. If we just varianted on python-major.minor,
yes that would be overly specific for many cases, but it would also avoid
this problem.

src/rez/package_maker__.py
nerdvegas: I think I've found the issue here. I'm going to address it in another PR, which aims to simplify code in pip.py by making further use of `packaging`, specifically for managing environment markers.

lambdaclan: Hello Allan, great news. If there is anything I can do to help please let me know. I will keep the PR open for now just in case.

nerdvegas:
I'm pretty confident that #667 addresses this, so will close this PR when that is merged.

nerdvegas:
Closing now, I'm certain that #667 fixes the issue, cheers

Issue #660, opened 2019-07-10
[Feature] Additional tools to add to PATH


bfloch:
## Goal 

In order to provide tools like `dot` to resolved rez environments without having to rely on the system I would like to extend the rezconfig with an additional mechanism to prepend/append to PATH within environments.

## Currently rezconfig provides options to ...

1. `parent_variables`: Inherit variables (PATH) from the current environment
2. `standard_system_paths`: Provide a static list of PATH variables, as opposed to the default system variables
3. `rez_tools_visibility`: Option to append/prepend the Script/rez or bin/rez tools to the path

## Problems with the available options
1. This means all variables are inherited, which may easily conflict as the comments show
2. You would need to duplicate all the PATHs and maintain them
3. This is ideal since we can append / prepend and the tools are closely related, but we need to provide our own paths.

## Proposed solutions

A) Create a rezconfig list `rez_additional_tools` which is expanded and added following the logic of `rez_tools_visibility` (after it for backwards compatibility).

B) Alternatively allow 2. to be prepended / append as opposed to replacing the current PATH discovery from the system. 
I propose a new config variable: `standard_system_path_visibility` that accepts: 

- `replace` (default/backwards compatible)
- `append`
- `prepend`

EDIT: This option affects how the `standard_system_paths` behave.

I would be fine with either option and would be willing to implement it.

bfloch:
I went forward to implement variant B) but I am struggling to do the right thing.

Currently:
`resolved_context` -> `rex.append_system_path()` + `rex.append/prepend_rez_tools()` -> `shell(instance)->get_syspath()`

get_syspath in turn is checking the config to either use the standard_system_path = [] or get it from an clean environment by any means possibly for the the particular shell.

Now I could nicely remove the logic from get_syspath() and move it up one level to rex / resolved_context to fit the api nicely.
This way the shell would ALWAYS return a system PATH and it is up to rex / resolved context to adapt it to the configuration's need.

However the shell also implements the `which` function which in turn only relies on its own get_syspath().
So it will not discover the tools added at all - it might even break existing functionality.

Note: As is, the which kind of acts wrong (by design?) because it has no notion of the added rez_tools. But that might be fine since rez_tools are a construct of a resolved_context (??).

My thought - in order to keep the functionality compatible and the which intact is to move all the new system_path logic down to the shell level. I could easily manage it with a decorator so we don't need to repeat ourselves. It also moves some of the class variable caching into a centralized location, which is nice.

[...]

PR #661, opened 2019-07-11
[Feature] Additional tools to Path

bfloch:
Implements #660.

- [x] Backwards compatible
- [x] Unit tests
## Compatibility 

1. It is fully backwards compatible. A new config variable is introduced, that defaults to current behavior `replace`
2. Reuses the `standard_system_paths` variable usage
3. External shell plugins will not be affected but can remove the rezconfig logic and have to decorate `get_syspaths` with `syspaths_composer` to make use of the new feature.

## What do we gain?

- Ability to prepend / append paths to shell's syspath as opposed to only replacing it. Compare #660
- Removed control logic of `standard_system_paths` out of plugin implementation into central decorator `syspaths_composer`

### DIFF
--- a/src/rez/config.py
+++ b/src/rez/config.py
@@ -234,6 +234,13 @@ class RezToolsVisibility_(Str):
         return Or(*(x.name for x in RezToolsVisibility))
 
 
+class StandardPathVisibility_(Str):
+    @cached_class_property
+    def schema(cls):
+        from rez.shells import StandardPathVisibility
+        return Or(*(x.name for x in StandardPathVisibility))
+
+
 class OptionalStrOrFunction(Setting):
     schema = Or(None, basestring, callable)
 
@@ -271,6 +278,7 @@ config_schema = Schema({
     "plugin_path":                                  PathList,
     "bind_module_path":                             PathList,
     "standard_system_paths":                        PathList,
+    "standard_system_paths_visibility":             StandardPathVisibility_,
     "package_definition_build_python_paths":        PathList,
     "implicit_packages":                            StrList,
     "platform_map":                                 OptionalDict,
--- a/src/rez/rezconfig.py
+++ b/src/rez/rezconfig.py
@@ -366,10 +366,18 @@ rez_tools_visibility = "append"
 # scripts (such as .bashrc). If False, package commands are sourced after.
 package_commands_sourced_first = True
 
+# Defines how the standard_system_paths configuration acts on the $PATH:
+# - "replace":           Replace $PATH if standard_system_paths is set
+# - "append":            Append standard_system_paths to determined $PATH
+# - "prepend":           Prepend standard_system_paths to determined $PATH
+standard_system_paths_visibility = "replace"
+
 # Defines paths to initially set $PATH to, if a resolve appends/prepends $PATH.
 # If this is an empty list, then this initial value is determined automatically
 # depending on the shell (for example, *nix shells create a temp clean shell and
 # get $PATH from there; Windows inspects its registry).
+# The way the list is combined with $PATH is controlled via
+# rez_standard_system_paths_visibility.
 standard_system_paths = []
 
 # If you define this function, it will be called as the *preprocess function*
--- a/src/rezplugins/shell/cmd.py
+++ b/src/rezplugins/shell/cmd.py
@@ -3,7 +3,7 @@ Windows Command Prompt (DOS) shell.
 """
 from rez.config import config
 from rez.rex import RexExecutor, literal, OutputStyle, EscapedString
-from rez.shells import Shell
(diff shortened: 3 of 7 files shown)


bfloch:
Tests passing:
- [x] Windows 10.0.17134
  - [x] cmd
  - [x] tcsh
  - [ ] sh (fails on test_rex_code_alias, possibly other bug, investigating)
  - [x] bash
  - [x] csh

nerdvegas:
Thanks Blazej, will get on this soon. Ignore previous email I didn't see
you had the PR in already!

Cheers
A

bfloch:
Btw. @nerdvegas:
What's the correct pattern to reset the rez.config for each test?
Currently the state is kind of bleeding-through, which is not necessarily a problem since I reset the two parameters I am interested in, but it would be great if the config was clean for each test.

JeanChristopheMorinPerso:
>What's the correct pattern to reset the rez.config for each test?

In theory, the `TestBase` class (which is subclassed by the `TestShells` class takes care of resetting the config for each test. TO change the settings, you can use `self.update_settings`. See https://github.com/nerdvegas/rez/blob/e1e135ad0c882b42ddedd5155e73a86ef41f8620/src/rez/tests/util.py#L61

nerdvegas:
Hey @bfloch this in on my radar, but I'm not sure I'll be able to get to it until after SIGGRAPH, just FYI.
Thx
A

bfloch:
No worries. I'll have it more tested next week and try to make the tests cleaner.
Have a great show.

src/rezplugins/shell/cmd.py
```diff
@@ -68,14 +68,11 @@ def get_startup_sequence(cls, rcfile, norc, stdin, command):
         )
 
     @classmethod
+    @syspaths_composer
```
nerdvegas: So the functionality of syspaths_composer is fine, however I don't think it's correct for it to be a decorator like this.

The intent of `Shell.get_syspaths` is to define what system paths are associated with that shell - not _how_ to then apply them. So I would just take the functionality from the decorator, and apply it only where syspaths are actually set into PATH (https://github.com/nerdvegas/rez/blob/master/src/rez/rex.py#L1124)

nerdvegas:
This is just a comment - much as mentioned in #698, this feature may be unnecessary when we implement better env-var management (as described in #707). I don't want to hold back this PR to wait for that, but I think it's worth noting that this could become deprecated in lieu of that at some point (which would be good, because it means less edge cases).

src/rez/rezconfig.py
```diff
@@ -366,10 +366,18 @@
 # scripts (such as .bashrc). If False, package commands are sourced after.
 package_commands_sourced_first = True
 
+# Defines how the standard_system_paths configuration acts on the $PATH:
+# - "replace":           Replace $PATH if standard_system_paths is set
+# - "append":            Append standard_system_paths to determined $PATH
```
nerdvegas: This is confusing, I think you mean:

```
Append 'standard_system_path' to automatically determined system paths
```

right?

src/rez/config.py
```diff
@@ -271,6 +278,7 @@ def _parse_env_var(self, value):
     "plugin_path":                                  PathList,
     "bind_module_path":                             PathList,
     "standard_system_paths":                        PathList,
+    "standard_system_paths_visibility":             StandardPathVisibility_,
```
nerdvegas: I think "standard_system_paths_mode" would be more intuitive.

src/rez/rezconfig.py
nerdvegas: Ok ignore that, I think I have an idea on how to make this clearer. It's jut inherently a little hard to explain I think.

src/rez/rezconfig.py
```diff
@@ -366,10 +366,18 @@
...
+# - "prepend":           Prepend standard_system_paths to determined $PATH
+standard_system_paths_visibility = "replace"
+
 # Defines paths to initially set $PATH to, if a resolve appends/prepends $PATH.
 # If this is an empty list, then this initial value is determined automatically
 # depending on the shell (for example, *nix shells create a temp clean shell and
 # get $PATH from there; Windows inspects its registry).
+# The way the list is combined with $PATH is controlled via
+# rez_standard_system_paths_visibility.
```
nerdvegas: Instead:
```
# ...$PATH from there; Windows inspects its registry).
# 
# To specify both your own systems paths, AND automatically determined system paths, use the 
# "standard_system_paths_visibility" setting. For example, setting this to "append" will initially
# set $PATH to the automated system paths, followed by the paths listed here.
```

bfloch:
I agree. Let's focus on #707.

PR #339, opened 2016-08-15
Added appveyor config

skral:
In the wake of the open source initiative I think a appveyor config might help.
Initial config which runs all the tests. Unfortunately one test fails right now. Didn't look into why.

PR #662, opened 2019-07-12
Installer updates

nerdvegas:
This PR updates the current installer. Changes include:

* patched distlib (in build_utils) has been removed. The patch we were relying on has since been made part of the main distlib release, which we already have vendored;
* virtualenv has been updated to latest;
* scripts have been removed, and entry points are used instead;
* installation occurs via wheel;
* install.py code has been cleaned up and simplified. Specifically, standard use of distlib.ScriptMaker has been put in place;
* INSTALL.md has been updated with a full explanation of the installer, and why a pip-based installation is not the same as using install.py. Reviewers are encouraged to read this file.

It would be great if others could test this, especially Windows users.


nerdvegas:
Note also that this PR supersedes #625, although I'm interested in incorporating parts of that into separate PRs (notably, automated pip uploads). I'm also interested in Appveyor, which also belongs in a separate PR (and there is one - #339 - which should be compared with what's available from #625 and potentially combined).

instinct-vfx:
Will test this today.

instinct-vfx:
I just gave this a quick run to test the pip installation. The behavior does not match INSTALL.md.

> However, this comes with a caveat - rez command line tools are disabled once inside a rez environment (ie after using the rez-env command). The reasons are given in the next section.

I did the following to install rez through pip:
```
python setup.py bdist_wheel
pip install rez-2.36.0-py2-none-any.whl
```

Now Rez executables still work inside resolved shells though:

image

This should not work, should it?

btw. please let me know if you prefer this to be comments in the diff section i can also move it there.

instinct-vfx:
Also there seems to be an issue with paths containing ~. As our deployment setup installs from a tempfolder generated location this can contain ~ if the corresponding user profile has certain properties (like mine does). Still investigating what change exactly causes that issue and if it is something that should be fixed in Rez or in our setup.

INSTALL.md
```diff
@@ -1,3 +1,86 @@
...
+# Why Not Pip For Production?
+
+Rez is not a normal python package. Although it can successfully be installed
+using standard mechanisms such as pip, this comes with a number of caveats.
+Specifically:
+
+* When within a rez environment (ie after using the `rez-env` command), the rez
+  command line tools would not be guaranteed to function correctly;
+* When within a rez environment, other packages' tools (that were also installed
+  with pip) would remain visible, but would not be guaranteed to work.
```
mottosso: This may be addressed using one of the methods you mentioned in #625.

I made a prototype of this a few weeks ago but haven't been testing it in production for more than a few days, and only on Windows. I've updated it to work on Linux just now, you can test it and see if it fits the bill like this:

```bash
$ pip install bleeding-rez
$ rez --version
bleeding-rez-2.33.4
$ rez env python --isolated
```

The environment you'll get is completely void of anything from the parent environment, including any PYTHONPATH or site-packages. Now, this is so barebones, that it also excludes `rez` itself from PATH. So if you wanted it there, you could add it [here](https://github.com/mottosso/bleeding-rez/blob/d99c7baf65dd4b82ab72e80961c0d77aaefa36be/src/rez/cli/_main.py#L152-L157), where `bash` and `lsb_release` are explicitly added as well.

Alternatively, the `REZ_ENVIRONMENT` variable takes a JSON-serialised environment to be used in place of the skeleton environment built into the source. Out of the box you're left with a Docker-like environment, with only the shell itself plus your resolve.

INSTALL.md
```diff
@@ -1,3 +1,86 @@
...
+
+When you enter a rez environment, the rez packages in the resolve configure
+that environment as they see fit. For example, it is not uncommon for a python
+package to append to PYTHONPATH. Environment variables such as PYTHONPATH
+affect the behaviour of tools, including rez itself, and this can cause it to
```
mottosso: This would possible be addressed by the above as well.

INSTALL.md
```diff
@@ -1,3 +1,86 @@
...
+crash or behave abnormally.
+
+When you use the `install.py` script to install rez, some extra steps are taken
+to avoid this problem. Specifically:
+
+* Rez is installed into a virtualenv so that it operates standalone;
+* The rez tools are shebanged with `python -E`, in order to protect them from
```
mottosso: If the `--isolated` method works, the `python -E` is embedded into Rez, which means we could skip this particular step.

INSTALL.md
```diff
@@ -1,3 +1,86 @@
...
+  environment variables that affect python's behaviour;
+* The rez tools are stored in their own directory, so that other unrelated tools
```
mottosso: And finally, due to explicit control over the environment, this would also go away. Pof!

nerdvegas:
Are you referring to the subprocess approach? I considered this, however
the install.py-based installation works without the need for an extra
subproc. With that in mind, it doesn't make sense to support two different
installation methods, when one results in differing behaviour like this.

However, if enough people feel a great need for production-ready pip
installable rez, we could take that approach, _if_ it's isolated to pip
installations only (ie, install.py should continue doing what it currently
does). If contributed, this belongs in a separate PR.

A

nerdvegas:
Ah, it may be that the tool disabling only works in a virtualenv-based pip
installation.

Come to think of it, yeah there's no reliable way to remove visibility of
rez tools without taking out a bunch of other stuff as well. I'll update
the docs in this PR.

install.py
```diff
@@ -103,52 +100,57 @@ def copy_completion_scripts(dest_dir):
     return None
 
 
+def install_rez_from_source(dest_dir):
+    _, py_executable = get_venv_executable(dest_dir, "python")
+    _, pip_executable = get_venv_executable(dest_dir, "pip")
+
+    # build wheel
+    run_command([py_executable, "setup.py", "bdist_wheel"])
```
JeanChristopheMorinPerso: I think we can simplify and solidify this by doing:

```python
run_command([py_executable, "-m", "pip", "install", "."])
```
1. Using `py_executable -m pip` you don't need to take care of finding pip. Python will do it for us.
2. `pip install .` will build the wheel for you and install it for you. It is a little cleaner and a preferred approach from what I know. The other benefit is that you don't have to look for the wheel in the `dist` directory which could have some development stuff in there is not cleaned-up before running the `install.py` file.

setup.py
```diff
@@ -13,11 +13,20 @@
     sys.exit(1)
 
 
-if sys.version_info < (2, 6):
-    print("install failed - requires python v2.6 or greater", file=sys.stderr)
+if sys.version_info < (2, 7):
```
JeanChristopheMorinPerso: While we are there, we should also check for python 3. And we should also set the `python_requires` argument in the `setup` function.

src/rez/cli/_entry_points.py
```diff
@@ -0,0 +1,215 @@
...
+@scriptname("rez-bind")
+def run_rez_bind():
+    from rez.cli._main import run
+    return run("bind")
+
+
+@scriptname("rez-build")
+def run_rez_bind():
+    from rez.cli._main import run
+    return run("bind")
```
JeanChristopheMorinPerso: I think you copy/paste a little too fast :stuck_out_tongue: It should be `build` and the function name should also contain `build` as right now it override the bind. Same for `rez-config`.

JeanChristopheMorinPerso:
>Ah, it may be that the tool disabling only works in a virtualenv-based pip
installation.
>
>Come to think of it, yeah there's no reliable way to remove visibility of
rez tools without taking out a bunch of other stuff as well. I'll update
the docs in this PR.

Yep, tool disabling only works when installing with `install.py`. We could potentially detect from the rez tools if we are not in _recommended deployment method_ and log a warning maybe? I know some tools that does this (`pipenv` is the first to come to my mind). Or simply block the usage. We could call `rez-system.is_production_rez_install` from https://github.com/nerdvegas/rez/blob/master/src/rez/cli/_main.py#L72 and spit out a warning if it's false: `We detected that Rez was installed with Pip. Please be aware the rez
  command line tools are not guaranteed to function correctly. See <link to wiki> for more information`.

What do you think?

INSTALL.md
```diff
@@ -1,3 +1,85 @@
 # Installation
 
 See https://github.com/nerdvegas/rez/wiki/Getting-Started#installation
+
+
+First, install Rez. Download the source, and, from the source directory, run:
```
JeanChristopheMorinPerso: Philosophical question (sorry in advance): Should we move this doc in the https://github.com/nerdvegas/rez/wiki/Getting-Started#installation doc? I'm also wondering if we should add a mention about how to install in the README as it's usually a common practice to do so.

Or we want to tackle that into a separate subject and wait for your roadmap before deciding?

install.py
nerdvegas: nice, changes coming

setup.py
nerdvegas: A py3 port is coming soon (another dev is working on it) so I think I'll leave this for now.

src/rez/cli/_entry_points.py
nerdvegas: thanks for the catch, fixes coming

INSTALL.md
nerdvegas: Yes to both, will make those changes.

instinct-vfx:
> Also there seems to be an issue with paths containing ~. As our deployment setup installs from a tempfolder generated location this can contain ~ if the corresponding user profile has certain properties (like mine does). Still investigating what change exactly causes that issue and if it is something that should be fixed in Rez or in our setup.

Please ignore this one, i found the issue and it is not within Rez. Sorry for that.

src/rez/cli/_entry_points.py
```diff
@@ -0,0 +1,257 @@
...
+            "https://github.com/nerdvegas/rez/wiki/Installation#why-not-pip-for-production "
+            " for futher details.\n"
+        )
+
+
+### Entry points
+
+@scriptname("rez")
+def run_rez():
+    check_production_install()
```
JeanChristopheMorinPerso: I'm wondering if you had a reason to not move that into the `rez.cli._main.run`. I have no problem if we do the check here. The only reason why I was mentioning `rez.cli._main` is because that would allow us to remove some duplication and have the check more centralized. Maybe you see a future where one command would not need the check?

nerdvegas:
The reason really is just to get this code into the callstack as early as
possible. The more rez code we import in a non-production-safe install, the
more chance something goes wrong before the warning is printed.
A

JeanChristopheMorinRodeoFX:
It makes a lot of sense.

I'm also wondering if we should write some tests to make sure it behaves as we want and that it stays like this in the future. With potentially all the cases we found in the last couple of weeks.

What do you think? Not too sure how we would go on testing that though.

nerdvegas:
Yeah perhaps, but let's leave it on ice for the moment. There's a lot to
do, we'll see if any problems arise.

A

nerdvegas:
Can anyone confirm that this has been tested on Windows? I think we're ready to merge if so.

instinct-vfx:
I tried it and it broke for me in various places. I am not sure if this is our setup though. Will do a vanilla install and report later today. Sorry for the delay. Too many things going on at once he.

PR #667, opened 2019-07-19
Pip improvements

nerdvegas:
Our existing pip integration had some fairly major shortcomings, hopefully this PR addresses many of them.

Main issues addressed:

* Conversion of pip version specifiers to equivalent rez Requirement is more robust
* Variant requirements are correctly constructed based on env markers if present
* Extras are properly managed
* Requirements are properly filtered out based on extras and env markers
* More robustly manages mix of PEP426 and PEP528 requirements specs
* Utility pip-related code has been moved to a separate utils/pip.py file


instinct-vfx:
I am working on a test-suite to test with lots of packages. Already found a kind of peculiar issue.
Installing pytest fails because the variant contains `<` which is one of the redirect characters in cmd.exe.
My main question would be the actual variant though as it creates a single variant with the following packages:
```
variants = \
   [['python-2.7',
     'funcsigs-1.0+',
     'pathlib2-2.2.0+',
     'more_itertools-4.0.0+<6.0.0']]
```

While the resulting path is only invalid on windows it seems, the variant still does not really make sense to me on first sight. Needs more investigation. The shell error may be solved by hashed variants i guess?

instinct-vfx:
Looking at the package (thanks @lambdaclan for the support on that one!) It seems the variant is indeed correct.
i just gave hashed variants a try by adding `pkg.hashed_variants = True` to rez pip here: https://github.com/nerdvegas/rez/blob/61e0a641b91b68fb8dca449c69c196c9f354328c/src/rez/pip.py#L324

That seems to work fine for pytest (and also for using the resulting package).

lambdaclan:
Adding the dist-info metadata for reference (pytest)

```
Requires-Python: >=2.7, !=3.0., !=3.1., !=3.2., !=3.3.
Requires-Dist: py (>=1.5.0)
Requires-Dist: six (>=1.10.0)
Requires-Dist: packaging
Requires-Dist: attrs (>=17.4.0)
Requires-Dist: atomicwrites (>=1.0)
Requires-Dist: pluggy (<1.0,>=0.12)
Requires-Dist: importlib-metadata (>=0.12)
Requires-Dist: wcwidth
Requires-Dist: funcsigs (>=1.0) ; python_version < "3.0"
Requires-Dist: pathlib2 (>=2.2.0) ; python_version < "3.6"
Requires-Dist: more-itertools (<6.0.0,>=4.0.0) ; python_version <= "2.7"
Requires-Dist: more-itertools (>=4.0.0) ; python_version > "2.7"
Requires-Dist: colorama ; sys_platform == "win32"
Provides-Extra: testing
Requires-Dist: argcomplete ; extra == 'testing'
Requires-Dist: hypothesis (>=3.56) ; extra == 'testing'
Requires-Dist: nose ; extra == 'testing'
Requires-Dist: requests ; extra == 'testing'
Requires-Dist: mock ; (python_version == "2.7") and extra == 'testing'
```

Rez pip now handles these:

```
Requires-Dist: funcsigs (>=1.0) ; python_version < "3.0"
Requires-Dist: pathlib2 (>=2.2.0) ; python_version < "3.6"
Requires-Dist: more-itertools (<6.0.0,>=4.0.0) ; python_version <= "2.7"
Requires-Dist: more-itertools (>=4.0.0) ; python_version > "2
```
and creates the variants accordingly but it leads to invalid paths unless hashed variants are used.

nerdvegas:
Yeah I expected this, didn't get a chance to mention.

So, hashed variants do work currently. The only downside is that the
variant 'handle' used to define the variant is still integer-based, but
that needs to be updated so it's based on the hash instead. It's still
ok-ish as it is today, but it'd be more robust under all conditions if it
were hash-based. This is a feature I need to get to soon.

I wonder if we should just enabled hashed variants on rez-pipified
packages. Thorsten if you've tested this and it seems to work, perhaps it's
good to go. It would solve the variant problem..

A

instinct-vfx:
I am in the progress of testing this against a bigger set of pip packages, will report on methodology asap. I do see another weird error that i can not reliably reproduce or track down. When installing packages i sometimes get the following:

```
ERROR: importlib-metadata 0.18 requires configparser>=3.5; python_version < "3", which is not installed.
ERROR: pytest 4.6.3 requires colorama; sys_platform == "win32", which is not installed.
ERROR: arrow 0.14.1 requires backports.functools-lru-cache>=1.2.1; python_version == "2.7", which is not installed.
```

So far it has always been these three (or a subset). I am not always getting it or at least fail to find the packages calling this. Every time i call rez-pip only on the packages that seem to call this it works :(

If anyone has ideas or pointers that would be really welcome. (Windows of course).

nerdvegas:
Note to self:

hey, so I found a problem. There was something odd in the original rez-pip code and I didn't know why it was there... now I know
8:32 PM
it seems that pip pkg names are non-case-sensitive... and there are cases where a package will list a required pkg with a name that is a different case to the pkg that pip downloads
8:32 PM
eg, Flask requires click, but the pkg pip downloads is actually Click

nerdvegas:
Fix for case sensitivity issues added

src/rez/utils/pip.py
```diff
@@ -0,0 +1,698 @@
...
+import pkg_resources
+
+from rez.vendor.distlib.util import parse_name_and_version
+from rez.vendor.packaging.version import (
+    parse as packaging_parse,
+    LegacyVersion as packaging_LegacyVersion,
+    InvalidVersion as packaging_InvalidVersion
+)
+from rez.vendor.packaging.requirements import Requirement as packaging_Requirement
+from rez.vendor.packaging.markers import _operators as marker_operators
```
JeanChristopheMorinPerso: Unused import.

nerdvegas:
So I've been thinking about it.. unless there are any objections, I think I'm going to just enable hashed variants on all rez-pipified packages, since that solves the varianting issue (ie ranges such as >N ending up in the variant root path, now that we're correctly constructing variants).

I'll also add some custom attributes such as `from_pip` and `is_pure_python`, that's useful info that we may as well keep intact in the rez package.

instinct-vfx:
That will also give hashed variants some thorough testing! I agree.

nerdvegas:
You know it also just struck me that it's a bit of a feat that simply setting `pkg.hashed_variants=True` works at all. We have a completely different case where `PackageMaker` is getting used to create a package (rez-pip, whereas the main one is of course rez-build)... and yet everything is working the same. That's fortunate given what we're trying to do.

src/rez/utils/pip.py
```diff
@@ -0,0 +1,705 @@
...
+        | !=1.2.*     | <1.2|1.3+   |
+
+    Args:
+        specifier (str): PEP440-compatible specifier string.
+
+    Returns:
+        `VersionRange`: Equivalent rez version range.
+    """
+    ranges = []
+    parts = specifier.split(',')
```
JeanChristopheMorinPerso: If you passed the `SpecifierSet` as mentioned lower, you could simply do `list(specifier)`. `packaging` will take care of removing all spaces.

Also, by passing the specifier set object, I think we could benefit from the packaging API here. For example, `list` on a `SpecifierSet` returns specifiers, from which you can get the operator with `.operator` and the version with `.version`. From the version we can do `packaging.version.Version(v)` and then use the methods to know if it's a pre/post release for example. We might still have to do some `endswith` but at least we will do way less than we do right now.

src/rez/utils/pip.py
```diff
@@ -0,0 +1,705 @@
...
+
+    Returns:
+        str: Pip version, eg "19.1.1", or None if version couldn't be detected
+    """
+    pattern = r"pip\s(?P<ver>\d+\.*\d*\.*\d*)"
+
+    if "Windows" in platform.system():
+        # https://github.com/nerdvegas/rez/pull/659
+        ver_str = subprocess.check_output(
+            pip_exe + " -V",
```
JeanChristopheMorinPerso: We might want to do `python -m pip` instead. That way we could be sure which pip will be used. I would even go one step further and use `sys.executable` to get the path to the correct python.

src/rez/utils/pip.py
```diff
@@ -0,0 +1,705 @@
...
+    #
+    requires = installed_dist.run_requires
+
+    # filter requirements
+    for req_ in requires:
+        reqs = normalize_requirement(req_)
+
+        for req in reqs:
+            # skip if env marker is present and doesn't evaluate
+            if req.marker and not req.marker.evaluate():
```
JeanChristopheMorinPerso: I have to admit that I'm confused here. Tell me if I have understood everything:

At this point, the extras have been removed from the marker. So the marker should only contain a python version and a platform?

src/rez/utils/pip.py
```diff
@@ -0,0 +1,705 @@
...
+    if "platform" in sys_requires:
+        sys_variant_requires.append("platform-%s" % _system.platform)
+
+    if "arch" in sys_requires:
+        sys_variant_requires.append("arch-%s" % _system.arch)
+
+    if "os" in sys_requires:
+        sys_variant_requires.append("os-%s" % _system.os)
+
+    if "python" in sys_requires:
```
JeanChristopheMorinPerso: Is this if necessary?

src/rez/utils/pip.py
nerdvegas: This code needs to be updated actually.. currently it causes pip to be run from outside of its rez context (if rezified pip was found), which is completely wrong. I'm gonna fix this quite soon so ignore this for now. I'll make it part of this PR though.

On that note, sys.executable would only get used if no rezified pip packages was found (but yes I agree that it should be used in that case, and that pip exe shouldn't be used directly).

src/rez/utils/pip.py
nerdvegas: The marker can refer to a number of things - platform, python, os (see https://www.python.org/dev/peps/pep-0508/#id24). I'm not sure what you are getting at though? The marker might specify python-3, and we might be running python-2.7, in which case the req should be skipped.

In any case, I think I see a problem - the marker is going to get evaluated wrt current python interpreter, even though a rezified pip may have been used (and thus, a potentially different python version). We do avoid a lot of potential problems if the system pip (ie the one rez gets installed with) is only ever used, but then you're restricted to rez-pipping only for that version of python.

src/rez/utils/pip.py
nerdvegas: I think it's cleaner to have it here, even though python is always a requirement. Then all the code that deals with converting a simple list of system requirements, into a more specific list, is all in the one place. Especially as later, it may get more complex and may not just default to python-MAJOR.MINOR.

src/rez/pip.py
```diff
@@ -385,47 +222,51 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
...
-                    if interpret(requirement["environment"]):
-                        requirements.extend(_get_dependencies(requirement, distributions))
-                elif "extra" in requirement:
-                    # Currently ignoring optional requirements
-                    pass
-                else:
-                    requirements.extend(_get_dependencies(requirement, distributions))
-
+    # get list of package and dependencies
+    distributions = list(distribution_path.get_distributions())
```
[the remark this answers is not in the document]
nerdvegas: I think this was just missed after code reorg (moving stuff into utils.pip). Will fix

src/rez/utils/pip.py
nerdvegas: I'm getting to this soon - yeah I'll redo this using native SpecifierSet as it looks like that'll simplify things

src/rez/utils/pip.py
```diff
@@ -0,0 +1,705 @@
...
+            # this will always be the entire version string
+            return pkg_version.base_version.lower()
+        else:
+            raise packaging_InvalidVersion(
+                "Version: {} is not compatible with PEP440.".format(dist_version)
+            )
+
+    rez_version = ""
+
+    if pkg_version.release:
```
JeanChristopheMorinPerso: This condition cannot be false right?

src/rez/utils/pip.py
```diff
@@ -0,0 +1,705 @@
...
+        if spec.startswith("!=") and spec.endswith(".*"):
+            return fmt("<{V}|{VNEXT}+")
+
+        # !=1 --> <1|1.1+; !=1.2 --> <1.2|1.2.1+
+        if spec.startswith("!="):
+            return fmt("<{V}|{VADJ}+")
+
+        raise PackageRequestError(
+            "Don't know how to convert PEP440 specifier %r into rez equivalent",
+            specifier
```
JeanChristopheMorinPerso: This gives an error: `TypeError: __init__() takes at most 2 arguments (3 given)`. It looks like you use the "lazy logger" technic her, which doesn't work for exceptions :stuck_out_tongue: .

src/rez/utils/pip.py
```diff
@@ -0,0 +1,705 @@
...
+
+    # AND together ranges
+    total_range = VersionRange(ranges[0])
+
+    for range_ in ranges[1:]:
+        range_ = VersionRange(range_)
+        total_range = total_range.intersection(range_)
+
+        if total_range is None:
+            raise PackageRequestError(
```
JeanChristopheMorinPerso: @nerdvegas I'm having difficulties to find when this happens. Could you enlighten me with an example please?

nerdvegas:
That would happen if you encountered a requirement like `foo (<2, >3)`.
There's every chance you would never encounter this (because why would
someone have a package published to PyPi with an impossible requirement
like this), but the check is here for completeness nonetheless. This is a
utility function that shouldn't make assumptions, so one can assume it may
be given a nonsensical specifier set.

src/rez/utils/pip.py
```diff
@@ -0,0 +1,705 @@
...
+        "python_implementation": [_py],
+
+        "sys.platform": [_plat],
+        "sys_platform": [_plat],
+
+        "os.name": [_plat, _arch, _os],
+        "os_name": [_plat, _arch, _os],
+
+        "platform.machine": [_plat, _arch],
+        "platform_machine": [_plat, _arch],
```
JeanChristopheMorinPerso: Technically the platform isn't necessary here. What was the thinking behind adding the platform? I don't mind to add it, but I would prefer it to be documented I think.

nerdvegas:
The thinking was that I don't think I've ever seen an only arch-dependent
package before. If that ever makes sense then I'm happy to change. However
I get the sense that python packaging requirements are somewhat loose,
perhaps it's good to be pessimistic - ie to assume that platform.machine,
os.name etc actually implies platform dependency? What do you think?

src/rez/utils/pip.py
JeanChristopheMorinPerso: We could override the python version all the time with the one we are running under or the one specified when `rez-pip` was invoked. So basically always do:

```python
req.marker.evaluate(
    environment={
        "python_full_version": str(python_version),
        "python_version": str(python_version.trim(2)),
        "implementation_version": str(python_version)
    }
)
```

src/rez/utils/pip.py
JeanChristopheMorinPerso: From Allan:
>The thinking was that I don't think I've ever seen an only arch-dependent
package before. If that ever makes sense then I'm happy to change. However
I get the sense that python packaging requirements are somewhat loose,
perhaps it's good to be pessimistic - ie to assume that platform.machine,
os.name etc actually implies platform dependency? What do you think?

I see what you mean. I have the feeling we shouldn't take guesses here and just do a one to one with the PEP. So basically each should only have one value. Like `os_name` should only map to `_os` (since the `os` package already requires the arch and platform package). `platform_machine` is really platform agnostic. It could weel be `x64` or `x86_64` or `x32` for example. That would be for a C extension pip package I guess.

Does that make sense?

src/rez/utils/pip.py
nerdvegas: I believe it can be, for versions like (eg) 1.a2, which is a pre-release, not a release.

src/rez/utils/pip.py
nerdvegas: Yeah you read my mind, was hoping it was possible to do this way and looks like it is, cheers.

src/rez/utils/pip.py
JeanChristopheMorinPerso: Actually, this can only happen if the version is a legacy version, like `a2`. If it's `1.a2`, the release would be `1`.

nerdvegas: ah yup I see, removing

src/rez/utils/pip.py
nerdvegas: yeah, will go with that, if we hit issues we can always revert to the more pessimistic implementation later.

src/rez/utils/pip.py
```diff
@@ -0,0 +1,689 @@
...
+            # this will always be the entire version string
+            return pkg_version.base_version.lower()
+        else:
+            raise packaging_InvalidVersion(
+                "Version: {} is not compatible with PEP440.".format(dist_version)
+            )
+
+    # the components of the release segment excluding epoch or any
+    # prerelease/development/postrelease suffixes
+    rez_version = pkg_version.base_version
```
JeanChristopheMorinPerso: Umm, I'm sorry but that introduces a bug. I thought the base version would be enough, but it includes the epoch... So we will need to go back to the previous version of the code. Sorry about that.

nerdvegas:
Ah cheers, haven't rerun the tests yet, thx

src/rez/utils/pip.py
```diff
@@ -0,0 +1,689 @@
...
+        "implementation_name": [_py],
+        "implementation_version": [_py],
+        "platform_python_implementation": [_py],
+        "platform.python_implementation": [_py],
+        "python_implementation": [_py],
+
+        "sys.platform": [_plat],
+        "sys_platform": [_plat],
+
+        "os.name": [_os],
```
JeanChristopheMorinPerso: Thinking about it even more, `os.name` can return `posix` (Linux, OSX, etc), `nt` (Windows) or `java` (Jython). We can't do much about Jython, but for the others, it's basically the platform, not the OS (OS would add a dependency on the arch).

nerdvegas:
Do you know of a reference anywhere that lists valid os.name values? I'll
update this with a comment.

nerdvegas:
Pls ignore.. I see that it maps to python os.name.
A

src/rez/utils/pip.py
JeanChristopheMorinPerso: > Do you know of a reference anywhere that lists valid os.name values? I'll
update this with a comment.

They are, here: https://docs.python.org/3/library/os.html#os.name
https://docs.python.org/3/library/platform.html#platform.system


These are the values I know that are mentioned in the docs.

nerdvegas:
Ok, I've merged in those tests and updated wrt PR changes here. Everything looks good. Give it a whirl and let me know how you go. If there aren't any problems ten I'll merge to master tomorrow.

Comprehensive tests btw! Would also be good to have a rez-pip install happen from the test installed_distributions dir also; you mention this should be possible but I don't know the details just yet.

Thx
A

instinct-vfx:
I have a first issue here. Have yet to investigate, but if i rez-pip install pip as a package rez-pip failes afterwards:
`16:33:16 INFO     Using pip-19.2.2 (c:\users\tkaufmann\rez_packages_local\pip\19.2.2\package.py[0])
  File "C:\Python27\Lib\runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "C:\Python27\Lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "c:\ProgramData\Mackevision\rez_local_development\rez\pip-improvements2\platform-windows\Scripts\rez\rez-pip.exe\__main__.py", line 7, in <module>
  File "c:\programdata\mackevision\rez_local_development\rez\pip-improvements2\platform-windows\lib\site-packages\rez\cli\_entry_points.py", line 187, in run_rez_pip
    return run("pip")
  File "c:\programdata\mackevision\rez_local_development\rez\pip-improvements2\platform-windows\lib\site-packages\rez\cli\_main.py", line 148, in run
    returncode = run_cmd()
  File "c:\programdata\mackevision\rez_local_development\rez\pip-improvements2\platform-windows\lib\site-packages\rez\cli\_main.py", line 140, in run_cmd
  File "c:\programdata\mackevision\rez_local_development\rez\pip-improvements2\platform-windows\lib\site-packages\rez\cli\pip.py", line 47, in command
    release=opts.release)
  File "c:\programdata\mackevision\rez_local_development\rez\pip-improvements2\platform-windows\lib\site-packages\rez\pip.py", line 194, in pip_install_package
    py_exe, context = find_pip(pip_version, python_version)
  File "c:\programdata\mackevision\rez_local_development\rez\pip-improvements2\platform-windows\lib\site-packages\rez\pip.py", line 115, in find_pip
    if int(pip_major) < 19:
ValueError: invalid literal for int() with base 10: ''`

JeanChristopheMorinPerso:
For me, having pip as a rez package should be prohibited as it's actually searching for quite a lot of trouble...

But anyway. Did you get this error while running `rez-pip install pip` or it worked when you did `rez-pip install pip`?

instinct-vfx:
`rez-pip install pip` worked fine, but all subsequent calls to rez-pip then fail. It tries to find_pip and fails when pip is present as a rez package it seems.

I wonder though: Don't we need pip as a rez package in order to allow installing python-3.x packages with rez-pip?

nerdvegas:
Hmm, so I've rez-pip'd pip before, and used the resulting pip, and it's
worked. I'll have a look at this.

Wrt whether this makes sense - absolutely it does. Whatever version of
python/pip rez itself has been installed with should be incidental, it
should have no bearing on how other packages are getting built. So whilst
it's convenient to pip-install packages using python-2.7, and rez's
2.7-based pip, it's also completely feasible that you'd want to pip-install
a python-3.x package... and vice versa.

A

nerdvegas:
I can `rez-pip -i pip`, then do further `rez-pip -i X`. It uses the
rezified pip and everything seems fine. Perhaps it's a Windows issue...
what happens specifically?
A

instinct-vfx:
I did a bit of research. We have a central python that is added to path through a wrapper. What happens is that rez falls back to this python executable when in a context. So if it finds the package, it somehow picks up the wrong python on the PATH. I will investigate further. If i wipe this central python from PATH it works as expected (note though that the central python is on path AFTER the local one).

instinct-vfx:
Okay, so the problem is related to system PATH vs. user PATH on windows. This only breaks if there is an additional python on the system path it seems. So for THIS ticket it can be considered out of scope. I will make an issue to track this down as it seems to be more of a general cmd issue.

JeanChristopheMorinPerso:
>I wonder though: Don't we need pip as a rez package in order to allow installing python-3.x packages with rez-pip?

In a normal python environment, pip is installed per python installation, either with --ensurepip (for python 2) when compiling python, by default in python 3 there is already pip installed, or by using the `get-pip.py` bootstrap file which by default will install in the python installation folder. All three options will also install `setuptools` and `wheel` and all three will install straight in the install directory by default. Of course, you can use `get-pip.py` with the `--user` option, though this can create problems if you have pip already installed in the python installation (corner cases, but still).

So to answer your question, if you would want to install python 3 packages, you would just have to call `python3.X -m pip install ...`, where python3.X is the version you specified with `rez-pip --python-version 3.X install <package>`. Doing this guarantees you which version of pip you use.

>So whilst it's convenient to pip-install packages using python-2.7, and rez's
2.7-based pip, it's also completely feasible that you'd want to pip-install
a python-3.x package... and vice versa.

When I'm saying that we shouldn't need a `pip` rez package, I'm not saying that we shouldn't use `rez-pip` to install a python X package even if rez was installed with python Y. I'm just saying that it is entirely possible to do without a pip rez package and that I wouldn't recommend having pip as a rez package just because we cannot guarantee that it will work (since pip is not supposed to be installed with the `--target` option. Yes it's possible to do it, but it's a little bit an offroad track).

instinct-vfx:
That would mean that rez-pip depends on specifically locally installed python interpreters that also need to have ensure pip. Wouldn't it be more stable/predictable to have a dependency on a standard pip package as installed by rez-pip itself?

instinct-vfx:
I have finished my testing and have good news and bad news. Good news first:
Of the 1000 most downloaded pypi packages, 942 installed fine with rez-pip. I threw the remaining packages at pip directly and the majority also failed with pip itself (for various reasons, like requiring python 3.x or not providing binary wheels (at least for python27).

Here are the remaining ones that did work with pip but not rez-pip:

```
mysql-python
pycrypto
pyinotify
suds-jurko
tensorflow-serving-api
ujson
```
I have not done thorough investigations yet, but mysql-python is giving me a headache straight away. I called it the same way that rez-pip does (`python -m pip install --use-pep517 --target PATH --no-deps`) but rez-pip pulls a source release whereas pip pulls a wheel.

maxnbk:
I can probably help with some suggestions for why some of these might fail.

Most likely, in some of these cases, you're seeing compilations fail, which are going to depend on finding an appropriate compiler on the host system, or appropriate flags to compile against other rezified pip or unrelated packages. Some also do not declare a dependency on a system-level devel library being installed, which is unfortunate.

There are also several packages which, in their setup.py, declare additional dependencies not a part of the pip-fetchable metadata. I can get a larger list of examples of these later, but a typical example is a need for vcversioner or setuptools_scm or pytest_runner.

In addition, we have a tool internally that has (at least until this work here can be more directly finished) supplanted rez-pip, which can be supplied an argument like "--links" that specifies packages to which CXXFLAGS / CFLAGS / LDFLAGS are meant to be appended at compile time, so that, for example, a rezified scipy can find and link to a rezified numpy at compile time, when building from a tarball and not just relying on wheels (The vast majority of our packages are compiled from source. Probably 99%+ of all packages that provide an sdist, as a direct result of this ability).

I have more ideas, but this is a good starter list of the most common reasons.

instinct-vfx:
I am fine if they fail in pip, then failure of rez-pip is a given. Just the ones that fail in rez-pip but work in pip are a concern.

I could check a few more of the list. I started the test in a different shell and forgot to wipe PIP_INDEX_URL which made it pull internal wheels for some of these packages. With these off the list rez-pip installed all but 2 packages (!!),

The only ones that fail with rez-pip are
```
suds-jurko
tensorflow-serving-api
```

src/rez/pip.py
```diff
@@ -218,63 +70,52 @@ def run_pip_command(command_args, pip_version=None, python_version=None):
...
-        pip_exe = which("pip")
-
-        if pip_exe:
-            print_warning(
-                "pip rez package could not be found; system 'pip' command (%s) "
-                "will be used instead." % pip_exe)
-            context = None
-        else:
-            raise
+        # fall back on system pip
```
JeanChristopheMorinPerso: Before falling back to system (so virtualenv's python), I would try to get the latest python version (if python was not provided) available as a rez package and assume that there might be pip installed into it. If this fails too, then fallback to rez's virtualenv pip.

Would that make sense?

nerdvegas:
> > I wonder though: Don't we need pip as a rez package in order to allow installing python-3.x packages with rez-pip?
> 
> In a normal python environment, pip is installed per python installation, either with --ensurepip (for python 2) when compiling python, by default in python 3 there is already pip installed, or by using the `get-pip.py` bootstrap file which by default will install in the python installation folder. All three options will also install `setuptools` and `wheel` and all three will install straight in the install directory by default. Of course, you can use `get-pip.py` with the `--user` option, though this can create problems if you have pip already installed in the python installation (corner cases, but still).
> 
> So to answer your question, if you would want to install python 3 packages, you would just have to call `python3.X -m pip install ...`, where python3.X is the version you specified with `rez-pip --python-version 3.X install <package>`. Doing this guarantees you which version of pip you use.
> 
> > So whilst it's convenient to pip-install packages using python-2.7, and rez's
> > 2.7-based pip, it's also completely feasible that you'd want to pip-install
> > a python-3.x package... and vice versa.
> 
> When I'm saying that we shouldn't need a `pip` rez package, I'm not saying that we shouldn't use `rez-pip` to install a python X package even if rez was installed with python Y. I'm just saying that it is entirely possible to do without a pip rez package and that I wouldn't recommend having pip as a rez package just because we cannot guarantee that it will work (since pip is not supposed to be installed with the `--target` option. Yes it's possible to do it, but it's a little bit an offroad track).

Ah yes that makes sense.

So what I would propose is to first assume that python itself provides pip, and only fall back to searching for a rezified pip if that is not the case. Really the reason to do this is just for backwards compatibility.

I think I'd rather do this in a separate PR though, as this one is dragging on, and the current behavior already (erroneously as @JeanChristopheMorinPerso points out) expects rezified pip. I would aim to do this directly after merging this PR, and probably will not do a release in the meantime.

How does that sound?

src/rez/pip.py
nerdvegas: I believe my latest comment addresses this - as in yes we would change things to be expecting python package to provided pip, and would be checking for this before falling back to rezified pip, then to system pip (and rezified pip just for backwards compatibility - current users could very well have python without pip, and rezified pip).

JeanChristopheMorinPerso:
>I think I'd rather do this in a separate PR though, as this one is dragging on, and the current behavior already (erroneously as @JeanChristopheMorinPerso points out) expects rezified pip. I would aim to do this directly after merging this PR, and probably will not do a release in the meantime.

That's good for me.

PR #668, opened 2019-07-19
Fix pydot regression

nerdvegas:
This fixes a regression introduced in `2.34.0`, which causes `rez-context -g` to
fail. The pydot vendor package was updated, and the newer version includes a
breaking change. Where `pydot.graph_from_dot_data` used to return a single graph
object, it now returns a list of graph objects.


nerdvegas:
Please note the included CHANGELOG update! We should begin enforcing this (cause I'm getting tired of doing changelog updates :P ).

PR #669, opened 2019-07-19
Adding (not latest) printout in resolves

smaragden:
This is something we have had in our local branch for a long time.
In the resolve printout, it will show packages that are not the latest version.

like this:
resolve

We have found this really useful for keeping our packages up to date.

I guess an extra `get_latest_package()` call could add a performance penalty and you might want to be able to configure this feature. We never bothered as we find the feature really useful and never felt any performance penalty.

Let me know if you want me to make it configurable or if you have any other concerns.

### DIFF
--- a/src/rez/resolved_context.py
+++ b/src/rez/resolved_context.py
@@ -8,7 +8,7 @@ from rez.system import system
 from rez.config import config
 from rez.util import shlex_join, dedup
 from rez.utils.sourcecode import SourceCodeError
-from rez.utils.colorize import critical, heading, local, implicit, Printer
+from rez.utils.colorize import warning, critical, heading, local, implicit, Printer
 from rez.utils.formatting import columnise, PackageRequest, ENV_VAR_REGEX
 from rez.utils.data_utils import deep_del
 from rez.utils.filesystem import TempDirs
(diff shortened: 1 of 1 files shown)


nerdvegas:
I like it, but there's something to consider.

In the not-too-distant future, there's going to be the ability to create 'standalone' contexts. These will contain all info needed to use their packages, and will sit alongside a localized package cache. You would be able to (for eg) create such a context on a server VM/container, and it'd run standalone.

The problem is, by introducing this not-latest feature, you necessitate access to the original package repo(s) once more.

Perhaps what we could do is identify these standalone contexts, and disable the not-latest feature in that case.

In any case I'm a little torn - rez-context currently shows you info on the given context, without needing to pull data from anywhere else. This changes that. But at the same time I can see how it's useful.

smaragden:
Yea I guess it pretty much defeats the purpose of the cache to go back and query the main repo.
We mainly use it as a debugging tool when we prepare the contexts for users. So I guess if we can disable it when not running against the main repo it would still be useful the way we use it today.

nerdvegas:
So, been thinking about this. Rez-context does run in different modes that show comparisons to external info - for eg, the `--fetch` option shows a diff against a re-resolved version of the context, which is a little like the info being shown here. 

So I think the best way forward is to do the same for this - provide some arg that changes the behaviour, and just lists the resolved packages, and if they are local/latest/not-latest etc.

I am thinking `--check`. Thoughts?

smaragden:
That should suffice. And would fulfill the purpose as a debugging tool.

PR #674, opened 2019-07-23
Fix missing import in powershell plugin

instinct-vfx:
A missing import slipped into the last MR. I do think that the powershell plugin needs some more work. Most notably i would like it to properly escape paths on alias() calls. But that is way less trivial than it sounds at first. I also plan adding a second plugin for poweshell6+. It will be separate because it actually has differing feature sets and it also has a different executable name (pwsh).

The only functional change is the added import on top. The rest is auto-formatting for pep8 in vscode.

### DIFF
--- a/src/rezplugins/shell/powershell.py
+++ b/src/rezplugins/shell/powershell.py
@@ -82,6 +82,8 @@ class PowerShell(Shell):
             whitespace = r"[\s]+"
             return whitespace.join(parts)
 
+        # TODO: Research if there is an easier way to pull system PATH from
+        # registry in powershell
         paths = []
 
         cmd = [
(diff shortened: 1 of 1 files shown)


nerdvegas:
Hey Thorsten, would you mind dropping some TODO comments in so that the existing problems are identified? Cheers!

instinct-vfx:
Not at all, will go ahead asap. The powershell 6+ stuff is more a new ticket rather than comments as it will be a separate file. I also forgot another thing: I can possibly replace the existing registry hacks to get the clean system env with something more elegant in powershell. But i have to research fist. Will add TODO notes and report here.

instinct-vfx:
Realized i had another mishap. Fixed and added TODOs for the two things i have in mind. Will create an issue for pwsh support.

The fix has been tested with a few powershell ready packages i have that make proper use of the tweaked alias() setup.

PR #675, opened 2019-07-23
Change _rez_fwd to _rez-fwd as the entry point was renamed in 2.38.0

### LINKED ISSUE #671: build scripts generated with incorrect shebang arg
After `rez build -s`, [v2.38.1](https://github.com/nerdvegas/rez/releases/tag/2.38.1) creates this first line in build-env:

`#!/usr/bin/env _rez_fwd`

However, the installed name is now `_rez-fwd`

I've fixed it locally by changing [src/rez/util.py:80](https://github.com/nerdvegas/rez/blob/e6217a09cc13656fbde69c0f057618fecc472aad/src/rez/util.py#L80) to:

`    create_executable_script(filepath, body, "_rez-fwd")`

instinct-vfx:
Fixes #671 

As the code prior to 2.38.0 already had a TODO to rename the executable i assume it was done on purpose. This fixes the reference in util.py to match the new name.

### DIFF
--- a/src/rez/util.py
+++ b/src/rez/util.py
@@ -56,8 +56,8 @@ def create_executable_script(filepath, body, program=None):
     # clean up the files once the test has run.  Temporarily we don't bother
     # setting the permissions, but this will need to change.
     if os.name == "posix":
-    	os.chmod(filepath, stat.S_IRUSR | stat.S_IRGRP | stat.S_IROTH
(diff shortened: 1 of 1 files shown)


nerdvegas:
Ah crap actually I think we need to revert the bin name instead.. the change in 2.38.0 wasn't intentional. Changing the name of _rez_fwd will break existing suites when they upgrade rez :/

Closing in lieu of my own PR to revert this regression.

Issue #678, opened 2019-07-23
Add shell plugin for poweshell 6+


instinct-vfx:
As a user i would like to be able to use powershell core 6.0/6.1. There is no feature parity between powershell 5.x and powershell core and the executable has been renamed to pwsh.

This issue aims to add support for powershell core. It is not yet clear to me how much of a difference (if any at all) there has to be except for the executable name.

instinct-vfx:
Initial PR created, functionality is there but not tested extensively yet.

nerdvegas:
Closed by #679 (https://github.com/nerdvegas/rez/releases/tag/2.39.0)

PR #679, opened 2019-07-25
Add powershell core 6+ support (pwsh)

instinct-vfx:
PowerShell core (6.x and newer) have quite some differences. For the shell plugin most notably is the way that the sourcing of the ps1 script works and the different name of the executable.

Issue: #678


nerdvegas:
Not sure if you thought this was ready for merging yet, but given that it's a new plugin that can't break anything previous, I've merged it anyway :P

PR #683, opened 2019-07-29
Add missing import subprocess

dsagilles:
resolving a powershell environment in windows would otherwise give me "NameError: global name 'subprocess' is not defined" (powershell.py line 108)

### DIFF
--- a/src/rezplugins/shell/powershell.py
+++ b/src/rezplugins/shell/powershell.py
@@ -9,6 +9,7 @@ from rez.backport.shutilwhich import which
 from functools import partial
 import os
 import re
+import subprocess
 
 try:
     basestring


nerdvegas:
Thanks. Seems related to #674. Will get this merged next week.

instinct-vfx:
Both PRs fix the same issue, mine also has the ToDo comment for updating the way PATH is fetched from the registry. If this PR is preferred (because it does not include autoformatting updates, not sure what your stance on these is) then please also include the ToDo note in line 85/86.

nerdvegas:
Closing in favor of #674.

RE autoformatting - I prefer not to have that mixed in with functional PRs. However at some point I would very much like to blacken the source (https://github.com/psf/black), at which point any discussion on formatting is moot (and there'd be a pre-commit hook to enforce).

Issue #688, opened 2019-08-08
[Regression - Version >= 2.39.0]  ConfigurationError: Error in Rez configuration under plugins.shell


lambdaclan:
The powershell addition seems to have caused a regression when using rez-pip on Windows.

```
C:\Users\Joe\Downloads\rez [(765f0a1...) +0 ~3 -0 !]
λ  rez -V
Rez 2.38.2
C:\Users\Joe\Downloads\rez [(765f0a1...) +0 ~3 -0 !]
λ  rez-pip -i pcapng
18:39:47 INFO     Using pip-19.2.1 (C:\Users\Joe\packages\pip\19.2.1\package.py[0])
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
Collecting pcapng
Installing collected packages: pcapng
Successfully installed pcapng-0.1.23
18:39:53 DEBUG    Removed 1.0 due to Requires-Dist
18:39:53 DEBUG    Removed 1.1 due to Requires-Dist
18:39:53 DEBUG    Removed 1.2 due to Provides-Extra
18:39:53 DEBUG    Found c:\users\joe\appdata\local\temp\pip-9mi4is-rez\rez_staging\python\pcapng-0.1.23.dist-info

1 packages were installed:
  pcapng-0.1.23: C:\Users\Joe\packages\pcapng\0.1.23\package.py (python-2.7)

C:\Users\Joe\Downloads\rez [(765f0a1...) +0 ~3 -0 !]
λ
```

```
C:\Users\Joe\Downloads\rez [(e55d68b...) +0 ~3 -0 !]
λ  rez -V
Rez 2.39.0
C:\Users\Joe\Downloads\rez [(e55d68b...) +0 ~3 -0 !]
λ  rez-pip -i pcapng
18:48:22 INFO     Using pip-19.2.1 (C:\Users\Joe\packages\pip\19.2.1\package.py[0])
18:48:24 ERROR    ConfigurationError: Error in Rez configuration under plugins.shell: missed keys set([Schema('pwsh')])
C:\Users\Joe\Downloads\rez [(e55d68b...) +0 ~3 -0 !]
λ 
```

jamesmclaren:
looks like the pwsh issue also affects 2.40.0

[...]

Issue #685, opened 2019-08-01
Wrong case requests lead to packages resolved twice with different versions


instinct-vfx:
Please excuse the weird subject line, but i had a hard time finding a one-liner for this one. Thanks to Dominic who discovered this one :)

It's windows only and most likely has to do with the fact that on windows file systems are not case sensitive. This causes an issue if someone requests a package with wrong case. That makes Rez see the package but the solver seems to interpret them as two different packages resulting in cases where the same package is resolved twice with potentially different versions.

Example:

We have a 3dsmax package and a maxtools package. The maxtools package requires the 3dsmax package. A correct request could be: `rez-env 3dsmax-2018 maxtools` and yields the expected result:
```
$ rez-env 3dsmax-2018 maxtools
INFO:__main__:Launching rez-env(rez_meta-0.5.7, rez-2.37.0, rez_config-0.4.3)
resolved by tkaufmann@WS-STR-DL135, on Thu Aug 01 11:55:14 2019, using Rez v2.37.0

requested packages:
3dsmax-2018
maxtools
~platform==windows  (implicit)
~arch==AMD64        (implicit)
~os==windows-10     (implicit)

resolved packages:
3dsmax-2018       \\data\rez-package-staging\3dsmax\2018\platform-windows\arch-AMD64
Qt.py-1.2.1       \\data\rez-package-staging\Qt.py\1.2.1\ff5a17a870e473adea6d65972631222d54a381e6
arch-AMD64        \\data\rez-package-staging\arch\AMD64
iniparse-0.4      \\data\rez-package-staging\iniparse\0.4\python-2.7
maxtools-3.0.5    \\data\rez-package-staging\maxtools\3.0.5\platform-windows\arch-AMD64
platform-windows  \\data\rez-package-staging\platform\windows
python-2.7.16     C:\ProgramData\Mackevision\rez_packages_local_staging\python\2.7.16\platform-windows\arch-AMD64
xlrd-1.2.0        \\data\rez-package-staging\xlrd\1.2.0\platform-windows\arch-AMD64\python-2.7
```

Now if someone requests with a different case than the actual package than both (the one created by maxtools' requires and the actually requested one) get added as different packages and resolve to different versions each pointing to the same package: `rez-env 3dsMax-2018 maxtools` (note the case change in 3dsmax)

```
$ rez-env 3dsMax-2018 maxtools
[...]

PR #689, opened 2019-08-08
[FIX] Make package resolve request respect case sensitivity -- Windows

lambdaclan:
# Description

[Relates to issue #685]

## Windows file systems are not case sensitive

Unlike Linux and Unix file systems, Windows paths are not case sensitive this means that requests to resolve packages via `rez-env` using the wrong case will go through and resolve successfully. This can cause potential issues on other platforms where using the same name with a different case for a package is valid.

This is [not a Python issue](https://stackoverflow.com/questions/17277566/check-os-path-isfilefilename-with-case-sensitive-in-python) as the path check to the package directory returns True since Windows ignores case sensitivity in path names.

## Issue

The root of the issue occurs at [`src/rezplugins/package_repository/filesystem.py`](https://github.com/nerdvegas/rez/blob/master/src/rezplugins/package_repository/filesystem.py#L787)
where the following check takes place

```
 if os.path.isdir(os.path.join(self.location, name)):
```
with location being the packages directories and name the requested package - this on Windows regardless of the case the name is using will return True if the names match.

### Considered equal

```
rez-env 3dsmax-2018
rez-env 3DSmax-2018
rez-env 3dsMAX-2018
rez-env 3dSmAx-2018
```
The issue goes through multiple channels and can be intercepted in a few places for example
[`src/rez/resolved_context.py`](https://github.com/nerdvegas/rez/blob/master/src/rez/resolved_context.py#L197) where the request takes place

```
self._package_requests = []        
for req in package_requests: - # name can also be checked here for case-sensitivity            
    if isinstance(req, basestring):               
        req = PackageRequest(req)            
    self._package_requests.append(req)
```

### Windows - Case insensitive file system

win_case_01

### Linux - Case sensitive file system

win_case_02

## Solution

This PR solves the issue by ensuring that the requested package exists as a directory in the
any of the packages directories. This is achieved through the use of `os.listdir` **which does give you the directory contents with their actual cases**.

~~In an attempt to get the same error message as case-sensitive systems, the fix has been added to src/rez/packages.py file in the get (package) families function. Failure to detect the package during to wrong case will return zero entries resulting in the expected flow.~~

See https://github.com/nerdvegas/rez/pull/689#pullrequestreview-273995063

## Breakdown

* Add case sensitivity check to package requests

## Type of change

# How Has This Been Tested?

I have tried resolving a series of packages using different case. The results (before, after) are outlined below:

### After

**Test Configuration**:
* Python version: 2.7.16
* OS: Linux, Windows 10
* Toolchain: rez 2.40.1 ~~2.38.2 (#688)~~, pip 19.2.1

## Point release
[Source](https://github.com/nerdvegas/rez/tree/2.40.0) | [Diff](https://github.com/nerdvegas/rez/compare/master...lambdaclan:fix/rez-env-case-sensitive-windows)

**Notes**

Add case sensitivity check to prevent issues during package resolve for OSes with case insensitive file systems.

**Merged pull requests:**

- [FIX] Make package resolve request respect case sensitivity -- Windows [\#689](https://github.com/nerdvegas/rez/pull/689) ([lambdaclan](https://github.com/lambdaclan))

**Closed issues:**

- Wrong case requests lead to packages resolved twice with different versions [\#685](https://github.com/nerdvegas/rez/issues/685)

### DIFF
--- a/src/rezplugins/package_repository/filesystem.py
+++ b/src/rezplugins/package_repository/filesystem.py
@@ -7,6 +7,7 @@ import os
 import stat
 import errno
 import time
+import platform
 
 from rez.package_repository import PackageRepository
 from rez.package_resources_ import PackageFamilyResource, VariantResourceHelper, \
@@ -20,7 +21,7 @@ from rez.utils.formatting import is_valid_package_name
 from rez.utils.resources import cached_property
 from rez.utils.logging_ import print_warning
 from rez.utils.memcached import memcached, pool_memcached_connections
-from rez.utils.filesystem import make_path_writable
+from rez.utils.filesystem import make_path_writable, get_case_canonical_path
 from rez.serialise import load_from_file, FileFormat
 from rez.config import config
 from rez.backport.lru_cache import lru_cache
@@ -461,6 +462,7 @@ class FileSystemPackageRepository(PackageRepository):
         Args:
             location (str): Path containing the package repository.
         """
+        location = get_case_canonical_path(location)
         super(FileSystemPackageRepository, self).__init__(location, resource_pool)
 
         global _settings
@@ -784,7 +786,16 @@ class FileSystemPackageRepository(PackageRepository):
 
     def _get_family(self, name):
         is_valid_package_name(name, raise_error=True)
-        if os.path.isdir(os.path.join(self.location, name)):
+        if "Windows" in platform.system():
+            if os.path.isdir(self.location):
+                if name in os.listdir(self.location):
+                    family = self.get_resource(
+                        FileSystemPackageFamilyResource.key,
+                        location=self.location,
+                        name=name)
+                    return family
+            return None
+        elif os.path.isdir(os.path.join(self.location, name)):
             family = self.get_resource(
                 FileSystemPackageFamilyResource.key,
                 location=self.location,
--- a/src/rezplugins/package_repository/filesystem.py
+++ b/src/rezplugins/package_repository/filesystem.py
@@ -786,16 +786,10 @@ class FileSystemPackageRepository(PackageRepository):
 
     def _get_family(self, name):
         is_valid_package_name(name, raise_error=True)
-        if "Windows" in platform.system():
-            if os.path.isdir(self.location):
-                if name in os.listdir(self.location):
-                    family = self.get_resource(
-                        FileSystemPackageFamilyResource.key,
-                        location=self.location,
-                        name=name)
-                    return family
-            return None
-        elif os.path.isdir(os.path.join(self.location, name)):
+        if os.path.isdir(os.path.join(self.location, name)):
+            # force case-sensitive match on pkg family, on case-insensitive platforms
+            if "Windows" in platform.system() and name not in os.listdir(self.location):
+                return None
             family = self.get_resource(
                 FileSystemPackageFamilyResource.key,
                 location=self.location,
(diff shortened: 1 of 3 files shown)


lambdaclan:
Hello Allan,

Thank you for your feedback. OK got it, I will look into it and make the necessary changes as suggested above.

nerdvegas:
One more subtlety that should be fixed at the same time:

Since paths on windows are case-insensitive, the filesystem package repo filepath should be `lower()`ed on windows, so that stuff like the unique repo identifier, and the local path that gets stored into variant handles, doesn't differ if the case differs.

In other words, currently /sw/packages and /sw/Packages would be getting treated as separate package repos, whereas on Windows they're actually the same. So, there's another bug currently present, it just hasn't caused any problems so far (that I'm aware of!).

To keep this clean and explicit, I'd recommend adding a function like `get_case_canonical_path` (or similar) in utils.filesystem, which would do the lower() in windows, but would return path unchanged on other platforms.

lambdaclan:
Indeed, I had in mind using `lower()` to make the requested package path correct but that would conflict with packages that are actually case-sensitive such as "PyInstaller" for example. Saying that, lowering the path up to the package location makes sense to avoid package repo mix-up. I will add the fix to this PR as well. Hopefully there are no other edge cases that can causes issues...

nerdvegas:
Yeah I meant specifically only up to the repo location, to be clear. Package names must be case sensitive regardless of filesystem.

lambdaclan:
Hello Allan, with regards to this

> In other words, currently /sw/packages and /sw/Packages would be getting treated as separate package repos, whereas on Windows they're actually the same. So, there's another bug currently present, it just hasn't caused any problems so far (that I'm aware of!).
To keep this clean and explicit, I'd recommend adding a function like get_case_canonical_path (or similar) in utils.filesystem, which would do the lower() in windows but would return path unchanged on other platforms.

If we went ahead and simply lowered the whole path wouldn't that cause other indirect issues? For example by default the packages path is ~/packages and as you mention ~/Packages would be treated the same on Windows but what if we do have a ~/Packages path defined on another platform and then try it to use it on Windows via the same rezconfig or via REZ_PACKAGES_PATH? We should possibly have a warning or something that Windows is case insensitive therefore all package paths will be lowered. 

This means on Windows check if the package location path is not lowered (contains uppercase) and if not give a warning that on Windows paths with different case are the same. Unless we take for granted that this is common knowledge.

nerdvegas:
Package repositories have a `_uid` method they need to implement, which
uniquely identifies the repo. It would probably be enough to lower
self.location on windows in this function. Just doing this will ensure that
~/packages and ~/Packages map to the same repo instance on Windows.

Hth
A

lambdaclan:
Sounds good, I will remove the warning then, everything else is implemented as discussed above.

src/rezplugins/package_repository/filesystem.py
```diff
@@ -784,7 +786,16 @@ def _get_families(self):
 
     def _get_family(self, name):
         is_valid_package_name(name, raise_error=True)
-        if os.path.isdir(os.path.join(self.location, name)):
+        if "Windows" in platform.system():
```
nerdvegas: I would simplify:

```
if os.path.isdir(os.path.join(self.location, name)):
    # force case-sensitive match on pkg family, on case-insensitive platforms
    if "Windows" in platform.system() and name not in os.listdir(self.location):
        return None
```

nerdvegas:
Please note my changes. I realised there's another case we didn't consider (unversioned packages); it also made sense to abstract case sensitivity into `platform_`, and remove explicit references to Windows instead.

Issue #694, opened 2019-08-19
Rex and expandable in other shells


bfloch:
This applies to any version up to 2.40.3 as far as I can tell.

Rex does `expandvars("${VAR}")` correctly regardless of the platform:
https://github.com/nerdvegas/rez/blob/master/src/rez/utils/formatting.py#L203

However, the default `expandable` will only work on shells that happen to support the `${VAR}` or `$VAR` syntax.

On windows the tests check for `%VAR%` but in fact the documentation make it sound as if the `${VAR}/$VAR` is the portable form that should work on any platform.

I assume that we should implement the necessary conversion for the expandable case on shells that require it.

So for cmd.exe we would replace `${VAR}` with `%VAR%` and for Powershell we would replace it with `${Env:VAR}`.

This way we can keep the package definitions platform independent. The risk is minimal since users are free to choose `literal`.

The question arises is expandvars should also replace the variables in the case where the variable is not found in the environment.

In both cases we might want to introduce rez configuration options for backwards compatibility.

bfloch:
Got feedback from @nerdvegas in Slack that this is incorrect behavior.

As such I wonder if it is even necessary to introduce backwards compatibility options.
The only case where I see a danger is when people relied on `${VAR}` syntax in variables without declaring them literal. One could argue that this is an inaccurate declaration.

bfloch:
Actually I found a duplicated reported by myself some time ago:
https://github.com/nerdvegas/rez/issues/443

nerdvegas:
So yes, I think we need to support $VAR and ${VAR} forms, which should be translated to their shell equivalents. Are you saying that currently, %VAR% is expected on windows, and that $VAR/${VAR} does not work? I am assuming so (and of course %VAR% would work because it simply ends up verbatim in the cmd source - but I think that that incidental behaviour is fine).

bfloch:
I have a branch which fixes this and many other shell related issues. This allowed to pass **all** shell tests cross platform. Still having some issues with zsh which might not be related to it, but I'll investigate before creating a PR.
https://github.com/bfloch/rez/tree/issue_694-rex-expandable-windows
Stay tuned.

Issue #691, opened 2019-08-12
Quotation marks issues on Windows.


KelSolaar:
Hi,

I'm testing 2.40.1 to upgrade from 2.16.0 and I'm having issues with quotation marks on Windows.

Specifically:

```cmd
C:\Users\thomas>rez-env substance_painter -- "Substance Painter"
'Substance' is not recognized as an internal or external command,
operable program or batch file.
```

It could be related to #664 but I'm not sure yet, I will try to take a look.

Cheers,

Thomas

nerdvegas:
Hi Thomas,

I wonder also if it's related to
https://github.com/nerdvegas/rez/releases/tag/2.35.0. Can you test that and
the version before it? (https://github.com/nerdvegas/rez/releases/tag/2.34.0
).

The changes in 2.35.0 do potentially affect how the command is escaped in
cmd. If that is the case, it's certainly worth looking at the merged PR
thread. Please let me know how you go!

Thanks,
A

KelSolaar:
Hi @nerdvegas,

Confirming that 2.35.0 introduced the issue, 2.34.0 is working as expected! I will look at what the PR changed.

Cheers,

KelSolaar:
So quickly comparing the generated *rez-shell.bat* and the command call:

### Rez 2.34.0

*rez-shell.bat*

```shell
set REZ_ENV_PROMPT=%REZ_ENV_PROMPT%$G
call c:\users\thomas\appdata\local\temp\rez_context_azh6ou\context.bat
set REZ_STORED_PROMPT=$P$G
set PROMPT=%REZ_ENV_PROMPT% $P$G
"Substance Painter"
exit %errorlevel%
```

*Command Call*

```shell
['c:\\windows\\system32\\cmd.exe', '/Q', '/C', 'call c:\\users\\thomas\\appdata\\local\\temp\\rez_context_azh6ou\\rez-shell.bat']
```

### Rez 2.35.0

*rez-shell.bat*

```shell
set REZ_ENV_PROMPT=%REZ_ENV_PROMPT%$G
call c:\users\thomas\appdata\local\temp\rez_context_mjdt6d\context.bat
set REZ_STORED_PROMPT=$P$G
set PROMPT=%REZ_ENV_PROMPT% $P$G
```

*Command Call*

```shell
[...]

PR #698, opened 2019-08-26
Enhancements for shell plugins

bfloch:
This fixes and improves the shell plugins, especially on Windows for cmd and PowerShell-like.
Formally excluded shell dependent tests are now passing.

I would love to have someone run the test-suite on MacOS.

### Future work
- Better Visual Studio handling in CMake for Windows
- Remove dependency of pwsh on SH-class for `get_syspaths` under Linux

### Fixes
- #694 Expand both generic environment variables `$ENVVAR / ${ENVVAR}` as well as shell specific ones correctly, like `%ENVVAR% ${Env:ENVVAR}` for cmd and PowerShell-like shells
- #691 Undoes regression in quotes handling of cmd

### Features / Enhancements
- Added `get_all_key_tokens` to allow for multiple key-forms
- Added `convert_tokens` to normalize token forms to shell specific form
- Allow pwsh shell on Linux (and pass tests)
- Implements optional script execution on windows via PATHEXT.
  - Windows shells may add custom PATHEXT values in plugin config file
  - `create_executable_script` now returns list of created scripts based on `config.create_executable_script_mode`:
     - `single` (default) Just create the requested script. Usually extension-less.
     - `py` create .py script that will allow launching scripts on windows,
     - `platform_specific` will create .py script on windows and requested on other platforms
    - `both` creates the requested file and a .py script for windows

- Introduced Shell specific regex parsing to handle ambiguous case like `set PATH=$ENV:VAR` in PowerShell vs. Bash within the NamespaceFormatter
- Introduced printout of shell name for `shell_dependent` test failure

### Tests passing
- [x] Windows 10.0.17134
  - [x] cmd*
  - [x] PowerShell 5.1.17134.165*
  - [x] pwsh 6.2.2*
  - *fails in `test_build_cmake` like in master. Will be addressed next.
  - [x] tcsh
  - [x] sh
  - [x] bash
  - [x] csh
  - [x] pwsh 6.2.1
  - [x] zsh ~~fails like on master~~ Passes based on #711

### Backwards compatibility
- Fully backwards compatible, new features use config options default to current behaviour
- Shell Plugin API has been extended slightly.
  - Implement `get_all_key_tokens(key)` (with safest/common form at index 0)
  - Implement `line_terminator()` for shell/platform specific line_terminator as expected by the shell
  - `join` now a class method
  - Extend `ENV_VAR_REGEX` if needed with shell specific token forms
  - You can make use of new `convert_tokens` in you implementation of `info` and `error`

### DIFF
--- a/src/rez/tests/util.py
+++ b/src/rez/tests/util.py
@@ -178,10 +178,12 @@ def shell_dependent(exclude=None):
                 try:
                     func(self, *args, **kwargs)
                 except AssertionError as e:
-                    # Add the shell to the exception message
-                    args = list(e.args)
-                    args[0] += " (in shell '{}')".format(shell)
-                    e.args = tuple(args)
+                    # Add the shell to the exception message, if possible.
+                    # In some IDEs the args do not exist at all.
+                    if e.args:
+                        args = list(e.args)
+                        args[0] += " (in shell '{}')".format(shell)
+                        e.args = tuple(args)
                     raise
         return wrapper
     return decorator


src/rez/rex.py
```diff
@@ -891,7 +896,11 @@ def escape_envvar(matchobj):
             value = (x for x in matchobj.groups() if x is not None).next()
             return "${{%s}}" % value
 
-        format_string_ = re.sub(self.ENV_VAR_REGEX, escape_envvar, format_string)
+        regex = ActionInterpreter.ENV_VAR_REGEX
+        if "regex" in kwargs:
+            regex = kwargs["regex"]
```
nerdvegas: `regex = kwargs.get("regex") or ActionInterpreter.ENV_VAR_REGEX`

src/rez/rezconfig.py
```diff
@@ -619,6 +619,29 @@
 suite_alias_prefix_char = "+"
 
 
+###############################################################################
+# Utils
```
nerdvegas: I would just add this to the `Misc` section, it's not that clear what "utils" really means in the context of rez settings.

src/rez/shells.py
```diff
@@ -160,6 +160,29 @@ def spawn_shell(self, context_file, tmpdir, rcfile=None, norc=False,
         """
         raise NotImplementedError
 
+    @classmethod
+    def get_key_token(cls, key, form=0):
```
nerdvegas: I find the use of `get_key_token`/`key_form_count` a bit confusing. How about we just update `get_key_token` to return a list, where the first item is the preferred form? It's not used in many places so not much code would need updating, and I'd be pretty surprised if anyone considered this part of the public API (so I don't expect backwards compatibility to be an issue here).

src/rez/tests/test_shells.py
```diff
@@ -69,27 +63,69 @@ def test_no_output(self):
...
-        # possible to run this test on Windows.  The 'hello_world' executable
-        # is not registered correctly on Windows so always returned the
-        # incorrect error code.
         sh = create_shell()
         _, _, _, command = sh.startup_capabilities(command=True)
 
         if command:
             r = self._create_context(["hello_world"])
-            p = r.execute_shell(command="hello_world",
+            script = "hello_world"
```
nerdvegas: `script` as a var here is a bit odd?

src/rez/tests/test_shells.py
```diff
@@ -145,13 +181,19 @@ def test_rcfile(self):
...
         # exactly what we expect.
+        sh = create_shell()
         echo_cmd = which("echo")
-        if not echo_cmd:
+
+        # Certain shells will not find echo.
+        # TODO: If this exception was created for the following shells then
+        #  it is redundant. Can we clarify which platforms this was meant for?
+        has_buildin_echo = sh.name() in ['cmd', 'powershell', 'pwsh']
+        if not echo_cmd and not has_buildin_echo:
```
nerdvegas: Could we keep this simple and just use whatever the `echo` equivalent is when on Windows? Are there any non-windows shells that wouldn't have `echo`?

src/rez/tests/test_shells.py
```diff
@@ -189,21 +231,29 @@ def _execute_code(func, expected_output):
 
             out, _ = p.communicate()
             self.assertEqual(p.returncode, 0)
-            token = '\r\n' if platform_.name == 'windows' else '\n'
+
+            # Powershell and Unix uses \n
```
nerdvegas: Perhaps it would be better to add a `line_terminator` property to Shell class, rather than hardcoding this behaviour here. I had probably assumed at the time that terminator was platform-dependent and not shell-dependent.

src/rez/tests/test_shells.py
```diff
@@ -189,21 +231,29 @@ def _execute_code(func, expected_output):
...
                 env.FOO = value
                 # Wrap the output in quotes to prevent the shell from
                 # interpreting parts of our output as commands. This can happen
                 # when we include special characters (&, <, >, ^) in a
                 # variable.
-                info('"%FOO%"' if windows else '"${FOO}"')
+                if is_powershell:
+                    info('`"{}`"'.format(sh.get_key_token("FOO")))
+                else:
+                    info('"{}"'.format(sh.get_key_token("FOO")))
```
nerdvegas: So the original code here should really just have been:

    info('"${FOO}"')

The test is showing two things - that ${X}/$X variable references are getting correctly translated to their shell equivalent; and that use of expandable/literal is working as expected. Rather than directly use `get_key_token` anywhere here, the idea was just to test the actual output from the shell.

Wrt the whole adding of double quotes here to avoid problems with '>' etc, it does seem like that should be dealt with by `info` itself (and by shell-specific string expansion in general). But that stuff gets tricky and a change related to that should be in a separate PR. Is Powershell a special case where this has to be done differently?

src/rez/tests/test_shells.py
```diff
@@ -225,19 +275,32 @@ def _print(value):
...
-            _print(literal("${WHO}").e(" %WHO%" if windows else " $WHO"))
+            _print(literal("${WHO}").e(" $WHO"))
 
             # Make sure we are escaping &, <, >, ^ properly.
             _print('hey & world')
             _print('hey > world')
             _print('hey < world')
             _print('hey ^ world')
 
+            # Platform dependent form of variables.
```
nerdvegas: I would just include unix shells here also, rather than have an exception at all. It should work also and it doesn't really matter if it re-tests some of the same things.

src/rez/tests/test_shells.py
```diff
@@ -294,6 +372,25 @@ def _rex_appending():
 
         _execute_code(_rex_appending, expected_output)
 
+    @shell_dependent()
+    def test_variable_encoding(self):
```
nerdvegas: I think we could do without this test actually - these tests are really only asserting internal implementation details of the shells. If you go down this road (eg, testing that powershell class acts like powershell) you'd have to write loads more tests, and I don't really see the value in that. I would argue that that isn't what this test suite is for - it's for checking that rex code (which is shell agnostic by design) gives the expected result in all shells - no more.

src/rez/util.py
```diff
@@ -10,11 +10,35 @@
...
 
 
+class ExecutableScriptMode(Enum):
+    """
+    Which scripts to create with util.create_executable_script.
+    """
+    # Start with 1 to not collide with None checks
+
+    # Requested shebang script only. Usually extension-less.
+    requested = 1
```
nerdvegas: I don't really understand what is meant by "requested" here? And it isn't clear what "requested" would mean when on Windows... isn't it the same as "platform_specific"?

Would "native" be a better term? I'm trying to understand what "requested" means as in the term by the way, not the implementation!

src/rez/util.py
```diff
@@ -46,18 +82,72 @@ def create_executable_script(filepath, body, program=None):
...
+    Evaluates the py_script_mode for the requested filepath on the given
+    platform.
+
+    Args:
+        filepath: requested filepath
+        py_script_mode (ExecutableScriptMode):
+        platform (str): Platform to evaluate the script files for
+
+    Returns:
+
```
nerdvegas: Missing docstring

src/rezplugins/shell/cmd.py
```diff
@@ -7,7 +7,7 @@
 from rez.system import system
 from rez.utils.system import popen
 from rez.utils.platform_ import platform_
-from rez.util import shlex_join
+from rezplugins.shell.powershell import list2cmdline
```
nerdvegas: This is nasty! If there's common functionality here then it should be moved into util code somewhere. There certainly should not be cross referencing of shell implementations like this

src/rezplugins/shell/powershell.py
```diff
@@ -1,39 +1,54 @@
...
     def executable(cls):
-        if cls._executable is None:
-            cls._executable = Shell.find_executable('powershell')
-        return cls._executable
+        raise NotImplementedError
 
     @classmethod
     def name(cls):
-        return 'powershell'
+        raise NotImplementedError
```
nerdvegas: just drop this method, no need to implement here

src/rezplugins/shell/powershell.py
```diff
@@ -1,39 +1,54 @@
...
 
     @classmethod
     def file_extension(cls):
-        return 'ps1'
+        raise NotImplementedError
```
nerdvegas: just drop this method, no need to implement here

src/rezplugins/shell/pwsh.py
```diff
@@ -1,25 +1,12 @@
...
-from rez.rex import RexExecutor, OutputStyle, EscapedString
 from rez.shells import Shell
-from rez.utils.system import popen
 from rez.utils.platform_ import platform_
-from rez.backport.shutilwhich import which
-from functools import partial
-from subprocess import PIPE
-import os
-import re
+from rezplugins.shell.powershell import PowerShellBase
```
nerdvegas: Yeah I see that we need to do this here

src/rezplugins/shell/pwsh.py
```diff
@@ -1,25 +1,12 @@
...
+from rezplugins.shell.sh import SH
```
nerdvegas: ..but this should be changed

src/rez/bind/hello_world.py
```diff
@@ -43,7 +43,15 @@ def bind(path, version_range=None, opts=None, parser=None):
     def make_root(variant, root):
         binpath = make_dirs(root, "bin")
         filepath = os.path.join(binpath, "hello_world")
-        create_executable_script(filepath, hello_world_source)
+
+        # In order for this script to run on each platform we create the
```
nerdvegas: While this comment is correct, I don't think it belongs here. This caveat is already pointed out in a very similar comment with the function definition.

src/rez/shells.py
nerdvegas: To add to this - the func should be allowed to return str or list of str, so that any existing external shell plugins won't break. So in that respect we need backwards compatibility, but I think that will be enough.

bfloch:
Thanks for the review. I'll dropped notes/questions and will look into the fixes.

src/rez/shells.py
bfloch: Did it as I proposed. This is way cleaner, the API stays compatible and the naming is way clearer.
Hope you agree.

src/rez/tests/test_shells.py
bfloch: I've added a `line_terminator` to the shell API.

src/rez/tests/test_shells.py
bfloch: This now passes and I was able to do the right thing by introducing `convert_tokens` to Shell.
This was actually simpler that both of us thought 👍

src/rez/tests/test_shells.py
bfloch: You are right.

src/rez/util.py
bfloch: I renamed it to "single". Hope that makes more sense.
I left `platform_specific`, but it you prefer `native` let me know.

src/rezplugins/shell/pwsh.py
bfloch: I've added a powershell_common module.

src/rezplugins/shell/pwsh.py
bfloch: Hope that comment makes it clearer why we use SH here.

bfloch:
I've updated the first post and addressed everything in your review.
One thing I could not solve is the SH reference in pwsh, but I limited the scope and added a proper comment with a reference how this might be solved upstream of pwsh.

Ran the tests again, pass as before.

src/rez/tests/test_shells.py
bfloch: Removed the platform checks. Way cleaner, thanks.

maxnbk:
FYI, zsh tests on master should be fixed when #711 is merged in.
Notifying in case it matters for this PR.

src/rez/shells.py
```diff
@@ -160,7 +176,46 @@ def spawn_shell(self, context_file, tmpdir, rcfile=None, norc=False,
...
+        Args:
+            key: Variable name to encode
+
+        Returns:
+            str of encoded token form
+        """
+        return cls.get_all_key_tokens(key)[0]
+
+    @classmethod
+    def get_all_key_tokens(cls, key):
```
nerdvegas: :+1:

nerdvegas:
I'm ready to merge this, but would like to hear RE OSX testing also. In the meantime I'm going to start looking at GitHub Actions, so we'll have this testing covered anyway (and across py2/3 also, which is going to become important pretty soon).

src/rezplugins/shell/cmd.py
```diff
@@ -182,12 +184,34 @@ def _create_ex():
             executor.interpreter._saferefenv('REZ_ENV_PROMPT')
             executor.env.REZ_ENV_PROMPT = literal(newprompt)
 
+        # Make .py launch within cmd without extension.
+        if self.settings.additional_pathext:
+            executor.command('set PATHEXT=%PATHEXT%;{}'.format(
```
nerdvegas: It's a fairly minor point, but in nested rez-envs this will cause PATHEXT to become `>PY;.PY;.PY...`. If it's easy, it may be worth updating this command to only add the additional pathext if it isn't already in PATHEXT. If it's a pain I wouldn't worry too much.

nerdvegas: Actually, note also that wrt #707 (see my comments there), that with sufficient env-var control added to the project, this special case may not be necessary at all. I'm imagining something like:

```
env_var_settings = {
  "PATHEXT": {
    "set_always": ".PY",  # always set PATHEXT=.PY in a new shell...
    "path_modify_mode": "retain_after"  # ..but also keep pre-rez value of PATHEXT appended as well
  }
}
```

bfloch: I personally can wait to have #707 done first. Everyone who has a need for better shell support may feel free to merge this branch for testing.
Up to you - I will fix the issue above as suggested, just in case.

bfloch:
> FYI, zsh tests on master should be fixed when #711 is merged in.
> Notifying in case it matters for this PR.

Cool will try merging this in. Thanks!

bfloch:
Will fix the conflict and I'll give the de-duplication on cmd another shot today.
Might have to do with shell=False again, not sure. The syntax certainly works within cmd sessions ...

src/rezplugins/shell/cmd.py
bfloch: For now I've implemented de-duplication as is for cmd and PowerShell.
I will move the discussion about the generic variable handling to the #707

bfloch:
> FYI, zsh tests on master should be fixed when #711 is merged in.
> Notifying in case it matters for this PR.

Notable changes since your last review:
- Fixed promp not expanded with cmd (probably because variable expansion now works properly)
- Supressing PowerShell header
- Added PATHEXT deduplication for cmd and PowerShell as suggested
- Rebased on current master. Fixed the zsh failure as advertised. Thanks.

So now it is just a question if we want #707 first.
I guess since it technically does not rely on it and we do have quite a lot of changes here we should probably go ahead and merge it. At least we get some real feedback early on, while there still might be some design discussions happening in #707.

All tests (except cmake related) pass on Win & Linux.
MacOS testing is dependent on the github actions setup, I presume. I'll leave that part to you guys since I don't have access to a Mac.

nerdvegas:
Thanks Blazej. I'm happy to merge this and will do soon - it stands on its own from #707 and fixes an outstanding bug on Windows.

nerdvegas:
Just an update - getting to this soon. We're near the end of the py3 work, I'm hoping to get that done first then merge this right after.

bfloch:
No worries, there is no rush with this merge. I would love to hear your feedback on #737 though since this is the last missing piece for our Rez deployment.

Btw. I haven't tested this PR with python 3, so let me know if there is any work to be done.
In fact, we as a Blender-based animation studio would love to run rez with python 3 so I might be running tests based on your effort soon.

nerdvegas:
Yeah I'm pretty slammed atm but I'm keen to get onto both #737 and #703 so
I'll do that soon.

We're maybe a few days/a week away from wrapping up on py3 port, so I
reckon if you could do some testing on it once that's done, that would be
very helpful.

Cheers
A

PR #705, opened 2019-08-30
fixed bez rezbuild.py breaking on old-style print

### LINKED ISSUE #704: bez rezbuild.py breaks on old-style print
https://github.com/nerdvegas/rez/blob/master/src/rez/cli/_bez.py#L41

This line forces new-style print function in rezbuild.py files. This has the effect of breaking some existing build files, and is a form of non-backwards compatibility.

nerdvegas:
print_function made rezbuild.py with old-style print broken
Closes #704


maxnbk:
I tested this commit change independently on my current working directory for rez, and it does solve the issue of rezbuild.py failures with print statements.

nerdvegas:
this is a test

PR #714, opened 2019-09-04
yaml upgrade

maxnbk:
we need to upgrade yaml to have py2 and py3 versions available. `__init__.py` in the yaml vendor folder selects between them, and could be collapsed when py2 support is dropped.
loaders specifications are added wherever loaders are used to avoid a deprecation warning that newer yamls provide when not specifying a loader.


src/rez/vendor/README.md
```diff
@@ -31,8 +31,8 @@ Note that the latest versions column is just to give us an idea of how far back
...
 | version                     | ?                          |                                         | This is actually part of rez. It's here because I've been intending on releasing it separately for ages.                                                                                                                                                                                                                  |
-| yaml (PyYAML)               | 3.10 (May 30, 2011)        | 5.1 (Mar 13, 2019)                      | Not much changes, mostly broken releases that never made it. Between 3.10 and 5.1 there is only 3 versions that made it to the public (3.11 and 3.12, 3.13), the rest were all removed or marked as pre-release. It's safe to update. |
-
+| yaml lib (PyYAML)           | 5.1 (May 30, 2011)         | 5.1.2  (Jul 5, 2018)                    | No changes but must maintain separate version between py2 and py3 for time being                                                                                                                                                      |
```
JeanChristopheMorinPerso: Thanks for updating the doc :)

PR #715, opened 2019-09-04
improve non-string iterable handling

maxnbk:


src/rez/util.py
```diff
@@ -175,6 +177,14 @@ def _atexit():
         pass
 
 
+def iterable(arg):
```
nerdvegas: should be `is_non_string_iterable`

PR #716, opened 2019-09-04
replace async with block to avoid py3 async keyword

maxnbk:
async is a py3 keyword, so we must use something else instead. block seems appropriate.


nerdvegas:
default vaue / if cond needs to be inverted

PR #718, opened 2019-09-04
swap 2.6 support for 3.x in version module

maxnbk:
py 2.6 support was being maintained through a backport of total_ordering, which firstly, now has an implementation maintained through future, but secondarily, appears not to be needed in general with some small implementation changes, along with a couple of print fixes, making the version module py3 compatible.


nerdvegas:
Just a note. I noticed this in the py docs on `total_ordering`:

```
While this decorator makes it easy to create well behaved totally ordered types, it does come at the cost of slower execution and more complex stack traces for the derived comparison methods. If performance benchmarking indicates this is a bottleneck for a given application, implementing all six rich comparison methods instead is likely to provide an easy speed boost.
```

We should leave as-is now, but given that the version module makes up the majority of where solve time goes, I think we should do as suggested and manually implement all operators in all classes, possibly doing away with `_comparable` class.

I've made a separate ticket: https://github.com/nerdvegas/rez/issues/719

PR #732, opened 2019-09-06
swap between getargspec and getfullargspec due to getargspec deprecation

maxnbk:
this is arguably the most dangerous change we need to make for py3 compatibility, and I'm not altogether sure what the consequences are, we could use some very careful review of this set of changes.


maxnbk:
closed in favor of #743

PR #734, opened 2019-09-06
refer to current python executable directly

maxnbk:


src/rez/tests/test_context.py
```diff
@@ -71,7 +72,7 @@ def test_execute_command_environ(self):
                   "print(os.getenv(\"BIGLY\")); "
                   "print(os.getenv(\"OH_HAI_WORLD\"))")
 
-        args = ["python", "-c", pycode]
+        args = [sys.executable, "-c", pycode]
```
nerdvegas: I don't think this is correct - we want to run "python" within the context, which almost certainly wouldn't match `sys.executable`, which is rez's own python interpreter installed into its virtualenv.

Issue #744, opened 2019-09-10
#712 merged in 2.43.0 caused external environ not to pass through to resolve


maxnbk:
testable between `2.42.2` and `2.43.0` by `rez-env`-ing a package with:
```
def commands():
    print(env.keys())
```
result in `2.42.2` is a list of variable names, 2.43.0 is `[]`

solved in incoming PR.

maxnbk:
addressed in #745

nerdvegas:
Closed by https://github.com/nerdvegas/rez/pull/745

PR #745, opened 2019-09-10
address #744 (rex dictmixin issue)

maxnbk:


bfloch:
Not a big issue but you should be able to skip the condition via:
```python
from six.moves import UserDict
```

nerdvegas:
Added.

PR #747, opened 2019-09-14
Issue 696 shell availability

nerdvegas:
Relevant: https://github.com/nerdvegas/rez/issues/696

### DIFF
--- a/src/rez/shells.py
+++ b/src/rez/shells.py
@@ -22,14 +22,37 @@ basestring = six.string_types[0]
 
 
 def get_shell_types():
-    """Returns the available shell types: bash, tcsh etc."""
+    """Returns the available shell types: bash, tcsh etc.
+
+    Returns:
+    List of str: Shells.
+    """
     from rez.plugin_managers import plugin_manager
     return list(plugin_manager.get_plugins('shell'))
 
 
+def get_shell_class(shell=None):
+    """Get the plugin class associated with the given or current shell.
+
+    Returns:
+        class: Plugin class for shell.
+    """
+    if not shell:
+        shell = config.default_shell
+        if not shell:
+            from rez.system import system
+            shell = system.shell
+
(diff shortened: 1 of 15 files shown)


maxnbk:
is the intention for the tests to pass or to show as skipped in the resulting selftest?
(because as implemented, they are passing rather than showing as skipped in the report, which is what I would have expected.)
If the intention is for them to pass, then consider this approved by me, if the opposite, there is more to do.
Tested on Cent.

nerdvegas:
You can't actually skip the test (ala unittest.skip/TestCase.skipTest). This is because the per_available_shell decorator just loops over the wrapped function - if you skip, the remaining runs in the loop would be skipped too. This may be solvable with some fancy meta programming, but IMO isn't worth the effort.

Now that you mention test skipping though, I see a couple of other things that could be done better. I'll update this PR shortly with those additional fixes.

JeanChristopheMorinPerso:
Will test this in a couple of hours on Arch Linux, so an ultra up-to-date system with Python 2 and 3.

And as I already said, we would highly benefit to start to use pytest. This would allow us to have more granular tests (parametrized) which would allow us to skip based on complex conditions. But that's probably gonna be later. Code wise for now it looks good.

nerdvegas:
I'm familiar with pytest and I agree it'd be great to update the tests to use it. One note on that before I forget - IIRC vendoring pytest is a little problematic as it brings in a bunch of dependencies. But I actually think pytest may be a case that doesn't have to be vendorized. The primary reason for vendoring is to make rezification of rez itself trivial. However, since the rez API itself does not require tests to be present, then pytest isn't required to be vendored. So, I think pytest could be included via standard requirements.txt so it gets installed into the rez virtualenv. It would actually make sense to include this via a "test" extra also.

If anyone wants to talk further about this then let's take it to slack/ticket rather that polluting this thread further :)

PR #749, opened 2019-09-20
[wiki] Changes to pip install command

j0yu:
`pip install rez` installs the ancient `2.0.ALPHA.52` [on pypi.org](https://pypi.org/project/rez/#history), see https://github.com/nerdvegas/rez/pull/579#issuecomment-526963146

Changed to `pip install` using source directory, see https://github.com/nerdvegas/rez/pull/579#issuecomment-527026855

Also added `rez-pip --install` suggestion from https://github.com/nerdvegas/rez/pull/579#issuecomment-526801383


nerdvegas:
Hey thanks, just getting to this now.

I like the idea of an extra workflow for testing all installation types (via install.py, pip with various args). But a couple of things:

* The name. What does wiki-doctest-installation mean? I would be happy to simply call this workflow 'Installation'
* The workflow doesn't need to be docker-based, I would drop that.
* I would also drop any use of or references to rez-bind, I want to deprecate this. I don't think we need to do the `rez-bind pip` in the workflow - we can just let rez-pip pickup the system pip install.
* I also think these tests should verify the installation, rather than just run `find`. Eg, after a `pip install .`, I would run python and test that rez imports successfully. After `rez-pip install .`, I would verify that `rez-env rez -- python -c 'import rez'` works.

Looking forward to getting this merged, it's been mentioned in the past that installation testing is missing.

j0yu:
@nerdvegas I've updated the code now with the changes and also the PR description to reflect notable changes:

- [x] Rebased and compacted commits on top of 2.50.0
- [x] Renamed to Installation
   > The name. What does wiki-doctest-installation mean? I would be happy to simply call this workflow 'Installation'
- [x] Changed to not be based off Docker Python image
   > The workflow doesn't need to be docker-based, I would drop that.
- [x] CI verifies using `rez-env rez -- python -c 'import rez'`
   > I also think these tests should verify the installation, rather than just run `find`. Eg, after a `pip install .`, I would run python and test that rez imports successfully. After `rez-pip install .`, I would verify that `rez-env rez -- python -c 'import rez'` works.
- [ ] Got rid of `rez-bind pip` but unfortunately for the tests, it looks like I still require `rez bind python` to setup the `python` rez package needed for the `rez` package. It seems [to be part of it's variants](https://github.com/wwfxuk/rez/commit/ecbbc5dc6402e44f95d7bdb0637fb028807c7504/checks?check_suite_id=367269447#step:6:84)
  > I would also drop any use of or references to rez-bind, I want to deprecate this. I don't think we need to do the `rez-bind pip` in the workflow - we can just let rez-pip pickup the system pip install.

j0yu:
Looks like I missed out a few things during refactor/rebase.

Will need to update again from https://github.com/wwfxuk/rez/tree/pip-fallback

j0yu:
Should be good to review again @nerdvegas , I've moved the CI into #878 as it has fix for pip in there

wiki/pages/Installation.md
```diff
@@ -26,15 +26,25 @@ source /opt/rez/completion/complete.sh
...
+- Add `/opt/rez` to `PYTHONPATH`
+
+Alternatively, if you already have `rez` setup and have `pip>=19` available, 
+you can then install `rez` as a `rez` package by using:
+
+    ]$ rez-pip --install .
+
+However, these methods comes with a caveat - rez command line tools _are not 
+guaranteed to work correctly_ once inside a rez environment (ie after using the 
+`rez-env` command). The reasons are given in the next section.
```
JeanChristopheMorinPerso, line 47: Really minor: I would add that a warning is outputted to the terminal when the rez commands are used in this context.

j0yu: It's mentioned a few lines below it but I'll tidy it up into 1 warning message

nerdvegas:
Lets get onto this once https://github.com/nerdvegas/rez/pull/878 is sorted.

JeanChristopheMorinPerso:
Umm ok, it seems like you pushed some commits. I'll have to find how to remove my approval...

wiki/pages/Installation.md
```diff
@@ -26,30 +26,50 @@ source /opt/rez/completion/complete.sh
...
 
-```
-Pip-based rez installation detected. Please be aware that rez command line tools
-are not guaranteed to function correctly in this case. See
-https://github.com/nerdvegas/rez/wiki/Installation#why-not-pip-for-production
-for futher details.
-```
+The environment variables to activate `rez` will be slightly different:
 
+- Add `/opt/rez/bin` to `PATH`
```
JeanChristopheMorinPerso, line 47: I don't think we should invite users to add this to their path as it could make them think they can use the rez cli. And here, it only points to the bin folder, which only contains the ``python` binaries...

j0yu: I'll strongly emphasis this and put the warning right after it

Issue #737, opened 2019-09-06
Better Environment Variable control in Rex


bfloch:
(Originated from #707)

## Settings and Ops

This is a rather complex topic since we want to support so many use cases so I'll build up to the suggestion based on every possible use case imaginable. It also gives us opportunity to migrate existing settings in a generic configuration.

I don't want to jump into implementation details yet (hope this is doable at all :) ), but let's see if we
can come up with use cases best demonstrated via an sample.

### References

#703 
#698 
#661 
(more ?)

### Use cases

(cases handled by the sample are crossed)

- Path operations:
  - [x] Handle paths in platform independent fashion
  - [x] Handle paths in explicit platform dependant fashion
  - [ ] Handle differences in Drive handling (C:\ -> C:/ vs /C/) (Needed?)
  - [ ] Handle platform dependent path prefix (C:\path -> /mnt/c/path) (Needed?)
- [x] Define platform dependent settings
- [x] Migrate `all_*_variables`
- [x] Define path separators (like ; and :)
- [ ] Stable de-duplication of separated values (Needed?)
- [x] Implicit variables, system variables and there placement with context variables

### Sample

(Note: This is a completely artificial example to showcase many possible use cases)

```python
env_var_settings = {

    # -------------------------------------------------------------------------
    # Global control inherited(!) by all other variables.

    "*": {

        # Migration of all_parent_variables
        #
        # Full control over variables is achieved given the following parent_variables & actions:
        #
        # clear (No parent variables)
        #  [prepend_all] [prepend_parent->FAILS] [append_parent->FAILS] [$REZ_VARS] [append_all]
        #
        # prepend (parent before rez variables)
        #  [prepend_all] [prepend_parent] [$PARENT_VARS] [append_parent] [$REZ_VARS] [append_all]
        #
        # append (parent after rez variables)
        #  [prepend_all] [$REZ_VARS] [prepend_parent] [$PARENT_VARS] [append_parent] [append_all]
        #
        "parent_variables": "clear"

        # Migration of all_resetting_variables
        # Note: Actually had a problem with understanding the naming, but kept for consistency
        "resetting_variables": False

        # Any variable (!!!) on this level may be "its expected type" OR dict of 
        # ("platform": "its expected type").
        # Note we are only handling platform.system differences.
        "pathsep": {
            "windows": ";",
            "linux":   ":",
            # Any other platform, except specific one …
            "*":       "🍎",
        },
        # Default append, prepend and set operation
[...]

Issue #703, opened 2019-08-29
CMake plugin overhaul


bfloch:
In order to support cmake on windows (and to improve the plugin in general) I am proposing the following changes to the cmake plugin:

1. Allow `build_system` and `child_build_system` to be `None` (new default)
   - In this case, instead of hard-coded logic, use `cmake` without specific generator and `cmake --build build_path --target {ALL}` where `ALL` is `all` in unix and `ALL_BUILD` on windows.
     - The `all` and `ALL_BUILD` distinction is unfortunate, but can be verified by listing all targets
   - This allows cmake to choose the best default for any platform (Visual Studio on Windows) and build without having to deal with Visual Studio and MSBuild discovery manually - which is really painful on windows. See the note below.

2. Deprecate the hardcoded generators
   - We should support all generators that the underlying platform & cmake version supports
   - For backwards compatibility I will keep the current dictionary, but I will alternate it with a dynamic list coming from `cmake --help` that will always match _a_ cmake-executable.
   - The only gotcha is that we don't have the context at class scope, so technically a different cmake version might run in context then what the arguments would show. I can deal with it by using the cmake bin path as key for the cached generators, but in this 10% case the bound arguments would not match the underlying generators. I will need to lift the validation in the command line parameter and let cmake deal with it. However platform and cmake version should be the only necessary inputs. Generators are compiled into cmake and can not be changed/extended otherwise.
      - If we really wanted to have accurate arguments we'd need to be creating a context for this part, but since the current method is not accurate either (hard-coded and may be wrong), I guess we could live with it. Also if you have an explicit cmake version it will match perfectly.

This would allow building on Windows (or other platforms?) with the default settings but still keep backwards compatibility.

I am already working towards a PR on this. Feedback appreciated.
Other necessary changes for better path handling will be in a separate issue.

**Related note for the curious**
Windows is very _special_.

[...]

PR #750, opened 2019-09-27
Gh actions - first pass

nerdvegas:
Notes:
* Windows is not passing, help appreciated! (Cmake related?)
* Badges don't work, I don't know why.

A couple of test-related things were changed:
* There was a bug causing only the first shell to be tested with funcs decorated in `per_available_shell`.
* Verbosity dropped on solver tests, was making it pretty hard to go through the Actions output.
* The "logging" test was removed. Failed logging won't actually raise an exception, see https://docs.python.org/2/howto/logging.html#exceptions-raised-during-logging. The existing test was actually failing and putting a lot of noise into the test output. Given the simplicity of the test, and the fact that exceptions won't get raised, I thought it best just to remove it.
* We were getting a bug where "decode" was called on the stdout arg from a subprocess.Popen().communicate() call. This is actually an issue with the powershell plugin. It was using universal_newlines=True, which changes the return type; see https://docs.python.org/3.4/library/subprocess.html#frequently-used-arguments. I am not clear on why this was added so I'm unsure of side effects. However it does seem as though if this is necessary, it could be passed through in `**Popen_args` instead anyway.

One thing we should definitely do is port to pytest and do away with the per_available_shell silliness. I'd be interested in adding a `--test` option to `install.py`, which would then install the rez-selftest tool, and pytest (as a native requirement in the installation venv, not a vendored lib). I think it'd be better to not install rez-selftest if `--test` wasn't specified, since it would require pytest in order to work.

Another thing we should do is update the rez-selftest `-s` arg to take multiple shells, and for each of those, to assert that the shell is available, rather than skip over it if not. Right now it isn't that clear whether a given shell is actually getting tested in the CI; for eg, rez could think the shell is unavailable (which in itself may be a bug). So the way I'd see it working is, we would do the apt-get install etc of the shells we want to test, then pass those shells to rez-selftest also, which would fail if any of those shells are not available / rez doesn't know about them.


KelSolaar:
> Badges don't work, I don't know why.

They will once you merge the changes into `master` :)

KelSolaar:
Btw are you planning to squash the history?

nerdvegas:
Bam! Of course I need a merge to master first. I'm not sure if there's even
an option for branch-specific badges yet either, there's no mention in the
docs.

History - you mean the run history..? Not even sure if that's possible yet.

KelSolaar:
> History - you mean the run history..? Not even sure if that's possible yet.

Nah, the commit one, it scorched my retina! :)

image

nerdvegas:
Ah right gotcha. Yeah I'll make sure I do that before merge. Had to keep
pushing in order to test.
A

KelSolaar:
Sure absolutely make sense, something I noticed before but they fixed during the last few days is that file parsing errors were creating junk on the left actions sidebar. It is gone now, for the better :)

nerdvegas:
Yeah I've noticed other stuff too. In the Actions tab, the available runs
on the left-hand side appear and disappear at odd times, and sometimes
you're not in the run you'd expect. Definitely still beta. But I bet it's
gonna get so much better. I'm a bit surprised the workflow language is so
limited though, I've already hit a few cases where something like Jinja2
would be really useful.

KelSolaar:
> I'm a bit surprised the workflow language is so
limited though

It is almost exactly the Azure Pipelines stuff, and yeah there are some very frustrating bits! It will get better though, can see things changing on a per-day basis almost!

JeanChristopheMorinPerso:
I'm wondering, should we test on CentOS instead of Ubuntu as CentOS is probably way more used than Ubuntu in VFX?

JeanChristopheMorinPerso:
Well screw my last comment, they don't provide anything other than Ubuntu :( .

bfloch:
Great work!

Windows failing is expected.
It is being addressed in #737 and #703. Waiting for feedback.

We could skip the cmake test on windows for now - I don't believe rez has a strict dependency on cmake. Everything else should pass through.

bfloch:
> windows.yaml
> `# TODO install shells (which ones need installing?)`

PowerShell (5.x) and cmd should be build in Windows. We just need pwsh (PowerShellCore 6.x):
https://github.com/PowerShell/PowerShell/releases

bfloch:
Worth checking if pwsh is not installed by default on all platforms:
https://help.github.com/en/articles/workflow-syntax-for-github-actions#using-a-specific-shell

If not something like might do the trick (just tested locally):
```
steps:
  - name: Install pwsh
    shell: PowerShell
    run: |
      [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
      Invoke-WebRequest http://github.com/PowerShell/PowerShell/releases/download/v6.2.3/PowerShell-6.2.3-win-x64.msi -OutFile PowerShell-6.2.3-win-x64.msi
      Start-Process -Wait -FilePath msiexec -ArgumentList /package, PowerShell-6.2.3-win-x64.msi, /quiet, ADD_EXPLORER_CONTEXT_MENU_OPENPOWERSHELL=1, ENABLE_PSREMOTING=1, REGISTER_MANIFEST=1
```

nerdvegas:
merged: https://github.com/nerdvegas/rez/releases/tag/2.47.3

PR #753, opened 2019-10-03
use subprocess in 'text' mode in most cases

nerdvegas:
This returns an str rather than bytes, which is typically what we want. The "universal_newlines" arg is equivalent, but misleading (hence the rename to 'text' in py3).

I haven't been able to repro, but I know that @maxnbk was hitting an issue related to this. I expect this PR to probably fix that.

### DIFF
--- a/src/rez/bind/_utils.py
+++ b/src/rez/bind/_utils.py
@@ -132,7 +132,8 @@ def _run_command(args):
         args,
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
-        shell=use_shell
+        shell=use_shell,
+        text=True
     )
 
     stdout, stderr = p.communicate()
--- a/src/rez/bind/_utils.py
+++ b/src/rez/bind/_utils.py
@@ -132,8 +132,7 @@ def _run_command(args):
         args,
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
-        shell=use_shell,
-        universal_newlines=True
+        shell=use_shell
     )
 
     stdout, stderr = p.communicate()
--- a/src/rez/package_py_utils.py
+++ b/src/rez/package_py_utils.py
@@ -169,7 +169,7 @@ def exec_command(attr, cmd):
     """
     import subprocess
 
-    p = popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    p = popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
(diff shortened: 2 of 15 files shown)


maxnbk:
Yes, this PR improves 1 of 2 most-likely related issues on py3.6/3.7.
However, around line 1425 in the rez-selftest log, (well, at least in my shell), I'm getting:
```
/rez/lib/python3.7/site-packages/rez/tests/test_shells.py:138: ResourceWarning: unclosed file <_io.BufferedReader name=5>
  p = r.execute_shell(command=cmd, stdout=subprocess.PIPE)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
/rez/lib/python3.7/site-packages/rez/tests/util.py:185: ResourceWarning: unclosed file <_io.BufferedReader name=6>
  func(self, *args, **kwargs)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
```
The error message is slightly different/improved in 3.7, as 3.7 provides the tracemalloc behavior. I've enabled that for the following improved warning:
```
/rez/lib/python3.7/site-packages/rez/tests/test_shells.py:140: ResourceWarning: unclosed file <_io.BufferedReader name=5>
  p = r.execute_shell(command=cmd, stdout=subprocess.PIPE)
Object allocated at (most recent call last):
  File "/sw/packages/external/python/3.7.4/platform-linux/arch-x86_64/os-CentOS-7/lib/python3.7/subprocess.py", lineno 759
    self.stdout = io.open(c2pread, 'rb', bufsize)
/rez/lib/python3.7/site-packages/rez/tests/util.py:185: ResourceWarning: unclosed file <_io.BufferedReader name=6>
  func(self, *args, **kwargs)
Object allocated at (most recent call last):
  File "/sw/packages/external/python/3.7.4/platform-linux/arch-x86_64/os-CentOS-7/lib/python3.7/subprocess.py", lineno 759
    self.stdout = io.open(c2pread, 'rb', bufsize)
```

nerdvegas:
Yeah we were already getting those actually, I noticed them last week.

JeanChristopheMorinPerso:
Yeah, I've seen this too maybe two weeks ago. I'm wondering if it's because we don't call `communicate` or `wait` in some places in the tests.

JeanChristopheMorinPerso:
But for real, this looks good to me (the changes). I've seen a couple of errors because of the returned value that was in bytes instead of str.

nerdvegas:
Was thinking the same. I'm out of town atm but should get these merged
Monday US.

PR #756, opened 2019-10-04
fix cores detection error on win due to bytes-like from subprocess

maxnbk:
Addresses:
`22:30:51 ERROR    Error detecting physical core count, defaulting to 1: cannot use a string pattern on a bytes-like object`


nerdvegas:
This is actually already fixed in #753 - see https://github.com/nerdvegas/rez/blob/popen_as_str/src/rez/utils/platform_.py#L540

PR #757, opened 2019-10-04
rez-pip: Assume pip provided by python package

### LINKED ISSUE #706: rez-pip should assume python provided pip
Newer python versions ship with pip, however rez-pip first searches for a rezified pip package, falling back to its own pip installation if one isn't found.

Instead, rez-pip should (in this order):
* Search for rezified python matching python version request;
* If found, test if pip is present;
* If pip is present, use it;
* If not present, search for rezified pip (this is for backwards compatibility);
* If rezified pip is found, use it;
* If not, fall back to rez's python installation.

The rez-pip `--pip-version` arg should also be marked as deprecated (though not removed just yet).

### LINKED ISSUE #764: rez-pip python 3 error
`from StringIO import StringIO`

I've got this fixed, but I can not push to remote.

JeanChristopheMorinPerso:
This PR closes #706 and closes #764 

This PR adds a new logic to find which `pip` will be used to install pip packages.

The logic is as follow:

1. Search for pip in the rezified `python` package specified with `--python-version`, or
the latest version if not specified;
2. If found, use it;
3. If not found, search for pip in the rezified `pip` package specified with `--pip-version`,
or latest version if not specified.
4. If rezified `pip` is found, use it;
5. If not found, fall back to pip installed in rez own virtualenv.

I also added a deprecation warning when `--pip-version` is used.

Lastly, I added a wiki page to explain what is `rez-pip` and how it really works under the hood. This, I hope, will save us some questions and will help us remember what are the good practices when it comes to pip + rez.

I've tested this with most (if not all) scenarios, but I invite everybody to also try on their side to verify that it works for everybody in all sort of environment (windows, osx, linux, etc).

### DIFF
--- a/src/rez/cli/pip.py
+++ b/src/rez/cli/pip.py
@@ -6,13 +6,14 @@ from __future__ import print_function
 
 def setup_parser(parser, completions=False):
     parser.add_argument(
-        "--pip-version", dest="pip_ver", metavar="VERSION",
-        help="pip version (rez package) to use, default is latest")
-    parser.add_argument(
         "--python-version", dest="py_ver", metavar="VERSION",
         help="python version (rez package) to use, default is latest. Note "
         "that the pip package(s) will be installed with a dependency on "
         "python-MAJOR.MINOR.")
+    parser.add_argument(
+        "--pip-version", dest="pip_ver", metavar="VERSION",
+        help="pip version (rez package) to use, default is latest."
+        " This option is deprecated and will be removed in the future.")
     parser.add_argument(
         "-i", "--install", action="store_true",
         help="install the package")
@@ -31,6 +32,7 @@ def setup_parser(parser, completions=False):
 def command(opts, parser, extra_arg_groups=None):
     from rez.pip import pip_install_package, run_pip_command
     import sys
+    import warnings
 
     if not (opts.search or opts.install):
         parser.error("Expected one of: --install, --search")
@@ -39,6 +41,15 @@ def command(opts, parser, extra_arg_groups=None):
         p = run_pip_command(["search", opts.PACKAGE])
         p.wait()
         return
+
+    if opts.pip_ver:
+        with warnings.catch_warnings():
+            # Cause all warnings to always be triggered.
+            warnings.simplefilter("always")
+            warnings.warn(
+                "The option --pip-version is deprecated and will be removed in a future version",
+                category=DeprecationWarning
+            )
 
     installed_variants, skipped_variants = pip_install_package(
         opts.PACKAGE,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,23 @@
 # Change Log
 
+## 2.48.0 (yyyy-mm-dd)
+[Source](https://github.com/repos/nerdvegas/rez/tree/2.48.0) | [Diff](https://github.com/repos/nerdvegas/rez/compare/2.47.4...2.48.0)
+
+**Notes**
+
+* Rez-pip: Add a new logic to find which pip will be used to install pip packages.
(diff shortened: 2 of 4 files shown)


src/rez/cli/pip.py
```diff
@@ -40,6 +42,15 @@ def command(opts, parser, extra_arg_groups=None):
         p.wait()
         return
 
+    if opts.pip_ver:
+        with warnings.catch_warnings():
```
nerdvegas: I feel like we should just trigger the warning here, and leave it up to elsewhere to set the filter. Since this is CLI code, we could safely just set to `always` in cli._main or something. The reason I say is because there are a bunch of other places where deprecation warnings should go, that aren't necessarily in CLi code. In those cases you would not be wanting to set the warnings filter - that would be up to the host application to set.

We might also consider wrapping this with a utils.deprecation_warning() func.

nerdvegas: ps - I also think we should do this in a separate PR though, so happy to leave this as-is

JeanChristopheMorinPerso: My thinking was that if we always let warnings be printed out, it could cause some noise. That's nice when doing development and when running tests, but it's not so nice for a user that just wants to use rez without getting warnings from other libs or python itself for example. That's why I haven't enabled them globally.

I also thought about a deprecation warning function but I wasn't sure if you would have like it since right now there is no deprecation policy in Rez (For example, see the policy of pip https://pip.pypa.io/en/stable/development/release-process/#deprecation-policy), as everything as to stay backward compatible... Introducing deprecation warnings would mean we would start to really deprecate things and break backward compatibility in some places... So I wasn't sure. But if we are fine with this (I am fully fine with breaking compatibility when necessary and when well documented with proper error messages/warnings that points to the right documentation), I'm all for a common function.

nerdvegas:
Fair enough that enabling all warnings in rez cli tools may not be the best
idea. But yeah it'd also be a good idea to push into a utility function.
The problem I see is that, if you're in non-cli code then you shouldn't be
mucking about with warnings filters (same idea as python logging - apps
should control logging settings, not the libs doing the logging). And yet
at the same time, we _do_ want to force these warnings on when in cli, as
in your code above. But it would be nice to have just the one
deprecation_warning() function that works correctly regardless of where it
gets run from (ie via rez cli tool, or something else using the rez API).

In order to do that, we need some way of knowing if a rez cli tool is
currently being used. If it's not already there, it should be trivial to
add some env var (_REZ_CLI_TOOL or something), so then
deprecation_warning() can do what it needs to. Ie, always print the warning
if a CLI tool is active, or just call warnings.warn if not.

JeanChristopheMorinPerso:
@nerdvegas I added an entry in the changelog. You will just need to update the date for the release (I had to rebase too).

JeanChristopheMorinPerso:
GH Actions (from my fork): https://github.com/JeanChristopheMorinPerso/rez/runs/261669537

JeanChristopheMorinPerso:
Though I haven't updated the version in `rez.utils._version`...

nerdvegas:
@JeanChristopheMorinPerso I notice you're using python's `-s` flag when using pip and I can't help but wonder if we should be doing this with rez's own CLI tools also. IIRC we only use `-E` for those.

JeanChristopheMorinPerso:
@nerdvegas `virtualenv` technically ignores the `user site directory` by default. The only thing I'm not sure about is if it also ignores `$PYTHONNOUSERSITE` which I think it does. I'll try to find a documentation that correlates what I'm saying... but so far I find no useful info apart from me knowing it works like this... But I feel I'm missing something. In any case, we can go ahead with this PR and do the `-s` changes if needed in another PR I think.

PR #758, opened 2019-10-05
Rez test release hook

fnaum:
A simple pre-release test hook that we have been using at AL for more than a year.

It just run `rez test packageName==Version` if the package that is being released contains the `tests` field

Related to https://github.com/nerdvegas/rez/issues/665

### DIFF
--- a/src/rez/config.py
+++ b/src/rez/config.py
@@ -373,6 +373,7 @@ config_schema = Schema({
     "warn_none":                                    Bool,
     "debug_file_loads":                             Bool,
     "debug_plugins":                                Bool,
+    "debug_hooks":                                  Bool,
     "debug_package_release":                        Bool,
     "debug_bind_modules":                           Bool,
     "debug_resources":                              Bool,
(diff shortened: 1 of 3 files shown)


nerdvegas:
Isn't this actually a post-release hook? The package is already installed before its tests are run.

I think we're gonna need to test via a temp install, and abort the actual release if test(s) don't pass.

fnaum:
Maybe this code does not make sense without the point #2 in this other PR https://github.com/nerdvegas/rez/pull/759
When we run rez test, we do a local install and use that package (contents and metadata) to run the test, when this fails this pre-release hook aborts the release.

fnaum:
actually reading the code now, I just realize that this is using `install_path` which in our case that is a temp area.

We do have a specialized rez-unleash command that does a rez-release in a temp area and then if the tests passes get the package and checks it in our asset management system that puts it in the final destination.
I'll have a look when I'm back to work if I can make this more general for the community

nerdvegas:
I have to decline this PR as it's superseded by https://github.com/nerdvegas/rez/pull/811

To give context - the way this is now managed in rez-test avoids the need to build/install the package twice. Unfortunately this cannot be implemented as a pre-release hook, because the hook is not called at the appropriate time (and changing that would not make sense for most hook plugins).

Please see that PR for more details.
Thx
A

PR #759, opened 2019-10-08
Rez test improvements

fnaum:
Some rez test improvements

- added the ability to run `rez test` in "developer mode" using `rez test .` from the root folder of the package.
- Automatically build the package if the package is not found in  the `REZ_LOCAL_PACKAGES_PATH` 
- Update the built package.py if the `tests` section changed
- added `--interactive` option, to resolve an interactive environment to run the tests. ie. (`rez test --interactive TESTNAME`). In case the package has multiple variants, you can also specify the variant using `--variant` 
- By default change to a `build/rez_test_result` directory to run the test and store the test artifacts (this way we can pick those results from tools like sonar, that expect the nose and coverage reports in a known place). If they need to be run on the root of the packages  `run_in_root` can be set in the `tests` field of the `package.py`.
- added `--clean` to clean the test artifacts created from a previous test run.
- Tests added along with the `mock` module under `vendor`


nerdvegas:
Update: Hoping to be working on this soon, we need the same functionality at Method. Stay tuned.

src/rez/cli/test.py
```diff
@@ -14,11 +25,17 @@ def setup_parser(parser, completions=False):
...
+        help="resolves an interactive environment for running the tests.")
+    parser.add_argument(
+        "--variant", dest="variant", default=0, type=int,
+        help="variant number which should be used to set the interactive env. Works only with --interactive option")
     PKG_action = parser.add_argument(
         "--extra-packages", nargs='+', metavar="PKG",
         help="extra packages to add to test environment")
     PKG_action = parser.add_argument(
-        "PKG",
+        "PKG", nargs='?',
```
nerdvegas: `-h` does not make it clear what it means if `PKG` is not provided. It also is not clear from the help that `PKG` can be a path (ie `.`). I'm tempted to say that we keep it simpler than this - PKG is either provided or not, and if not, then we're in "developer" mode.

src/rez/cli/test.py
```diff
@@ -14,11 +25,17 @@ def setup_parser(parser, completions=False):
     parser.add_argument(
         "--nl", "--no-local", dest="no_local", action="store_true",
         help="don't load local packages")
+    parser.add_argument(
+        "--interactive", dest="interactive", default=None,
+        help="resolves an interactive environment for running the tests.")
+    parser.add_argument(
+        "--variant", dest="variant", default=0, type=int,
+        help="variant number which should be used to set the interactive env. Works only with --interactive option")
```
nerdvegas: This should work in the general case also, it makes sense to specify a variant to run tests on.

src/rez/cli/test.py
```diff
@@ -67,3 +95,75 @@ def command(opts, parser, extra_arg_groups=None):
...
+
+def is_dev_run(name):
+    """
+    Decides if is a dev_run, based on the package name
+    Args:
+        name: name of package
+
+    Returns: True or False
+    """
+    return not name or name == '.'
```
nerdvegas: Too fragile. As mentioned earlier, I think it'd be better if developer mode is enabled only if PKG is not provided. Right now is odd because PKG can be a package string, or ".", but not any other valid path. There is too much variability in what PKG can be.

src/rez/cli/test.py
```diff
@@ -67,3 +95,75 @@ def command(opts, parser, extra_arg_groups=None):
...
+
+
+def get_package(path):
+    """
+    Function validates pkg name and indicates if tests were run in developer mode
+    Args:
+        path: path to the package
+
+    Returns: Tuple of values (package_name, is_dev_run)
```
nerdvegas: Wrong return docstring. Also not sure we need this function.

src/rez/cli/test.py
```diff
@@ -67,3 +95,75 @@ def command(opts, parser, extra_arg_groups=None):
...
+    """
+    try:
+        return get_developer_package(path)
+    except PackageMetadataError:
+        print("Not a valid rez package. Make sure you are in package's root directory", file=sys.stderr)
+        return None
+
+
+def prepare_dev_env_package(package):
```
nerdvegas: I'm actually pretty hesitant about rez-test possibly implicitly installing a package like this. Package installation is always explicit everywhere else. Furthermore this feels arbitrary - the package gets installed if it wasn't there, but what if it's there but put of date due to a sourcecode change? We can't detect that and I don't think we should.

It seems simpler and less error prone to simply give an error if the current developer package isn't present as a local package, and leave it up to the user if and when to rebuild/reinstall the package.

src/rez/cli/test.py
```diff
@@ -67,3 +95,75 @@ def command(opts, parser, extra_arg_groups=None):
...
+    sys.argv = [original_args[0]] + ['-i']
+    try:
+        run("build")
+    except SystemExit as exit_code:
+        if exit_code.code:
+            raise
+    sys.argv = original_args
+
+
+def update_package_in_local_packages_path(package, installed_package):
```
nerdvegas: Again this feels arbitrary - the package is being kept up-to-date wrt its tests, but not its actual payload (ie if source was changed). I think we should just leave it up to the developer to reinstall the package.

Having said that, something that I think _would_ be acceptable in developer mode is if the test definitions are always taken from the developer package, rather than from the installed package.

src/rez/package_maker__.py
```diff
@@ -31,7 +31,8 @@
         Or(basestring, [basestring]),
         {
             "command": Or(basestring, [basestring]),
-            Optional("requires"): [package_request_schema]
+            Optional("requires"): [package_request_schema],
+            Optional("run_in_root"): bool
```
nerdvegas: `run_in_root` is much too general a setting name for this case. I'd suggest `tests_artifact_path` or similar.

src/rez/package_test.py
```diff
@@ -209,11 +215,20 @@ def print_header(txt, *nargs):
 
                 print_header("\nRunning test command: %s\n", cmd_str)
 
-            retcode, _, _ = context.execute_shell(
-                command=command,
-                stdout=self.stdout,
-                stderr=self.stderr,
-                block=True)
+            test_result_dir = os.path.abspath('build/rez_test_result' if not test_info["run_in_root"] else '.')
```
nerdvegas: 'build/rez_test_result' should be configurable. As mentioned earlier, we should have a `tests_artifact_path` setting rather than `run_in_root`. Then, to run tests in root for a given package, you do a package-level config override, like so:

```
with scope("config") as c:
    c.tests_artifact_repository = '.'
```

This is more consistent with how settings are managed in other parts of the codebase.

src/rez/utils/system.py
```diff
@@ -36,3 +37,18 @@ def popen(args, **kwargs):
             kwargs["stdin"] = subprocess.PIPE
 
     return subprocess.Popen(args, **kwargs)
+
+
+@contextmanager
+def change_dir(dir_path):
```
nerdvegas: This is very similar to existing `rez.utils.filesystem.retain_cwd`. I would update this existing func to be able to chdir to a specified path, and use that instead.

nerdvegas:
So I'm freed up to work on this next week for Method. I think it would make sense to use this PR and https://github.com/nerdvegas/rez/pull/758 as a basis for that work, rather than updating these PRs though.

Reasons are:
* I need to make immediate use of the time I'm getting from Method (rez dev time doesn't happen often!)
* We need to make these changes with https://github.com/nerdvegas/rez/issues/665 in mind, so there will probably be some significant differences to these PRs.

Fede I wanted to let you know in case you thought I'm throwing this work away! Rest assured a chunk of the work is already done here and I'll be incorporating it into the upcoming PRs. Expect to see more next week.

nerdvegas:
Note that some of this PR has been superseded by https://github.com/nerdvegas/rez/pull/807. However, I'm still open to adding the dev mode functionality shown here, as long as there's consensus. It's not clear to me how this should work exactly - as mentioned above, the way the package is automatically re-installed if the tests definition changes, feels inconsistent.

Wrt `--interactive`, it definitely makes sense to do this, and I'm aiming to make this available in a separate PR which mimics the behaviour in this PR.

JeanChristopheMorinPerso:
@nerdvegas @fnaum The way I see the dev mode is just to avoid some typing. So instead of doing `rez-build -ci && rez-test packageA==x.x.x`, you could just do `rez-build -ci && rez-test .`. It would make things simpler to use IMO. So it's just a "shorthand".

nerdvegas:
Yeah I don't mind that functionality, but in
https://github.com/nerdvegas/rez/pull/759 'dev mode' is not quite that
simple, as it does things like automatically re-install the package if its
tests change. I am not so sure about that, because it feels a bit arbitrary
as to whether a package is reinstalled (ie yes if tests change, no if
anything else changes).

In #759, test byproducts are also written to a specific directory when in
dev mode. So the question is also whether this should happen and if so,
what that path should be.

A

JeanChristopheMorinPerso:
>in #759 'dev mode' is not quite that
simple, as it does things like automatically re-install the package if its
tests change.

Yeah, I wouldn't go with the auto-install. Hence why I was giving `rez-build -ci && rez-test .` as an example. I'd almost prefer if the user was manually building (as he probably knows more what he needs to do for his tests to run then us).

>In #759, test byproducts are also written to a specific directory when in
dev mode. So the question is also whether this should happen and if so,
what that path should be.

I still haven't made my mind on this. To be frank, the first time I reviewed this PR I didn't like the idea that rez would know where my tests artifacts are. And as of today I have the same feeling. You could have 5 different test target that each run different tools that will all generate tests artifacts in a different way (and sometimes if would even be impossible to choose where to put them). That means each test would have to define it's artifact directory. Though a possible case, it might not be a super common one. It would be easy to implement though. I think the hardest part will be to agree on how to handle that...

nerdvegas:
Marking as ON HOLD - these features are desired (interactive mode, dev mode), however the code has changed too much for this PR to be viable. Keeping for reference until an equivalent is merged.

src/rez/cli/test.py
fnaum: agree

src/rez/cli/test.py
fnaum: probably not needed, I'm guessing was added to make the code more testable, but  I'm pretty sure you can call directly `get_developer_package`

src/rez/cli/test.py
fnaum: yes, we can not detect that. ATM we are only detecting changes on the packages.py and reinstalling the package when the tests are out of date. 
I'll do a poll tomorrow when I'm at work, I'm guessing not many devs are really relying on this feature. So, I guess it can go away

src/rez/cli/test.py
fnaum: In that case, if they are taken from the developer package, I think people will be happy

src/rez/package_test.py
fnaum: agree

fnaum:
Sorry, I was not active when you reviewed this, we lost the only resource that was working with rez and all the other tools and we have not found a replacement yet. :(

PR #762, opened 2019-10-15
Subproc wrapper part2

### LINKED ISSUE #761: ResourceWarning with ResolvedContext.execute_shell (py3)
In test_shells, the following simplified test is giving the warnings shown:

```
    def test_wtf(self):
        import tracemalloc
        tracemalloc.start()

        r = ResolvedContext([])

        for i in range(2):
            p = r.execute_shell(command='ls', stdout=subprocess.PIPE)
            p.wait()
```

Output:
```
:ajohns@turtle:~/tmp$ PYTHONTRACEMALLOC=1 ~/rez3/bin/rez/rez-selftest --shells
/home/ajohns/rez3/lib/python3.7/site-packages/rez/tests/test_shells.py:62: ResourceWarning: unclosed file <_io.BufferedReader name=4>
  p = r.execute_shell(command='ls', stdout=subprocess.PIPE)
Object allocated at (most recent call last):
  File "/home/ajohns/packages/python/3.7.4/platform-linux/lib/python3.7/subprocess.py", lineno 759
    self.stdout = io.open(c2pread, 'rb', bufsize)
/home/ajohns/packages/python/3.7.4/platform-linux/lib/python3.7/unittest/case.py:628: ResourceWarning: unclosed file <_io.BufferedReader name=5>
  testMethod()
Object allocated at (most recent call last):
  File "/home/ajohns/packages/python/3.7.4/platform-linux/lib/python3.7/subprocess.py", lineno 759
    self.stdout = io.open(c2pread, 'rb', bufsize)
```

These warnings are only being shown when the SECOND execute_shell call is made. If range(2) is changed to range(1), no warnings occur.

This ONLY HAPPENS IN THE TEST. If I take the same code and run it in a script via rez-python, I don't get any warnings.

nerdvegas:
-added new `Popen` class (subclassed `subprocess.Popen`) to allow for use as context in py2/3. This fixes a ResourceWarning in py3.
-removed old `open` and `py23.subprocess_Popen` functions
-moved a bunch of subproc/script- related functions to new rez.utils.execution module (want to deprecate rez.util at some point)
-fixed DEV_NULL file object left open in some cases

Closes #761


JeanChristopheMorinPerso:
I'm not sure I get the point of subclassing `Popen` if it's just to not call `communicate`... The cost of `_, _ = p.communicate()` is the same as `p.wait()`, but having this py2/3 wrapper cost (maintenance) more than calling `communicate` IMO. The current implementation of the `__exit__` for py2 also doesn't match 100% what's done in the stdlib in py3.

Note that I'm not against the subclass for the `stdin` and the `text` thing (well I find the text thing limit, but I can live with it).

The way I see things, the less py2/3 hacks we have, the better the health of the project will be.

nerdvegas:
I don't think there's anything wrong with subclassing Popen as a means of
adding the support for 'text', and the extra stuff we do wrt stdin. And I
don't see any specific need for py2's Popen.__exit__ to match py3 - it's
just enabling use as a context in both cases. Wrt 'text', that was done
because it's really just a renaming of universal_newlines in py3, which
they did because it's become misleading in py3 (it isn't intuitive that
'universal_newlines=True' would change the return type).

I can see that calling communicate() avoids the problem, but I don't like
that because it's misleading - communicate is there to use when you want to
get the output of a process. What I don't understand is why use of wait()
in py3 would be causing a ResourceWarning when that must surely be a very
common pattern.

Given all of the above, I opted for adding support for, and using, Popen as
a context where necessary, to avoid the leak. It fixes the issue and keep
the intent of the code clear. I also think it's generally cleaner to use
contexts for temporary resources anyway, so why not move towards that in
all cases where Popen is used?

maxnbk:
testing on cent with py-2.7, 3.6, 3.7, 3.8 has succeeded for me with no ResourceWarning.

PR #772, opened 2019-10-23
Skip `test_build_cmake` on Windows.

bfloch:
The goal is to have better visibility of failure on Windows in our
automated tests, while the CMake issues are being addressed.
Added a known issues and limitations section in the README.

### DIFF
--- a/src/rez/tests/test_build.py
+++ b/src/rez/tests/test_build.py
@@ -9,6 +9,7 @@ from rez.exceptions import BuildError, BuildContextResolveError,\
 import unittest
 from rez.tests.util import TestBase, TempdirMixin, find_file_in_path, \
     per_available_shell, install_dependent, program_dependent
+from rez.utils.platform_ import platform_
 import shutil
 import os.path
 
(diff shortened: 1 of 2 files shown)


bfloch:
Should pass Windows after #770

src/rez/tests/test_build.py
```diff
@@ -155,6 +156,11 @@ def test_builds_anti(self):
     @program_dependent("cmake")
     def test_build_cmake(self):
         """Test a cmake-based package."""
+        if platform_.name == "windows":
```
JeanChristopheMorinPerso: I'm wondering if we could potentially do:

```python
    @unittest.skipIf(platform_.name == "windows", "This ...")
    @program_dependent("cmake")
    def test_build_cmake(self):
        """Test a cmake-based package."""
```
Would that work?

nerdvegas: I'm merging this anyway, given that it's a temporary measure. If it works then it works - better to get traction on the source of the problem (and yes Blazej I will get up to speed on that ticket soon).

bfloch: Looks good @JeanChristopheMorinPerso.
I just took it from some other location that does it with a condition, so we might want to unify all of them.
But it's also true that we should try to get rid of these guys completely :)

Issue #773, opened 2019-10-24
Windows actions fail


bfloch:
Our actions fail on windows. Only the cmake test is supposed to fail
But it was skipped in #772 so we should actually pass.

In reality we fail additional 4 tests:

1. `test_build_custom (powershell)` - Usually is not being run due to a dependency on g++ and make, which is kindly provided by the github actions instance. We have to probably deal with it or try to skip it no matter what (assuming that mingw is not a supported path)

2. `test_build_whack (pwsh)` - Something wrong with the environment!?

3. `test_command (cmd)` - I assume that python is not the default application for .py and hence the PATHEXT variable has not effect.

4. `test_command_returncode (cmd)` Same as before

Logs as seen on https://github.com/nerdvegas/rez/runs/272706733

```
2019-10-24T04:36:52.0179972Z ======================================================================
2019-10-24T04:36:52.0180850Z SOLVE #6 (2 fails so far)...
2019-10-24T04:36:52.0200508Z ERROR: test_build_custom (rez.tests.test_build.TestBuild)
2019-10-24T04:36:52.0200671Z SUCCESS
2019-10-24T04:36:52.0201957Z Test a make-based package that uses the custom_build attribute.
2019-10-24T04:36:52.0208387Z checking that permutations also succeed...
2019-10-24T04:36:52.0235668Z ----------------------------------------------------------------------
2019-10-24T04:36:52.0237804Z SOLVE #1 (0 fails so far)...
2019-10-24T04:36:52.0238380Z Traceback (most recent call last):
2019-10-24T04:36:52.0238470Z SOLVE #2 (0 fails so far)...
2019-10-24T04:36:52.0240808Z   File "d:\a\rez\rez\build\lib\site-packages\rez\tests\util.py", line 158, in wrapper
2019-10-24T04:36:52.0240999Z SOLVE #3 (1 fails so far)...
2019-10-24T04:36:52.0241903Z     return func(self, *args, **kwargs)
2019-10-24T04:36:52.0242041Z SOLVE #4 (1 fails so far)...
2019-10-24T04:36:52.0244123Z   File "d:\a\rez\rez\build\lib\site-packages\rez\tests\test_build.py", line 177, in test_build_custom
2019-10-24T04:36:52.0244302Z SOLVE #5 (2 fails so far)...
2019-10-24T04:36:52.0244720Z     proc = context.execute_command(['hai'], stdout=PIPE)
[...]

PR #775, opened 2019-10-25
Fixes some failing tests on windows

willjp:
I took a stab at addressing some of the failing tests under python3+windows.

* corrects failing tests in `rez/tests/test_packages.py` on windows
* adds `%SYSTEMROOT%` envvar to Python(ActionInterpreter) so `rez/tests/test_context.py` tests pass on windows


bfloch:
Haven't tried your PR yet, but I think version and changelog are handled by @nerdvegas

willjp:
oh! Sorry, I was trying to follow the directions in `CONTRIBUTING.md` . I can undo the commit though if it is a problem.

bfloch:
I think you might be right. It's just that this was recently updated and I misunderstood the change:
https://github.com/nerdvegas/rez/blob/master/RELEASE.md

bfloch:
Do you think you change can in future be represented by the mechanism proposed here:
https://github.com/nerdvegas/rez/issues/737

Not saying that we don't need the workaround for the time being, but just checking if the proposal
might miss a feature.

willjp:
Oh that looks very nice! No sense hiding conditionals all over the place :)

If I have understood your proposal correctly:

* The first commit's fixes would be addressed by your ``"modify_op": PlatformDependent({"windows": "windowspath"})`` to format the path consistently. It would just need to be used on both ends, or have a special object that can be used to test equality.

* The second commit is kind of weird, and specific python-3 on windows. I'm curious if it is even needed in later versions of python. A cleaner workaround within your proposal might define a whitelist of variables to inherit from the parent process unless they are assigned by the user? 

Sorry, I got excited and wanted that green tests-passing badge across the board :)

bfloch:
Me, too. Other then your fix I am working towards fixing the other Windows tests in #773

nerdvegas:
Just two general points:

1. Yes the contributing guidelines were changed yesterday, but I only
mentioned on slack, nw.
2. Generally if we need to put in a fix that we also know will change at
some point in future (eg as Blazej mentions about #737), I'm fine with
that, but please please put a big fat comment there highlighting the fact.

Cheers!
A

src/rez/rex.py
```diff
@@ -633,6 +634,7 @@ def error(self, value):
     def subprocess(self, args, **subproc_kwargs):
         if self.manager:
             self.target_environ.update(self.manager.environ)
+        self.target_environ = self.adjust_env_for_platform(self.target_environ)
```
nerdvegas: I would drop the return value, and avoid the unnecessary dict copy.  `adjust_env_for_platform` should just alter self.target_environ in-place, asthe previous code already does

src/rez/rex.py
```diff
@@ -684,6 +686,59 @@ def get_key_token(self, key):
         # here because the API requires it.
         return "${%s}" % key
 
+    def adjust_env_for_platform(self, env):
+        """ Make required platform-specific adjustments to env.
+        """
+        if sys.platform.startswith('win'):
+            env = self._add_systemroot_to_env_win32(env)
```
nerdvegas: Same as above, in-place rather than copy

src/rez/rex.py
```diff
@@ -684,6 +686,59 @@ def get_key_token(self, key):
...
+                                 'SYSTEMROOT': 'C:\Windows'})
+
+                # failure
+                Popen(cmds, env={'PATH': 'C:\\Python-3.6.5'})
+
+                #> Fatal Python Error: failed to get random numbers to initialize Python
+
+        """
+        # 'SYSTEMROOT' only relevant on windows
+        if not sys.platform.startswith('win'):
```
nerdvegas: this check is already done, and this is a non-public func, so this can be removed

src/rez/rex.py
```diff
@@ -684,6 +686,59 @@ def get_key_token(self, key):
...
+        if env is None:
+            return env
+        # leave SYSTEMROOT alone if set by user
+        if 'SYSTEMROOT' in env:
+            return env
+        # not enough info to set SYSTEMROOT
+        if 'SYSTEMROOT' not in os.environ:
+            return env
+
+        new_env = env.copy()
```
nerdvegas: as above, in-place rather than copy

src/rez/tests/test_packages.py
```diff
@@ -309,11 +316,17 @@ def preprocess(this, data):
 
     def test_6(self):
         """test variant iteration."""
+        # ensure that differing case doesn't get interpreted as different repos
+        # on case-insensitive platforms (eg windows)
+        base = os.path.join(self.py_packages_path, "variants_py", "2.0")
+        if not platform_.has_case_sensitive_filesystem:
+            base = base.lower()
```
nerdvegas: should move `_format_platformpath` and reuse here instead

nerdvegas: Actually, this sort of code already exists in another place, see https://github.com/nerdvegas/rez/blob/master/src/rezplugins/package_repository/filesystem.py#L473.

It would be good to add a utility function to rez.utils.filesystem, something like so:

```
def canonical_path(path, platform=None):
    if platform is None:
        from rez.utils.platform_ import platform_
        platform = platform_

    path = os.path.realpath(path)
    if not platform.has_case_sensitive_filesystem:
        path = path.lower()
    return path
```

I think this may also fix a potential bug currently, where two package repositories could be the same, but if one uses symlinks to the same path on disk, rez will think they are two different repos.

To give some context: The lower() is there to fix a previous bug that was caused by differing cases of the same path, on windows (non-case-sensitive), which then led rez to believe that there were two different packages that were in fact the same package ('foo' vs 'Foo' for eg).

At least if there's just the one utility function, this can be documented in one place.

willjp:
Thank you very much for the code review! 

I'm updating the code now, and I'll post again once I hve it fixed.

src/rez/rex.py
willjp: fixed!

src/rez/rex.py
willjp: fixed!

src/rez/rex.py
willjp: fixed!

thank you, I'm never sure how much I should worry about accidents. this seems like a good rule of thumb.

src/rez/rex.py
willjp: fixed!

src/rez/tests/test_packages.py
willjp: Thank you, I agree with this idea.

I wrote `canonical_path` in addition to tests for it.

There are some *frustrating* pain points here for testing however.

``` python
# python-3.6.5, windows-10
ntpath.realpath('/a/b/c')      #>> 'C:\\a\\b\\c'
posixpath.realpath('C:/a/b/c') #>> 'C:\\Users\\willjp\\progs_shared\\vendor\\rez/C:/a/b/c'
```

``` python
# python-3.7.4, linux
ntpath.realpath('/a/b/c')     #>> '\\a\\b\\c'
posixpath.realpath('/a/b/c')  #>> '/a/b/c'
```

This change in python is not represented in the docs, so I am uncertain when they were introduced:  https://docs.python.org/3/library/os.path.html?highlight=realpath#os.path.realpath

willjp:
Firstly, thanks very much for the code-review! I really appreciate the help.

I updated the code:

* new `rez.utils.filesystem.canonical_path()`
* new test file `rez.tests.test_utils.test_filesystem` to test `canonical_path`
* `rez.tests.test_context` uses `canonical_path`
* `rezplugins.package_repository.filesystem.FileSystemPackageRepository` uses `canonical_path`

Unrelated to previously mentioned:

* `rez.utils.filesystem.to_nativepath()` was broken (`os.path.join` requires an expanded list). I also adjusted it to use `replace('\\', '/')` .

Misc Questions:

* I noticed from old imports that you previously had a `rez.vendor.mock` but removed it. None of the other tests use mocks, so I'm guessing that you made an intentional decision to use `TestCase.skipTest()` instead. I followed this pattern, but just wanted to confirm this is how I should be proceeding.

* Out of curiosity, what is the goal of`rez.utils.filesystem.{to_nativepath, to_ntpath, to_posixpath}` versus `{os.path, ntpath, posixpath}.normpath` ? I feel like the latter might accomplish more, but I didn't want to break anything you already have in place.

Finally:

Thanks for clearing up about `CONTRIBUTING.md` - In the future, I'll leave the version and changes files alone. :)

bfloch:
Together with this PR I get green tests on Windows via docker for all three python version.

nerdvegas:
Nice, will get this merged soon.

I've asked ASWF about their container hosting, haven't heard back yet.

A

JeanChristopheMorinPerso:
@willjp For the test file you added, instead of `rez/tests/test_utils/test_filesystem.py`, we will need a `rez/tests/test_utils_filesystem.py`. This is so that the test can be shown when using `rez-selftest`.

willjp:
sure, fixed!
 
Actually while on the subject of tests - do you want this testfile? None of the other utils are tested directly, I wasn't sure if this was something you guys would want, or if this was unecessary.

JeanChristopheMorinPerso:
Tests are always welcome. We are lagging behind in terms of tests, so yes multiple parts of rez are not really tested. In general, I think we have about 43% of coverage. Thanks for the changes and adding the new tests :)

nerdvegas:
```
Out of curiosity, what is the goal ofrez.utils.filesystem.{to_nativepath, to_ntpath, to_posixpath} versus {os.path, ntpath, posixpath}.normpath ? I feel like the latter might accomplish more, but I didn't want to break anything you already have in place.
```

Good question. I don't remember, it's worht investigating. Could be that py2.6 didn't have that support? Not sure.

Made a ticket: https://github.com/nerdvegas/rez/issues/783

bfloch:
@nerdvegas @willjp Wondering if we could make the `test_utils_filesystem.py` -> `test_utils.py` for now.
I have literally a single test from data_utils (HashableDict) and I did not want to open a new test suite for it.
Since you also have a couple of tests I wonder if we could share that namespace and rather diverge in future when we have more tests.

Thoughts?

nerdvegas:
Yeah I don't think it's unreasonable to have a general utils test, so go
ahead. Rule here should just be that stuff tested in utils is low level and
standalone.
Cheers
A

willjp:
yeah absolutely, that sounds good to me. Give me a few minutes to finish up at work and I'll push an update.

PR #779, opened 2019-10-28
Popen UnicodeDecodeError partial fix

### LINKED ISSUE #776: rez-release UnicodeDecodeError (windows)
``` yaml
rez:  2.47.9
os: windows-10.0.17763
python: 3.7.4
```

~~Occurs when rez-releasing a package on windows, if the git repostiory already has the tag wanted by the rez-release.~~

Occured during `rez-release`, when it runs `git log` in subprocess with `text=True`, and log contains `chr(141)` .

```
Checking state of repository...
Exception in thread Thread-23:
  File "C:\Python-3.7.4\Lib\threading.py", line 926, in _bootstrap_inner
    self.run()
  File "C:\Python-3.7.4\Lib\threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Python-3.7.4\Lib\subprocess.py", line 1238, in _readerthread
    buffer.append(fh.read())
  File "c:\users\willjp\documents\rez\install\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8d in position 1021688: character maps to <undefined>
```

willjp:
Fixes #776 

~~Overrides `subprocess.Popen(..., text=True).communicate()` to use `stdout.decode('utf-8')` instead of default behaviour. *does not fix in stdout/stderr attributes* .~~

~~**Has anyone else encountered this? If not I will close this PR**. I'm guessing it came from a copy/paste, I don't know how common this type of issue would be. This fix feels wrong, but It's the only idea I could come up with.~~

~~Do you guys have any thoughts or suggestions?~~

sets `subprocess.Popen(..., encoding='utf-8')` when `text=True` to address decoding issues.

### DIFF
--- a/src/rez/utils/execution.py
+++ b/src/rez/utils/execution.py
@@ -42,7 +42,6 @@ class Popen(_PopenBase):
     Allows for Popen to be used as a context in both py2 and py3.
     """
     def __init__(self, args, **kwargs):
-
         # Avoids python bug described here: https://bugs.python.org/issue3905.
         # This can arise when apps (maya) install a non-standard stdin handler.
         #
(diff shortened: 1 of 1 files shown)


willjp:
reverts original commit, replaces with param `subprocess.Popen(..., encoding='utf-8')`. *MUCH* cleaner, simpler, and works with stdout/stderr attributes.

Thank you @bfloch  and @nerdvegas for finding a much better solution.

willjp:
oh dear, that didn't go my way... I apparently broke everything. I'll take another look at this tomorrow night.

nerdvegas:
What's latest with this, what issue was caused?
Cheers
A

willjp:
I'm just comparing the issues that came back with the CI server, with another branch - but I think it is good to go.

The issue appeared on windows in `python-3+`, during a rez-release, when my git log contained a reverse-linefeed-character. I replaced my original ugly fix with the param `Popen(..., encoding=True)`, but it is a parameter that was only introduced in `python-3.6` .

So it should now work for all of python2, and python3.6+.

willjp:
If you'd like I can write a test as a proof of the break/solution.

I started but gave up, after wrestling with trying to get the character to print at all.

nerdvegas:
I don't think a test for this very specific case is needed, but yeah we
should think about incorporating unicode into some tests, for eg putting
unicode into author/description of a package in the tests for eg. But I
think we can leave that for another REP (improving tests).

Thx!
A

src/rez/utils/execution.py
```diff
@@ -59,15 +58,34 @@ def __init__(self, args, **kwargs):
...
+        kwargs["universal_newlines"] = True
 
-        super(Popen, self).__init__(args, **kwargs)
+        # fixes py3/cmd.exe UnicodeDecodeError() with some characters.
+        #    UnicodeDecodeError: 'charmap' codec can't decode byte
+        #    0x8d in position 1023172: character maps to <undefined>
+
+        # NOTE: currently no solution for `python3+<3.6`
+        #
+        if sys.version_info[:2] >= (3, 6):
```
nerdvegas: I would do this only if `encoding` is not already in kwargs

willjp: oh thank you! fixing now!

willjp:
Thanks again for the help/patience. You guys are all super nice :)

PR #780, opened 2019-10-30
Unwanted debug printing

predat:
Prevent rez from printing debug messages such as:

> 09:25:21 DEBUG searching plugin path <path_to_plugin>

even when debug_plugins is False.

(Introduce in 6f0e5db8ff50dbe8c32982dd57db374b2b679d1d).

### DIFF
--- a/src/rez/plugin_managers.py
+++ b/src/rez/plugin_managers.py
@@ -116,7 +116,8 @@ class RezPluginType(object):
         paths = reversed(paths)
 
         for path in paths:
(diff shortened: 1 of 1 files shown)


nerdvegas:
No problem with the PR, but please always include a description.

alefeve:
If I may propose a description:

Prevent rez from printing debug messages such as:
09:25:21 DEBUG    searching plugin path <path_to_plugin>
even when debug_plugins is False.

PR #781, opened 2019-10-30
Windows Tests via Docker

bfloch:
This PR includes two fixes for cmd/powershell on windows and requires https://github.com/nerdvegas/rez/pull/775 to succeed future Windows tests.

Currently relying on hub.docker.com as the github service is broken:
https://github.community/t5/GitHub-Actions/Docker-Push-to-Package-Registry-from-Windows/m-p/36393#M2560

**Setup Before Merging**
_(as configured in the PR, feel free to change if needed)_
1. Create a hub.docker.com account matching the github name: nerdvegas.
2. Create an empty repository called rez-test-win
3. In hub.docker.com go to Account/Security/New Access Token
   Choose a name (like "Github Access Token") and copy the token
4. In github, repo settings create a secret `DOCKERHUB_TOKEN` and paste the TOKEN from docker

Once merged the `Compose Windows Docker Image` action should run.
The `Windows Docker` tests will fail at this point, but future pushes should succeed.

In future any required change to the docker image should be a separate PR as documented.
As such the regular actions will not run when only the docker image related files have been touched.
Similarly the image will not be rebuild when none of its files have been touched. Since the image is tagged with the latest revision on these files everything should always be in sync and use the correct images moving forward.

There is some ugliness in the duplication of code throughout the workflow files. GH Actions pros may find better ways to pass the environments instead of recalculating them.

Fingers crossed!


bfloch:
Quick note: I left cmake out of the docker image. 
We could enable it for the future, it was just pretty slow to install during my countless tests.

.github/docker/windows/entrypoint.ps1
```diff
@@ -0,0 +1,60 @@
...
+$ErrorActionPreference = "Stop"
+
+echo "Testing $repo at $commit"
+
+# Fixes encoding issue on Windows 10 local docker run.
+#
+${ENV:PYTHONIOENCODING} = "UTF-8"
+
+# Only 2.x respects the install dir. 3.x will install to C:\Python3x
+choco install python --yes --version=$python --installargs='TargetDir=""C:\python""'
```
JeanChristopheMorinPerso: I feel like we should do the install into the Dockerfile. The goal of docker is to have a reproducible build, we we won't have with the install in the entrypoint.

Could we install all version in their respective folder in the dockerfile and append the one we test into $PATH here?

So in the Dockerfile:

```
RUN choco install python --yes --version=2.7.11 --installargs='TargetDir=""C:\python2.7""' \
    && choco install python --yes --version=3.6.8 --installargs='TargetDir=""C:\python3.6""' \
    && choco install python --yes --version=3.7.5 --installargs='TargetDir=""C:\python3.7""'
```

And here:

```
${ENV:PATH} += ";C:\python"
${ENV:PATH} += ( $python.Split(".")[0..1] -Join '' )
```
(note my noobeness in Windows here on the last line, I don't know if we can reference a variable inside a string...).

This proposition would speed up our builds and would make the tests really reproducible. What do you think?

bfloch: Good point. I guess the reason I did as I did is so we would need to maintain the images each time we want to change python Versions. The install does not add that much to the test time - but I was originally considering three images (since the cost is small due to layering in docker). In retrospect installing all three versions would work for me, too.

Btw. I was not able to control the target dir for 3.x python, but the default were reasonable so we can make it work by just installing to c:\python27

instinct-vfx: I am leaning towards putting it into the Dockerfile. Making the builds as reproducible and deterministic as possible would beat slightly lower maintenance efforts. Looking at the recent python release dates that looks bearable to me.

.github/workflows/windows-docker.yaml
```diff
@@ -0,0 +1,57 @@
+#
```
instinct-vfx: What do the errors below mean?

JeanChristopheMorinPerso: I don't know, but they are annoying. They've always been there when a build fails....

instinct-vfx: Thanks. Was just wondering.

bfloch:
Thanks for the feedback.
I will also enable the cmake install in the Docker image. We're not using it but it's good we have it once we work on the cmake support.

bfloch:
Please ignore the pushes for a while since I'm testing on this branch on my fork :-/ Sorry for the noise.

bfloch:
Another thing I should check: If the image building and usage is in the same workflow I could express the dependency and we would not need to separate the process into two commits. I'll check if this is possible with github actions. Essentially I would need the 'paths:' checks within a conditional of a step.

nerdvegas:
Not that this should be part of the same PR, but I can't help but think it'd be clean to go the 'multi python installed to docker image' route for the non-windows builds as well... including cmake install also. Thoughts?

.github/workflows/windows-docker-image.yaml
```diff
@@ -0,0 +1,102 @@
...
+  #
+  # State depends on .github\docker\rez-win-base\** and this file
+  base:
+    runs-on: windows-2019
+
+    steps:
+      - name: Login to docker repository
+        run: |
+          ${gh_user} = ("${{ github.repository }}" -Split '/')[0]
+          docker login -u ${gh_user} -p ${{ secrets.DOCKERHUB_TOKEN }}
```
JeanChristopheMorinPerso: I would use the `--password-stdin` option here. So something like `echo ${{ secrets.DOCKERHUB_TOKEN }} | docker login -u ${gh_user} --password-stdin`. THis will be safer (see https://docs.docker.com/engine/reference/commandline/login/#provide-a-password-using-stdin)

We could also probably use a credential helper to be even more secure (see https://docs.docker.com/engine/reference/commandline/login/#credentials-store).

.github/docker/rez-win-base/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+FROM mcr.microsoft.com/windows/servercore:$WINDOWS_VERSION
+
+SHELL ["powershell.exe", "-NoLogo", "-NoProfile", "-ExecutionPolicy", "ByPass"]
+
+# ------------------------------------------------------------------------------------------------------------
+# Chocolatey
+#
+ENV chocolateyUseWindowsCompression false
+RUN iex ((new-object net.webclient).DownloadString('https://chocolatey.org/install.ps1'))
+RUN choco feature disable --name showDownloadProgress
```
JeanChristopheMorinPerso: We should do all the installs into one `RUN` command. That will save us some layers, and so reduce the image weight. Also, I suspect choco downloads files in a cache of some sort that will stay in the image if we don't remove this cache. We should make sure to clean any files that was created by choco. Note that the deletion of such files needs to be done in the same layer that they were created. So in the end, we will have one `RUN` that installs stuff and that also deletes stuff.

.github/docker/rez-win-py/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+#
+ARG PYTHON_VERSION
+
+# ------------------------------------------------------------------------------------------------------------
+# Python
+#
+# Installs given Python version to C:\Python
+# Python 2.x uses msi while 3.x has an exe with separate arguments
+#
+RUN ${PYTHON_INSTALL_PATH} = 'C:\Python';                                                                                            `
```
JeanChristopheMorinPerso: I'm curious, what are the floating back quotes for?

bfloch: Essentially it is just like a backslash for multi-line commands.
Here's the deal with quoting on windows based Dockerfiles. First you should define this:
```
#escape=`
```
https://docs.docker.com/engine/reference/builder/#escape

This way you can use backslash in script as expected by PowerShell for paths and backticks for line ends with multi-line commands or quoting.

FYI backtick is the native escape character in Powershell since backslash is (wrongly) reserved for paths.

You still need to escape `\"` for `"` which you need when you want to expand a variable, while in most cases you can get away with a single `'` when using static strings.

One also has to be careful because variable expansion changes in docker AFTER the `shell` like, but in my best attempts there is no way to use a variable in the `FROM` statement with PowerShell convention.

Hence only this schema works:
```
ARG ONE
FROM $ONE<- bash-like expansion
SHELL powershell
ARG TWO <- Powershell expansion
RUN ${ENV:TWO}
```
While this will just drive you mad:
```
SHELL powershell
ARG ONE
FROM $ONE<- NO
FROM ${ENV:ONE}<- NO
What happened to my ARG value? ¯\_(ツ)_/¯
```
So the lesson, to my surprise, is that the commands may not necessarily affect global state.

.github/workflows/windows-docker-image.yaml
bfloch: Tried this without success:
```
Write-Output ${{ secrets.DOCKERHUB_TOKEN }} -NoNewLine | docker login -u ${gh_user} --password-stdin
```
Do you happen to know where to find a PowerShell sample to do this properly?

JeanChristopheMorinPerso: I am unfortunately a complete noob in Windows... So I can't help on that front. But if it's too hard, then I guess we have to give up on this...

bfloch: I think I got it. Apparently you just need the string on Powershell:
```
'${{ secrets.DOCKERHUB_TOKEN }}' | docker login -u ${gh_user} --password-stdin
```
However - without going through a file like the example in the docs show - how would this help security? Am I not still throwing the token at the log file?
Also I was hoping that the whole 'secrets' mechanism would do the right thing to not expose it to logs. Regardless - we got it now through stdin - so at least the message does not come up.

JeanChristopheMorinPerso: The actions are supposed to take care of not dumping the value of secrets in the logs, but we never know, so better be safe than sorry.

As for security, yes the password still ends up in plain text in the daemon config. That's basically why I was pointing at the different credential stores for docker.

.github/docker/rez-win-base/Dockerfile
bfloch: I've merged runs where possible.
I've added a cleaning job at the end of the images. Not sure how much is saves vs. not having it - I believe choco does cleanup temporary downloads. It is a separate commit in case we want to revert it. The images are still somewhat big but I guess that is the the windows image that mostly counts.

bfloch:
History cleaned up. Verified it working. The PR is open again.

.github/docker/rez-win-base/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN iex ((new-object net.webclient).DownloadString('https://chocolatey.org/install.ps1'));             `
+    choco feature disable --name showDownloadProgress;                                                 `
+    choco install git.install --yes --version=${ENV:GIT_VERSION};                                      `
+    choco install cmake --yes --version=${ENV:CMAKE_VERSION} --installargs 'ADD_CMAKE_TO_PATH=System'; `
+    choco install pwsh --yes --version=${PWSH_VERSION};
+
+# ------------------------------------------------------------------------------------------------------------
+# Reduce image size
+#
+RUN choco install --yes choco-cleaner;                    `
```
JeanChristopheMorinPerso: If you really want the image size to go down, the cleanup has to be done in the same layer as where the temp files are created.

PR #787, opened 2019-11-06
Fix for windows docker login error

bfloch:
It seems like the windows image composing works correctly, but the subsequent pulls fail to login to docker. That is with the exact same code!

In worst case I can just use the -u -p options directly and rely on github doing the right masking in the logs due to the variable being a secret.

**WIP**
Trying to find a fix that would only affect the windows image. The only difference I can spot is the checkout action being triggered earlier in this workflow.

### DIFF
--- a/.github/workflows/windows-docker.yaml
+++ b/.github/workflows/windows-docker.yaml
@@ -36,12 +36,12 @@ jobs:
       fail-fast: false
 
     steps:
-    - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


bfloch:
Ok I was able to verify that this is an issue with the secrets.DOCKERHUB_TOKEN not being available in this workflow. I'll open a ticket instead since this might be a github actions issue.

PR #791, opened 2019-11-08
Windows ci fail on forked pr

### LINKED ISSUE #790: windows Actions workflow fails on PRs from fork
This is because windows test is docker-based and requires access to secretes. Because forked PRs run on the forker's account, they do not have access to the github secret necessary to access dockerhub.

nerdvegas:
Fixes #790 (sort of)

This is a fairly hacky fix for failing windows tests, but there don't seem to be many good options.

Windows test is failing on forked PRs, because forks don't have access to secrets, and thus `secrets.DOCKERHUB_TOKEN` evals to empty string.

This PR forces the windows test to be a no-op in this situation. It's not ideal, but neither is removing pull-requests as an event, because this _should_ be working for collaborators and it would be a shame to lose that.

The clumsy approach you see wrt setting DOCKERHUB_TOKEN envvar in each step is explained here: https://github.community/t5/GitHub-Actions/If-expression-with-context-variable/td-p/34556. Basically, you can't use `jobs.{jobname}.if` because this doesn't have access to the `secrets` context (why??).

Anyway the aim of this PR is to have windows test in as good a shape as it can be currently. Happy to revert this to something else later. 

ps:

It's also not great because the fork user may have an unrelated DOCKERHUB_TOKEN secret. Chances of that are low, and this is hopefully a semi-temporary fix, so I think we can live with this for now. Unfortunately there's no such thing as a 'user' context in github Actions where you might check user permissions and have an if condition on that :/

### DIFF
--- a/.github/workflows/windows-docker-image.yaml
+++ b/.github/workflows/windows-docker-image.yaml
@@ -22,9 +22,9 @@ on:
   push:
     # Will only trigger building a docker image if any of these changed
     paths:
-    - '.github/docker/rez-win-base/**'
-    - '.github/docker/rez-win-py/**'
-    - '.github/workflows/windows-docker-image.yaml'
+      - '.github/docker/rez-win-base/**'
+      - '.github/docker/rez-win-py/**'
+      - '.github/workflows/windows-docker-image.yaml'
 
 jobs:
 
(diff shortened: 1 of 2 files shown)


bfloch:
Sad to see this happening. Do we actually need to **login** to pull public images from docker hub at all? I am not an expert here but maybe we can get rid of the login all together for the pull.

bfloch:
Let's close this in favor of #794 and #795 
I made it two commits so that we can see if the general workflow would work.
I'll address a better fall back solution soon.

nerdvegas:
Fixed by #795

PR #793, opened 2019-11-08
Fixes #792 cmd empty echo

### LINKED ISSUE #792: cmd handles empty echo incorrectly
With `cmd` any `info('')` call results in `echo` which in `cmd` shows the status of the echo.
To print an empty line `echo.` is needed instead.

Reported by @instinct-vfx, introduced at:
https://github.com/nerdvegas/rez/commit/b277d26d8d76f28c0e05b0a60158b5d03009c96d#diff-b1aa7ab463cecaa539089580fe418dc2R164

bfloch:
Without it `info('')` generates `echo is ON.` on cmd's stdout.

EDIT: More context
In `cmd` and `info('')` results in `echo ""` which is interpreted as command to show the current state of echo.
The way to echo an empty line is `echo.`.

I've also added a test that makes sure that info('') is interpreted as empty line.

### DIFF
--- a/src/rez/tests/test_shells.py
+++ b/src/rez/tests/test_shells.py
@@ -250,6 +250,7 @@ class TestShells(TestBase, TempdirMixin):
             _print("ello")
             _print(literal("ello"))
             _print(expandable("ello"))
+            info('')
             _print("\\")
             _print("\\'")
             _print("\\\"")
@@ -291,6 +292,7 @@ class TestShells(TestBase, TempdirMixin):
             "ello",
             "ello",
             "ello",
+            "",
             "\\",
             "\\'",
             "\\\"",
@@ -331,8 +333,9 @@ class TestShells(TestBase, TempdirMixin):
 
         # We are wrapping all variable outputs in quotes in order to make sure
         # our shell isn't interpreting our output as instructions when echoing
-        # it but this means we need to wrap our expected output as well.
-        expected_output = ['"{}"'.format(o) for o in expected_output]
+        # it but this means we need to wrap our expected output as well. Only
+        # exception is empty string, which is just passed through.
+        expected_output = ['"{}"'.format(o) if o else o for o in expected_output]
 
         _execute_code(_rex_assigning, expected_output)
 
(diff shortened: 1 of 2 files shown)


src/rezplugins/shell/cmd.py
```diff
@@ -77,7 +77,7 @@ def get_syspaths(cls):
 
         # detect system paths using registry
         def gen_expected_regex(parts):
-            whitespace = "[\s]+"
+            whitespace = r"[\s]+"
```
bfloch: Eliminates a warning (invalid escape sequence)

bfloch:
I reran the tests since we did not have the PR workflow fixed back then.
Ready for review @nerdvegas and @instinct-vfx

src/rez/tests/test_shells.py
```diff
@@ -331,8 +333,9 @@ def _print(value):
 
         # We are wrapping all variable outputs in quotes in order to make sure
         # our shell isn't interpreting our output as instructions when echoing
-        # it but this means we need to wrap our expected output as well.
-        expected_output = ['"{}"'.format(o) for o in expected_output]
+        # it but this means we need to wrap our expected output as well. Only
+        # exception is empty string, which is just passed through.
+        expected_output = ['"{}"'.format(o) if o else o for o in expected_output]
```
bfloch: This was necessary since the interpretation of empty string in quotes would not match throughout the shells.
Not sure if this is correct but the data itself is correct (empty) and this is what I wanted to compare here.

instinct-vfx:
I want to take a closer look at the code. While this does strip out the ECHO output it still leaves empty lines for me. Here is a comparison of 2.47.11, 2.47.13 and the current state of this MR.

While the empty lines are no show stopper they are still not needed and i am a bit confused by the difference. Why were the newlines introduced? And why does 2.47.11 not have the "you are now in a rez environment"?

**2.47.11**
image

**2.47.13**
image

**This MR**
image

bfloch:
I addressed the message in a recent fix, I just matched it to the other shells.
https://github.com/nerdvegas/rez/commit/b277d26d8d76f28c0e05b0a60158b5d03009c96d

As to the newline:
`info('')` not meant to generate a newline. Else you would just leave it out.
I think this matches the other shells implementations.
So the MR is conforming to other shells and probably what we want.

bfloch:
Here in cmd:
https://github.com/nerdvegas/rez/blame/master/src/rezplugins/shell/cmd.py#L165
Here in the UnixShells:
https://github.com/nerdvegas/rez/blob/master/src/rez/shells.py#L370

PR #798, opened 2019-11-12
rez.pip: Support python 2 executable on Windows (796)

### LINKED ISSUE #796: find_pip_from_context failing on Windows platform
The recent changes to the rez pip module have broken pip installs on Windows,

https://github.com/nerdvegas/rez/blob/6db191f4aa01903e22955cf38e8342cb8697a499/src/rez/pip.py#L162

When doing a `context.which` call to find the python executable the major version token is being appended to 'python' which on Windows returns None.

This causes an issue with the subsequent `subprocess` call which attempts to find a pip install.

JeanChristopheMorinPerso:
Fixes #796.

On Windows, versions of Python less than 2 doesn't have versioned executable... So I'm basically just adding a check for windows and only append the version when not under windows.

### DIFF
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -18,6 +18,7 @@ from rez.exceptions import BuildError, PackageFamilyNotFoundError, \
 from rez.package_maker__ import make_package
 from rez.config import config
 from rez.system import System
+from rez.utils.platform_ import platform_
 
 from tempfile import mkdtemp
 from pipes import quote
@@ -159,7 +160,12 @@ def find_pip_from_context(python_version, pip_version=None):
         print_debug("No rez package called %s found", target)
         return None, None, None
 
-    py_exe = context.which("python%s" % python_major_minor_ver.trim(1))
+    py_exe_name = "python"
+    if platform_.name != "windows":
+        # Python < 2 on Windows doesn't have versionned executable.
+        py_exe_name += str(python_major_minor_ver.trim(1))
+
+    py_exe = context.which(py_exe_name)
 
     proc = context.execute_command(
         # -E and -s are used to isolate the environment as much as possible.


bfloch:
Another way to deal with it is https://www.python.org/dev/peps/pep-0397/
```
Python Launcher for Windows Version 3.7.5150.1013

usage:
py [launcher-args] [python-args] script [script-args]

Launcher arguments:

-2     : Launch the latest Python 2.x version
-3     : Launch the latest Python 3.x version
-X.Y   : Launch the specified Python version
     The above all default to 64 bit if a matching 64 bit python is present.
-X.Y-32: Launch the specified 32bit Python version
-X-32  : Launch the latest 32bit Python X version
-X.Y-64: Launch the specified 64bit Python version
-X-64  : Launch the latest 64bit Python X version
-0  --list       : List the available pythons
-0p --list-paths : List with paths
```

Not sure if it is guaranteed to be installed but once it is there is solves a bunch of issues.
It also respects `#! shebang` to some degree.

JeanChristopheMorinPerso:
@bfloch thanks for pointing this out. I just don't know if it's available in all Python versions. I think we can go with the current solution as it's quite simple and won't bring us much support I think.

instinct-vfx:
I think python launcher was added with 3.3 but i am not sure. 2.7 does not ship with it.

bfloch:
On windows both Python 2.7 and 3.x will be plain `python.exe`.
If you have and 3.4+ (I believe) installed you will have the launcher.

My thought was that a more reliable detection would be on windows to try `py.exe` and let it handle it. They already seem to have accounted for all the cases: 
https://github.com/python/cpython/blob/master/PC/launcher.c

If it is not available just roll as is - since in this case it is likely you only have python 2.x installed anyway.

I just can't see how `which` + `python.exe` would have a chance to find the right version on Windows. It pretty much depends on the order of installation, no?

nerdvegas:
Happy to go with this until proven otherwise that we need something more complex.

This and some other recent PR merges coming soon.

Cheers all
A

bfloch:
Fair. I have not yet studied `py.exe` thoroughly to give a definite recommendation.

PR #802, opened 2019-11-12
Feature/add 'prefix' argument to rez-pip

predat:
This PR offers the option to customize the installation location of a pip package using rez-pip.

### DIFF
--- a/src/rez/cli/pip.py
+++ b/src/rez/cli/pip.py
@@ -25,8 +25,8 @@ def setup_parser(parser, completions=False):
         help="install as released package; if not set, package is installed "
         "locally only")
     parser.add_argument(
-         "-p", "--install_path", type=str, metavar='PATH',
-         help="install package to a custom location")
+         "-p", "--prefix", type=str, metavar='PATH',
+         help="install to a custom package repository path.")
     parser.add_argument(
         "PACKAGE",
         help="package to install or archive/url to install from")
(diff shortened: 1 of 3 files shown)


src/rez/cli/pip.py
```diff
@@ -24,6 +24,9 @@ def setup_parser(parser, completions=False):
         "-r", "--release", action="store_true",
         help="install as released package; if not set, package is installed "
         "locally only")
+    parser.add_argument(
+         "-p", "--install_path", type=str, metavar='PATH',
```
instinct-vfx: I think it would make sense to keep the arguments consistent across the different CLI tools. rez-build uses `-p` and `--prefix`. Is there any reason to use `--install-path` instead?

predat: You're right @instinct-vfx (done in 262fe07).

JeanChristopheMorinPerso:
@predat, I'm curious to know your use case here. Do you have any example of when this would be useful?

predat:
@JeanChristopheMorinPerso, here are the use cases I think about:
- in my studio, we have a very deep packages structure and the REZ_RELEASE_PACKAGES_PATH variable is not always adequate
- to do some tests
- in case the packages path is dynamic
- in case the packages path is contextual
- because the rez-build command already has this argument
- and I'm probably forgetting some...

PR #804, opened 2019-11-13
Migrate rezgui.qt imports to Qt.py

douglaslassance:
Qt.py is abstracting the various implementations of Qt in Python allowing the rez GUIs to run with PySide2, PyQt5, PySide and PyQt4.


nerdvegas:
Hi Douglas,

I'm happy to merge this PR, but I just want to check that it's been tested fairly extensively? Clearly it's more difficult to have tests cases for UI so I just want to make sure first. If you've been running this without issue then I'm fine with merging.

Thx
A

PR #811, opened 2019-11-29
Rep001 2 hooks

nerdvegas:
Implements pre-install/release running of package tests.

Note that this is not done as a pre-release hook plugin, because the way this works doesn't match up to the places in which these hooks are run.

The way pre-install (and release) tests are run is like so:

* variant (ie its payload) is installed to correct location (package.py is not yet updated)
* separate temp package is created, containing just the installed variant
* this tmp pkg uses a new internal attrib (`_redirected_base`) to point back to the true package install location
* the tests are run on this temp pkg
* if tests fail, true variant payload is deleted
* if tests succeed, package.py is then updated as per usual, and new variant becomes visible.

Note that https://github.com/nerdvegas/rez/pull/807 must be merged before this PR.

### DIFF
--- a/src/rez/package_repository.py
+++ b/src/rez/package_repository.py
@@ -2,7 +2,6 @@ from rez.utils.resources import ResourcePool, ResourceHandle
 from rez.utils.data_utils import cached_property
 from rez.plugin_managers import plugin_manager
 from rez.config import config
-from rez.backport.lru_cache import lru_cache
 from rez.exceptions import ResourceError
 from contextlib import contextmanager
 import threading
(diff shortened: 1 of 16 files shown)


src/rez/package_repository.py
```diff
@@ -353,18 +370,26 @@ def _uid(self):
 class PackageRepositoryManager(object):
```
nerdvegas: Note: These minor changes weren't really needed as part of this PR - this work was done when I started investigating a different direction. In any case, these changes fix a bug, where repositories would've been cached into an LRU pool that's shared across all instances of this class.

nerdvegas:
Merged to https://github.com/nerdvegas/rez/releases/tag/2.51.0

PR #815, opened 2019-12-03
removed odd case of _Bound instantiation with Version

nerdvegas:
Also added some minor optimisations - the results of version range parse can be ifelse'd rather than if'd, since the regex can only match one of the OR'd statements in the regex at any one time.

### DIFF
--- a/src/rez/vendor/version/version.py
+++ b/src/rez/vendor/version/version.py
@@ -567,62 +567,76 @@ class _VersionRangeParser(object):
     # The regular expression for a version - one or more version tokens
     # followed by a non-capturing group of version separator followed by
     # one or more version tokens.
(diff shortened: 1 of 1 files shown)


src/rez/vendor/version/version.py
```diff
@@ -567,62 +567,76 @@ class _VersionRangeParser(object):
...
+    # string to each token class, and that could be used to construct
+    # `version_group` as below. We need to keep a dict of these parser instances,
+    # to avoid recompiling the large regex every time a version range is
+    # instantiated. In the cpp port this would be simpler - VersionRange could
+    # just have a static parser that is instantiated when the version range
+    # template class is instantiated.
+    #
+    version_group = r"([0-9a-zA-Z_]+(?:[.-][0-9a-zA-Z_]+)*)"
+
+    version_range_regex = (
```
nerdvegas: Note: formatting change only, no change to regex here

PR #827, opened 2019-12-18
[Feature] Add rez-pip extra args passthrough

### LINKED ISSUE #816: rez-pip creates .pyc files by default
This is an issue if the installed package is a "pure" package (no binary components) because the pyc files might not work with other interpreters.

My proposal would be to add `--no-compile` to pip.py and not generate pyc files at all when using rez-pip.

lambdaclan:
# Description

closes #816 

Enable users to customize the `pip install` command by explicitly defining additional settings to be passed to *rez-pip* in the form of [extra options](https://pip.pypa.io/en/stable/reference/pip_install/#options). Extra arguments can be defined either by using the new *rez-pip* cli argument `[-e/--extra..]` or by adding items to the new *rezconfig* entry `pip_extra_args` in the form of:

```
pip_extra_args: ["--no-compile", "--prefer-binary"]
```

Implementation is based on the ticket above and the discussion between JC, Thorsten and Allan up on slack.

## Breakdown

* Add new rez-pip argument -e/--extra to pass additional options to pip install command
* Additional options can be defined in a rezconfig file entry pip_extra_args to ensure they are always applied
* The command line argument -e/--extra overrides the configuration file options (if any)
* Repeating options automatically override previous one by pip (without extra processing) only last one counts
* Invalid options are handled by pip (no extra care was taken to prevent them)

## Type of change

## Possible issues

As discussed [below](https://github.com/nerdvegas/rez/pull/827#discussion_r363997813) regarding repeating options since through extra args the user can override the target and use pep517 options which are prerequisites for correct package install we will need to handle that scenario somehow.

The way I see it there are two methods:

~a. Always ignore the target option and anything that can affect use pep517 such as no-binary etc~
~b. Have a list of allowed options that can be handled by extra args (filter out invalid ones)~

**Edit:**

It has been agreed to assume user responsibility for overriding the wrong options. It needs to be made clear that extra args adds to or overrides the pre-configured args.

# How Has This Been Tested?

* Tried defining multiple option combinations between CLI and rezconfig
* I will look into the test-suite and see if I can add tests if they already exist for other commands

**Test Configuration**:
* Python version: 2.7.17, 3.8
* OS: Linux, Windows 10
* Toolchain: rez 2.50.0, pip 19.3.1

## Demo

* rez-pip cli -e --extra argument

pip_cli_extra_args_demo

* rez-pip cli extra args via rezconfig

pip_cli_extra_args_conf_demo

## Point release
[Source](https://github.com/nerdvegas/rez/tree/2.50.0) | [Diff](https://github.com/nerdvegas/rez/compare/master...lambdaclan:feature/rez-pip-extra-args-passthrough?expand=1)

**Notes**

Add rez-pip extra args passthrough

**Merged pull requests:**

- [Feature] Add rez-pip extra args passthrough [\#827](https://github.com/nerdvegas/rez/pull/827) ([lambdaclan](https://github.com/lambdaclan))

**Closed issues:**

- rez-pip creates .pyc files by default [\#816](https://github.com/nerdvegas/rez/issues/816)

### DIFF
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -251,17 +251,29 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
         context.print_info(buf)
         _log(buf.getvalue())
 
+    # preconfigured options
+    pre_configured = ["--use-pep517", "--target=%s" % destpath]
+
     # gather additional pip install options
     pip_extra_opts = extra_args if extra_args else config.pip_extra_args
 
+    # make sure that the target option is overridden if supplied
+    if any("--target=" in opt for opt in pip_extra_opts):
+        # extra args redefines the target option so ignore pre-configured one
+        pre_configured = pre_configured[0:1]
+    # disable pep517 if forced
+    if "--no-use-pep517" in pip_extra_opts:
+        pre_configured = []
+
+    # keep unique opts
+    opts = list(set(pre_configured + pip_extra_opts))
+
     # Build pip commandline
     cmd = [
-        py_exe, "-m", "pip", "install",
-        "--use-pep517",
-        "--target=%s" % destpath
+        py_exe, "-m", "pip", "install"
     ]
 
-    cmd.extend(pip_extra_opts)
+    cmd.extend(opts)
 
     if mode == InstallMode.no_deps:
         cmd.append("--no-deps")
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -198,7 +198,8 @@ def find_pip_from_context(python_version, pip_version=None):
 
 
 def pip_install_package(source_name, pip_version=None, python_version=None,
-                        mode=InstallMode.min_deps, release=False, prefix=None):
+                        mode=InstallMode.min_deps, release=False, prefix=None,
+                        extra_args=None):
     """Install a pip-compatible python package as a rez package.
     Args:
         source_name (str): Name of package or archive/url containing the pip
@@ -212,6 +213,7 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
             managed.
         release (bool): If True, install as a released package; otherwise, it
             will be installed as a local package.
+        extra_args (List[str]): Additional options to the pip install command.
 
     Returns:
         2-tuple:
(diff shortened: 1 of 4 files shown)


src/rez/pip.py
```diff
@@ -249,13 +251,18 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
...
         _log(buf.getvalue())
 
+    # gather additional pip install options
+    pip_extra_opts = extra_args if extra_args else config.pip_extra_args
+
     # Build pip commandline
     cmd = [
         py_exe, "-m", "pip", "install",
         "--use-pep517",
         "--target=%s" % destpath
```
nebukadhezer: What happens if someone specifies --target as an extra arg ? Or --user ? Should that be filtered out ?

lambdaclan: Hello Johanness, as mentioned in the PR description:

> Repeating options are automatically ignored by pip (without extra processing)

I also thought that some sort of filtering will be required but after a few tests I confirmed (via the --target option) that any repeating options simply override previous ones, only the last one is taken into account.

Saying that, allowing overriding target can break the installation so I guess it is better to always ignore it same goes for --no-binary since we want pep517 to be always on. That is the problem of allowing extra args, the user can break the rez package install if they override certain options.

*I propose that we have a list of allowed options to be specified and only handle those.* I am waiting for Allan's feedback on this.

Thank you for pointing it out though, I will update the PR info to make it clear.

JeanChristopheMorinPerso: >I propose that we have a list of allowed options to be specified and only handle those.

When we discussed about it, the conclusion was that if someone uses the extra args, he must know what he is doing. At least from what I remember. Instead of having a list of accepted arguments, I would more document which argument should not be specified in the extra args.

lambdaclan: Hell JC, thank you very much for your input. Any way is fine with me. Actually documenting it somewhere sounds like a good idea. I could even add it to the help text if it makes sense. I will see how I can make it apparent of what needs to be avoided when using extra args. A warning can also be shown if target or pep517 are overridden.

nerdvegas:
+1 for putting onus on the user, if they're using this then they should
know what they're doing.

It also needs to be clear if specifying the extra args _adds_ to the
configured extra args, or overwrites them (should overwrite imo).

Thx
A

lambdaclan:
Hello @nerdvegas when you get a chance can you please give me an update on the status of this PR? I did the changes as discussed above so if there are no other stoppers maybe we can merge?

Thank you :pray:

nerdvegas:
Hey Ira,

No problem will get this merged really soon.

Thx
A

PR #834, opened 2020-01-11
added new env vars - REZ_SHELL_INIT_TIMESTAMP, REZ_SHELL_INTERACTIVE

### LINKED ISSUE #833: add env-var to record shell init time
I have a case where I need to know the time when a tool was launched, which is done via `ResolvedContext.execute_shell` (eg, when you run `rez-env blah -- mycommand`). I would like to write this to an env-var.

The existing `REZ_USED_TIMESTAMP` (https://github.com/nerdvegas/rez/wiki/Environment-Variables#context-environment-variables) is similar to what I need, but not quite right. REZ_USED_TIMESTAMP is set to `--timestamp` if present, rather than the true time of resolve. Furthermore, it's time the context was _resolved_, not the time that a command was run, which can occur later (a context can be resolved once and used many times after - eg, suites).

I propose adding a new `REZ_SHELL_INIT_TIMESTAMP` env-var, which is set to epoch time every time a shell is created (via `execute_shell`).

Related - it would also be useful to set a var that indicates whether a command was run or not (ie `command` arg in `execute_shell`). The reason - I need to determine tool startup time, and if command is None, this generally means that an interactive shell was started, and I need to filter out those cases.

nerdvegas:
Closes #833

### DIFF
--- a/src/rez/resolved_context.py
+++ b/src/rez/resolved_context.py
@@ -1240,7 +1240,7 @@ class ResolvedContext(object):
         context_file = context_filepath or \
             os.path.join(tmpdir, "context.%s" % sh.file_extension())
 
-        # interpret this context and write out the native context file
+        # interpret this context
         executor = self._create_executor(sh, parent_environ)
         executor.env.REZ_RXT_FILE = rxt_file
         executor.env.REZ_CONTEXT_FILE = context_file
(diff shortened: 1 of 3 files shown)


maxnbk:
the variables are being generated correctly, at the very least
(whether that solves the reason we went down this road, I can't be totally sure yet, as we discussed, but the PR does no harm at the worst, so I think this is fine to merge if desired)

PR #845, opened 2020-02-01
Install as package part2

nerdvegas:
Working off https://github.com/nerdvegas/rez/pull/579, but with following changes (@j0yu ):

* Use of PackageMaker code in install.py was too fragile. Only a few rez modules are imported here, because rez isn't built yet. Importing PackageMaker greatly increases the amount of code we import in the installer, which means that any changes (like compiled source - which is coming) will probably break it. I've taken basically the same approach, but moved that functionality into the installation instead and done it that way.
* Removed cli tools from the rez package. I don't want to encourage that use case - the production install is there for that, the rez package is intended for using the rez API.


j0yu:
Looks good, maybe you can add some Ci/tests to it, like the ones from #579 (which I could close in favour of this one)

nerdvegas:
Hey Joe,

Yeah I agree. It's particularly good to have the tests run via the rez
package, as this will pick up any inconsistencies between that and the
production installation.

There's two things I think we need to address first though:
- the python rez-bind. I'd really like to get rid of this. I wonder if we
should create the rez package in a special mode (for this workflow) that
ditches the python requirement. The end result would be the same (since the
rez-bind interfaces with system python anyway) but without the bind stuff
(which we want to deprecate). We could achieve this by having an env-var
that the workflow sets, which then causes the created rez package to drop
the python requirement.
- the rez package has no cli tools (by design as mentioned in PR), so we
need to expose a way to run the tests via API, but that seems reasonable to
do anyway and should be trivial.

What are your thoughts? What I'm thinking is we put your PR "ON HOLD", and
ticket up something to add the changes as outlined above, and I'l merge
this new PR in the meantime. I want to keep your workflow around for
reference until we sort this out.

Cheers
A

install.py
```diff
@@ -98,26 +101,117 @@ def copy_completion_scripts(dest_dir):
...
+    Args:
+        repo_path (str): Full path to the package repository to install into.
+    """
+    from tempfile import mkdtemp
+
+    # do a temp production (venv-based) rez install
+    tmpdir = mkdtemp(prefix="rez-install-")
+    install(tmpdir)
+
+    try:
```
j0yu: Maybe this temp install setup can be turned into a context so it can be re-used in the future:

<details><summary>temp_install context</summary>

```python
import contextlib


@contextlib.contextmanager
def temp_install(cleanup=True, print_welcome=False):
    """Do a temp production (venv-based) rez install.
    
    Args:
        cleanup (bool): Cleanup temp install when exiting context.
    """
    from tempfile import mkdtemp

    tmpdir = mkdtemp(prefix="rez-install-")
    install(tmpdir, print_welcome=print_welcome)

    try:
        yield tmpdir
    finally:
        if cleanup:
            try:
                shutil.rmtree(tmpdir)
            except Exception:
                pass


def install_as_rez_package(repo_path):
    """Installs rez as a rez package.
    
    Note that this can be used to install new variants of rez into an existing
    rez package (you may require multiple platform installations for example).
    
    Args:
        repo_path (str): Full path to the package repository to install into.
    """
    with temp_install() as tmpdir:
        # This extracts a rez package from the installation. See
        # rez.utils.installer.install_as_rez_package for more details.
        #
        args = (
            os.path.join(tmpdir, "bin", "python"), "-E", "-c",
            r"from rez.utils.installer import install_as_rez_package;"
            r"install_as_rez_package('%s')" % repo_path
        )
        print(subprocess.check_output(args))
```
</details>

PR #862, opened 2020-03-04
Add case for ../../include/... files

### LINKED ISSUE #861: rez-pip - no case for ../../include/... file
I found an issue with rez-pip trying to install greenlet.
This is due to a similar case as here:
https://github.com/nerdvegas/rez/blob/49cae49a9dd4376b9efb6d571713b716d315b32b/src/rez/pip.py#L426-L432

We need to check for packages reporting files as ../../include/... as well.

To replicate:
`rez-pip -i greenlet`

smaragden:
fixes #861


smaragden:
This is obviously more complicated than this. Will reopen when i have figured it out.
See #861 for more info

PR #866, opened 2020-03-22
pip install path remap

### LINKED ISSUE #861: rez-pip - no case for ../../include/... file

j0yu:
Closes #861 
Closes #863

New configuration setting to allow people to map and configure `rez-pip --install` of unknown `RECORD` paths.

For example, for #861, a new entry like so can be added to capture and re-map and move all `../../*` paths that are unknown/not handled by `rez-pip` to just live under the root rez package installation like so:

```python
# Inside rezconfig.py...
pip_install_remaps = [

    # ...add these before the end

    # # Fix for greenlet in https://github.com/nerdvegas/rez/issues/861
    # Path in record          | pip installed to    | copy to rez destination
    # ------------------------|---------------------|--------------------------
    # usr/local/*             | *                   | python/*
    {
        "record_path": r"(.*){sep}usr{sep}local{sep}(.*)",
        "pip_install": r"\1{sep}usr{sep}local{sep}\2",
        "rez_install": r"\2",
    },

    # # Fix for Qt.py in https://github.com/nerdvegas/rez/issues/861
    # Path in record          | pip installed to    | copy to rez destination
    # ------------------------|---------------------|--------------------------
    # ../../NOT PARDIR        | *                   | *
    {
        "record_path": r"^{p}{s}{p}{s}(?!{p})(.*)",
        "pip_install": r"\1",
        "rez_install": r"\1",
    },
]
```

@smaragden, want to give these a test?

----

### Added

- `rez.config.PipInstallRemaps` setting to extend `validate` to perform token substitutions for...
- ...new `rez.rezconfig.pip_install_remaps` setting

### Changed

- More detailed error message for unknown source file
- Tidied imports in `rez.pip`

### DIFF
--- a/src/rez/config.py
+++ b/src/rez/config.py
@@ -125,23 +125,19 @@ class StrList(Setting):
 class PipInstallRemaps(Setting):
     """Ordered, pip install remappings."""
     PARDIR, SEP = map(re.escape, (os.pardir, os.sep))
-    TOKENS = {'sep': SEP, 's': SEP, 'pardir': PARDIR, 'p': PARDIR}
+    RE_TOKENS = {'sep': SEP, 's': SEP, 'pardir': PARDIR, 'p': PARDIR}
+    TOKENS = {'sep': os.sep, 's': os.sep, 'pardir': os.pardir, 'p': os.pardir}
+    KEYS = ["record_path", "pip_install", "rez_install"]
 
-    schema = Schema(
-        [
-            {
-                "record_path": And(str, len),
-                "pip_install": And(str, len),
-                "rez_install": And(str, len),
-            }
-        ]
-    )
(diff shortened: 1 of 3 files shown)


zachlewis:
Thanks @j0yu, I think this is pretty effective. The error message with the detailed instructions is really helpful, and the tokens are quite nice. This helped me rez-pip install ipykernel, which was having similar trouble with stuff in '../../share/*'

nerdvegas:
Real clean @j0yu cheers,
A

Issue #871, opened 2020-04-02
override git plugin config in package.py


alexxbb:
What is the proper way of overriding **allow_no_upstream** config property in package.py?

I found [this code](https://github.com/nerdvegas/rez/blob/9aaf75c16a2705d50b44e3ef08e6737b2066357d/src/rezplugins/release_vcs/git.py#L103)

And tried various ways with no success:
```python
with scope("config") as c:
    c.plugins.release_vcs.git.allow_no_upstream = True
```
```python
# found this in some of the test files, it fails with an error
with scope("config") as c:
    c.override("plugins.release_vcs.git.allow_no_upstream", True)
```

```python
with scope("plugins"):
  with scope("release_vcs"):
    with scope("git"):
      allow_no_upstream = True
```

nerdvegas:
Hey, I can't repro this unfortunately, it's working for me (using the first form listed). What os/rez-version? What was the specific error message you got?

alexxbb:
Hi Allan, 
I'm getting this error:
```
ReleaseVCSError: command failed: /usr/bin/git rev-parse --abbrev-ref --symbolic-full-name '@{u}'
fatal: HEAD does not point to a branch
```
```rez-release --skip-repo-errors``` works, I thought I keep the settings within the package.py
This is needed for a release with a gitlab CI job, it checks out a tag, and therefore the HEAD is missing.
```
>> rez --version
Rez 2.53.1
>>python --version
Python 2.7.5
```
```
CentoOS 7
```
We have REZ_CONFIG_FILE point to our studio config, and there is a plugins dictionary with just the emailer plugin configured, I tried this too, but got the same error.
```
env REZ_CONFIG_FILE="" rez-release
```

```python
# -*- coding: utf-8 -*-
name = '...'
version = '1.6.12'
authors = ['arusev <arusev@imageworks.com>']

requires = []
variants = [['houdini-18.0', 'python-2']]
private_build_requires = ['fxBuildTools-4']

with scope("config") as c:
    c.release_packages_path = "/shots/fx/home/software/packages"
    c.plugins.release_hook.emailer.recipients = []
    c.plugins.release_vcs.git.allow_no_upstream = True

def commands():
    env.HOUDINI_PATH.prepend("{root}/houdini")
    env.PATH.prepend("{root}/bin")
    env.PYTHONPATH.prepend("{root}/python")


uuid = '173c976d-43f4-460f-924b-0abbb41a04e1'

```

[...]

PR #872, opened 2020-04-03
Fix for git rev-parse error out before checking for allow_no_upstream

alexxbb:
Fixed an issue where git rev-parse would error out before checking the config.allow_no_upstream.

I built locally and was able to release without issues


alexxbb:
Fixes #871

PR #878, opened 2020-04-18
Fix issue 826 - correct python and pip fallback

### LINKED ISSUE #826: rez-pip issues finding pip executable
Two related issues:

## 1
If I have a python package that provides a 'python' exe but not (eg) 
python2', rez can end up using the wrong py exe:

```
15:22:51 INFO     Found pip-8.1.1 inside /home/ajohns/packages/python/2.7.12/package.py. Will use it with /usr/bin/python2
```

## 2
rez-pip finds pip in the following order (from docstring):
        1. Search for rezified python matching python version request;
        2. If found, test if pip is present;
        3. If pip is present, use it;
        4. If not present, search for rezified pip (this is for backwards compatibility);
        5. If rezified pip is found, use it;
        6. If not, fall back to rez's python installation.

However, if pip is found, but its version is <19, then rather than continue on with the search for a valid pip, it errors out with:
```
rez.exceptions.RezSystemError: pip >= 19 is required! Please update your pip.
```

I have a case where the rez installation itself has a valid pip, but rez-pip will not work because it's found a rezified pip that is <v19.

## TODO
* Check for both `python` and `python2` available from the context (if pip is being used from a context), rather than assuming that 'python2' will be supplied (on non-windows). It would probably be enough to check that the exe is within the python/pip rez package's _root_. Given this change, I am not sure we would require checking for Windows anymore.
* Continue in the search order if a <v19 pip is found
* Improve logging to explain steps being taken.

j0yu:
[Installation](https://github.com/j0yu/bleeding-rez/actions?query=workflow%3AInstallation+branch%3Afix-826-pip-fallback) Closes #826

Ran across #826 while trying to refactoring pip wiki fix PR #749. The CI for #749 is also added here since it requires the pip fallback fix

1. [x] Use `python2` or `python3` rather than `python`

   [Currently with disgusting hack][hack] and thinking about either (need feedback!):

      - a new `add_sys_path` kwargs for [`ResolvedContext.get_environ`][get_environ] and [`ResovledContext._execute`][exe], or 
      - a new `ResolvedContext.append_sys_path` attribute which [`ResovledContext._execute`][exe] and any other methods can use

2. [x] Fallback to more compatible `pip` version if possible

[hack]: https://github.com/nerdvegas/rez/blob/ce3c9da8fda027e5182b06650df21e7b7cbd2bdc/src/rez/pip.py#L122-L186
[get_environ]: https://github.com/nerdvegas/rez/blob/ce3c9da8fda027e5182b06650df21e7b7cbd2bdc/src/rez/resolved_context.py#L932
[exe]: https://github.com/nerdvegas/rez/blob/ce3c9da8fda027e5182b06650df21e7b7cbd2bdc/src/rez/resolved_context.py#L1599

### DIFF
--- a/.github/workflows/installation.yml
+++ b/.github/workflows/installation.yml
@@ -33,7 +33,7 @@ jobs:
       run: |
         ${{ matrix.method }}
         find /opt/rez/ -maxdepth 2
-    
+
     - name: rez status
       run: |
         export ${{ matrix.exports }}
@@ -44,14 +44,14 @@ jobs:
         export ${{ matrix.exports }}
         rez-pip --install .
         rez view rez
-    
+
     - name: Import rez package in Python
       run: |
         export ${{ matrix.exports }}
 
         # Still needed as there's no fallback to use system's python
-        rez bind python  
-        
+        rez bind python
+
         echo "Checking rez as python package: ==========================================="
         rez env rez -- python -c 'import rez;print(rez.__file__)'
 
@@ -87,7 +87,7 @@ jobs:
       run: |
         ${{ matrix.method }}
         find /opt/rez/ -maxdepth 2
-    
+
     - name: rez status
       run: |
         export ${{ matrix.exports }}
@@ -98,13 +98,13 @@ jobs:
         export ${{ matrix.exports }}
         rez-pip --install .
         rez view rez
-    
+
     - name: Import rez package in Python
       run: |
         export ${{ matrix.exports }}
 
         # Still needed as there's no fallback to use system's python
-        rez bind python  
-        
+        rez bind python
+
         echo "Checking rez as python package: ==========================================="
         rez env rez -- python -c 'import rez;print(rez.__file__)'
--- a/.github/workflows/installation.yml
+++ b/.github/workflows/installation.yml
@@ -1,0 +1,110 @@
+name: Installation
+on: [push]
+
+jobs:
+  linux-vm:
+    name: ${{ matrix.python-version }} - ${{ matrix.method }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version:
+          - '2.7'
+          - '3.6'
+          - '3.7'
+        method:
+          - 'python ./install.py'
+          - 'pip install --target /opt/rez .'
+        include:
+          - method: 'python ./install.py'
+            exports: 'PATH=${PATH}:/opt/rez/bin/rez'
+          - method: 'pip install --target /opt/rez .'
+            exports: 'PATH=${PATH}:/opt/rez/bin PYTHONPATH=${PYTHONPATH}:/opt/rez'
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@master
+
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install
+      run: |
+        ${{ matrix.method }}
+        find /opt/rez/ -maxdepth 2
+    
+    - name: rez status
+      run: |
+        export ${{ matrix.exports }}
+        rez status
+
+    - name: rez-pip --install .
+      run: |
+        export ${{ matrix.exports }}
+        rez-pip --install .
+        rez view rez
+    
+    - name: Import rez package in Python
+      run: |
+        export ${{ matrix.exports }}
+
+        # Still needed as there's no fallback to use system's python
+        rez bind python  
+        
+        echo "Checking rez as python package: ==========================================="
+        rez env rez -- python -c 'import rez;print(rez.__file__)'
+
+  linux-containers:
+    name: python:${{ matrix.python-version }} - ${{ matrix.method }}
+    runs-on: ubuntu-latest
+    container: python:${{ matrix.python-version }}
+    strategy:
+      matrix:
+        python-version:
+          - '2.7'
+          - '3.6'
+          - '3.7'
+        method:
+          - 'python ./install.py'
+          - 'pip install --target /opt/rez .'
+        include:
+          - method: 'python ./install.py'
+            exports: 'PATH=${PATH}:/opt/rez/bin/rez'
+          - method: 'pip install --target /opt/rez .'
+            exports: 'PATH=${PATH}:/opt/rez/bin PYTHONPATH=${PYTHONPATH}:/opt/rez'
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@master
+
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install
+      run: |
+        ${{ matrix.method }}
+        find /opt/rez/ -maxdepth 2
+    
+    - name: rez status
+      run: |
+        export ${{ matrix.exports }}
+        rez status
+
+    - name: rez-pip --install .
+      run: |
+        export ${{ matrix.exports }}
+        rez-pip --install .
+        rez view rez
+    
+    - name: Import rez package in Python
+      run: |
+        export ${{ matrix.exports }}
+
+        # Still needed as there's no fallback to use system's python
+        rez bind python  
+        
+        echo "Checking rez as python package: ==========================================="
+        rez env rez -- python -c 'import rez;print(rez.__file__)'
--- a/.github/workflows/installation.yml
+++ b/.github/workflows/installation.yml
@@ -9,8 +9,8 @@ jobs:
       matrix:
         python-version:
           - '2.7'
-          - '3.6'
-          - '3.7'
+          # - '3.6'
+          # - '3.7'
         os:
           - windows
           - macos
@@ -19,7 +19,7 @@ jobs:
             method: 'python .\install.py "C:\Program Files\rez"'
             check-parent: 'dir "C:\Program Files"'
           - os: macos
-            method: 'python ./install.py /Applications/Utilities/rez'
+            method: 'sudo python ./install.py /Applications/Utilities/rez'
             check-parent: 'ls -lah /Applications/Utilities'
 
     steps:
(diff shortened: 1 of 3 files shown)


src/rez/pip.py
```diff
@@ -81,35 +86,106 @@ def find_pip(pip_version=None, python_version=None):
...
-        python_version,
-        pip_version=pip_version
-    )
-
-    if not py_exe:
-        py_exe, pip_version, context = find_pip_from_context(
-            python_version,
-            pip_version=pip_version or "latest"
-        )
+    for version in [pip_version, pip_version or "latest"]:
```
nerdvegas, line 91: won't this attempt to find the same pip version twice, when `pip_version` is specified?

src/rez/pip.py
```diff
@@ -81,35 +86,106 @@ def find_pip(pip_version=None, python_version=None):
...
+            py_exe, found_pip_version, context = find_pip_from_context(
+                python_version, pip_version=version
+            )
+            if _check_found(py_exe, found_pip_version):
+                break
+        except BuildError as error:
+            print_warning(str(error))
 
-    if not py_exe:
+    else:  # No break from: for version in [pip_version,...]
```
nerdvegas, line 101: I can't believe I've been writing python this long and have never seen this. There's nothing technically wrong with this, but tbh I'd avoid it because it's so esoteric.

src/rez/pip.py
```diff
@@ -81,35 +86,106 @@ def find_pip(pip_version=None, python_version=None):
...
+
+    # 1.
+    py_packages = (v for v in context.resolved_packages if v.name == name)
+    python_package = next(py_packages)
+
+    # Filthy hack to not have system paths added in ResovledContext._execute
+    # by RexExecutor.append_system_paths(), which calls Shell.get_syspaths()
+    shell_class = create_shell().__class__
+    current_syspaths = shell_class.syspaths
+    try:
```
nerdvegas, line 154: Yeah I don't think we should do this. Not because it's a hack per se, but because it isn't threadsafe. Ie this could alter syspaths in a context being created in another thread.

nerdvegas:
So wrt find_python_from_package:

1. Why recreate a context containing just python, rather than possibly [python, pip]? The pip package isn't going to provide a python exe so this seems unnecessary?

2. In terms of what to do about avoid system paths. We could fix this by providing some arg to execute_shell/etc like `no_append_sys_paths`. However it's not the right fix in the long run, when we (hopefully) have the more granular env-var control in place - once that's there, you should be able to do an execute_shell and redefine how any given var behaves anyway (eg inherits original value etc).

2. is basically what you'er suggesting (your 1.) but I think no_append_sys_paths is a bit better naming since no_append_sys_paths=False is definitely default behaviour. I would go with this. Definitely not with the hack in place currently, see comments.

Cheers
A

src/rez/pip.py
j0yu: True, I also find it fairly susceptible to indent/whitespace messing up logic if code was reformatted/refactored incorrectly hence the comment. I'll see if I can change it up a bit

j0yu:
> 1. Why recreate a context containing just python, rather than possibly [python, pip]? The pip package isn't going to provide a python exe so this seems unnecessary?

Good point, the logic can be much simpler now too!

>  but I think no_append_sys_paths is a bit better naming since `no_append_sys_paths = False` is definitely default behaviour. I would go with this. 

Cool, I'll go with 
 - a new `ResolvedContext.non_append_sys_path` attribute which [`ResovledContext._execute`][exe] and any other methods can use

[hack]: https://github.com/nerdvegas/rez/blob/ce3c9da8fda027e5182b06650df21e7b7cbd2bdc/src/rez/pip.py#L122-L186
[get_environ]: https://github.com/nerdvegas/rez/blob/ce3c9da8fda027e5182b06650df21e7b7cbd2bdc/src/rez/resolved_context.py#L932
[exe]: https://github.com/nerdvegas/rez/blob/ce3c9da8fda027e5182b06650df21e7b7cbd2bdc/src/rez/resolved_context.py#L1599

j0yu:
Actually, I still prefer the name without the `no_` prefix to avoid ~~DNEG~~ double negatives like in your example and also, to avoid statements like `if not no_append_sys_path:`

.github/workflows/installation.yml
```diff
@@ -0,0 +1,110 @@
...
+        # Still needed as there's no fallback to use system's python
+        rez bind python
+
+        echo "Checking rez as python package: ==========================================="
+        rez env rez -- python -c 'import rez;print(rez.__file__)'
+
+  linux-containers:
+    name: python:${{ matrix.python-version }} - ${{ matrix.method }}
+    runs-on: ubuntu-latest
+    container: python:${{ matrix.python-version }}
```
JeanChristopheMorinPerso, line 61: That runs the test in a docker container right? What's the reason for doing this? I can't think of a case where install rez would fail in a container but not on a bare metal/vm machine.

.github/workflows/installation.yml
```diff
@@ -0,0 +1,110 @@
...
+
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install
+      run: |
+        ${{ matrix.method }}
+        find /opt/rez/ -maxdepth 2
```
JeanChristopheMorinPerso, line 35: I'm tempted to suggest to use [check-manifest](https://pypi.org/project/check-manifest/) to test that pip would install the correct files. Though, we would probably have to work a little bit on the `MANIFEST.in` to make it work. But that should be done in another PR IMO if we ever decide to go that way.

.github/workflows/installation.yml
```diff
@@ -0,0 +1,110 @@
...
+    name: ${{ matrix.python-version }} - ${{ matrix.method }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version:
+          - '2.7'
+          - '3.6'
+          - '3.7'
+        method:
+          - 'python ./install.py'
```
JeanChristopheMorinPerso, line 15: We technically already test this type of install quite a lot for all platforms and all python versions. I think it would make sense to test this here, but only if we test `install.py` more in depth, like make sure it installs the rez scripts in the `rez` folder correctly, and on all platforms. We should also test https://github.com/nerdvegas/rez/blob/master/install.py#L81.

Without testing these things, I have the feeling we will just have jobs that will take CPU time without much benefits.

What do you think?

.github/workflows/installation.yml
j0yu: yeah, sounds like a good idea. Let's get a new PR/issue opened for that

.github/workflows/installation.yml
j0yu: I was just double checking that if the system already has Python it wasn't doing anything different from using [actions/setup-python](https://github.com/actions/setup-python)

It didn't, [see here](https://github.com/j0yu/bleeding-rez/runs/621905284?check_suite_focus=true), so I'll remove it

.github/workflows/installation.yml
j0yu: I agree with testing `python -E`, let's do it in a separate PR. Originally these tests were written just to double check what is said in the [Installation Wiki page](https://github.com/nerdvegas/rez/wiki/Installation) actually holds true and is [not misleading beginners](https://github.com/nerdvegas/rez/pull/749).

So far, this is just a start and we can expand on it to more in-depth and [multi-platform tests](https://github.com/j0yu/bleeding-rez/runs/622011791?check_suite_focus=true) in the future to ensure whenever someone discovers this project and tries to follow the wiki step by step, it should _just work_.

src/rez/pip.py
```diff
@@ -81,35 +86,73 @@ def find_pip(pip_version=None, python_version=None):
...
+        context (ResolvedContext): Resolved context with Python and pip.
+        name (str): Name of the package for Python instead of "python".
+        default (str): Force a particular fallback path for Python executable.
+
+    Returns:
+        str or None: Path to Python executable, if any.
+    """
+    py_exe_path = default
+    context = context.copy()
+    context.append_sys_path = False  # GitHub nerdvegas/rez/pull/826
```
nerdvegas, line 137: confusing comment: 826 is an issue not a pr

src/rez/pip.py
```diff
@@ -81,35 +86,73 @@ def find_pip(pip_version=None, python_version=None):
...
+
+    py_packages = (v for v in context.resolved_packages if v.name == name)
+    python_package = next(py_packages)
+
+    if platform_.name == "windows" and python_package.version < Version("3"):
+        py_exe_path = context.which("python")  # GitHub nerdvegas/rez/pull/798
```
nerdvegas, line 143: confusing comment: 798 is an issue not a pr

nerdvegas:
Ok cheers, pretty much ready to merge this. Ran out of time just now but this will be done soon.
Thx
A

ps - let's keep separate issues (eg installation CI) in separate PRs ;) The smaller the PR the easier it is for me to digest and merge it.

src/rez/pip.py
j0yu: https://github.com/nerdvegas/rez/pull/798 is a pr which fixes issue https://github.com/nerdvegas/rez/issues/796

j0yu:
Cheers for the review! I've updated with changes to code comments. 

> ps - let's keep separate issues (eg installation CI) in separate PRs ;) The smaller the PR the easier it is for me to digest and merge it.

Sure thing, I'll keep that in mind next time!

src/rez/pip.py
```diff
@@ -81,35 +86,73 @@ def find_pip(pip_version=None, python_version=None):
...
+        str or None: Path to Python executable, if any.
+    """
+    py_exe_path = default
+    context = context.copy()
+    context.append_sys_path = False  # GitHub nerdvegas/rez/issue/826
+
+    py_packages = (v for v in context.resolved_packages if v.name == name)
+    python_package = next(py_packages)
+
+    if platform_.name == "windows" and python_package.version < Version("3"):
```
nerdvegas, line 142: I don't think we need this special case anymore? You've removed system paths from the context, so we should be able to safely search (py3.7, py3, py) as per code block below, regardless of win/non-win.

I'm keen to get this merged so I'll just change this myself on the merge, but let me know if you see an issue with this thx.

j0yu: Sounds good to me

PR #880, opened 2020-04-19
Fix ISSUE-879: AttributeError: 'Namespace' object has no attribute 'func'

rfletchr:
This merge handles the attribute error raised when trying to get `func` when no sub parser has been selected.


rfletchr:
thanks for getting that merged so fast @nerdvegas

PR #882, opened 2020-04-27
Initial multi-platform install CI

j0yu:
Awaiting for https://github.com/nerdvegas/rez/pull/878 and https://github.com/nerdvegas/rez/pull/749

To Do:
- [ ] Update wiki page to match CI result

See also:
- [Basics of OS X folder structure](https://www.dummies.com/computers/macs/mac-operating-systems/basics-of-the-os-x-folder-structure/)
- [Mac OS X directory structure explained](https://osxdaily.com/2007/03/30/mac-os-x-directory-structure-explained/)
- [AppData folder in Windows](https://www.howtogeek.com/318177/what-is-the-appdata-folder-in-windows/)

### DIFF
--- a/.github/workflows/installation.yml
+++ b/.github/workflows/installation.yml
@@ -2,9 +2,38 @@ name: Installation
 on: [push]
 
 jobs:
+  os-vm: 
+    name: ${{ matrix.python-version }} - ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}-latest
+    strategy: 
+      matrix:
+        python-version:
+          - '2.7'
+          # - '3.6'
+          # - '3.7'
+        os:
+          - windows
+          - macos
+        include:
+          - os: windows
+            method: 'python .\install.py "C:\Program Files\rez"'
+            check-parent: 'dir "C:\Program Files"'
+          - os: macos
+            method: 'sudo python ./install.py /Applications/Utilities/rez'
+            check-parent: 'ls -lah /Applications/Utilities'
+
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions/setup-python@v1
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/installation.yml
```diff
@@ -0,0 +1,82 @@
...
+
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - run: ${{ matrix.check-parent }}
+      - run: ${{ matrix.method }}
+
+  linux-vm:
```
nerdvegas, line 33: I think we need to split this into two jobs - one for install.py-based installation, and another for pip-based install. The main reason is because I don't want to promote any use of cli tools in the pip-installed rez, in fact I think the tools should be removed in that case.

So what should happen currently imo is:

install.py job:
- existing steps, unchanged

pip install job:
- do a pip-install, but standard (no --target /opt/rez)
- do a python -c 'import rez;print(rez.__file__)' test
- do a 'rez-pip --install .' equivalent via API
- test the resulting rez package via API (ie create a ResolvedContext, run a 'rez-status' or similar in it)

Furthermore, I think we should remove the need for the rez-bind python, like so:
- set an internal env-var for this CI, perhaps `_REZ_CI_RUNNING`
- have rez-pip drop the python requirement if it's installing the rez package

This would remove the requirement of a rez python package from CI, which is good because that isn't what we want to be testing here. I think it's sufficient that rez fall back to system python for the purposes of what we actually want to test here (that the installation works at all; that the installation is capable of creating a rez package from its setup.py).

I think it would be fine to leave that second part for a follow-up PR, and to leave the 'rez-bind python' here for now.

PR #893, opened 2020-05-26
Package cache

nerdvegas:
Package caching implementation, previously referred to as "package localization".

I wanted to steer clear of the term "localization" because it could insinuate something else:
* Package caching: Copying variant payloads onto local disk in a non-intrusive way (asynchronously via a separate process in this case) in order to improve performance, where package contents would otherwise be fetched over shared storage;
* Package localization: The process of creating a set of variants in a repo specifically to meet the requirements of a context, so that that context and repo could be copied to a server/container (for eg) and then run in a standalone fashion (ie, no access to package repo on shared storage required).

See the new docs in markdown for further details.


KelSolaar:
@nerdvegas : Looks promising, we have a system that caches/uncaches/updates packages locally. Main question I have is: How do you handle package updates, technically a package once released should be immutable but in reality, it might not be, so how does the system cope up with such updates?

nerdvegas:
Simply put, it doesn't. Local packages are not cached by default, and you
have the default_cachable_per_repository setting to blacklist any package
repos that may not be immutable.

Are there non-local cases where you might need this? Bear in mind that what
matters here is variant immutability, not package.

Cheers
A

KelSolaar:
It happens quite regularly that I need to update a network package and that the changes need to be replicated to the artist's local cache. For example, I updated all our Maya packages so that they require the new Autodesk Maya Security Tools, which ensures that they are picked at all times, even by old projects.

We have a process that hashes the network packages and the artist's workstations then check periodically if their local cache hashes are up-to-date and update the relevant packages accordingly.

instinct-vfx:
@KelSolaar This is not accounted for because variants are designed to be immutable. The usual approach would be to add an internal versioning scheme that allows to keep the upstream version but add internal versions for package changes.

@nerdvegas Great to see this happening! Will try to have a look at it today. One thing that occurred to me recently (and without yet having looked at this PR so please excuse if this is off): What would you think about an addition to the concept where the payload is a zip instead and HAS to be localized in order to be used? We have quite a few packages that you can't really use at all across the network. Or as a slight variation allows the addition of a zipped copy of the payload that is created on first caching and can be used in subsequent caching?

nerdvegas:
Do you mean changes to the package.py, or to the installed variant itself?
To be clear package caching does not cache the package definitions
themselves. All it does is switch REZ_XXX_ROOT to the local cache copy if
it exists.

A

nerdvegas:
Hey Thorsten,

In terms of forcing a package to be cached before it can be used, I think
it would be fairly straightforward to do that. Perhaps a 'pre_cache' bool
package attribute would enable this.

Wrt zip - is the intent to speed up copying of the package to local disk?
If so, what kind of difference are you seeing between that and copying the
payload natively?

Cheers
A

instinct-vfx:
Yes the reasons is speeding up the copy. I think we are in a bit of a worse spot than others there but the main problem is many small files. I did a quick test with our vraymax-4.30.00 package. That's a bit over 1GB and has around 1500 files many of which are very small header and source files from the SDK. The raw package folder took around 90s to copy to local disc. I packed it (with store compression ratio so the size is the same) and copied from the same location and to the same local disk in 10s.

instinct-vfx:
Having thought about it some more i think that if we do it it would justify it's own PR. What i had in mind is approx:
- Add a new attribute to package.py that indicates the package should have a zip
- rez-build would then (after successfully building the package) zip up the payload and place it in a defined location with a defined name (can follow the logic of hashed variants for local caches i guess)
- Add support to the package cache feature to pull the zip to a temp folder and extract rather than just copy
- Bonus points for adding support to rez-pip to also create these cache zips. I just checked and localizing numpy took 10s. That can scale up quite a lot with many dependencies.

Also @lambdaclan had an interesting idea today in a call: Could this be extended to do multi-threaded copies (as in spawn sub-threads per variant copy task up to a configurable amount of copy threads? Especially with the small files bandwidth is VERY low and that would also benefit a lot.

nerdvegas:
Yes I think any work in improving caching speed would be in a separate PR,
I did not want to optimise early. But tbh I also wasn't thinking about that
so much, because the idea is that packages get cached asynchronously
anyway, and your first resolve probably won't get the cached variant
payloads. In this scenario you can see how caching speed is less important.

I'm interested in these cases where you need the caching to happen before
the resolve is completed. What are some use cases? Is it common?

I think it'll be fairly easy to use threading to speed up caching, but
adding the zip workflow does add more complexity and I think we need to
weigh up whether it's worth doing. It seems like that may only be the case
for these must-cache packages.

Cheers
A

instinct-vfx:
Examples would be Maya, Houdini or Unity and Unreal. For Maya the problem is, that it crawls to a halt also while using, not only during load. That's also mostly a windows problem as far as i understand. Houdini just takes ages too load afaik. It's also genuinely a releasing problem. It's around 3GB and has around 50k files. With many nightlies and releasing them often that creates a lot of load. the VFX dept already has a custom tooling to have zipped payloads because it was unbearable.
For Unreal and Unity i am not 100% sure they can even be properly wrapped in a package but my try failed for two reasons:
1. They are huge. Around 30GB and insane amounts of files. Both releasing and local caching takes ages.
2. They have very deep path hierarchy. Even without any variants i ran into paths that were too long for Windows+SMB.

There are more but these are some prominent examples.

nerdvegas:
Ok, so breaking it down:

* Some apps are simply too slow over nfs, we would like to force caching on
these, and live with the initial resolve delay while the variant is copied
over
* Some apps are very large, and copying a zip is significantly faster than
copying the payload as-is
* We need a way to build/specify this zipfile

I think these are all valid, but should be addressed as separate PRs.

In terms of building/registering said zip, my current thoughts are:
- introduce 'cache_zip' package attr, eg "cache_zip = '{root}/payload.zip'"
- have a step in the build that will create this zipfile if 'cache_zip' is
specified and doesn't already exist after the build has run.

Thoughts?
A

nerdvegas:
One more thing, just playing devil's advocate here.

Note that it is possible to warm any given package cache via "rez-pkg-cache
-a <variant-uri>". For these apps where you need caching ahead of time, you
might want to have a scheduled process that does this. If that were the
case, then the time it takes to copy the variant over is a lot less
important, and you may not really see benefit from the zip feature.

maxnbk:
Just fyi, it may be useful to be able to use the actual package root rather
than variant root for that zipfile, as well. ie, supply all variants as a
zip.

instinct-vfx:
Only variants are guaranteed to be immutable though, no? So that zip would need updating which seems very easy to break in subsequent releases. Also, for example with a big package like maya or houdini you would not want to download the linux variants on windows and vice versa.

PR #935, opened 2020-08-13
Fix #934, no hash string in include script file name

### LINKED ISSUE #934: Installed package not including latest module
### Problem

I have my `rezconfig.py` set `package_definition_python_path="/path/to/shared/code"` and I include the shared module with `include` decorator for my package's `commands` function.

The included module was copied correctly into `{package.root}/.rez/include` directory when I run `rez-build --install`. But the installed package not able to pick up the correct copy of include module after I re-build with source have changed.

### Cause

I found that it was because the copied shared module was named with hash string, which is fine but then Rez did not pick the right one, only the first one it `glob`ed.

https://github.com/nerdvegas/rez/blob/f81c721c079abe99746dce80853da45351a2b683/src/rez/utils/sourcecode.py#L324-L328

Maybe we could drop naming shared module with hash string ? Just hash it again when trying to access them.
I'll try submit a PR as a proposal of solution.

davidlatwe:
This PR fix #934. Installed package will be able to pick up updated included module.

### What's changed

* Python module that being included in `package.py` with `@include`, will be copied without hash string.
* When an installed package try to load the includes, will firstly find the module that is not named with hash string, if NOT found, fallback to previous behavior. And if found, the module file will be hashed again for finding previous imported one.

### DIFF
--- a/src/rez/utils/sourcecode.py
+++ b/src/rez/utils/sourcecode.py
@@ -292,9 +292,11 @@ class IncludeModuleManager(object):
         self.modules = {}
 
     def load_module(self, name, package):
+        from hashlib import sha1
         from rez.config import config  # avoiding circular import
         from rez.developer_package import DeveloperPackage
 
+        hash_str = None
         # in rare cases, a @late bound function may get called before the
         # package is built. An example is 'requires' and the other requires-like
         # functions. These need to be evaluated before a build, but it does also
@@ -303,8 +305,6 @@ class IncludeModuleManager(object):
         # copied into package payload) location.
         #
         if isinstance(package, DeveloperPackage):
-            from hashlib import sha1
-
             # load sourcefile from original location
             path = config.package_definition_python_path
             filepath = os.path.join(path, "%s.py" % name)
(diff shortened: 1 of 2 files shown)


nerdvegas:
Hey good find!

I think the PR takes the right approach, but I'd make one change. To avoid recalculating the source hash at package load time (since we should avoid anything that may incur more cost when resolving an env), we should store it at package install time. Ie, whereas before it was stored in the filename (eg "utils-<sha1>.py"), now we should install two files instead (eg 'utils.py' and 'utils.sha1').

The idea behind this was that it may be likely that many installed packages use the same include sources, hence the hash to lookup already-compiled modules. The implemented was flawed of course and that's what you've found.

One other thing - I would make the backwards compatibility code painfully obvious - eg in a nested func with a comment referencing this issue/PR (I'm a big fan of leaving breadcrumbs, and often have I cursed myself for not doing so in the past). But thanks for taking compatibility into account, that's a priority in this project.

davidlatwe:
> To avoid recalculating the source hash at package load time (since we should avoid anything that may incur more cost when resolving an env), we should store it at package install time. Ie, whereas before it was stored in the filename (eg "utils-{sha1}.py"), now we should install two files instead (eg 'utils.py' and 'utils.sha1').

I had thought about that, but not sure how to do it at the time being.
Adding a file `<name>.sha1` to keep the hash is a good idea !

src/rez/utils/sourcecode.py
```diff
@@ -312,26 +311,32 @@ def load_module(self, name, package):
...
                 hash_str = filepath.rsplit('-', 1)[-1].split('.', 1)[0]
-            else:
-                filepath = pathname
-
-        if hash_str is None:
-            with open(filepath, "rb") as f:
-                hash_str = sha1(f.read().strip()).hexdigest()
+                # End, for details of backward compat,
+                # see https://github.com/nerdvegas/rez/issues/934
+                # and https://github.com/nerdvegas/rez/pull/935
```
davidlatwe, line 339: > I would make the backwards compatibility code painfully obvious

XD
Not sure about the *nested function* though, but this would be enough ?

PR #938, opened 2020-08-19
Raise unversioned error when config not allowed

davidlatwe:
### Problem

When building unversioned package with `config.allow_unversioned_packages = False`, the error raised and says :
```
  ...
  File "..\rezplugins\package_repository\filesystem.py", line 715, in _create_variant
    overrides=overrides
  File "..\rezplugins\package_repository\filesystem.py", line 1228, in _create_variant
    raise RezSystemError("Internal failure - expected installed variant")
rez.exceptions.RezSystemError: Internal failure - expected installed variant
```
Which isn't clear enough to know that the failure was due to building a package which is unversioned while Rez has been configured to not allow it.

### Propose

Check package data and config settings between data override applying and data dumping.
And the output would be :
```
Invoking custom build system...
20:48:55 ERROR    PackageMetadataError: Unversioned package is not allowed in current configuration.
```

### DIFF
--- a/src/rezplugins/package_repository/filesystem.py
+++ b/src/rezplugins/package_repository/filesystem.py
@@ -1184,7 +1184,8 @@ class FileSystemPackageRepository(PackageRepository):
         # Stop if package is unversioned and config does not allow that
         if (not package_data["version"]
                 and not config.allow_unversioned_packages):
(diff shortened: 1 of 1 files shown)


nerdvegas:
Thanks again for yet another concise and explanatory PR!

PR #939, opened 2020-08-19
Improvements to bind module

pablogipi:
These changes are mainly targeted to improve the bind module in order to make easier to automatically create packages for installed software.
It mainly creates some new options in rezconfig.py and add some new functions into rez-bind._utils.
the main concept is to support sofware installed following the next convention:
- There is a root folder for all applications. By default is assumed to be /opt .
- Every application has it's own folder using the same name as the app.
For instace for houdini: /opt/houdini  .
Then inside the app folder there are two options.
If we have only one version of the app installed it can live inside the app folder, so we can have an app like cmder installed in: /opt/cmder  . Then is assumed that version is the only one installed.
The usual case, several versions are installed, every version is installed in a subfolder, the name for these subfolders really doesnt matter but it is suggested to use something meaninful like the version number or a convention used by the app, examples for houdini and python.
- Using apps convention:
```
/opt/houdini/hfs17.5.886
/opt/houdini/hfs18.0.427
/opt/python/python27
/opt/python/python38
```
- Using version number/name:
```
/opt/houdini/17.5.886
/opt/houdini/18.0.427
/opt/python/27
/opt/python/38
```
These are some new options supported in rezconfig.py: 
```
bind_use_folders_vers
bind_quickstart_tools
bind_apps_tools
release_packages_root
release_packages_types
```
I have attached a zip file with an example of a rezconfig.py using the new options, full of comments for every new option, they are also included in the rezconfig.py in the PR. And some examples of how a bind module can use the new features.
The idea in production is to be able to use something like:
rez bin --release --quickstart 
rez bin --release --apps
And get packages installed for modules included in the bind_quickstart_tools and bind_apps_tools.

[RezPR.zip](https://github.com/nerdvegas/rez/files/5097900/RezPR.zip)


nerdvegas:
Hi,

I cannot review this PR as it's target to develop branch and there are 480 changed files; you need to PR to master.

Having said that - rez-bind is to be deprecated so I would rather any further development in that area to stay forked. Nowdays package.py's are more expressive than they used to be, and it should be possible to model 'binding' packages that way instead. For example, see https://github.com/nerdvegas/rez-recipes/blob/master/python/_native/package.py.

Thx
Allan

Issue #708, opened 2019-08-30
Alias can't be used on the same line as rez-env


JanRiebeAxis:
We are observing some unexpected behaviour with the alias feature.

For the examples let's assume we are setting up a package 'alias_test' in which we define alias('alias_command', 'echo alias successfully excecuted')

When entering a rez shell that includes a package which sets an alias the alias can afterwards be used as expected.
Example:
C:\Users\me>rez-env alias_test
> C:\Users\me>alias_command
> C:\Users\me>alias successfully executed

However when running the alias command in the same line the alias is not recognised.
Example:
C:\Users\me>rez-env alias_test -- alias_command
> C:\Users\me>'alias_command' is not recognized as an internal or external command,
operable program or batch file.

This was mentioned before in issue #389 but appears to still occur.

instinct-vfx:
This is currently expected because if you use the command syntax with -- (or -c) there is no actual shell created and hence doskey is never run. This works with the powershell and pwsh plugins. I am actually not sure if there is a really good way to support this.

JanRiebeAxis:
Thanks for the insight. That makes sense, though it would be a good feature to have.
As a workaround we are now considering to wrap commands into batch scripts, to act as an alias would, and append these to path in the rez package.

JeanChristopheMorinPerso:
Fixed by #948

PR #948, opened 2020-09-02
Support rez-env -c <alias> or -- <alias> (Windows CMD shell)

davidlatwe:
### Problem

As #708 stated, `rez-env pkg -c alias-cmd` or `rez-env pkg -- alias-cmd` currently not supported (on Windows).

### Solution

Since the `Alias` object will be saved in `ActionManager` if any alias setup been found in package's commands by `RexExecutor`, look into that list and matching them with input `command`, swap it back to actual command if matched, would be the simplest solution.

### DIFF
--- a/src/rezplugins/shell/cmd.py
+++ b/src/rezplugins/shell/cmd.py
@@ -23,6 +23,7 @@ class CMD(Shell):
     # http://ss64.com/nt/cmd.html
     syspaths = None
     _doskey = None
+    _doskey_alias = None
     expand_env_vars = True
 
     _env_var_regex = re.compile("%([A-Za-z0-9_]+)%")    # %ENVVAR%
@@ -152,6 +153,7 @@ class CMD(Shell):
                     stdin=False, command=None, env=None, quiet=False,
                     pre_command=None, add_rez=True, **Popen_args):
 
+        command = self._reveal_alias(command)
         startup_sequence = self.get_startup_sequence(rcfile, norc, bool(stdin), command)
         shell_command = None
 
@@ -312,6 +314,7 @@ class CMD(Shell):
                 self._doskey = "doskey"
 
         self._addline("%s %s=%s $*" % (self._doskey, key, value))
+        self._alias_map(key, value)
 
     def comment(self, value):
         for line in value.split('\n'):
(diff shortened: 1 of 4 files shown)


instinct-vfx:
I think this would rather belong into the cmd shell plugin instead of resolved_context. It is a shell specific issue, not platform specific. (Powershell handles it fine).

davidlatwe:
That's weird, just tested on current master branch with my powershell and it didn't work :O

davidlatwe:
Okay, I have changed my implementation into CMD specific. ☺️

davidlatwe:
Just checking in, is this a valid/solid solution to you guys ? :)

src/rezplugins/shell/cmd.py
```diff
@@ -152,6 +153,7 @@ def spawn_shell(self, context_file, tmpdir, rcfile=None, norc=False,
                     stdin=False, command=None, env=None, quiet=False,
                     pre_command=None, add_rez=True, **Popen_args):
 
+        command = self._reveal_alias(command)
```
nerdvegas, line 156: I think `_expand_alias` would be more intuitive

src/rezplugins/shell/cmd.py
```diff
@@ -352,6 +355,16 @@ def join(cls, command):
     def line_terminator(cls):
         return "\r\n"
 
+    def _alias_map(self, key, value):
+        if self._doskey_alias is None:
+            self._doskey_alias = dict()
+        self._doskey_alias[key] = value
+
+    def _reveal_alias(self, command):
```
nerdvegas, line 363: This could be a little more robust I think. The problem's gonna be if args are passed to the alias, the alias expansion won't happen in that case. Even just taking the first word in `command` and expanding just that, is probably going to work in 99% of cases.

I would also include in a docstring somewhere, an explanation as to why this is here in the first place. Without knowing about the doskey issue it's gonna be difficult for someone coming in fresh to know why this alias expansion is being done.

davidlatwe: Thanks for review :D

> The problem's gonna be if args are passed to the alias, the alias expansion won't happen in that case.

Yes, maybe `str.startswith` would do the trick. Will test it later.

> I would also include in a docstring somewhere

🙆🏼‍♂️ 👌🏼

davidlatwe:
Just improved the implementation, the `command` that has alias + arguments can be matched and *expand* now. 

Also, one more simple test case is added.

PR #951, opened 2020-09-07
Fixes package cache stall on osx.

reformstudios:
Using shutil on osx causes the package cache to stall. (unable to delete the .copying file maybe? )
Replacing shutil with cp appears to fix the issue.
"copy" command added to retain windows compatibility.

Resolves nerdvegas/rez/#950


reformstudios (commented):
Fixed missing module from call.

nerdvegas:
Have to close, this change has broken the Windows build.

We need more specific details as to why OSX is failing.

PR #956, opened 2020-09-16
Fix unicode vcs changelog encode err

davidlatwe:
### Problem

`rez-release` raise `UnicodeDecodeError` when releasing package with non-ascii characters in change log.

### Solution

* Explicitly set `encoding="utf-8"` when fetching results from vsc with subprocess. (but not if `python<3.6`)
* Allow non-ascii character when formatting with `json`.

### DIFF
--- a/src/rez/utils/execution.py
+++ b/src/rez/utils/execution.py
@@ -73,8 +73,8 @@ class Popen(_PopenBase):
         #
         # NOTE: currently no solution for `python3+<3.6`
         #
-        if sys.version_info[:2] >= (3, 6) and "encoding" in kwargs:
-            kwargs['encoding'] = 'utf-8'
+        if sys.version_info[:2] < (3, 6) and "encoding" in kwargs:
+            kwargs.pop("encoding")
 
         super(Popen, self).__init__(args, **kwargs)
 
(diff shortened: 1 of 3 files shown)


src/rez/utils/execution.py
```diff
@@ -73,8 +73,8 @@ def __init__(self, args, **kwargs):
         #
         # NOTE: currently no solution for `python3+<3.6`
         #
-        if sys.version_info[:2] >= (3, 6) and "encoding" in kwargs:
-            kwargs['encoding'] = 'utf-8'
+        if sys.version_info[:2] < (3, 6) and "encoding" in kwargs:
+            kwargs.pop("encoding")
```
nerdvegas, line 77: I'm not sure this is right. Looking at the original code, I think there may have been an existing bug, and L76 should be:

```
if sys.version_info[:2] >= (3, 6) and "encoding" not in kwargs:
```

davidlatwe: Ah, my original intention was to not pin the encoding to "utf-8" (no specific reason), only removes it when the Python version is not compat, since `encoding` kwarg has been explicit set in commit d38e873, which should be removed if Python version incompat.

Now after #961, I think commit d38e873 should be reverted or it would be Python 2.7 incompat ?

PR #957, opened 2020-09-17
Fix repo location false mismatch

davidlatwe:
### Problem

I have my `release_packages_path` set to a mounted network drive e.g. `'x:/rez-studio/release'`, and when I run the following script to get current resolved context, I got `ResourceError` that saying my repository location mismatched.

```
>>> from rez.status import status
>>> status.context
...
rez.exceptions.ResourceError: location mismatch - requested 'x:\\rez-studio\\release',
    repository location is '\\\\server\\technical\\rez-studio\\release'
```

<details><summary>Full error</summary>

```shell
>>> from rez.status import status
>>> status.context
  File "c:\users\davidlatwe.lai\rez\packages\install\rez\2.67.0\platform-windows\payload\rez\resolved_context.py", line 615, in read_from_buffer
    return cls._read_from_buffer(buf, identifier_str)
  File "c:\users\davidlatwe.lai\rez\packages\install\rez\2.67.0\platform-windows\payload\rez\resolved_context.py", line 1666, in _read_from_buffer
    context = cls.from_dict(doc, identifier_str)
  File "c:\users\davidlatwe.lai\rez\packages\install\rez\2.67.0\platform-windows\payload\rez\resolved_context.py", line 1511, in from_dict
    variant = get_variant(variant_handle)
  File "c:\users\davidlatwe.lai\rez\packages\install\rez\2.67.0\platform-windows\payload\rez\packages.py", line 671, in get_variant
    variant_resource = package_repository_manager.get_resource_from_handle(variant_handle)
  File "c:\users\davidlatwe.lai\rez\packages\install\rez\2.67.0\platform-windows\payload\rez\package_repository.py", line 499, in get_resource_from_handle
    resource = repo.get_resource_from_handle(resource_handle)
  File "c:\users\davidlatwe.lai\rez\packages\install\rez\2.67.0\platform-windows\payload\rez\package_repository.py", line 346, in get_resource_from_handle
    raise ResourceError("location mismatch - requested %r, "
rez.exceptions.ResourceError: location mismatch - requested 'x:\\rez-studio\\release', repository location is '\\\\server\\technical\\rez-studio\\release'
```
</details>

Those two paths were actually pointing to the same location, one is a mounted path, another is an Windows UNC path.

The UNC path was coming from the filesystem plugin, which convert the path with `rez.utils.filesystem.canonical_path`,
at here 👇🏼 
https://github.com/nerdvegas/rez/blob/77ada5b1a9a1793fd14a2403ebc8c5f6385599ba/src/rezplugins/package_repository/filesystem.py#L478-L480

But the other path (from `ResourceHandle`) that put to varified with, does not made it canonical.
https://github.com/nerdvegas/rez/blob/77ada5b1a9a1793fd14a2403ebc8c5f6385599ba/src/rez/package_repository.py#L345-L349

Hence the false alarm.

### Solution

Is this PR.

### DIFF
--- a/src/rez/package_repository.py
+++ b/src/rez/package_repository.py
@@ -337,14 +337,14 @@ class PackageRepository(object):
             # always be made from repo.make_resource_handle... for now,
             # at least, error to catch any "incorrect" construction of
             # handles...
-            if resource_handle.variables.get("repository_type") != self.name():
+            repository_type = resource_handle.variables.get("repository_type")
+            if repository_type != self.name():
                 raise ResourceError("repository_type mismatch - requested %r, "
                                     "repository_type is %r"
-                                    % (resource_handle.variables["repository_type"],
-                                       self.name()))
+                                    % (repository_type, self.name()))
 
             location = resource_handle.variables.get("location")
-            if self.name() == "filesystem":
+            if repository_type == "filesystem":
                 location = canonical_path(location)
 
             if location != self.location:
--- a/src/rez/package_repository.py
+++ b/src/rez/package_repository.py
@@ -1,5 +1,6 @@
 from rez.utils.resources import ResourcePool, ResourceHandle
 from rez.utils.data_utils import cached_property
+from rez.utils.filesystem import canonical_path
 from rez.plugin_managers import plugin_manager
 from rez.config import config
 from rez.exceptions import ResourceError
@@ -342,11 +343,14 @@ class PackageRepository(object):
                                     % (resource_handle.variables["repository_type"],
                                        self.name()))
 
-            if resource_handle.variables.get("location") != self.location:
+            location = resource_handle.variables.get("location")
+            if self.name() == "filesystem":
+                location = canonical_path(location)
+
+            if location != self.location:
                 raise ResourceError("location mismatch - requested %r, "
                                     "repository location is %r "
-                                    % (resource_handle.variables["location"],
-                                       self.location))
+                                    % (location, self.location))
 
         resource = self.pool.get_resource_from_handle(resource_handle)
         resource._repository = self


nerdvegas:
The strange thing about this is that the handle's `location` is set from the repo's `location` in the first place, so how did one end up non-canonical? Possibly you've gotten a cached resolve, and the handle was constructed on a different machine, and for whatever reason, the path on that host did not resolve to the same canonical path as it does on your localhost?

In any case, I'll take the fix but I'll move this filesystem- specific code into the filesystem repo plugin itself (ie override `get_resource_from_handle`).

davidlatwe:
>... handle's `location` is set from the repo's `location` in the first place, so how did one end up non-canonical ?

Yes,.. oh, but the handle's `location` in this case wasn't set from PackageRepository I think.

So I had another close look :

It was the `status` object loading ResolvedContext file,

https://github.com/nerdvegas/rez/blob/05fd08b90052b120598f7f58fd8b02338ed3f462/src/rez/status.py#L44

then parsing `"resolved_packages"` as `variant_handle`, (this is the part that I didn't know at the time for this PR)

https://github.com/nerdvegas/rez/blob/05fd08b90052b120598f7f58fd8b02338ed3f462/src/rez/resolved_context.py#L1504-L1511

and use it to get resource.

https://github.com/nerdvegas/rez/blob/05fd08b90052b120598f7f58fd8b02338ed3f462/src/rez/packages.py#L671

Indeed, from my RXT file, the `location` of `resolved_packages` were not canonical, so maybe the proper fix should be around here ?

https://github.com/nerdvegas/rez/blob/05fd08b90052b120598f7f58fd8b02338ed3f462/src/rez/resolved_context.py#L1511

davidlatwe:
> ~so maybe the proper fix should be around here ?~

Just saw the commit  72d3771, good place for the fix !

nerdvegas:
It would still be good to figure out where the non canonical path is coming
from! Ultimately all variant handles should be getting generated from the
package repo it came from... Even a handle in an rxt has come from a repo
at some point.

Issue #965, opened 2020-10-20
io.UnsupportedOperation when using rez api with pytest


spsalefeve:
Hi,

I have the following issue when using rez API in my tests with pytest:

```bash
2020-10-08 09:27:59,859|setcontext|solve_util|l. 465| ERROR| Context couldn't be resolved.
Traceback (most recent call last):
  File "/app/package/int/setcontext/current/python/setcontext/solve_util.py", line 460, in _resolve
    resolved_context = ResolvedContext(package_list, **kwargs)
  File "/app/package/ext/rez/current/python/rez/resolved_context.py", line 208, in __init__
    for x in config.implicit_packages]
  File "/app/package/ext/rez/current/python/rez/utils/data_utils.py", line 191, in __get__
    result = self.func(instance)
  File "/app/package/ext/rez/current/python/rez/utils/data_utils.py", line 587, in getter
    return self._validate_key(key, attr, key_schema)
  File "/app/package/ext/rez/current/python/rez/config.py", line 614, in _validate_key
    return key_schema.validate(value)
  File "/app/package/ext/rez/current/python/rez/config.py", line 54, in validate
    data = expand_system_vars(data)
  File "/app/package/ext/rez/current/python/rez/config.py", line 783, in expand_system_vars
    return _expanded(data)
  File "/app/package/ext/rez/current/python/rez/config.py", line 778, in _expanded
    return [_expanded(x) for x in value]
  File "/app/package/ext/rez/current/python/rez/config.py", line 778, in <listcomp>
    return [_expanded(x) for x in value]
  File "/app/package/ext/rez/current/python/rez/config.py", line 776, in _expanded
    return scoped_format(value, system=system)
  File "/app/package/ext/rez/current/python/rez/utils/scope.py", line 262, in scoped_format
    return formatter.format(txt, pretty=pretty, expand=expand)
  File "/app/package/ext/rez/current/python/rez/utils/formatting.py", line 182, in format
    return formatter.format(s)
  File "/usr/local/lib/python3.7/string.py", line 186, in format
    return self.vformat(format_string, args, kwargs)
  File "/usr/local/lib/python3.7/string.py", line 190, in vformat
    result, _ = self._vformat(format_string, args, kwargs, used_args, 2)
  File "/usr/local/lib/python3.7/string.py", line 230, in _vformat
    obj, arg_used = self.get_field(field_name, args, kwargs)
[...]

PR #966, opened 2020-10-20
965| Fix io.UnsupportedOperation

spsalefeve:
when using rez api with pytest.

This Pull Request is to fix Issue #965.


nerdvegas:
fixes #965

PR #968, opened 2020-10-27
Fix forwarding script on Windows (suite supporting)

davidlatwe:
### Problem

The *forwarding script* is basically an executable yaml file that has shebang which made it can be executed by `_rez_fwd`.

But this doesn't work on Windows, which means that tools in `rez-suite` cannot be found nor executed, and the build script that generated by `rez-build --scripts` won't work either.

### Solution

Adding extension `.cmd` to the forwarding script, makes it a semi-batch-script, e.g.

```cmd
@echo off
_rez_fwd.exe %~dpnx0
goto :eof
:: YAML
func_name: _FWD__spawn_build_shell
kwargs:
    ...
```

The line `goto :eof` will skip the YAML content when this script is being called in terminal, and the first 4 cmd script lines will be stripped before parsing into dict by `yaml.load` in `forward.py`.

I also fixed the missing import statement in custom build system on my way out.


davidlatwe:
Sorry for the force push, was trying to get rid of commits which turns out to be bad approach in test.

Mainly because Rez is not activated (rez bin is not in `PATH`) in the CI test, and the suite tools require `_rez_fwd` to work or test fail. But if I change the Action workflow yaml to add Rez in `PATH`, Windows docker image will need to be rebuilt by @nerdvegas which may cause too much trouble. Anyway, based on the README in `rez.test` dir, not activating Rez seems to be deliberately, so I changed to inject rez bin path solely in that test case.

Now, with that test case, should be enough to prove `rez-suite` can work cross platform. 🤔

src/rez/tests/test_suites.py
```diff
@@ -135,6 +139,31 @@ def test_3(self):
...
+        s.save(suite_path)
+
+        bin_path = os.path.join(suite_path, "bin")
+        env = os.environ.copy()
+        # To access _rez_fwd
+        env["PATH"] = os.pathsep.join([system.rez_bin_path, env["PATH"]])
+        # activate suite
+        env["PATH"] = os.pathsep.join([bin_path, env["PATH"]])
+
+        output = subprocess.check_output(["hunny"], shell=True, env=env,
```
davidlatwe, line 164: Should have change this to use tool's `context.execute_shell(block=True)`, then I won't have to put `rez_bin_path` manually. Didn't know that before.

Will fix that later.

davidlatwe (commented):
Has anyone tried this on Windows ? 🤓

src/rez/tests/test_suites.py
davidlatwe: Sorry, scratch that. The purpose of this test case is to test suite tool can be executed from a rez activated shell, not a resolved context.
Doc string added in f4ca471 to be more clear.

Issue #974, opened 2020-11-11
  include could not find load file:      RezBuild error on Windows


spsalefeve:
We're currently using an old version of rez (2.49.0) and would like to update to 2.68.4. Everything wees to be fine with the new version on Linux, but we can't build with Cmake anymore on Windows.
The first error is:

```cmd
CMake Error at CMakeLists.txt:7 (include):
  include could not find load file:

    RezBuild
```
and other error follows indicating the Rez environment variables for Cmake are not set.
It appears the reasons for this error are:

- our **cmake** package has  **visualstudio** in its `build_requires`
- our  **visualstudio** contains the following code:

```python
 if building:
        cmd = 'call "{executable}" x64'
        cmd = cmd.format(executable=os.path.join(expandvars("{root}"), "bin", "msvcEnv.bat"))
        command(cmd)
```
- rez now sets the build environment in `post_actions_callback` instead of `actions_callback` and it's too late

Would it be reasonable to use again `actions_callback` at [this line](https://github.com/nerdvegas/rez/blob/b21516589933afeed1e1a1a439962d2e20151e2d/src/rezplugins/build_system/cmake.py#L166) when building?
Or do yo know any other way to run cmake via rez on Windows? 
I tried to use `vswhere` as mentionned [here](https://github.com/nerdvegas/rez/issues/703) but it doesn't return any installPath.

spsalefeve:
Hello,
Could this PR be merged? It would enable us to update our very old rez on Windows!

PR #975, opened 2020-11-12
Fix running VisualStudio env when building

spsalefeve:
This PR fixes Issue #974 .

### DIFF
--- a/src/rezplugins/build_system/cmake.py
+++ b/src/rezplugins/build_system/cmake.py
@@ -163,7 +163,7 @@ class CMakeBuildSystem(BuildSystem):
         retcode, _, _ = context.execute_shell(command=cmd,
                                               block=True,
                                               cwd=build_path,
-                                              post_actions_callback=callback)
+                                              actions_callback=callback)
         ret = {}
         if retcode:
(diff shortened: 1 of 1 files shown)


nerdvegas:
I can't merge this PR yet, but it's on the right track. You are right in that CMAKE_MODULE_PATH should be getting set earlier, I think.

I think the reason post_actions_callback was being used is because this is where `pre_build_commands` is getting run from. With this PR, pre_build_commands would get run before package commands, instead of after, and I can see that breaking existing behaviour for some users (it also feels correct that pre_build_commands should be run as late as possible, but before the build - this is intuitive).

I think what we need to do is:
* split BuildSystem.add_standard_build_actions so that pre_build_commands setup is done in another method (eg `add_pre_build_commands`)
* keep the change in this PR that uses actions_callback instead, but also add a post_actions_callback that calls add_pre_build_commands.

In other words, we want what this PR is doing, but pre_build_commands should always be run via post_actions_callback instead.

Please note that this change needs to be done in all the build system plugins, not just cmake, for reasons of consistency. Fortunately it looks like a pretty trivial change.

Thx
A

PR #979, opened 2020-11-30
Remove bez

nerdvegas:
Special case code has been added to detect existing rezbuild.py files; the raised exception provides instructions for porting to build_command instead.


src/rez/build_system.py
```diff
@@ -73,6 +73,36 @@ def create_build_system(working_dir, buildsys_type=None, package=None, opts=None
...
+            # which expects a rezbuild.py file. Include info in error showing
+            # how to port to a custom build command.
+            #
+            if os.path.exists(os.path.join(working_dir, "rezbuild.py")):
+                msg = (
+                    "No build system is associated with the path %s.\n"
+                    "\n"
+                    "There is a rezbuild.py file present, suggesting you were "
+                    "using the deprecated bez build system. You need to use a "
+                    "custom build command instead. You port your existing "
```
JeanChristopheMorinPerso, line 86:
```suggestion
                    "custom build command instead. You can port your existing "
```
I would suggest `You can port` instead of `You port`. Or `You can do that like so:`.

nerdvegas: updated

JeanChristopheMorinPerso (approved):
It's nice to see bez finally going away. 👍

PR #981, opened 2020-12-14
Fixes release hook for Python 3

bfloch:
Python 3 does not have a long type.


bfloch:
I edited this in nerdvegas/rez instead of my fork by mistake. Sorry for that.

PR #989, opened 2020-12-15
fix: exit file write retry loop after successfull write

bpabel:
This prevents the `open_file_for_write` function from attempting to
always write the file twice, even when the first write is successful.
It also solves a bug where if the first write was successful,
but the second raises a permission error, the loop will fail to catch
the error it was designed to catch and raise an exception.


nerdvegas:
Yeesh dunno how we missed this one for so long, cheers

Issue #858, opened 2020-02-24
'rez build --install --prefix' error 


ClementHector:
Hi there!

We encountered the following error while deploying a library of ours in a specific directory with rez on a python's docker image through Gitlab CI/CD.

We'd be grateful if you could tell us how to resolve this issue by ourselves or if you could fix it!
We might have missed some context or information in this comment; please tell us so if necessary.

Many thanks in advance.

```
rez build --install --prefix /mnt/rnd/PIPELINE/USER/HC
Building into: '/mnt/rnd/PIPELINE/USER/HC'

--------------------------------------------------------------------------------
Building smurfs-0.0.27...
--------------------------------------------------------------------------------
Resolving build environment: ellipsanime '~maya==2018.6' '~houdini==17.0.352' '~nuke==11.3.4' ellipsanime rez_utils-1
resolved by root@runner-c392f0d9-project-56-concurrent-0, on Mon Feb 24 17:08:30 2020, using Rez v2.40.4

requested packages:
ellipsanime         
~maya==2018.6       
~houdini==17.0.352  
~nuke==11.3.4       
ellipsanime         
rez_utils-1         
~platform==linux    (implicit)
~arch==x86_64       (implicit)
~os==debian-10      (implicit)

resolved packages:
ellipsanime-0.0.3  /mnt/rnd/PIPELINE/__PACKAGES__/project/ellipsanime/0.0.3  
python-3.7.6       /mnt/rnd/PIPELINE/__PACKAGES__/app/python/3.7.6           
rez_utils-1.0.3    /mnt/rnd/PIPELINE/__PACKAGES__/int/rez_utils/1.0.3        

Invoking custom build system...
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/rez/serialise.py", line 71, in open_file_for_write
    f.write(content)
  File "/usr/local/lib/python3.7/contextlib.py", line 119, in __exit__
    next(self.gen)
  File "/usr/local/lib/python3.7/site-packages/rez/vendor/atomicwrites/__init__.py", line 154, in _open
    self.commit(f)
  File "/usr/local/lib/python3.7/site-packages/rez/vendor/atomicwrites/__init__.py", line 185, in commit
    replace_atomic(f.name, self._path)
  File "/usr/local/lib/python3.7/site-packages/rez/vendor/atomicwrites/__init__.py", line 92, in replace_atomic
    return _replace_atomic(src, dst)
[...]

PR #991, opened 2020-12-16
Handle non atomic write method when writing on network fs with linux

jlorrain:
Linux doesn't provide a sync method for network mounted fs, which cause atomicwrites to fail.
I added a failsafe to write with a basic (non atomic) method for that edge case

Fix issue #858

### DIFF
--- a/src/rez/serialise.py
+++ b/src/rez/serialise.py
@@ -73,13 +73,18 @@ def open_file_for_write(filepath, mode=None):
             with atomic_write(filepath, overwrite=True, **encoding) as f:
                 f.write(content)
 
-        except WindowsError as e:
-            if attempt == 0:
+        except OSError as e:
+            # If we are writing to the network, we can't ensure atomicity
+            if e.errno == 22:
+                with open(filepath, "w", **encoding) as f:
+                    f.write(content)
+
+            elif attempt == 0:
                 # `overwrite=True` of atomic_write doesn't restore
                 # writability to the file being written to.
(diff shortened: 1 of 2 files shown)


src/rez/serialise.py
```diff
@@ -73,13 +73,18 @@ def open_file_for_write(filepath, mode=None):
             with atomic_write(filepath, overwrite=True, **encoding) as f:
                 f.write(content)
 
-        except WindowsError as e:
-            if attempt == 0:
+        except OSError as e:
+            # If we are writing to the network, we can't ensure atomicity
+            if e.errno == 22:
```
bpabel, line 78: Error no. 22 is an invalid filepath, right? What does that have to do with network paths?

jlorrain: The goal is to handle the sync method error of atomicwrites: 

>  File "/usr/local/lib/python3.7/site-packages/rez/vendor/atomicwrites/__init__.py", line 44, in _sync_directory
>   _proper_fsync(fd)
>OSError: [Errno 22] Invalid argument

This is ugly, but atomicwrites doesn't handle properly this error itself

nerdvegas:
Hey @jlorrain thanks for finding this and providing a fix. I think this `open_file_for_write` has gotten a bit messy, I'm gonna take what you have here and put in a new PR that incorporates the same approach, so don't be alarmed if this gets closed. I'll comment with new PR details here.

nerdvegas:
Closing in lieu of https://github.com/nerdvegas/rez/pull/998

Issue #994, opened 2020-12-22
wiki workflow broken


nerdvegas:
Wiki workflow doesn't properly deal with PRs/branches and fails when master isn't being merged.

* artifact 'wiki-markdown' could get overwritten by unrelated branches (artifact scope does not appear to be local to a workflow run)
* "Release: xxx" title link at top of sidebar is broken. It's "refs/head/master" when it should be current release version
* always updates master branch wiki, meaning that PRs end up overwriting current wiki docs before they're merged

nerdvegas:
closed by https://github.com/nerdvegas/rez/releases/tag/2.69.7

PR #995, opened 2020-12-22
Issue 994 wiki workflow fixes

nerdvegas:
This PR limits the workflow to master merges only. It would be possible to have it work for any branch, but would involve tracking a matching remote branch in the rez.wiki project. However, that is hard (or not possible?) to see in the github UI anyway and therefore of very limited value.

### DIFF
--- a/.github/workflows/wiki.yaml
+++ b/.github/workflows/wiki.yaml
@@ -1,8 +1,8 @@
 name: Wiki
 on:
   push:
-    #branches:
-    #  - master
+    branches:
+      - master
     paths:
       - '.github/workflows/wiki.yaml'
       - 'wiki/**'
--- a/.github/workflows/wiki.yaml
+++ b/.github/workflows/wiki.yaml
@@ -1,10 +1,8 @@
 name: Wiki
 on:
   push:
-    paths:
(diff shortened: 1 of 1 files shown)


nerdvegas:
Fixes #994

.github/workflows/wiki.yaml
```diff
@@ -1,11 +1,10 @@
 name: Wiki
 on:
   push:
+    branches:
+      - master
```
JeanChristopheMorinPerso, line 5: @nerdvegas This will limit us to see if the wiki builds correctly after changes to it. What we could do is allow all branches, but only run the publish step if the current branch is master. What do you think?

nerdvegas:
Hey JC, I did consider that, however from my own experience I'd say that
the wiki build never really fails, but instead generates problematic source
(but valid markdown) that can only be validated by actually looking at it.
And since you can't actually see it, there doesn't seem much point!

In any case if there's good reason then yes we could run the wiki build
regardless. If the artifact is still required (to somehow validate locally,
for eg) we could still do that too, but it'd have to be qualified (perhaps
with commit sha) to avoid potential overwrite across branches.

A

PR #1011, opened 2021-01-15
Improve get_variant_from_uri on Windows

davidlatwe:
When using Rez bin tool like `rez-pkg-cache` on Windows, the package path of variant URI must be in same letter case and same path sep with the path in `rezconfig` when manually adding variant for cache. Or it won't be able to find the variant.

This PR improves that.

### DIFF
--- a/src/rez/utils/filesystem.py
+++ b/src/rez/utils/filesystem.py
@@ -343,6 +343,7 @@ def is_subdirectory(path_a, path_b):
     try:
         relative = os.path.relpath(path_a, path_b)
     except ValueError:
+        # paths are on different mount drives
         return False
     else:
         return not relative.startswith(os.pardir + os.sep)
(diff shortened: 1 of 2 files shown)


src/rez/utils/filesystem.py
```diff
@@ -340,8 +340,12 @@ def is_subdirectory(path_a, path_b):
     """Returns True if `path_a` is a subdirectory of `path_b`."""
     path_a = os.path.realpath(path_a)
     path_b = os.path.realpath(path_b)
-    relative = os.path.relpath(path_a, path_b)
-    return (not relative.startswith(os.pardir + os.sep))
+    try:
+        relative = os.path.relpath(path_a, path_b)
+    except ValueError:
```
nerdvegas, line 345: what's this about..?

davidlatwe: Just added a comment.
It's Windows thing, when two paths are on different mount drives, e.g. :
```python
os.path.relpath("c:/hello", "d:/hello")
# ValueError: path is on mount 'c:', start on mount 'd:'
```

PR #1013, opened 2021-01-22
Adds more variables to the custom build system.

bfloch:
This patch adds more commands to the custom build system.

The reason is that we might be able to define more commands directly without intermediate bash or python scripts
if certain data is available in the build_command. Most prominently `install_path` so that we can for example satisfy build systems like:
```
build_command = "python setup.py --install-dir {build_path}"
``` 
This also eliminates the need to grab the environment should you prefer to handle build scripts via arguments only, which is a little more verbose to understand the action from the package.py

The full list of supported variables to the build_command are now:
- build_path
- install (retained)
- install_path
- name
- root (retained)
- variant_index (or empty string if non is given)
- version

Also adressed minor spelling/formatting issue and removed unnecessary temporary variables.

This will not break compatibility but due the the simplicity I did not add unit-tests. Let me know if you would prefer any.

### DIFF
--- a/src/rezplugins/build_system/custom.py
+++ b/src/rezplugins/build_system/custom.py
@@ -23,7 +23,6 @@ from rez.utils.logging_ import print_warning
 from rez.vendor.six import six
 from rez.config import config
 
-
 basestring = six.string_types[0]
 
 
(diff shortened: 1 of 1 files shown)


nerdvegas:
Looks good, however for completeness https://github.com/nerdvegas/rez/wiki/Package-Definition-Guide#build_command needs to be updated also to include the new fields.

PR #1029, opened 2021-02-23
context bundles

### LINKED ISSUE #1009: bundled contexts ("bundles")
A mechanism for storing both a context, and the variants that context uses, into a single relocatable directory. This directory could then be copied to a server, or stored into an image, for standalone deployment. Note that packages need to be relocatable.

The process should be broken down into the following smaller operations, which can be useful in their own right:

1. Copy the variants in a context into a target pkg repo
2. "Retarget" a context to a nominated repo(s)
3. Allow relative-path pkg repo references in contexts
4. Explicitly disable memcaching in a context (and store this option as a context attribute). And/or, infer non-cachability when pkg repo path is relative.

In particular, separating out (1) and (2) will allow for scenarios where you may want multiple contexts to be retargeted onto the same pkg repo. This would allow for a new suite schema, where the variant payloads could all be stored into one repo within the suite directory, thus making the entire suite relocatable and standalone.

A bundled context would look like this:

```
/my_bundle
  /context.rxt  # retargeted to relative repo "./packages"
  /packages
    /python/2.7/...
    /foo/1.0.0/...
```

To create this bundled context in manual steps you would:
```
]$ rez-env foo bah -o tmp.rxt
]$ rez-context --show-uris --print-resolve | xargs -i rez-cp --variant-uri {} --dest-path ./my_bundle/packages
]$ rez-context tmp.rxt --retarget ./my_bundle/context.rxt ./my_bundle/packages 
]$ rm -f tmp.rxt
```

For convenience, a new rez-bundle tool is available, and does the equivalent to above:
```
]$ rez-env foo bah -o tmp.rxt
]$ rez-bundle tmp.rxt ./my_bundle
```

nerdvegas:
Closes #1009

### DIFF
--- a/src/rez/cli/_main.py
+++ b/src/rez/cli/_main.py
@@ -8,7 +8,7 @@ import importlib
 from argparse import _StoreTrueAction, SUPPRESS
 from rez.cli._util import subcommands, LazyArgumentParser, _env_var_true
 from rez.utils.logging_ import print_error
-from rez.exceptions import RezError, RezSystemError
+from rez.exceptions import RezError, RezSystemError, _NeverError
 from rez import __version__
 
 
@@ -146,7 +146,7 @@ def run(command=None):
         extra_arg_groups = []
 
     if opts.debug or _env_var_true("REZ_DEBUG"):
-        exc_type = None
+        exc_type = _NeverError
     else:
         exc_type = RezError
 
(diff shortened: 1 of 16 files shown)


src/rez/cli/_main.py
```diff
@@ -146,7 +146,7 @@ def run(command=None):
         extra_arg_groups = []
 
     if opts.debug or _env_var_true("REZ_DEBUG"):
-        exc_type = None
+        exc_type = _NeverError
```
nerdvegas, line 149: small py3 compat issue I ran into

PR #1034, opened 2021-03-03
Issue 1032 pypi

### LINKED ISSUE #1032: Pypi entries are out of date
Add a workflow to automate new releases to Pypi, even though pip is not the recommended installation procedure.

nerdvegas:
Fixes #1032 
Also changed the wiki workflow a little by splitting into build/publish workflows.


.github/workflows/wiki-publish.yaml
```diff
@@ -20,46 +21,44 @@ jobs:
...
+      - uses: actions/upload-artifact@v2
         with:
           name: wiki-markdown
           path: ${{ env.TEMP_WIKI_DIR }}
 
-
   publish:
     name: Publish to GitHub Wiki
     runs-on: ubuntu-latest
     needs: build
```
j0yu, line 50: Would it be possible to still keep the build and publish as separate jobs in the same workflow YAML file, but for the `publish` job, include an extra `if` to only run it during a `release` event?

```yaml
on:  # Merge in this at the top of file
  release:
    types: [released]


  publish:
￼    name: Publish to GitHub Wiki
￼    runs-on: ubuntu-latest
     if: github.event_name == 'release'
￼    needs: build
```

.github/workflows/pypi.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Build rez
+        run: |
+          python setup.py sdist bdist_wheel
+
+      - name: Upload to PyPi
+        run: |
+          python -m twine upload dist/*
+        env:
+          TWINE_USERNAME: '${{ secrets.PYPI_USERNAME }}'
+          TWINE_PASSWORD: '${{ secrets.PYPI_PASSWORD }}'
```
j0yu, line 33: I've spotted [pypa have setup a CI action](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#publishing-the-distribution-to-pypi-and-testpypi) that could be used here to do the upload if you prefer instead of `twine`

nerdvegas:
Thanks for these suggestions Joe, I'll try them out now. Good idea RE the
wiki workflow.
A

nerdvegas:
Winner, those changes are released.

j0yu:
awesome, cheers!

PR #1036, opened 2021-03-05
add optionvars

davidlatwe:
### Feature propose

Add `rezconfig.optionvars` to store arbitrary key-value data and retrieve them in package.

```python
# rezconfig.py
optionvars = {
    "hello": "world"
}
```

```python
# package.py
name = "foo"
def commands():
    print(optionvars("hello"))
```

```shell
> rez-env foo
world
```

### Use case

I currently would use this feature to setup local payload path for package development across developer machines.

```python
# package.py
name = "foo"
def commands():
    if intersects(ephemerals.get("dev.foo", "dev.foo-0"), "1"):
        env.PYTHONPATH.append(optionvars("foo_dev_src"))
    else:
        env.PYTHONPATH.append("{root}/payload")
```

I know my use case could have many alternatives, but I thought this would be the most convenient approach with ephemeral packages and rezconfig involved.

What do you think ?


davidlatwe:
Just to be more clearer.

The main idea of this feature, is like adding *user preferences* on package or the value of corresponding ephemeral, which may not easy to define in package's definition file due to the differences between machines/users.

Although we can define those as environment variables in a shell launching script, but I think it would be better if those can be managed by `rezconfig`, and traced with `rez-config` command. In result, the startup environment can be simplified.

nerdvegas:
What if we were just to expose the `config` object instead? I like this idea as it may be helpful to expose other exisitng config settings; but also a studio isn't forced to use a specific `optionvars` key and could instead choose their own.

davidlatwe:
Thanks !!

> isn't forced to use a specific `optionvars` key and could instead choose their own.

Not sure I understand this, do you mean that one could pick their own name instead of "optionvars" ?
If so, I think we cannot do this actually, rezconfig only accept key that is in it's schema.

nerdvegas:
Ah yes - I was thinking about package definitions, not the config.

Ok, I think this is fine.
A

src/rez/rex.py
```diff
@@ -947,6 +947,10 @@ def expandable(value):
     return EscapedString(value, False)
 
 
+def optionvars(name, default=None):
+    return config.optionvars.get(name, default)
```
nerdvegas, line 951: `(config.optionvars or {}).get(name, default)`

davidlatwe:
Great, thanks !
Will start updating Wiki doc for this.

src/rez/rex.py
davidlatwe: Oh yes ! It could be None !

src/rez/rezconfig.py
```diff
@@ -914,6 +914,8 @@
     },
 ]
 
+optionvars = None
```
nerdvegas, line 917: There needs to be a docstring here, see the comments on all other settings. These get parsed and turned into the `Configuring Rez` wiki page.

nerdvegas:
Note that when updating the wiki, you need a section in https://github.com/nerdvegas/rez/wiki/Package-Commands#objects. In this case I think it'd be fine for that to be very brief and link to the `optionvars` section in `Configuring-Rez`. Something like

```
### optionvars
*Function*

Arbitrary data from the rez config, that can be accessed by packages. See the [config setting](...) for more details.
```

davidlatwe:
Just checking, is this able to be merged ?

nerdvegas:
It's on my list yes. I can probably get to it next week.
A

Issue #960, opened 2020-09-21
RezPluginError: Unrecognised package repository plugin: 'filesystem'


davidlatwe:
Shoule be a better title for this, but I cannot came up one that is short enough and understandable, so just take the error message for now.

This is an edge case, but the error was confusing so I though this should be worth to report.

### Problem (Reproducible)

I have a package that ships a custom package repository plugin, the directory structure looks like this:

```
my_package
  │ package.py
  │
  └─rezplugins
      │  __init__.py
      │
      └─package_repository
         │  __init__.py
         └─ my_plugin.py
```
And in my terminal, `cd` to `my_package`, with `rez` and `python` pacakge being resolved in the context, run Python and :

```python
> ..\my_package>python
Python 3.7.9 (default, Aug 31 2020, 17:10:11) [MSC v.1916 64 bit (AMD64)] :: Anaconda, Inc. on win32

>>> from rez.config import config
>>> from rez.package_repository import package_repository_manager
>>> for path in config.packages_path:
...     repo = package_repository_manager.get_repository(path)
...
15:34:06 DEBUG    searching plugin path ..\my_package\rezplugins\package_repository...
Traceback (most recent call last):
  File "..\pacakges\install\rez\2.67.0\platform-windows\rez\plugin_managers.py", line 176, in get_plugin_class
    return self.plugin_classes[plugin_name]
KeyError: 'filesystem'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "..\pacakges\install\rez\2.67.0\platform-windows\rez\package_repository.py", line 441, in get_repository
    repository = self._get_repository(normalised_path)
  File "..\pacakges\install\rez\2.67.0\platform-windows\rez\package_repository.py", line 512, in _get_repository
    cls = plugin_manager.get_plugin_class('package_repository', repo_type)
  File "..\pacakges\install\rez\2.67.0\platform-windows\rez\plugin_managers.py", line 297, in get_plugin_class
    return plugin.get_plugin_class(plugin_name)
  File "..\pacakges\install\rez\2.67.0\platform-windows\rez\plugin_managers.py", line 179, in get_plugin_class
    % (self.pretty_type_name, plugin_name))
rez.exceptions.RezPluginError: Unrecognised package repository plugin: 'filesystem'
>>>
```
Notice the debug message :
```python
15:34:06 DEBUG    searching plugin path ..\my_package\rezplugins\package_repository...
```
[...]

PR #1040, opened 2021-03-12
Extension plugins

davidlatwe:
### Feature Propose
Allowing people registering their own sub-command, forwarding function, configs.

### What's Added (updated)
* New type plugin "Command"
* New plugin loading mechanism : scan `sys.path` for module that has `rezplugins` sub-module
* Warning messages for plugin name clashing related issue

### Why scan modules for plugins ?
The new command type plugin enables registering custom sub-command, which may need to access additional module to do it's job. Hence we have to install the module which is required by the sub-command into Rez's venv, so that it could be imported in production installed Rez session which is ignoring `PYTHONPATH`. With these in mind, installing custom rezplugins as a sub-module would be the most desirable way. Now, how to let Rez register plugin that has been installed into `lib/site-packages` ? Instead of registering full path into `rezconfig.plugin_path`, simply scan modules from `sys.path` is most friendly.

### Note
When name clashing happen, plugins loaded by the new loading mechanism will take precedence over plugins that are registered in `rezconfig.plugin_path`.

~Please have a visit to my example plugin and have a look how this works 👉🏼 https://github.com/davidlatwe/rezbefoo~ (outdated)


nerdvegas:
Hey David, I really like this. Some thoughts:

* "Application" isn't the right terminology I think, "extension" would be better. The main reason I say this is because rez is actually missing a construct that I think of as an "app instance." Ie, somewhere to store package repo cache states, current config and so on (which is global state presently, and unfortunately). So I'd rather keep the term 'app' spare for if/when we make a change like this.
* "plugin_module" is too vague. With the above in mind I would simply go with "extensions".
* In a production install, you're going to want to install this into the rez venv. Even though doing that is straightforward (just pip install into the venv) it requires knowledge of rez internals (ie that it has its own venv in the first place). With that in mind, we should probably have a rez-install-ext tool which does this.

Thoughts?

instinct-vfx:
Hey,

i also really like this. I do agree with Allan's points. Also i wonder, as discussed before things like rez-pip might (should?) become separate projects, would this be the mechanism to install them?

nerdvegas:
Absolutely my thoughts Thorsten yes, rez-pip should be a separate project
and yes should be installed as an extension. Downside being that 'rez pip'
exists but 'rez-pip' doesn't, however that's something that rez-install-ext
would be able to fix, which is another reason for having it.

instinct-vfx:
Great stuff. I am totally in love with the idea of a more modular Rez and an ecosystem of extensions that you can use if you have a use case (think other package managers like nuget, chocolate or vcpkg or the likes on windows)!

@davidlatwe  what do you think about the changes Allan suggested?

nerdvegas:
Yeah it needs to happen. Rez is already more monolithic than I'd like,
modularity ftw.
A

nerdvegas:
Just by chance I ran into this today, illustrating my point about an Application class: https://github.com/nerdvegas/rez/issues/1033

davidlatwe:
Thanks @nerdvegas and @instinct-vfx !!

About naming, I agree we should avoid using the term "Application" for this, but not so sure about using "Extension" for that plugin type instead. Renaming the config entry "plugin_module" into "extensions" make sense to me as well, since the plugin manager can load all types of plugin from there. With this in mind, naming the new type of plugin to "Extension" makes me think the config "extensions" can only use to load that specific plugin type. Maybe "Supplement" or "Add-on" would be better name for new plugin type ? I will push the commit that reflect the points above first, and rename again if my point make sense to you guys as well.

About additional "rez-install-ext" tool, I was hoping that if we could just use venv's `pip` to do this, or with `rez-python -m pip`, hence the PR #1039 and #1041. Since it would be better if rez and it's extensions can use same installation method/standard. Still, what will "rez-install-ext" do ? Like, checking target environment if is a valid Rez production install before running pip ?

nerdvegas:
Hey I don't quite follow... do you mean that you're thinking ahead to where all plugins are loaded this way, and that 'plugin_modules' would encompass all plugin types, not just extensions? In any case that brings up another point.. it'd be great if that config entry were not necessary at all tbh. It's an odd extra step - with most tools, installing a new plugin is all you need to do, the system then knows about it. It'd be good if we could do the same.

Wrt rez-install-ext. I would not even want the user to have to run `rez-python -m pip foo` to install an extension, imo this is already too much detail. Providing rez-install-ext (or rez-install-plugin let's say) is future proofing - there may be other things we need to do as part of that process. The fact that pip is used as part of plugin installation at all should be of no interest to the user.

What if then, we start thinking about a new plugin architecture at the same time; and we think about how to ditch the config setting. We could have this:

* rez-install-plugin is used to install plugins
* it pip installs them into rez's venv
* we introduce a new type of plugin called an "extension". This type of plugin provides a cli tool
* as well as supporting "rez foo" extension form, we also support "rez-foo". Rez-install-plugin takes care of this
* we don't have to update a config setting; rez-install-plugin tracks this itself
* initially we don't support plugin installation in pip-based rez installs perhaps? Or perhaps we can, but the initial emphasis would be on getting this to work for production installs.
* this all has to work in combination with old-style (ie existing) plugins.

?

ps - one way we might track installed plugins is to iterate over all visible modules and expect certain standard attributes (rez_plugin_type, rez_plugin_name for eg) that identify them as rez plugins.

davidlatwe:
> do you mean that you're thinking ahead to where all plugins are loaded this way, and that 'plugin_modules' would encompass all plugin types, not just extensions?

No, not thinking ahead, it already does that. Since the current "extension" layout like what I have in my example [rezbefoo](https://github.com/davidlatwe/rezbefoo), the `rezplugins` is a sub-module of `rezbefoo`, so whatever plugin type it contains, will get loaded from the "plugin_modules".

> we don't have to update a config setting; rez-install-plugin tracks this itself

That would be awesome :D

> as well as supporting "rez foo" extension form, we also support "rez-foo". Rez-install-plugin takes care of this

I'll have a few more tests on #1039 since I thought this might still a bit related, e.g. to use `entry_points` or not in the `setup.py`.

nerdvegas:
Oh right, apologies for my misunderstanding. So is this compatible with
existing plugin types? Ie if I have an old-style plugin (found via
'plugin_path') and a new one, will they both function correctly and both
appear in the list shown by `rez -i`?

In terms of avoiding the config setting and just detecting plugins
automatically, that should be doable I think. We'd just use
`pkgutil.iter_modules` and find those modules with a `rezplugins` submodule.

davidlatwe:
No worries ! And yes, old-style plugin that found via "plugin_path" can live with new that found via "plugin_modules" (renamed into "extensions" at this point), and all listed by `rez -i` 🥳

nerdvegas:
Ok coooool. Giving the existing interoperability then, I reckon dropping
the config setting is a must (because old-style plugins don't require doing
this, so neither should new-style ones!) However I _ think_ the module
iteration approach suggested should do the trick.

Looking forward to seeing this wrapped up, I've been wanting to rework
plugins for a while now. And when we add rez-install-plugin, we'll have a
bonafide way to support plugins that can be implemented as separate,
standard pip packages. Lovely. When we get there, it'd be great to have an
example installable plugin as well. Maybe we add an examples/ dir, move
that example_packages/ dir into it, and add sometihng like
examples/plugins/foo.

Thx
A

davidlatwe:
> I reckon dropping the config setting is a must (because old-style plugins don't require doing this, so neither should new-style ones!)

Sorry, just to be sure. :P
Are you saying that to drop new config setting "plugin_module" and replace it with module iteration approach or maybe other, as long as no additional setting required ?

nerdvegas:
Yep. Basically these new-type plugins need to be automatically detected,
because that's already the case for old-style plugins.

Another thought - once we also have an example plugin in the repo, we could
use this in a new plugin.yaml github workflow to test that plugin
installation works.

Incidentally, I'm currently working on a workflow to automatically run the
benchmarking suite on every release, and keep an historic list in the repo.

Cheers
A

davidlatwe:
> because that's already the case for old-style plugins.

:O
I didn't know about this, I thought the only way to load them is from "pluing_path" config setting !

Anyway, I'll get started :D

nerdvegas:
Mm well, I suppose you do end up with a path per plugin so I guess you are
explicitly listing them. In any case, not having to would be great. Let's
see if it's doable.

davidlatwe:
> > because that's already the case for old-style plugins.
>
> :O
> I didn't know about this ...

Just found out that my old-style plugins they indeed are being auto-loaded (multiple plugins are auto-loaded per rezplugins path in config "pluing_path"), I misunderstanding what Allan have said above as we don't even need to register any path into the "pluing_path" config settings for loading old-style plugins. Anyway, I am sure that I am on the same page now. 😅

---

### Spec

So here's the spec based on [Allan's comment above](https://github.com/nerdvegas/rez/pull/1040#issuecomment-799891819) :

1. Introduce a new type of plugin called "extension". Which can be used to provide sub-command, forwarding functions.

2. Implement a new command tool for installing plugins

	* Naming the tool, say `rez-plugin` ?
    * Via command `rez-plugin --install <plugin-dir>`
    * The tool pip installs them into rez production env.
    * As well as supporting "rez foo" extension form, should also support "rez-foo".

3. No need to update config for each newly installed plugin specifically

4. All has to work in combination with old-style (ie existing) plugins.


Spec.1 is what this PR is for. Spec.3 has been implemented in commit 765fffb.
Now I am going to work on Spec.2 while keeping Spec.4 in mind.

davidlatwe:
I think issues like #960 is somewhat related, which may happen when developing/installing extension. And to avoid that kind of confusion, I propose a fix in here altogether.

In #960, ignoring current working directory when loading plugins was the original proposed fix. However, I find it's not solid enough to be bulletproof nor able to provide any clear error message.

So instead, I propose to add a file named `MAIN` in `rezplugins` module to indicate that it is the main plugin extending point. And from there, we could provide much stable plugin loading process and clearer log messages.

nerdvegas:
I don't think adding a MAIN file is really an option, simply put it's not backwards compatible. What's the issue specifically? Can we not just infer from the location of the 'rezplugins' submodule in any given plugin?

src/rez/plugin_managers.py
```diff
@@ -52,18 +53,23 @@ def extend_path(path, name):
...
-        if subdir not in path and os.path.isfile(initfile):
-            path.append(subdir)
+        elif config.debug("plugins"):
+            print_debug("skipped nonexistant rez plugin path: %s" % dir_)
+
+    # Extend old-style plugins
+    for dir_ in config.plugin_path:
+        append_if_valid(dir_)
+    # Extend new-style plugins
+    for loader, name, ispkg in pkgutil.iter_modules():
```
nerdvegas, line 70: I don't quite follow why this is here? Do new-style plugins require an `extend_path`, as old-style do?

In any case I don't think we can do this. This is causing an iteration over all modules, per plugin. I'm already a little concerned about doing that once, let alone once per plugin load. We need to come up with a different approach that doesn't do this here - module iteration should happen once only. Also imo new-style plugins shouldn't require use of `extend_path` at all.

setup.py
```diff
@@ -77,6 +77,7 @@ def find_files(pattern, path=None, root="rez"):
             find_files('*', 'tests/data') +
             find_files('*.exe', 'vendor/distlib'),
         'rezplugins':
+            ['MAIN'] +
```
nerdvegas, line 80: I would much rather not have this. Can we not infer from presence of `rezplugins`?

src/rez/cli/_util.py
```diff
@@ -60,6 +60,52 @@
...
+
+            except Exception:
+                if config.debug("plugins"):
+                    import traceback
+                    from rez.vendor.six.six import StringIO
+                    out = StringIO()
+                    traceback.print_exc(file=out)
+                    print_debug(out.getvalue())
+
+        elif config.debug("plugins"):
```
nerdvegas, line 99: I would default command_behaviour to something sensible rather than fail to load. Less barrier to entry, and most extensions are only going to require standard (non grouped) arg behaviour.

src/rez/plugin_managers.py
```diff
@@ -259,8 +265,43 @@ def register_plugin():
             'rezplugins' is always found first.
     """
     def __init__(self):
+        self._plugin_types_root = None
         self._plugin_types = {}
 
+    def locate_main_rezplugins(self):
```
nerdvegas, line 271: Let's not do this, and keep it simple instead. We expect a topmost 'rezplugins' submodule within the plugin package, and that's all.

src/rez/plugin_managers.py
nerdvegas, line 70: Deal with plugins with the same name here. If there's a clash, new-style plugin should take precedence, and a warning message should be printed. This will help migration.

nerdvegas: Ps - this will also match existing behaviour wrt a recent change, which ensured that plugins with the same name as a builtin (such as package_repository.filesystem) override the builtin.

nerdvegas, line 70: There is a problem here we need to address I think. I'm already a bit nervous about iterating over all modules (from a performance pov), but this is actually causing iter_modules to be called once *per plugin type*. We need to do something here to avoid that. I think a simple module list cache would do the trick.

nerdvegas:
We're getting there! One thing though - this PR is begging for a full example plugin, which we'd then install and test as part of a workflow.

One other thing - is this intentionally missing the rez-extensions tool? If so, perhaps we do that and the example plugin, in a separate PR - I think that'd be ok. Another note - rez-extension isn't a great name, because it can install all plugins, not just extensions. However, rez-plugins is already taken :(. I am not sure what to do about this one.

This is annoying, but what if we change the terminology a bit:
* "extensions" generally applies to all rez plugins
* we rename the new "extension" plugin type to "command" instead.

I think this fixes a few things - rez-extension makes sense then; a rez "command extension" is pretty intuitive also. I don't think we need to go so far as renaming 'plugins' to 'extensions' in the code (indeed this would be backwards compatible anyway). The Wiki needs a page on plugins however, ad if we called them extensions there, I think that'd get the message across enough.

src/rez/plugin_managers.py
davidlatwe: Oh sorry, this and the "MAIN" file were for fixing #960. Although #960 only happens on old-style plugins, but I think it's still a bit related to this PR since we are gonna implementing a new plugin installation method.

setup.py
davidlatwe: For old-style plugin and to provide a proper fix on issue #960, I think we have to 🤔 
Or instead of as "package_data" file "MAIN", just an empty python script `_main.py` ?

src/rez/cli/_util.py
davidlatwe: Ah, the extension type plugin can provide more than just sub-command, like forwarding functions, plugin configs, so it was meant to be like this to enable sub-command specifically by providing attribute `command_behaviour`. Since it would not make sense to register a sub-command when the plugin only want to register forwarding functions.

src/rez/plugin_managers.py
davidlatwe: Yeah, a module path list getter function with cache in instance of `RezPluginManager` should be it.

davidlatwe: > Deal with plugins with the same name here.

Hmmm, I think we should deal it in `RezPluginType.load_plugins`, since here we only get the python module name, not the name of each plugin.

davidlatwe:
> this PR is begging for a full example plugin, which we'd then install and test as part of a workflow.

Yeah, I'll try to make one today :)

> is this intentionally missing the rez-extensions tool?

No, just having some implementation detail might need more discussion, like, what other function that tool would have other then install plugins ? Checking `setup.py` to ensure the module that is being installed have valid rez plugin ?

I agree we could implement the tool in separate PR, since I think there are still something that need to be addressed and may make this thread a big one.

> we rename the new "extension" plugin type to "command" instead.

I like this new terminology, despite this new plugin type could have no command parser but only forwarding functions in my original propose. I mentioned that in this [comment](https://github.com/nerdvegas/rez/pull/1040#discussion_r599244713), I now retract that and provide a default `command_behavior` if there isn't one. Sub-command should be the first citizen.

davidlatwe:
Oh and, sorry about mixing the fix for #960 here, I thought it was related but seems bring in a bit confusion instead. Shall I revert it or make a force push ?

nerdvegas:
Yeah I would not include the fix for #960 here, it's confusing the issue.

I feel like the problem is a bit of an edge case and that requiring a MAIN
in every plugin root is overkill. Perhaps instead there may be a way to
detect that a plugin has been found via the local directory and to ignore
it in that case.

Agreed @ keeping test + rez-extensions for separate PR.

Cheers
A

src/rez/plugin_managers.py
davidlatwe: reverted.

setup.py
davidlatwe: reverted.

src/rez/plugin_managers.py
davidlatwe: implemented in 2b9cd2b

src/rez/cli/_util.py
davidlatwe: implemented in 8d0af73 (see this [comment](https://github.com/nerdvegas/rez/pull/1040#issuecomment-804600475))

src/rez/plugin_managers.py
davidlatwe: warning message added in 8e820b6

davidlatwe:
> I feel like the problem is a bit of an edge case and that requiring a MAIN
in every plugin root is overkill.

Ah no, we won't need that MAIN file in every plugin root, only the main `rezplugins` module. Anyway, will re-visit this in another PR. :)

Test failed on latest commit. Looks like there's something wrong with the test data, seems the material that required for the new tests were not being installed and leads to failed test. Will investigate this later.

davidlatwe:
Okay, that's it.
Requested changes have been addressed, tests added. And there's an example extension which contains a simple functional command type plugin that can be installed, but owning to we haven't implement the extension installing tool, I only left a brief pip install instruction in the README.

davidlatwe:
Top comment of this PR has been updated to reflect changes.

nerdvegas:
Hey David, cheers! I will get to this, but bear in mind I have an FMX talk coming up and it may be derailed by that for a bit as I'll have to spend time over a few weeks prepping for it.

A couple of thoughts (no need to act on these now! It's so I don't forget):
* It may be good to update existing (builtin) commands so they are also Command- based, simply for sake of consistency.
* When we have a rez-extensions tool, it might be good to have it update some command list stored in the rez installation. This would allow us to skip the module scan, which might be good if this takes appreciable time to perform. But if it's quick then I probably wouldn't bother.
* We should consider renaming Command to CLICommand to avoid confusion, as there are "package commands" already.

maxnbk:
I just want to jump in to say, awesome work here so far @davidlatwe .

nerdvegas:
Hey I'm just looking at this now and I see an issue. Currently, the plugin manager searches all packages in sys.path and will assume its found a plugin type if it finds a subdir in the package with matching name (eg "build_system"). There are two problems with this though:
* it's a little broad - there could be a package present that has nothing to do with rez, that has a submodule with a name like this;
* it's searching too many paths. For example, it will search a path like `/home/ajohns/rez/2.90.0/local/lib/python2.7/site-packages/coverage` 7 times (one for each plugin type) - it'd be good to discard this path just once instead.

To that end, how about we mandate that a `__rez__.py` file must be present in the package root? That way we can discard packages earlier, and avoid potentially picking up modules that aren't rez plugins at all.

davidlatwe:
I have thought about this earlier before, and was considering using python package's `entry_points` to [register them as rez's python plugin](https://packaging.python.org/guides/creating-and-discovering-plugins/#using-package-metadata).

But adding `__rez__.py` seems much simpler. :)

instinct-vfx:
Reading Allan's comment i was actually going to suggest using entry_points. They seem very well suited for this task and are specifically designed to support plugin-like discovery mechanisms. I would actually prefer that route.

davidlatwe:
But with `entry_points`, one would need to actually install it as a python package, which...
which is required right ? Because in Rez production install, PYTHONPATH won't work so you have to install it.

Oh, but `entry_points` may not work if that Rez is as a rez package. 🤔

instinct-vfx:
Could you elaborate on the last one? Why would it not work if Rez is a rez package? PYTHONPATH is a good point hm.

davidlatwe:
Sorry, that was wrong. It doesn't matter if Rez is as a rez package or not, but which Python is being used. Because that will affect where the python package metadata will be looked for entry points, correct ?

But if it's using `__rez__.py` then as long as it can be found in `sys.path`, it can be loaded.

nerdvegas:
Oh yeah you may be right. Since entry points are specific to python's own
packaging ecosystem, it's gonna be more difficult to manage this in
rezified packages (eg if you wanna use rez API + plugins within a resolved
env).

instinct-vfx:
That makes sense.

davidlatwe:
Just re-read the comment and,

> Currently, the plugin manager searches all packages in sys.path and will assume its found a plugin type if it finds a subdir in the package with matching name (eg "build_system").

I found that isn't entirely true, there still need `rezplugins` namespace on top of `build_system` for rez to include the module. But the current implementation indeed wasting search time on packages that are obviously has nothing to do with rez.

So, I will just adding `rezplugins` folder existence check for now, since I think it works the same as `__rez__.py`.

nerdvegas:
You're right, I missed this. Sound good.

PR #1051, opened 2021-03-30
Let rez-selftest try using pytest

davidlatwe:
For better test log reviewing, I propose to opt-in `pytest` (still use `unittest` if `pytest` not installed in Rez venv).
See how tidy the test reports in github-actions are.

I have changed the `selftest` sub-command arg mode to `passthrough`, so it can accept additional `pytest` args. But the existing [`/pytest.ini`](https://github.com/nerdvegas/rez/blob/master/pytest.ini) contains `pytest-cov` options, so one might have to install `pytest-cov` for the arguments like `--cov` can be recognized.

### DIFF
--- a/.github/workflows/core.yaml
+++ b/.github/workflows/core.yaml
@@ -44,6 +44,10 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Install test dependencies
+        run: |
+          python -m pip install pytest
+
       - name: Install Rez
         run: |
           mkdir ./build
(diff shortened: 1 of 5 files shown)


davidlatwe:
Can we have this merged ?

nerdvegas:
Expect this merged soon yes. I'm keen to as the current test output is very
verbose and it can be pretty difficult to track down where a given bug
actually is. I've been away so the usual rez dev didn't happen this week.

Cheers
A

Issue #673, opened 2019-07-22
REP-002: Public Rez Repositories (2019-2020 Roadmap)


nerdvegas:
# **WORK IN PROGRESS, DO NOT COMMENT YET**

This REP outlines the work necessary to enhance Rez so that it is able to install a package and its dependencies from public rez package repositories.

# Motivation

A major goal of rez is the ability to install packages from a public repository. Currently, studios that use rez often need to spend significant resources "rezifying" existing third party packages, so they can be integrated into their pipeline.

Currently, due to the programmatic nature of rez package definitions, it's possible to author a package that installs third party software, and this could be shared from a public repo (on GitHub for eg). For a example, see [this python package](https://github.com/nerdvegas/rez-recipes/blob/master/python/2.7.4/package.py). However, there are significant limitations to this approach currently:

* There is no dependency tracking. Ie, a built package's requirements are not installed also. If its requirements don't already exist in a rez package repository local to the studio, then the build will fail.
* There is no build configuration tracking. Ie, it is not possible to build multiple variants of a package that have been built in different ways (eg, debug vs release).
* There is no way to consume packages in a predictable way wrt build configuration. For example, even if we used stub packages to represent the build mode of a package, there still is not a reliable way to guarantee that the correct variant is chosen in a resolved environment.

# Goals

The goal of this REP is to overcome these shortcomings and to provide a tool and API for installing a rez package from a public repository.

A practical aspect of this goal is to provide the entirety of the [VFX Reference Platform](https://vfxplatform.com/) as a rez repository, so that any given package in the platform can be built easily, and its build configuration managed easily.

It is _not_ the goal of this REP to provide the ability to automatically download public repositories (ala pip etc) nor to provide support for artifact repositories, where packages have already been built and their binaries are downloaded instead. While it may make sense to extend to this functionality, they should be introduced in a separate REP(s). An installation workflow that requires downloading a public package repository first, then building packages from source, already achieves the goals described in this REP.

PR #1060, opened 2021-04-13
Implement REP-002.006: directive request syntax

davidlatwe:
This PR implements `REP-002.006` in #673. Note this, the new syntax isn't meant to replace/deprecate the current requirements expansion feature as what `REP-002.006` and `REP-002.001` have said, but try to convert those wildcarded requests into new syntax internally while building package.

### Delaying Requirement Expansion

Current requirements expansion feature expands requirements on package load, by scanning currently installed packages. In contrast, this new feature expands requirements with resolved package build context.

### Storing Additional Information

Information that parsed from directive syntax on package-load time, e.g. directive function name and arguments, need to be stored somewhere during package build. One possible option would be to attach them to `PackageRequest` or other related classes, which means to store them in package itself like attributes' metadata.

But the `PackageRequest` and other related classes are low level stuff, which isn't a good fit for this high level feature. Plus, by the nature of how package attributes are loaded before build, how attributes will be processed afterward, and how package variants will be spawned multiple times in different places, those information is hard to guarantee to be retrieved or written correctly at any given point without fighting more details in low level.

So instead, I end up storing them outside of package, and retrieve or modify those data by a key that is hashed from variant.

### Caveats

Currently, for old style expansion syntax requests (those have wildcards), this new feature only ***try*** to convert them into new syntax, and fallback to early expansion route if the conversion result is not defined in new syntax.

For example, how request like `foo-1+<2` will be harden is not yet been defined in `REP-002.006`, so for request like `foo-1.*+<2.*` can not be converted.

Also for case like `foo<**`, although it doesn't make sense in package requires but still can be converted, however the conversion result is `foo//harden` which means differently than the old style syntax.

### DIFF
--- a/src/rez/build_process.py
+++ b/src/rez/build_process.py
@@ -6,7 +6,7 @@ from rez.exceptions import BuildProcessError, BuildContextResolveError, \
     ReleaseVCSError, _NeverError
 from rez.utils.logging_ import print_warning
 from rez.utils.colorize import heading, Printer
-from rez.utils.request_directives import process_directives
+from rez.utils.request_directives import resolve_directives
 from rez.resolved_context import ResolvedContext
 from rez.release_hook import create_release_hooks
 from rez.resolver import ResolverStatus
@@ -266,7 +266,7 @@ class BuildProcessHelper(BuildProcess):
         if context.status != ResolverStatus.solved:
             raise BuildContextResolveError(context)
 
-        process_directives(variant, context)
+        resolve_directives(variant, context)
 
         return context, rxt_filepath
 
--- a/src/rez/build_process.py
+++ b/src/rez/build_process.py
@@ -6,6 +6,7 @@ from rez.exceptions import BuildProcessError, BuildContextResolveError, \
     ReleaseVCSError, _NeverError
 from rez.utils.logging_ import print_warning
 from rez.utils.colorize import heading, Printer
+from rez.utils.request_directives import process_directives
 from rez.resolved_context import ResolvedContext
 from rez.release_hook import create_release_hooks
 from rez.resolver import ResolverStatus
@@ -264,6 +265,9 @@ class BuildProcessHelper(BuildProcess):
 
         if context.status != ResolverStatus.solved:
             raise BuildContextResolveError(context)
+
+        process_directives(variant, context)
+
         return context, rxt_filepath
 
     def pre_release(self):
--- a/src/rez/utils/request_directives.py
+++ b/src/rez/utils/request_directives.py
@@ -1,0 +1,280 @@
+
+from rez.vendor.schema.schema import Schema, Optional, Use, And
+from rez.vendor.version.version import VersionRange
+from rez.vendor.version.requirement import Requirement
+from rez.vendor.version.util import dewildcard
+from rez.utils.formatting import PackageRequest
+from copy import copy
+import inspect
+import sys
+
+
+# directives
+#
+
+class DirectiveBase(object):
+    """Base class of directive request handler"""
+    @classmethod
+    def name(cls):
+        """Return the name of the directive"""
+        raise NotImplementedError
+
+    def parse(self, arg_string):
+        """Parse arguments from directive syntax string"""
+        raise NotImplementedError
+
+    def to_string(self, args):
+        """Format arguments to directive syntax string"""
+        raise NotImplementedError
+
+    def process(self, range_, version, rank=None):
+        """Process requirement's version range"""
+        raise NotImplementedError
+
+
+class DirectiveHarden(DirectiveBase):
+    """Harden directive request version to specific rank"""
+    @classmethod
+    def name(cls):
+        return "harden"
+
+    def parse(self, arg_string):
+        if arg_string:
+            return [int(arg_string[1:-1].strip())]
+        return []
+
+    def to_string(self, args):
+        if args and args[0]:
+            return "%s(%d)" % (self.name(), args[0])
+        return self.name()
+
+    def process(self, range_, version, rank=None):
+        if rank:
+            version = version.trim(rank)
+        hardened = VersionRange.from_version(version)
+        new_range = range_.intersection(hardened)
+        return new_range
+
+
+# helpers
+#
+
+def parse_directive(request):
+    """Parsing directive requests when creating package
+    """
+    if "//" in request:
+        request_, directive = request.split("//", 1)
+    elif "*" in request:
+        request_, directive = _convert_wildcard_to_directive(request)
+        if not directive:
+            return request
+    else:
+        return request
+
+    # parse directive and save into anonymous inventory
+    _directive_args = directive_manager.parse(directive)
+    directive_manager.loaded.put(_directive_args,
+                                 key=request_,
+                                 anonymous=True)
+
+    return request_
+
+
+def _convert_wildcard_to_directive(request):
+    ranks = dict()
+
+    with dewildcard(request) as deer:
+        req = deer.victim
+
+        def ranking(version, rank_):
+            wild_ver = deer.restore(str(version))
+            ranks[wild_ver] = rank_
+        deer.on_version(ranking)
+
+    cleaned_request = str(req)
+    # do some cleanup
+    cleaned_request = deer.restore(cleaned_request)
+
+    if len(ranks) > 1:
+        # should we support this ?
+        return None, None
+    else:
+        rank = next(iter(ranks.values()))
+
+        if rank < 0:
+            directive = "harden"
+        else:
+            directive = "harden(%d)" % rank
+
+        return cleaned_request, directive
+
+
+def bind_directives(package):
+    """Bind previously parsed directives to each variants
+    """
+    anonymous = directive_manager.loaded.storage(anonymous=True)
+
+    for variant in package.iter_variants():
+        requires = copy(variant.variant_requires)
+        requires += variant.get_requires(build_requires=True,
+                                         private_build_requires=True)
+        requires_str = set(map(str, requires))
+
+        data = {r: v for r, v in anonymous.items() if r in requires_str}
+        directive_manager.loaded.put(data, key=variant)
+
+    directive_manager.loaded.clear(anonymous=True)
+
+
+def process_directives(variant, context):
+    """Evaluate directives with resolved context
+    """
+    package = variant.parent
+    package_data = package.validated_data()
+
+    directives = directive_manager.loaded.retrieve(key=variant) or dict()
+    resolved_packages = {p.name: p for p in context.resolved_packages}
+
+    processed = dict()
+    directed = list()
+
+    def visit_directive(request):
+        directive_name, args = directives.get(str(request)) or (None, None)
+        resolved = resolved_packages.get(request.name)
+
+        if directive_name and resolved:
+            request = PackageRequest(str(
+                Requirement.construct(
+                    name=resolved.name,
+                    range=directive_manager.process(
+                        request.range, resolved.version, directive_name, args,
+                    ))
+            ))
+            directed.append(request)
+
+        return request
+
+    visitor = And(PackageRequest, Use(visit_directive))
+    schemas = {
+        "requires": [visitor],
+        "build_requires": [visitor],
+        "private_build_requires": [visitor],
+        "variants": [[visitor]],
+    }
+
+    for key, value in schemas.items():
+        if key not in package_data:
+            continue
+        schema = Schema({Optional(key): value})
+        data = schema.validate({key: package_data[key]})
+        if directed:
+            processed[key] = copy(data[key])
+            directed.clear()
+
+    directive_manager.processed.put(processed, key=variant)
+
+
+def apply_directives(variant):
+    """Patch evaluated directives to variant on install
+    """
+    directed_requires = directive_manager.processed.retrieve(key=variant)
+
+    # Just like how `cached_property` caching attributes, override
+    # requirement attributes internally. These change will be picked
+    # up by `variant.parent.validated_data`.
+    for key, value in directed_requires.items():
+        setattr(variant.parent.resource, key, value)
+        if key == "variants":
+            setattr(variant.resource, "variant_requires", value[variant.index])
+
+
+class DirectiveManager(object):
+
+    def __init__(self):
+        self._loaded = VariantDataInventory()
+        self._processed = VariantDataInventory()
+        self._handlers = dict()
+
+    @property
+    def loaded(self):
+        return self._loaded
+
+    @property
+    def processed(self):
+        return self._processed
+
+    def register_handler(self, cls, name=None, *args, **kwargs):
+        name = name or cls.name()
+        self._handlers[name] = cls(*args, **kwargs)
+
+    def parse(self, string):
+        for name, handler in self._handlers.items():
+            if string == name or string.startswith(name + "("):
+                return name, handler.parse(string[len(name):])
+
+    def to_string(self, name, args):
+        handler = self._handlers[name]
+        return handler.to_string(args)
+
+    def process(self, range_, version, name, args):
+        handler = self._handlers[name]
+        return handler.process(range_, version, *args)
+
+
+class VariantDataInventory(object):
+
+    def __init__(self):
+        self._anonymous = dict()
+        self._identified = dict()
+
+    def _hash(self, key, anonymous):
+        if anonymous:
+            return key
+        else:
+            variant = key
+            return (
+                variant.name,
+                str(variant.version),
+                variant.uuid,
+                variant.index,
+            )
+
+    def storage(self, anonymous):
+        return self._anonymous if anonymous else self._identified
+
+    def put(self, data, key, anonymous=False):
+        key = self._hash(key, anonymous)
+        storage = self.storage(anonymous)
+        storage[key] = data
+
+    def retrieve(self, key, anonymous=False):
+        key = self._hash(key, anonymous)
+        storage = self.storage(anonymous)
+        if key in storage:
+            return copy(storage[key])
+
+    def drop(self, key, anonymous=False):
+        key = self._hash(key, anonymous)
+        storage = self.storage(anonymous)
+        if key in storage:
+            storage.pop(key)
+
+    def clear(self, anonymous=False):
+        storage = self.storage(anonymous)
+        storage.clear()
+
+
+def anonymous_directive_string(request):
+    """Test use"""
+    name, args = directive_manager.loaded.retrieve(request, anonymous=True)
+    return directive_manager.to_string(name, args)
+
+
+directive_manager = DirectiveManager()
+
+# Auto register all subclasses of DirectiveBase in this module
+for obj in list(sys.modules[__name__].__dict__.values()):
+    if not inspect.isclass(obj):
+        continue
+    if issubclass(obj, DirectiveBase) and obj is not DirectiveBase:
+        directive_manager.register_handler(obj)
--- a/src/rez/utils/request_directives.py
+++ b/src/rez/utils/request_directives.py
@@ -126,7 +126,7 @@ def bind_directives(package):
     directive_manager.loaded.clear(anonymous=True)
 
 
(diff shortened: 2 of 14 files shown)


davidlatwe:
CI test in latest commit has failed but is unrelated to this PR. See [log](https://github.com/nerdvegas/rez/pull/1060/checks?check_run_id=2335550234#step:10:5403).

davidlatwe:
### Update

* Refactored the test cases to prove the functionality and possible use case.
* Moved `bind_directives` function from `PackageMaker` to `DeveloperPackage`, because the directives require the build process and `PackageMaker` generated package doesn't always ends up there.

src/rez/vendor/version/util.py
```diff
@@ -24,3 +27,126 @@ def dedup(iterable):
...
+
+
+@contextmanager
+def dewildcard(request):
+    deer = WildcardReplacer(request)
+    yield deer
+    deer.clean()
+
+
+class WildcardReplacer(object):
```
nerdvegas, line 39: This really needs docstrings. I am having a lot of trouble determining what this class actually does by looking at the code here. It isn't obvious what on_wildcard/on_version should be, nor when they are called or what they expect to be passed.

Would you not have a class that does this:

```
>>> foo = WildcardReplacer("foo-1.*+<2")
>>> foo.get_unexpanded()
Requirement('foo-1+<2')
>>> foo.get_expanded(Version("1.2.3"))
Requirement('foo-1.2+<2')
```
?

nerdvegas: I also realised that there are expansions that are no longer possible when we aren't expanding pre-build-env. For example, consider "foo-1+<2.**" - that makes sense with pre-build-env expansion but is completely nonsensical post-build-env expansion. I have an inkling that we're not taking this kind of thing into consideration here? I could be wrong though.

src/rez/utils/request_directives.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+def bind_directives(package):
+    """Bind previously parsed directives to each variants
+    """
+    anonymous = directive_manager.loaded.storage(anonymous=True)
+
+    for variant in package.iter_variants():
+        requires = copy(variant.variant_requires)
+        requires += variant.get_requires(build_requires=True,
```
nerdvegas, line 115: you're doubling up here - get_requires() already includes variant_requires

src/rez/utils/request_directives.py
```diff
@@ -0,0 +1,280 @@
...
+        return None, None
+    else:
+        rank = next(iter(ranks.values()))
+        directive = "harden(%d)" % rank
+
+        return cleaned_request, directive
+
+
+def bind_directives(package):
+    """Bind previously parsed directives to each variants
```
nerdvegas, line 109: Requires more explanation, what does 'bind' mean specifically?

src/rez/utils/request_directives.py
```diff
@@ -0,0 +1,280 @@
...
+    for variant in package.iter_variants():
+        requires = copy(variant.variant_requires)
+        requires += variant.get_requires(build_requires=True,
+                                         private_build_requires=True)
+        requires_str = set(map(str, requires))
+
+        data = {r: v for r, v in anonymous.items() if r in requires_str}
+        directive_manager.loaded.put(data, key=variant)
+
+    directive_manager.loaded.clear(anonymous=True)
```
nerdvegas, line 122: What's the `anonymous` stuff about?

Also, this seems like it may not be threadsafe?

src/rez/utils/request_directives.py
```diff
@@ -0,0 +1,280 @@
...
+        return
+
+    package = variant.parent
+    package_data = package.validated_data()
+    resolved_packages = {p.name: p for p in context.resolved_packages}
+
+    processed = dict()
+    directed = list()
+
+    def visit_directive(request):
```
nerdvegas, line 139: docstring and further explanation needed

src/rez/utils/request_directives.py
```diff
@@ -0,0 +1,280 @@
+
```
nerdvegas, line 1: The functions and classes here really require more explanation, with examples. It is quite hard to follow what's going on.

src/rez/utils/request_directives.py
```diff
@@ -0,0 +1,280 @@
...
+    def parse(self, arg_string):
+        """Parse arguments from directive syntax string"""
+        raise NotImplementedError
+
+    def to_string(self, args):
+        """Format arguments to directive syntax string"""
+        raise NotImplementedError
+
+    def process(self, range_, version, rank=None):
```
nerdvegas, line 30: I'm not sure what the idea is here. We have the abstract notion of a 'directive', which is anything added to a pkg request via `//`, but then `rank` is included here in the process method, even though the idea of rank is specific to the harden directive. It seems like there is the abstraction here but at the same time it's too specific to be useful for anything other than the harden directive?

Should it not work something like:
* directive instances are parsed out of a package.py, then somehow associated so that they can be applied back onto the package later on
* the package is updated to replace the original package request with whatever the directive says to do (so harden for eg would replace a req like foo-1.* with foo-1)
* the build resolve occurs
* the package is then updated with the directives applied (so foo-1.* becomes foo-1.4 for eg)

There needs to be some explanation like what I've just done here, giving an overview of what's going on here and how the process works as a whole. I'm having trouble assembling the bigger picture from reading the code at the moment.

src/rez/utils/request_directives.py
davidlatwe: Yeah, sorry about that 😅
doc string coming !

src/rez/utils/request_directives.py
nerdvegas: I'm also confused too because it seems like the old wilcard-based syntax, and the harden syntax, aren't quite the same thing.

What I mean is:
* foo-1.2+<2//harden(2) might expand to foo-1.2.3 (is that right?)
* but foo-1.2.*+<2 might expand to foo-1.2.3+<2 (?)

They're fairly different things, as one (harden) converts to a specific resolved version, whereas the other (wildcards) can retain the original request range. Does it work like this currently? I am not clear on what behaviour to expect.

davidlatwe: > Does it work like this currently?

Currently this kind of *multi-rank* version range is not supported and should fall back to old-style expansion.
Because, just like the example you have above, I thought the result of `foo-1.2+<2//harden(2)` could be something like `foo-1.2+<2.0.0`, but wasn't sure at the time being.

src/rez/utils/request_directives.py
davidlatwe: The `anonymous` was because, the requirement string must be normalized by the schema and turn into the Requirement object. And at that moment, there seems no way to know which package that requirement string belongs to, hence the *anonymous* state.

src/rez/utils/request_directives.py
davidlatwe: The `bind` is the process to retrieve previously parsed *anonymous* requirement string (see https://github.com/nerdvegas/rez/pull/1060#discussion_r625543219) and bind to the created package object as being identified. Maybe should be renamed to `identify_directives` :/

src/rez/utils/request_directives.py
nerdvegas: Ah right. Yeah I think this is a little trickier than I expected.

I think it only makes sense for harden() to resolve to a specific version and nothing else. Eg foo-1//harden(2) -> foo-1.3, foo-1.3+<2//harden(3) -> foo.1.3.5. You can't be guaranteed to non-ambiguously 'harden' any version range, and even if you could do it somehow, it wouldn't be intuitive. For that I think wildcard syntax is _required_. So I actually think we need both. Eg, you can do this (but not with harden directive): foo-1.*+<2 -> foo-1.3+<2. But other times you might actually want foo-1.*+<2 -> foo-1.3, and that's when you need harden. Maybe this does actually call for two different Harden directive classes.

I don't think we should ever fall back to pre-build expansion, mixing that up with post-build expansion is just going to be confusing and could lead to bugs that are difficult to track down.

nerdvegas:
Hey David,

I've had more time to think about this and here's where I'm at, let me know what you think.

**Old/new syntax (wildard vs //harden)**

Thinking about it more, I think we could avoid a lot of complexity by aiming to deprecate wildcards, and not attempt to convert them to new post-build hardening at all. We'll still support it but will print a deprecation warning. I'm not really sure that existing requirements expansion is used much, nor am I convinced that people will actually find it useful to support something like `foo-1.*+<2 -> foo-1.3+<2` (for eg).

**Directive class**

Basically it should look something like (simplified eg of harden here):
```
class HardenDirective(DirectiveBase):
    name = "harden"

    def __init__(self, request_str, rank=None):
        super(HardenDirective, self).__init__(request_str)
        self._can_harden_request():  # raise error if not possible, eg it's a conflict request (uses !)
        self.rank = rank

    def get_pre_build_request(self):
        return self.request_str

    def get_post_build_request(self, build_context):
        pkg_name = self._get_pkg_name_from_req_str()
        variant = build_context.get_resolved_package(pkg_name)
        return '%s-%s' % (pkg_name, variant.version.trim(self.rank))
```

So, a string like "foo-1.2+<2//harden(3)" would cause a HardenDirective instance to be created, and init'd with request_str="foo-1.2+<2", rank=3.

**Directive Manager**

I think where I'm getting confused is that there's a global directive manager in the first place. A developer package gets eval'd and built all within a single build process, so managing directives should happen there. Basically all the relevant directive instances need to be created there, and then you could just track them in a simple dict (ie, keyed by original request string. Then every matching request in the package's requires/build_requires/variants gets replaced with the hardened equivalent, post build resolve).


In summary I think we could simplify things a lot by:
* not converting wildcards to harden, and marking as deprecated
* create harden directives within a build process instance, and update the package based on simple find/replace of items in requires/variants/etc.

What this means is that a developer package needs to be eval'd twice for two different cases - pre and post build. However this already happens, so we can take a similar approach (I need to look back at this code to refresh myself on the details): https://github.com/nerdvegas/rez/wiki/Package-Definition-Guide#available-objects. See how early-bound funcs are already eval'd twice. This is a kinda similar situation.

There are some other sutble details that we'll have to figure out, I won't go over them just yet as I need to think about them a bit more :)

Cheers
A

davidlatwe:
Thanks Allan,

> I think we could avoid a lot of complexity by aiming to deprecate wildcards, and not attempt to convert them to new post-build hardening at all.

Yeah, that would simplified a lot :D
Just to be sure, the wildcards stay as *pre-build* expansion, and only `//harden` gets expands in post-build time, correct ?

> ..., nor am I convinced that people will actually find it useful to support something like foo-1.*+<2 -> foo-1.3+<2 (for eg).

Agree, I was struggling for this kind of cases, wondering if even anyone using this kind of expansion and how would that even get harden as.

About the global directive manager,

> A developer package gets eval'd and built all within a single build process, so managing directives should happen there.

Yes, that's the most ideal. But the issue *was*, that the developer package gets loaded before the creation of build process, and we have to normalize those directive/wildcard syntax or the schema validation won't pass due to it's not a valid syntax for `PackageRequest`.

But. If now we are going to let wildcard expansion stay as is, and defining the harden behavior on ranged request as `foo-1.3+<2//harden(3) -> foo.1.3.5`, then the implementation will simplified a lot and trimming the `//` mark would be the only work when loading the developer package before build, which can be implemented in `PackageRequest` class in minimum performance cost and drop the global directive manager and do all the rest in build process.

nerdvegas:
Ok thanks for the explanation that makes sense.


"""Just to be sure, the wildcards stay as pre-build expansion, and only
//harden gets expands in post-build time, correct ?"""

Yup correct.


"""...But the issue *was*, that the developer package gets loaded before
the creation of build process..."""

So then, what if directives (and their mappings to the original request
they correspond to) are stored in the DeveloperPackage itself (ie the dev
pkg actually parses its request strings and creates directives itself)?
That seems less hacky - at load time the package can replace its //harden
requests with whatever the harden directive tells it to (ie, just drop the
//harden(...)). Then requests can be replaced when within the build context
later, as part of that second evaluation phase. I think really this is the
way to do it, that way nothing changes in PackageRequest (or anywhere
outside of DeveloperPackage really).

davidlatwe:
I'll be damned.

> So then, what if directives (and their mappings to the original request
they correspond to) are stored in the DeveloperPackage itself (ie the dev
pkg actually parses its request strings and creates directives itself)?

Just found that I can get the request string before schema validation from here, right before sending `data` into `create_package`
https://github.com/nerdvegas/rez/blob/3f66e9ca1a0db227af608757e8dced08e88bd461/src/rez/developer_package.py#L105-L106

I didn't realize that !

> I think really this is the
way to do it, that way nothing changes in PackageRequest (or anywhere
outside of DeveloperPackage really).

Yeah, now I learnt the magic, will re-implement this soon !

src/rez/developer_package.py
```diff
@@ -152,6 +170,14 @@ def get_reevaluated(self, objects):
...
             return self.from_path(self.root)
 
+    # TODO: should have a better name
+    def evaluate_directives(self, build_context, variant_index=None):
+        variant = self.get_variant(variant_index)
+        override = evaluate_directive_requires(variant, build_context)
+        # TODO: if no directive, skip re-evaluate variant
+        package = self._from_path(self.root, override=override)
+        return package.get_variant(variant_index)
+
```
davidlatwe, line 180: Not sure if there is a better way to re-apply resolved directives back to variant.
I have tried to override the variant data/resource but failed, it looks like we must re-evaluate it again with passing resolved directive request as *override*.

src/rezplugins/build_process/local.py
```diff
@@ -414,7 +417,12 @@ def cancel_variant_install():
         # add release info to variant, and install it into package repository
         release_data = self.get_release_data()
         release_data["release_message"] = release_message
+
+        variant = variant.parent.evaluate_directives(context,
+                                                     variant.index)
+
         variant_ = variant.install(release_path, overrides=release_data)
+
```
davidlatwe, line 425: I have also tried to pass the resolved directive data to `variant.install(path, overrides=overrides)`, but won't work.

davidlatwe:
So now the new approach is much simpler :
1. parse requirement strings and extract directives before developer package's creation
2. redo step 1. in package pre-processing
3. build process run
4. After *local* build process has completed the build successfully, re-evaluate the variant to apply directives with previously resolved build context. The variant re-evaluation will re-run Step 1. & 2. again but with the evaluated directive data.

Wasn't sure if that variant re-evaluation can be avoided, but currently cannot see an alternative that is safe and simple.

davidlatwe:
Found a way to apply harden directives back to variant before building variant payload and without re-running whole developer package loading process. 🎉

src/rez/build_process.py
```diff
@@ -223,9 +223,26 @@ def get_package_install_path(self, path):
...
+        # Note that this re-evaluated variant is ONLY used here, for the purposes
+        # of creating the build context. The variant that is actually installed
+        # is the one evaluated where 'building' is False.
+        #
+        re_evaluated_package = variant.parent.get_reevaluated({
+            "building": True,
+            "build_variant_index": variant.index or 0,
+            "build_variant_requires": variant.variant_requires
+        })
+        re_evaluated_variant = re_evaluated_package.get_variant(variant.index)
```
davidlatwe, line 242: Moving this package and variant re-evaluation into build-process base class `create_build_context` seems better and post-context-resolve re-evaluation can also happen here as well. Pre- and post-re-evaluations are in the same code block now.

src/rez/developer_package.py
```diff
@@ -152,6 +164,34 @@ def get_reevaluated(self, objects):
...
+        # re-evaluate variant
+        #
+        package_cls = type(self)
+        re_evaluated_package = create_package(self.name, data, package_cls)
+        re_evaluated_variant = re_evaluated_package.get_variant(variant.index)
+        # swapping resource
+        variant._parent = re_evaluated_package
+        variant.context = re_evaluated_package.context
+        variant.wrapped = re_evaluated_variant.wrapped
+
```
davidlatwe, line 194: About this *variant resource swapping*, it is here because, the `Variant` object doesn't really carry the variated data by itself, but asking for it's parent package that has been evaluated for that variant. This behavior is okay for installed package, but not for developer package in build-time.

Maybe we should re-implement `DeveloperPackage.iter_variants` ? That could spawn variant with re-evaluated package as it's parent. With that, I think the code above could change to just retrieve the variant's parent package and update it's data.

src/rezplugins/build_process/local.py
davidlatwe: Update:
This actually works, but the unittest is building same package twice, one in `clean=True`, and one not. Which means in second build, there's an existing package and the `overrides` cannot have the precedence to update `requires`.

src/rez/developer_package.py
davidlatwe: Sorry, scratch that.

Got some time to revisit this again with a bit more understanding about how underlaying works and re-try to avoid current resource swapping hack for applying directives.

I have re-tried with the `variant.install(overrides=data)`, but the overrides is applied after checking existing package data, so the directives cannot be applied if previous installed exists.

https://github.com/nerdvegas/rez/blob/3f66e9ca1a0db227af608757e8dced08e88bd461/src/rezplugins/package_repository/filesystem.py#L1366-L1369

Then, I re-tried another route, updating `variant.parent` (aka package) data.

In `variant.install`, the package data is retrieved by `variant.parent.validated_data`, which does this
https://github.com/nerdvegas/rez/blob/3f66e9ca1a0db227af608757e8dced08e88bd461/src/rez/utils/data_utils.py#L627-L637

Based on schema key, the data was firstly collected from the object attributes, then from `_data`, but only if the key is not yet been collected.

So the directives must be applied back as object attributes :
```python
        # self.resource._data.update(data)  # won't work
        for k, v in data.items():
            setattr(self.resource, k, v)
```

After knowing these, using `overrides` to apply directives is not a good option, but using `setattr` is still not clear enough.

Re-implement `DeveloperPackage.validated_data` seems better, in there, we could clearly define which data in `self.resource._data` can take precedence, for example:
```python
def validated_data():
    data = super(DeveloperPackage, self).validated_data()
    data.update(data.pop("buildtime_updates", {}))
    return data
```

davidlatwe:
Somehow, the implementation pass the test, but not working with actual `rez-build` command. In `rez-build` command, `DeveloperPackage.directives` become an empty dict when the build process is going to evaluate them. But everything fine in unit-test. Investigating this later.

Side note, late bound supports are not added into schema yet.

davidlatwe:
> Somehow, the implementation pass the test, but not working with actual rez-build command.

Found the problem, it was my `rezconfig.package_preprocess_function`, which triggered the bug and now it's fixed.

> Side note, late bound supports are not added into schema yet.

late bound support added.

src/rez/tests/test_directives.py
```diff
@@ -0,0 +1,180 @@
...
+
+        self.dev_repo.add("bar", version="1.0.0", **_k)
+        self.dev_repo.add("foo", version="1.0.0", requires=requires, **_k)
+
+        self._test_build("bar", "1.0.0")
+        self._test_build("foo", "1.0.0")
+        foo = get_package("foo", "1.0.0", paths=[self.install_root])
+        # no harden, because there were no requires when building is True
+        self.assertEqual("bar-1", str(foo.requires[0]))
+
```
davidlatwe, line 132: Currently not sure how to deal with this, the early bound requires has a directive requirement but not appeared in build.
What I have in mind right now is to re-evaluate the package once more with `building=False`.

davidlatwe: And, harden requirement in late bound should be consider as missed used, correct ? wildcard require could use `expand_requirements` function, but since the harden is based on resolved packages, which means it will being harden anyway.

maxnbk:
I just want to say this work is amazing @davidlatwe . I hope this effort gets completed soon. Great work.

nerdvegas:
hey @davidlatwe the need for this feature is gonna become more pressing soon (more on this soon!). I plan on diving back into this, we'll get it merged. Thanks for putting in the work, this is a tough one.

PR #1067, opened 2021-05-04
Fix #1065

### LINKED ISSUE #1065: windows package cache root switch does only work with "{root}" not this.root
The package cache root switch does not work with this.root in an envvar under windows.
Only if the the var is being set with "{root}/whatnot" the root switch to the local payload works.

The payload copy does work, only when resolving it wont use the local data even with REZ_READ_PACKAGE_CACHE=1
Changing the package from this.root to "{root}" will make it work.

davidlatwe:
This close #1065, as a propose fix.

Side note, was trying to update the test, but somehow I just cannot pass this one as package never being cached in resolve. However `pkg-cache` does work in my production setup.
https://github.com/nerdvegas/rez/blob/3f66e9ca1a0db227af608757e8dced08e88bd461/src/rez/tests/test_package_cache.py#L121-L123

### DIFF
--- a/src/rez/packages.py
+++ b/src/rez/packages.py
@@ -337,6 +337,7 @@ class Variant(PackageBaseResourceWrapper):
         _check_class(resource, VariantResource)
         super(Variant, self).__init__(resource, context)
         self._parent = parent
+        self._cached_root = None
 
     # arbitrary keys
     def __getattr__(self, name):
@@ -347,6 +348,10 @@ class Variant(PackageBaseResourceWrapper):
 
     def arbitrary_keys(self):
         return self.parent.arbitrary_keys()
+
+    @property
+    def root(self):
+        return self._cached_root or self.resource.root
 
     @cached_property
     def qualified_package_name(self):
@@ -426,6 +431,27 @@ class Variant(PackageBaseResourceWrapper):
             requires = requires + (self.private_build_requires or [])
 
         return requires
+
+    def is_cached(self):
+        """True if variant payload has been redirected to the cached copy
+
+        Returns:
+             Bool
+        """
+        return bool(self._cached_root)
+
+    def set_cached(self, cached_root):
+        """Redirect variant payload root to the cached copy
+
+        This should be set by `ResolvedContext`.
+
+        Args:
+            cached_root (str): Cached variant root path
+
+        Returns:
+            None
+        """
+        self._cached_root = cached_root
 
     def install(self, path, dry_run=False, overrides=None):
         """Install this variant into another package repository.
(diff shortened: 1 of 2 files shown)


nerdvegas:
I'm gonna take a look at this now. I'm a bit concerned about that test not passing in all cases, and I wonder if it's related to the fact that `Variant.root` has been changed to match the cached root. 

This is a tricky one - I'm not sure we should actually change Variant this way. Root is one of the attributes that define a variant's 'handle' (and this is how it's uniquely identified as a resource). Changing root may introduce subtle bugs since it may then not be recognised as the same resource object, and I have a feeling this may account for the test fail you're seeing.

I'm thinking we should do this instead:
* Have Variant not be aware of caching at all. Its root should always be the non-cached root
* Change VariantBinding instead, so that `some_variant.root` in a commands section _does_ equate to the cached root.

It does raise the question - if you have a Variant from a context, via rez API, should its root be the cached root, or not? I think it's a much safer bet to say no to this. If someone does actually need this for whatever reason, we could add a `ResolvedContext.get_variant_root`, and explain there that its return value may differ from `Variant.root` due to package caching.

Even if it seems a little odd that Variant.root wouldn't match the cached root for a context, consider that it would be even more strange to have a Variant whose attributes do not match its handle, which is what we have here (and even if that's not accounting for the issue you're getting with the test, I think there's just too much scope for it causing issues elsewhere).

I'm gonna give this alternate approach a go in a separate branch just now, I'll let you know how it goes.

cheers
A

nerdvegas:
Pls see https://github.com/nerdvegas/rez/pull/1076

davidlatwe:
Hey Allan,
Thanks ! And yeah, #1076 seems a much better fix, good point on changing `VariantBinding`.

> if you have a Variant from a context, via rez API, should its root be the cached root, or not? I think it's a much safer bet to say no to this

Agree. :)

> If someone does actually need this for whatever reason, we could add a `ResolvedContext.get_variant_root`

Or :
```python
from rez.package_cache import PackageCache
from rez.config import config

pkg_cache = PackageCache(config.cache_packages_path)
pkg_cache.get_cached_root(variant)
```

But the method `PackageCache.get_cached_root` will update file access/modify time if cache exists, maybe `ResolvedContext.get_variant_root` can skip that when querying ?

nerdvegas:
Closing as fixed by https://github.com/nerdvegas/rez/pull/1076 instead.

davidlatwe:
The merged branch showing in release note is incorrect. :O

nerdvegas:
Ah cheers, I use the wrong PR to generate that report, whoops.

PR #1076, opened 2021-05-18
switch to cached root in variant binding

### LINKED ISSUE #1065: windows package cache root switch does only work with "{root}" not this.root

nerdvegas:
-added test to verify that refs like resolve.mypkg.root are correct in commands
-changed ResolvedContext.execute_rex_code() to be post-actions callback

Fixes #1065 

This is an alternative fix to https://github.com/nerdvegas/rez/pull/1067 please see comment there.

### DIFF
--- a/src/rez/resolved_context.py
+++ b/src/rez/resolved_context.py
@@ -1313,7 +1313,7 @@ class ResolvedContext(object):
                                   parent_environ=parent_environ,
                                   command='',  # don't run any command
                                   block=False,
-                                  actions_callback=_actions_callback,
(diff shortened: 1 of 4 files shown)


nebukadhezer:
Thanks for the fix !

PR #1077, opened 2021-05-18
Refactor: Split add_standard_build_actions introducing add_pre_build_commands

Tilix4:
This PR fixes Issue #974.

Based on the indications from https://github.com/nerdvegas/rez/pull/975#issuecomment-728327928

### DIFF
--- a/src/rez/build_system.py
+++ b/src/rez/build_system.py
@@ -284,7 +284,6 @@ class BuildSystem(object):
 
         This includes:
         - Setting a standard list on env-vars;
-        - Executing pre_build_commands(), if the package has one.
         """
         # set env vars
         env_vars = cls.get_standard_vars(
@@ -302,10 +301,9 @@ class BuildSystem(object):
     @classmethod
     def add_pre_build_commands(cls, executor, variant, build_type,
                                    install, build_path, install_path=None):
-        """Perform build actions common to every build system.
+        """Perform build commands common to every build system.
 
         This includes:
-        - Setting a standard list on env-vars;
         - Executing pre_build_commands(), if the package has one.
         """
         from rez.utils.data_utils import RO_AttrDictWrapper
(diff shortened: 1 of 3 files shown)


Strangenoise:
Would be greate, we have the same build problem on windows.

src/rezplugins/build_system/custom.py
```diff
@@ -207,6 +207,19 @@ def _add_build_actions(cls, executor, context, package, variant,
             install_path=install_path
         )
 
+    @classmethod
+    def _add_pre_build_commands(cls, executor, variant, build_type,
```
nerdvegas, line 211: drop these `_add_pre_build_commands` methods and just call `add_pre_build_commands` directly.

It's different for `_add_build_actions` because those may or may not just comprise of the 'standard' build actions (it's feasible that a specific build system may want to set an env var for eg). However, in the case of `pre_build_commands`, there's nothing about that that is specific to a build system.

src/rez/build_system.py
```diff
@@ -301,6 +299,17 @@ def add_standard_build_actions(cls, executor, context, variant, build_type,
         for var, value in env_vars.items():
             executor.env[var] = value
 
+    @classmethod
+    def add_pre_build_commands(cls, executor, variant, build_type,
+                                   install, build_path, install_path=None):
+        """Perform build actions common to every build system.
+
+        This includes:
```
nerdvegas, line 307: This docstring is wrong, it describes what `add_standard_build_actions` used to do.

src/rez/build_system.py
```diff
@@ -286,8 +286,6 @@ def add_standard_build_actions(cls, executor, context, variant, build_type,
         - Setting a standard list on env-vars;
         - Executing pre_build_commands(), if the package has one.
```
nerdvegas, line 287: This comment is no longer true

nerdvegas:
Some minor issues but I should be able to get this merged once they're resolved, cheers

Tilix4:
I corrected what you requested :)

Issue #865, opened 2020-03-19
fail-graph not showing true root of the conflict


brentVilla:
We found cases where the graph generated to show conflicts doesn't show the root cause of the conflict.  Suppose we have three products, foo, bar, and baz.  bar depends on foo and baz depends on both foo and bar.  In other words, we have package.py with the following:
```
% cat foo/1.0/package.py
name = 'foo'
version = '1.0'

% cat foo/2.0/package.py
name = 'foo'
version = '2.0'

% cat bar/1.0/package.py
name = 'bar'
version = '1.0'
requires = ['foo-1.0']

% cat bar/2.0/package.py
name = 'bar'
version = '2.0'
requires = ['foo-2.0']

% cat baz/1.0/package.py
name = 'baz'
version = '1.0'
requires = ['foo-1.0', 'bar-1.0']

% cat baz/2.0/package.py
name = 'baz'
version = '2.0'
requires = ['foo-2.0', 'bar-2.0']
```
The following generates a conflict as expected because baz-1.0 requires foo-1.0 and bar-1.0 which conflicts with the foo-2.0 and bar-2.0 in the request:
rez-env foo-2.0 bar-2.0 baz-1.0 --fail-graph

However, the fail graph won't show baz at all although it's the one in conflict.
![resolve-dot-JeEyIg](https://user-images.githubusercontent.com/6012216/77092030-986ddf80-69c6-11ea-9ef6-78d362bb085d.png)

Now, if I drop the intermediate bar-2.0 request, I will get a more helpful failure graph that shows baz as being part of the conflict.
rez-env foo-2.0 baz-1.0 --fail-graph
![resolve-dot-h9lgdr](https://user-images.githubusercontent.com/6012216/77092279-da972100-69c6-11ea-8f16-855af9ef3e9f.png)

This is a trivial case for this issue.  But in more complex resolves, it can be quite difficult and time consuming to modify the request to get at the root cause of the conflict.

nerdvegas:
Thanks very much for the repro. There has been a graph bug lurking for a long time, and this simple repro should help a lot in tracking it down.

davidlatwe:
Just saw this issue and thought this might worth mentioning.

Changing request order from
```
rez-env foo-2.0 bar-2.0 baz-1.0 --fail-graph
```
into
```
rez-env baz-1.0 foo-2.0 bar-2.0 --fail-graph
```
will also get more helpful graph.

[...]

Issue #856, opened 2020-02-21
.ignore pkg repo functionality not documented in wiki


nerdvegas:
See title.
https://github.com/nerdvegas/rez/blob/master/src/rezplugins/package_repository/filesystem.py#L456

PR #1086, opened 2021-05-25
Maximize fail graph

davidlatwe:
This is a workaround for #865. Which will try to resolve a failing context with a re-ordered request for maximizing fail graph, take the example in #856, this PR will generate the following graph when `--fail-graph` is set.

image


davidlatwe:
Closing this for the [comment](https://github.com/nerdvegas/rez/issues/865#issuecomment-848287114).

PR #1087, opened 2021-05-26
Fix conflict fail graph #865

### LINKED ISSUE #865: fail-graph not showing true root of the conflict

davidlatwe:
This fix #865.

As the example in #865, here's the new fail graph :

image

But different from the workaround #1086 which has been pointed out is not reliable, this PR is done by fixing resolve graph generation.

The root cause was, scope node in graph is being selected to form a conflict edge, without checking if scope's request is really conflicted. Which may leads to the case like #865 had, scope node takes the precedence and the real conflicting request node gets left out and pruned.

### DIFF
--- a/src/rez/solver.py
+++ b/src/rez/solver.py
@@ -1580,6 +1580,7 @@ class _ResolvePhase(_Common):
         failure_nodes = set()
         request_nodes = {}  # (request, node_id)
         scope_nodes = {}  # (package_name, node_id)
+        scope_requests = {}  # (node_id, request)
 
         # -- graph creation basics
 
@@ -1670,6 +1671,7 @@ class _ResolvePhase(_Common):
 
             id_ = _add_node(label, color, style)
             scope_nodes[scope.package_name] = id_
+            scope_requests[id_] = scope.package_request
             return id_
 
         def _add_reduct_node(request):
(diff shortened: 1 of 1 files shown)


nerdvegas:
Fantastic work, cheers David

PR #1092, opened 2021-06-03
Allow trailing comma in Legacy Metadata

bfloch:
This fixes rez-pip install for packages that have a trailing comma on
version requirements.

For example openpyxl will now install properly with this line:
```
Requires-Python: >=3.6,
                      ^ Parser will fail without the patch
```

Not sure if you rather want to wait for upstream.


bfloch:
Upstream issue:
https://bitbucket.org/pypa/distlib/issues/148/trailing-comma-results-in-syntax-error

PR #1094, opened 2021-06-04
Using tox to run flake8

j0yu:
:eyes:  _Look, there's a tox.ini! Maybe we can use that!_

Just a side thing to implement and use `py37-flake8` routine as it's currently partially implemented and not actively used.

I've made sure to update the:
- Ignore rules as per the Flake8 CI
- Check that the file path exclusion is fully compatible with the previous `find ...` rules (hacked a local venv's `flake8` source code to inspect and compare the file paths found are the same)

Not even sure if this is a good idea, heck, do we even want to go down the `tox` route for test, docs building and linting when developing rez source code locally? Perhaps `tox.ini` wasn't supposed to be used anyways?

### DIFF
--- a/.github/workflows/flake8.yaml
+++ b/.github/workflows/flake8.yaml
@@ -33,19 +33,5 @@ jobs:
         with:
           python-version: 3.7
 
-      - name: Install Dependencies
-        run: |
-          pip install flake8
-
-      - name: Run flake8
-        run: >-
-          find -name '*.py'
-          -not -path './build_utils/*'
-          -not -path './rez/vendor/*'
-          -not -path './rez/data/*'
-          -not -path './support/*'
-          -not -path './rez/backport/*'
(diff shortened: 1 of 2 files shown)


nerdvegas:
I dunno I feel like tox is a bit redundant here, it's not really solving any issue we were having. Imo we should drop tox presently, but we could introduce it later if it proves useful. Thoughts?

j0yu:
Yep, dropping tox sounds good to me. We could maybe introduce [nox](https://nox.thea.codes/en/stable/) later replacing tox too when the time comes to setup a single test/docs/lint entrypoint

I'm considering closing this PR now, it was just a quick throw-away suggestion anyways

nerdvegas:
Closing, let's keep it in mind if tox makes sense at a later date though, cheers

PR #1111, opened 2021-07-29
Ensure command string is escaped correctly when executing in windows cmd shell

afrocircus:
### Problem:
The following does not execute correctly when run in a windows cmd prompt.
```
rez-env python -- echo "<hey>"
The system cannot find the file specified.
```

This is because the batch file that is written out is without the escaped chars.
```
set REZ_ENV_PROMPT=%REZ_ENV_PROMPT%$G
echo %PATHEXT%|C:\Windows\System32\findstr.exe /i /c:".PY">nul || set PATHEXT=%PATHEXT%;.PY
(call )
call C:\..\AppData\Local\Temp\rez_context_fiqpunrh\context.bat
set REZ_STORED_PROMPT_CMD=$P$G
set PROMPT=%REZ_ENV_PROMPT% $P$G
echo <Hey>
exit %errorlevel%
```

what we need is the batch file to contain.
`echo ^<Hey^>`

### Solution:
Ensure that the command run via rez-env is escaped correctly when executing in a windows cmd shell.

### DIFF
--- a/src/rez/tests/test_shells.py
+++ b/src/rez/tests/test_shells.py
@@ -197,6 +197,21 @@ class TestShells(TestBase, TempdirMixin):
 
     @per_available_shell()
     @install_dependent()
+    def test_rez_env_output_special_chars(self):
+        # here we are making sure that running a command with special chars
+        # via rez-env prints exactly what we expect.
+
+        # Assumes that the shell has an echo command, build-in or alias
+        cmd = [os.path.join(system.rez_bin_path, "rez-env"), "--", "echo", "<hey>"]
+        process = subprocess.Popen(
+            cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE, universal_newlines=True
+        )
+        sh_out = process.communicate()
+        self.assertEqual(sh_out[0].strip(), "<hey>")
+
+    @per_available_shell()
+    @install_dependent()
     def test_rez_command(self):
         sh = create_shell()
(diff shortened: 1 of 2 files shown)


nerdvegas:
Ok I've looked into this, it is more nuanced that it appears.

I think escaping the command like this is not the correct approach. The problem is that you cannot know which part of any given command may need to be escaped. Consider the command on linux: `rez-env -- bash -c 'anything-in-here'`. If we escape any part of `anything-in-here`, we could break the command, or backwards compatibility, or both.

Furthermore, the existing bash shell works even without escaping. Ie this works: `rez-env -- echo "!hey"` (but running `echo !hey` directly will fail).

The reason this works is because the Shell class's `join` function is applied to the command, see https://github.com/nerdvegas/rez/blob/master/src/rez/resolved_context.py#L1388. I think the problem is cmd's implementation of this, see https://github.com/nerdvegas/rez/blob/master/src/rezplugins/shell/cmd.py#L356. I think this function is not adding quotes to strings that need it, such as `<hey>`.

I don't recall why cmd has a different `join` implementation and why it doesn't use `shlex_join` as other shells do. Interestingly though I did find a bug in the unix shell impl, this fails: `rez-env -- echo '$>'`.

I'm going to close this PR due to the issues I've raised here, but the issue isn't resolved, I think Cmd.join needs fixing (I don't work on Windows though so my own attempt at a fix will probably be hamfisted).

nerdvegas:
https://github.com/nerdvegas/rez/pull/1115

PR #1115, opened 2021-08-03
quoting fix

### LINKED ISSUE #1114: bug in command quoting
The following fails and should not:
```
]$ rez-env -- echo '$>'
```

nerdvegas:
Fixes #1114 

Note that Windows test is now failing and that is intended, there's a preexisting bug there that we need to fix.

Related: https://github.com/nerdvegas/rez/pull/1111

### DIFF
--- a/src/rez/shells.py
+++ b/src/rez/shells.py
@@ -285,7 +285,7 @@ class Shell(ActionInterpreter):
         Returns:
             A string object representing the command.
         """
-        raise NotImplementedError
+        return shlex_join(command)
 
 
 class UnixShell(Shell):
@@ -515,10 +515,6 @@ class UnixShell(Shell):
     @classmethod
     def get_all_key_tokens(cls, key):
         return ["${%s}" % key, "$%s" % key]
-
-    @classmethod
-    def join(cls, command):
-        return shlex_join(command)
 
     @classmethod
     def line_terminator(cls):
--- a/src/rez/tests/test_shells.py
+++ b/src/rez/tests/test_shells.py
@@ -195,6 +195,12 @@ class TestShells(TestBase, TempdirMixin):
             sh_out = process.communicate()
             self.assertEqual(sh_out[0].strip(), txt)
 
+        # please note - it's no coincidence that there are no substrings like
+        # '$you' here. These would expand to the equivalent env-var (as
+        # intended), which would be an empty string. We're not testing that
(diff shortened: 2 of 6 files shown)


davidlatwe:
Continuing https://github.com/nerdvegas/rez/pull/1116#issuecomment-893049448 here. :)

In #1116, I changed the test to match result with shlex joined input, and it passed in Windows but actually failed in other OS.

```python
import subprocess
from shlex import join  # new in Python 3.8, just for quick testing

>>> subprocess.check_output(join(['echo', 'hey you']), shell=True, text=True).strip()
# Windows CMD
"'hey you'"
# In my WSL (Ubuntu)
'hey you'

# however..
>>> join(['echo', 'hey you'])
"echo 'hey you'"  # same result for both platform
```

So maybe we should match the result with actual shell output plus shlex joined input ? Like
```python
def popen(args, shell=False):
    process = subprocess.Popen(
        stderr=subprocess.PIPE, universal_newlines=True, shell=shell
    )
    return sh_out[0].strip()

a = popen([rez_env_exec, "--", "echo", txt])
b = popen(shlex_join(["echo", txt]), shell=True)

self.assertEqual(a, b)
```
Kind of stupid but seems fair ?

Just for proving the concept, I have updated my [branch](https://github.com/davidlatwe/rez/commits/issue_1114-command-quoting) and CI testing passed in both [Ubuntu](https://github.com/davidlatwe/rez/actions/runs/1102761242) and [Windows](https://github.com/davidlatwe/rez/actions/runs/1102761240).

nerdvegas:
I dunno, that seems oddly circular... all we're really doing at that point is verifying that rez-env is running the given command in a shell. Also isn't there no guarantee that the current shell (ie whatr raw subprocess will use if shell=True) will match the shell rez is using?

I'm starting to think we simply need a cmd caveat here.

davidlatwe:
Yeah.. :(
Or, how about just validate the test cases with hard coded results ? Like..

```python
    @per_available_shell()
    @install_dependent()

        cases = {
            "hey": {},
            "hey you": {"cmd": '"hey you"'},
            ...
        }

        def _test(txt):
            ...
            expect = cases[txt].get(config.default_shell, txt)
            self.assertEqual(sh_out[0].strip(), expect)

        _test("hey")  # simple case
        _test("hey you")  # with a space
```

PR #1116, opened 2021-08-04
quoting fix (continue on #1115, tested on Windows)

### LINKED ISSUE #1114: bug in command quoting

davidlatwe:
As title, this should fix #1114 based on #1115, and tested on my Windows machine.

* Base on the finding in https://github.com/nerdvegas/rez/pull/1111#issuecomment-891506931, I replaced `subprocess.list2cmdline` with `rez.util.shlex_join` for Windows shells.

* In test `rez.tests.test_shells.TestShells.test_rez_env_output`, I also changed the assertion to compare process output with shlex joined input. Thought that input string is being shlex joined on the way into subprocess so we should compare the same thing. Here's the original (#1115) testing output on my Windows :
```
___________________________________________________ TestShells.test_rez_env_output ____________________________________________________

self = <rez.tests.test_shells.TestShells testMethod=test_rez_env_output>

    @per_available_shell()
    @install_dependent()
        def _test(txt):
            # Assumes that the shell has an echo command, build-in or alias
            binpath = os.path.join(system.rez_bin_path, "rez-env")
            args = [binpath, "--", "echo", txt]

            process = subprocess.Popen(
                stderr=subprocess.PIPE, universal_newlines=True
            )
            self.assertEqual(sh_out[0].strip(), txt)

        # please note - it's no coincidence that there are no substrings like
        # '$you' here. These would expand to the equivalent env-var (as
        # intended), which would be an empty string. We're not testing that
        # here though.
        #

        _test("hey")  # simple case
>       _test("hey you")  # with a space

test_shells.py:205:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
test_shells.py:196: in _test
    self.assertEqual(sh_out[0].strip(), txt)
E   AssertionError: '"hey you"' != 'hey you'
E   - "hey you"
E   ? -       -
E   + hey you
E    (in shell 'cmd')
-------------------------------------------------------- Captured stdout call ---------------------------------------------------------

testing in shell: cmd...
======================================================= short test summary info =======================================================
FAILED test_shells.py::TestShells::test_rez_env_output - AssertionError: '"hey you"' != 'hey you'
========================================================== 1 failed in 1.28s ==========================================================
```

### DIFF
--- a/src/rez/tests/test_shells.py
+++ b/src/rez/tests/test_shells.py
@@ -195,6 +195,12 @@ class TestShells(TestBase, TempdirMixin):
             sh_out = process.communicate()
             self.assertEqual(sh_out[0].strip(), txt)
 
+        # please note - it's no coincidence that there are no substrings like
+        # '$you' here. These would expand to the equivalent env-var (as
+        # intended), which would be an empty string. We're not testing that
+        # here though.
+        #
+
         _test("hey")  # simple case
         _test("hey you")  # with a space
         _test("<hey>")  # special characters
--- a/src/rez/tests/test_shells.py
+++ b/src/rez/tests/test_shells.py
@@ -183,17 +183,25 @@ class TestShells(TestBase, TempdirMixin):
     @per_available_shell()
     @install_dependent()
     def test_rez_env_output(self):
-        # here we are making sure that running a command via rez-env prints
-        # exactly what we expect.
+        def _test(txt):
+            # Assumes that the shell has an echo command, build-in or alias
+            binpath = os.path.join(system.rez_bin_path, "rez-env")
+            args = [binpath, "--", "echo", txt]
 
-        # Assumes that the shell has an echo command, build-in or alias
-        cmd = [os.path.join(system.rez_bin_path, "rez-env"), "--", "echo", "hey"]
-        process = subprocess.Popen(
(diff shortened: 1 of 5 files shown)


nerdvegas:
Hey David, this isn't working, see https://github.com/nerdvegas/rez/pull/1116/checks?check_run_id=3245129015.
Question is, why did the test succeed? Actually it looks like pytest-based `rez-selftest` is exiting with 0, that'll have to be fixed(!!).

In any case this doesn't work because the output should not match the shlex-joined output. 

Example (bash):

```
]$ echo "hey you"
hey you
```

When rez creates the runtime to run a command, it writes shell script out to a `rez-shell.sh` file. So, it needs to literally print `echo "hey you"` into that script, which is why shlex_join is used. However, the output should be `hey you`, not `"hey you"`.

I am assuming that the required escaping is different in cmd, it's picking up the double quotes as part of the string to be printed, which is unintentional.

I may get a look in at 1117 today.

https://github.com/nerdvegas/rez/issues/1117

Cheers
A

Issue #892, opened 2020-05-22
rez-pip installs dist-info dir to root, causes probs with some pkgs


nerdvegas:
Some pip packages (eg `importlib_metadata`) expect the dist-info dir to be present in the root dir of the python package, which would be under `python/` in a rez-pip'd package. Rez currently installs dist-info dir one level above this, at the variant root.

Since there is no real need to store dist-info at the variant root, we should move it to python/ in order to fix these cases.

zachlewis:
I discovered the same thing, and came here suggest doing exactly this. You beat me to it by two hours.

On a related note, I've discovered that tricking pip into doing a --user install to the rez build install path seems to automatically solve many (most? all?) of the package deployment problems I've encountered with rez-pip in one fell swoop, like this dist-info placement issue, and others that I'm controlling for via #866. 

I've been using this pattern a lot lately when I need to manually rezzify certain python packages :

```
...

def pre_build_commands():
    env.PYTHONUSERBASE = '{build.install_path}'

build_command = "python -m install {root} --no-deps --user --no-compile --global-option='build_ext'"

def commands():
    env.PATH.append('{root}/bin')
    env.PYTHONPATH.append('{root}/lib/python{resolve.python.version.major}.'
                           '{resolve.python.version.minor}/site-packages')

```

As you can see, the tradeoff is a kind of gnarly relative location for the python packages themselves; although I suppose it would suffice to symlink the site-packages path to the pkg_root/python at build time and continue appending '{root}/python' to PYTHONPATH, and  nobody would be the wiser...

Maybe you guys have already considered this, but the $PYTHONUSERBASE thing was news to me.


Cheers!

instinct-vfx:
We are using PYTHONUSERBASE in our python package together with generating a .pth file. That makes python start way quicker with a lot of Packages. I still owe a write up on that. The path could be turned into the standard path in rez-pip easily enough though. Anyone else looked into that already?

spsalefeve:
Hey, there. Can you tell me if someone is working on this issue?

ColinKennedy:
Hello all, I'm having the same issue with rez-pip'ed packages due to this issue and would like a fix for this. @instinct-vfx as far as I know, [.pth files are possibly being deprecated](https://bugs.python.org/issue33944) so I'd like to move forward with @nerdvegas initial suggestion of moving the dist-info folder into `python` folder, unless there's a better approach.

[...]

PR #1123, opened 2021-09-07
Added a configuration option for rez-pip dist-info remapping

smaragden:
I created this PR to get some traction on #892 .

This PR keeps the current behavior, but lets us configure the current "special-case" in rez-pip that always copies dist-info files to the rez-package root.

By setting `pip_install_override_dist_info_remap = False`, dist-info files will end up alongside the python package as some tools expect.

In my opinion the better solution would be to just change the default behavior and copy the files to where they are expected.


nerdvegas:
I'm fine with this being default behaviour rather than configurable, unless there are objections?

nerdvegas:
Closing in favor of https://github.com/nerdvegas/rez/pull/1128 as there are no objections to hardcoding this behaviour.

PR #1136, opened 2021-09-29
Better shell specific testing

### LINKED ISSUE #1135: rez-env output test is not testing on specified shell
In test case `test_shells.TestShells.test_rez_env_output`, it uses decorator `per_available_shell`, which iterates all available shells and use `config.override` to set testing target.

https://github.com/nerdvegas/rez/blob/8f965f6ed7ac438c1ea331940e6ad14f766564fa/src/rez/tests/util.py#L218-L221

However, when testing `rez-env` output, subprocess is called which actually won't take the config override and always testing on system default shell.

https://github.com/nerdvegas/rez/blob/8f965f6ed7ac438c1ea331940e6ad14f766564fa/src/rez/tests/test_shells.py#L183-L190

The var `arg` in above code should changed into
```python
   ...
     def _test(txt): 
         ...
         args = [binpath, "--shell", config.default_shell, "--", "echo", txt]
```

I found this issue when I was tweaking code in CMD shell class and somehow the test result of PowerShell also gets affected and made me confused.

davidlatwe:
* fix #1135,
* also, able to exclude specific shells from testing with `@per_available_shell()`. This is currently as a helper feature for working on `test_rez_env_output`.
    
    Example usage as follow:

    ```python
        @per_available_shell(exclude=["cmd", "powershell"])
        @install_dependent()
            ...
    ```

### DIFF
--- a/src/rez/tests/test_shells.py
+++ b/src/rez/tests/test_shells.py
@@ -11,6 +11,7 @@ from rez.utils.execution import ExecutableScriptMode, _get_python_script_files
 from rez.tests.util import TestBase, TempdirMixin, per_available_shell, \
     install_dependent
 from rez.bind import hello_world
+from rez.config import config
 import unittest
 import subprocess
 import tempfile
@@ -183,16 +184,20 @@ class TestShells(TestBase, TempdirMixin):
     @per_available_shell()
     @install_dependent()
     def test_rez_env_output(self):
+        target_shell = config.default_shell  # overridden by test util
+
         def _test(txt):
             # Assumes that the shell has an echo command, build-in or alias
             binpath = os.path.join(system.rez_bin_path, "rez-env")
-            args = [binpath, "--", "echo", txt]
+            args = [binpath, "--shell", target_shell, "--", "echo", txt]
 
             process = subprocess.Popen(
                 args, stdin=subprocess.PIPE, stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE, universal_newlines=True
             )
             sh_out = process.communicate()
+            if sh_out[1]:
+                raise Exception("Command failed:\n%s" % sh_out[1])
             self.assertEqual(sh_out[0].strip(), txt)
 
         # please note - it's no coincidence that there are no substrings like
--- a/src/rez/tests/util.py
+++ b/src/rez/tests/util.py
@@ -198,8 +198,10 @@ def program_dependent(program_name, *program_names):
     return decorator
 
 
-def per_available_shell():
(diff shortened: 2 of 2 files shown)


src/rez/tests/test_shells.py
```diff
@@ -183,16 +184,20 @@ def test_rcfile(self):
...
-            args = [binpath, "--", "echo", txt]
+            args = [binpath, "--shell", target_shell, "--", "echo", txt]
 
             process = subprocess.Popen(
                 args, stdin=subprocess.PIPE, stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE, universal_newlines=True
             )
             sh_out = process.communicate()
+            if sh_out[1]:
+                raise Exception("Command failed:\n%s" % sh_out[1])
```
davidlatwe, line 200: I didn't use `process.returncode` is because, it seems that powershell always return `0` even the command actually failed.

davidlatwe:
Hmmm, looks like checking stderr isn't a good idea ? `test_rez_env_output` failed on `csh` shell :(

```
txt = '!hey>$'

    def _test(txt):
        # Assumes that the shell has an echo command, build-in or alias
        binpath = os.path.join(system.rez_bin_path, "rez-env")
        args = [binpath, "--shell", target_shell, "--", "echo", txt]
    
        process = subprocess.Popen(
            stderr=subprocess.PIPE, universal_newlines=True
        )
        if sh_out[1]:
>           raise Exception("Command failed:\n%s" % sh_out[1])
E           Exception: Command failed:
E           hey: Event not found.
E            (in shell 'csh')
```

src/rez/tests/test_shells.py
bfloch: Isn't this a case of powershell -File vs. -Command?
Only the latter passes the error code.
https://stackoverflow.com/questions/50200325/returning-an-exit-code-from-a-powershell-script

Or maybe `$ErrorActionPreference = "Stop"`:
https://stackoverflow.com/questions/9948517/how-to-stop-a-powershell-script-on-the-first-error

The checks also need to happen form `$LastExitCode` I believe:
https://stackoverflow.com/questions/10666035/difference-between-and-lastexitcode-in-powershell

What an over-engineered little shell :)

davidlatwe:
Github action cannot find any runner that matches `ubuntu-16.04`, due to :
https://github.blog/changelog/2021-04-29-github-actions-ubuntu-16-04-lts-virtual-environment-will-be-removed-on-september-20-2021/

Ubuntu version bumped in test matrix.

Issue #1175, opened 2021-12-10
Environment variables containing brackets break powershell shell


hutchinson:
Hi All,

Hope you're all well! First off, thanks for a great library, have really enjoyed getting onboard with it!

We've noticed a small bug when trying to create packages that contain environment variables with brackets in them and launching it in powershell. Here is a small repro case:

```
name = 'testpackage'
version = '1.0.0'

requires = [
    'platform-windows'
]

build_command = False

def commands():
    env['ENVVARWITH(x86)'] = 'This_breaks_powershell'
```

Running it with: `rez build -i & rez-env testpackage --shell=powershell` gives the following error:

```
[...snip...]
Temp\rez_context_68oyvs1b\context.ps1:53 char:16
+ $Env:ENVVARWITH(x86) = "This_breaks_powershell"
+                ~
Unexpected token '(' in expression or statement.
```

Looks like using the alternative syntax to set environment variables should work (although I haven't tried it):

```
Set-Item "env:ENVVARWITH(x86)" "This is fine"
dir env:

Name                           Value
----                           -----
ALLUSERSPROFILE                C:\ProgramData
[...snip...]
ENVVARWITH(x86)                This is fine
```

Will look to get a PR over to you ASAP.

PR #1176, opened 2021-12-12
Uses Set-Item rather than $Env: to configure powershell environment variables.

hutchinson:
Updates `powershell_base.py` to use `Set-Item` when setting environment variables to ensure environment variables with special characters such as `()` don't cause script failures.

Repro:

```
name = 'testpackage'
version = '1.0.0'

requires = [
    'platform-windows'
]

build_command = False

def commands():
    env['ENVVARWITH(x86)'] = 'This_breaks_powershell'
    env['ENVVARWITH(x86)'].append('Another value')
````

`rez build -i & rez-env testpackage --shell=powershell`

With this patch the above use case should work.

### DIFF
--- a/src/rezplugins/shell/_utils/powershell_base.py
+++ b/src/rezplugins/shell/_utils/powershell_base.py
@@ -282,14 +282,15 @@ class PowerShellBase(Shell):
 
     def setenv(self, key, value):
         value = self.escape_string(value)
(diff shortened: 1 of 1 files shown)


hutchinson:
Addresses issue #1176

nerdvegas:
closes #1175

PR #1179, opened 2021-12-16
Whichcraft now cares about executable symlinks on windows that might not end with typical executable extensions

### LINKED ISSUE #1178: Whichcraft not caring about executable symlinks on windows that do not have the extension of an executable
Rez 2.103.2
Windows 10

Hey,

in a previous commit we added whichcraft to care about rez pip for windows. 
In some cases however there might be executables in the packages that are just symlinks to executables. And these symlinks can have any kind of or no extension.

I have a fix ready to be reviewed.

Cheers,
Manuel

manuelkoester:
Fixes #1178 

Hey,

in a previous commit we added whichcraft to care about rez pip for windows. 
In some cases however there might be executables in the packages that are just symlinks to executables. And these symlinks can have any kind of or no extension.

To deal with that I am checking if the checked files are a symlink to an executable with the [listed extensions](https://github.com/nerdvegas/rez/pull/1157/files#diff-364b6b1f7924a21c8cce24cd56ea6e8a077572f6faf5dfb249a92c5cac28e5bcR11).

Cheers,
Manuel

### DIFF
--- a/src/rez/vendor/whichcraft/whichcraft.py
+++ b/src/rez/vendor/whichcraft/whichcraft.py
@@ -62,6 +62,15 @@ def which(cmd, mode=os.F_OK | os.X_OK, path=None, env=None):
         # have to try others.
         if any(cmd.lower().endswith(ext.lower()) for ext in pathext):
             files = [cmd]
+        # See if the file is a symlink. A symlink can be an executable on 
+        # windows without an extension. If it is, see if its target's extension
+        # matches any of the expted path extensions.
+        elif any(
+            os.path.realpath(os.path.join(p, cmd)).lower().endswith(ext.lower())
+            for ext in pathext
+            for p in path
+        ):
+            files = [cmd]
         else:
             files = [cmd + ext.lower() for ext in pathext]
     else:


nerdvegas:
This PR addresses a legitimate issue, but there are a couple of problems:

- iterating over `path` when using realpath to account for symlinks isn't taking into account that "%systemroot%" might be in a path;
- the logic could find a file called "foo" in one path, that symlinks to an executable, but the second loop (`for dir in path`) could then match to a different "foo" in a different path, incorrectly.
- according to https://docs.python.org/3/library/os.path.html, realpath also only resolves symlinks in py3.8+ - this should be noted.

The thing is though, addressing these issues changes the code enough that I think use of whichcraft is hardly warranted anymore.

I think I'll take what you've done here and address the concerns I've raised (as well as dropping whichcraft), and put in a separate PR which deprecates this one.

Thx
A

nerdvegas:
See https://github.com/nerdvegas/rez/pull/1182

manuelkoester:
Thanks for the additional information and also the new implementation Allan.
Greatly appreciated!

Issue #1150, opened 2021-11-09
AttributeError: 'NoneType' object has no attribute 'conflicts_with' when rez-env a package


frankchen211:
File "/mnt/pipeline/thirdparty/linux/rez/2.97.0/lib/python3.8/site-packages/rez/solver.py", line 2270, in get_fail_graph
    return phase.get_graph()
  File "/mnt/pipeline/thirdparty/linux/rez/2.97.0/lib/python3.8/site-packages/rez/solver.py", line 1748, in get_graph
    and scope_r.conflicts_with(conflicting_request):
AttributeError: 'NoneType' object has no attribute 'conflicts_with'

nerdvegas:
Thanks! I'm aware of this one and will be looking at it soon.
A

PR #1184, opened 2021-12-23
Test issue 1150 none conflicts with

davidlatwe:
Was lucky that I have not yet updating my running Rez to the latest and happen to spot a reproducible for the fixed issue #1150. Thought it would be useful to have this case written in tests.

The new test case is committed on top of a branch that has not yet fixed the issue, so you may see it fail if you run the test in that branch.

### DIFF
--- a/src/rez/tests/test_context.py
+++ b/src/rez/tests/test_context.py
@@ -17,8 +17,8 @@
 test resolved contexts
 """
 from rez.tests.util import restore_os_environ, restore_sys_path, TempdirMixin, \
-    TestBase
-from rez.resolved_context import ResolvedContext
+    TestBase, MemoryPkgRepo
+from rez.resolved_context import ResolvedContext, ResolverStatus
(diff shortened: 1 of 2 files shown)


nerdvegas:
On looking into this PR I think I need to decline it.

The reason is that this looks like an edge case that's specific to contexts, but it isn't - we can cover this same case by adding another test to the existing test_solver.py class, specifically in: https://github.com/nerdvegas/rez/blob/master/src/rez/tests/test_solver.py#L178

Specifically, see the line: https://github.com/nerdvegas/rez/blob/master/src/rez/solver.py#L1768-L1769,
which was added in fix: https://github.com/nerdvegas/rez/pull/1173

All we need to do is add a test solve where "an existing conflict request conflicts with a pkg requirement". It appears I just happened not to cover this specific case in the original solver tests.

nerdvegas:
Closing, see https://github.com/nerdvegas/rez/issues/1225

PR #1185, opened 2021-12-24
windows CI upgrade

### LINKED ISSUE #1181: windows gh actions issue
Windows builds are failing on pulling docker image. It appears to be some mismatch between the images that exist, and the commit-hash- based tag that our workflow expects to be there. It may be better to simply use the latest image available - all that really matters is that the image for the corresponding python version is being used (any image updates aren't tied into the rez version being tested - the entrypoint installs rez from the current checkout anyway).

There's also this error in the wondiws-docker-image workflow:
```
C:\ProgramData\chocolatey\bin\choco-cleaner.ps1 : The term 
'C:\ProgramData\chocolatey\bin\choco-cleaner.ps1' is not recognized as the 
name of a cmdlet, function, script file, or operable program. Check the 
spelling of the name, or if a path was included, verify that the path is 
correct and try again.
At line:1 char:527
+ ... hoco-cleaner; C:\ProgramData\chocolatey\bin\choco-cleaner.ps1; choco  ...
+                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:\ProgramData\...oco-cleaner.p 
   s1:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```
- although this doesn't appear to be failing the image build.
-

nerdvegas:
Fixes #1181 

This PR introduces a bunch of changes to improve the situation wrt windows CI tests. Specifically:

* Previously, windows.yaml would use the tagged image generated by windows-docker-image.yaml (where tag is the latest commit hash of relevant workflow/docker files). This breaks when a change is made that triggers both an image rebuild, _and_ the windows workflow, since the new image isn't built yet. Now, windows.yaml just uses latest image instead.
* Updated windows-docker-image.yaml to generate artifacts that contain docker inspect info, for the images generated. This is for debugging purposes
* Updated windows.yaml to print the specific tag of the image it's using, to be able to verify that it is actually using the latest image as expected
* reworked path/path-ignore filters across all workflows - now uses combined include/exclude form to better target files to trigger workflow runs from
* fixed bug causing choco-cleanup to fail. It's reporting deletion of ~300Mb of files now, so images should be considerably smaller
* general cleanup of workflow run code to be cleaner and easier to follow
* avoid using env-vars when standard pwsh var will do
* run pytest-based selftest on windows to match other os workflows
* drop use of DOCKER_CLI_EXPERIMENTAL feature, not supposed to be used in production
* moved py version into the image name, required in order for windows.yaml to be able to pull latest image

### DIFF
--- a/.github/docker/rez-win-base/Dockerfile
+++ b/.github/docker/rez-win-base/Dockerfile
@@ -23,13 +23,13 @@ ARG PWSH_VERSION=6.2.2
 # - PowerShellCore
 #
 ENV chocolateyUseWindowsCompression false
-RUN iex ((new-object net.webclient).DownloadString('https://chocolatey.org/install.ps1'));             `
-    choco feature disable --name showDownloadProgress;                                                 `
-    choco install git.install --yes --version=${ENV:GIT_VERSION};                                      `
+RUN iex ((new-object net.webclient).DownloadString('https://chocolatey.org/install.ps1')); `
+    choco feature disable --name showDownloadProgress; `
+    choco install git.install --yes --version=${ENV:GIT_VERSION}; `
     choco install cmake --yes --version=${ENV:CMAKE_VERSION} --installargs 'ADD_CMAKE_TO_PATH=System'; `
-    choco install pwsh --yes --version=${PWSH_VERSION};                                                `
-    choco install --yes choco-cleaner;                                                                 `
-    C:\ProgramData\chocolatey\bin\choco-cleaner.bat;                                                   `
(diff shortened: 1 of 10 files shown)


davidlatwe (commented):
Nice work 🍻
I also notice that `windows.yaml` workflow isn't testing with Python 3 for a long while, maybe we could re-enable it now ?

PR #1192, opened 2022-01-27
Fix aliases in windows powershell/pwsh

### LINKED ISSUE #1191: Aliases broken on windows when using powershell
Between versions 2.98.0 and 2.101.0 there has been a regression on aliases when using powershell on windows.

In 2.101.0 calling an alias will print out the command with `@args` postfix added and nothing will happen.

If cmd is used instead, everything works as expected.

koaleksa:
fixes #1191


nerdvegas:
Good find. I tracked this down, it was introduced in v2.98.1. It was an unintended side-effect of removing a small amount of code (2 lines) for licensing reasons, and neglecting to realise that the `.format(...)` (which was replaced by old-school %s replacement) was collapsing those double curly braces.

PR #1207, opened 2022-02-11
Issue 1206 wiki gpl rm

### LINKED ISSUE #1206: remove GPL3 code from update-wiki.py
Alas I've just found some lurking GPL code. It's in code that's only used during a workflow run however (it's a script that generates the wiki content). But still we should figure out how to replace this as a matter of urgency.

https://github.com/nerdvegas/rez/blob/master/wiki/update-wiki.py#L55

nerdvegas:
This PR removes some GPL3-licensed code from the ipdate-wiki.py script. It also simplifies said script, and fixes an issue where some pages did not have a ToC.

Fixes #1206


wiki/update-wiki.py
```diff
@@ -27,309 +10,66 @@
...
 import os
 import re
 import subprocess
 import shutil
 import sys
 
 
+# py3.7+ only
+if sys.version_info[:2] < (3, 7):
+    print("update-wiki.py: ust use python-3.7 or greater", file=sys.stderr)
```
Minkiu, line 22: typo: `ust`

PR #1208, opened 2022-02-11
Replace the version API with Rust

scott-wilson:


nerdvegas:
Just a note on this - if at all possible it'd be good to keep some form of `--no-rust` rez install option available (and great if this could translate to an extra for the rez pip package). Rez is _very_ backwards compatible by design and I'd hate to introduce a new requirement that studios may not be ready to meet.

In the same vein, we should keep both the py- and rust- based version submods present - perhaps making rust available via import of `rversion` rather than `version`, and choosing between them in rez based on availability. We'd also then run the same unit test over rversion as we do version already.

I'm noting this stuff now (even though we're so early in) so we're thinking about it earlier on and hopefully avoiding more work later.

nerdvegas:
Also want to make you aware of the benchmarking CI we already have. We'll probably have to tweak it to show rust numbers alongside py or something.

https://github.com/nerdvegas/rez/blob/master/metrics/benchmarking/RESULTS.md

KelSolaar:
Hi @scott-wilson,

To @nerdvegas's point, it would be great to describe what this PR does and why. There is no description at the moment which is mildly surprising.

Cheers,

Thomas

scott-wilson:
> Hi @scott-wilson,
> 
> To @nerdvegas's point, it would be great to describe what this PR does and why. There is no description at the moment which is mildly surprising.
> 
> Cheers,
> 
> Thomas

That's fair. I am mostly just getting something done before I turn this into a proper PR.

sirfuzzalot:
> Hi @scott-wilson,
> 
> To @nerdvegas's point, it would be great to describe what this PR does and why. There is no description at the moment which is mildly surprising.
> 
> Cheers,
> 
> Thomas

Here's an initial take at the PR description.

---

Increase solver speed by porting the **version** submodule to Rust.

`src/rez/vendor/version/`

Conditions:
1. Rust implementation must benchmark faster than equivalent Python resolve in most cases
2. Python implementation must remain available. Rust is optional.
3. Rust implementation should be written to allow easy porting/addition of other portions of **rez** in Rust.
4. Build process and code should be documented, targeting those who are unfamiliar with Rust
5. Rust code must contain passing unittests
6. Rust code must pass the existing Python test suite.
7. Rust code must have the **rez** license at the top of each file

setup.py
```diff
@@ -96,7 +95,10 @@ def find_files(pattern, path=None, root="rez"):
         "Programming Language :: Python",
         "Programming Language :: Python :: 2",
         "Programming Language :: Python :: 3",
+        "Programming Language :: Rust",
+        "Programming Language :: Python :: Implementation :: CPython",
+        "Programming Language :: Python :: Implementation :: PyPy",
```
JeanChristopheMorinPerso, line 100: I haven't reviewed this PR and I know it's still very WIP, but we don't really test nor advertise that it works with PyPy, so I'd better not add the PyPy classifier.

herronelou:
I know this is a bit old now but I'm wondering what kind of performance increase was coming out of this?
As our Rez repos are increasing in size resolves are getting slower and slower so I'm curious about anything that could speed it up at this point.

instinct-vfx:
> I know this is a bit old now but I'm wondering what kind of performance increase was coming out of this? As our Rez repos are increasing in size resolves are getting slower and slower so I'm curious about anything that could speed it up at this point.

Are you running memcached? We hardly see any performance issue with the solver with memcached.

herronelou:
> > I know this is a bit old now but I'm wondering what kind of performance increase was coming out of this? As our Rez repos are increasing in size resolves are getting slower and slower so I'm curious about anything that could speed it up at this point.
> 
> Are you running memcached? We hardly see any performance issue with the solver with memcached.

That's likely derailing the thread a little bit, but no, right now cache was disabled because we had issues with the resolve not always grabbing the latest versions if it has old stuff cached already. If you or someone else has time for a chat I'd love to go through our setup to see if there are things we could do to speed it up.

instinct-vfx:
You may want to jump onto the Slack if you can, that will give perspective and support by multiple parties

vvzen:
Would support for Rust ever be reconsidered to simplify shipping `rez` without having to depend on a Python interpreter at runtime? If done carefully, many things could be ported _gently_ using https://github.com/PyO3/pyo3 via a piecemeal transition.

Currently (unless I misunderstand how the caching works, which could well be!) you'd need to opt-in to memcached to get the performance benefits, but at that point (imho, of course!) you might as well just run your own infra to solve package management in a more efficient and tailored way. While I think there'd be value for some people in having `rez-env` be a standalone binary that has very little runtime requirements (so that it depends only on the glibc/libSystemB/etc of the current host). You'd potentially just need to have the binary in the $PATH on any machine and you'd be able to solve environments, which would be great for keeping production environments as lightweight as possible.

instinct-vfx:
There are multiple parts of rez mixed up i would say.
I'd say there is likely benefit in having self contained executables (keep in mind though that rez is WAY more than just rez-env).
Memcached helps with file IO bottlenecks. Which is mostly separate from resolve or version performance from my understanding.
Memcached is a really low-entry central caching solution. You can deploy it as a single-binary deployment. You could even run separate memcached instances locally on each machine if you wanted to skip any central infra.

That said, we should keep discussion here focused on the scope of the PR which is replacing "only" the version submodule with a rust implementation.

maxnbk:
> Could support for Rust ever be reconsidered to simplify shipping `rez`

There isn't really any objection to "support" rust, but to suggest that it somehow simplifies shipping rez, at least in the short/medium term, is not something the TSC would probably agree with. Currently, rez is able to ship entirely with sdist dependencies and does not require any amount of compilation (unless someone has customized their installation, which is fine for their own needs obviously). To support rust (or C++) modules, we would need to begin delivering platform-specific binaries (whether in a single package or multiple), or require that installation involve a compiler, which it doesn't currently, and I expect many people would object to.

The most compelling argument to use rust re-implementations of modules would be for performance reasons, and the previously discussed Version module was selected from benchmarks and is a definite target for such a swap. We received a couple of WIP attempts at doing so, but none that are actually complete or testable.

I think what the group is most in-favor of in the short-term is a way to enable testing of separate experimental rust-based module replacements. That way, we can evaluate if the added complications are worth it, maintain a stable API, and learn to maintain it as a rust project in the process. None of the rez maintainers are even vageuely rust-savvy at the moment, so that is a tradeoff as well.

Yes, rez-env is often used as a binary, but many studios require leveraging the python API as well, so replacing the entire set of binaries wholesale is a non-starter as well.

vvzen:
Thanks both for providing more context!

> Yes, rez-env is often used as a binary, but many studios require leveraging the python API as well, so replacing the entire set of binaries wholesale is a non-starter as well.

Makes sense. In that case then maybe there's less of a need since all production machines would require to have rez installed anyway.

> To support rust (or C++) modules, we would need to begin delivering platform-specific binaries (whether in a single package or multiple), or require that installation involve a compiler, which it doesn't currently, and I expect many people would object to.

I personally I would object that depending on the Python interpreter and a list of third party dependencies can be heavier than depending on rustup or cargo to provide per-platform binaries. In a way, the rez project project would take more of a toll in terms of maintenance burden with the hope that the user would take less of a toll (having less runtime dependencies and thus less moving parts in terms of package management).

I also wouldn't put C++ and Rust in that same mental bucket ("compiled languages") because, unlike C++, generating Rust binaries is arguably as simple as a `cargo build` for the majority of projects, without further configuration (no CMake/Ninja/etc). But I can understand that there's a lot of nuance that I'm probably missing.

> I think what the group is most in-favor of in the short-term is a way to enable testing of separate experimental rust-based module replacements. That way, we can evaluate if the added complications are worth it, maintain a stable API, and learn to maintain it as a rust project in the process. None of the rez maintainers are even vageuely rust-savvy at the moment, so that is a tradeoff as well.

That sounds great! If that ever happens, I'd heavily recommend to let somebody more experienced on Rust side drive the research phase, because I can see a serious risk of "failing before even starting" in letting only folks with a Python/C++ mindset attempt to port some code. A lot of ideas don't map 1:1 and without somebody capable of acting as an higher level interpreter I can see how things would hit roadblocks pretty quickly. 

> The most compelling argument to use rust re-implementations of modules would be for performance reasons, and the previously discussed Version module was selected from benchmarks and is a definite target for such a swap. We received a couple of WIP attempts at doing so, but none that are actually complete or testable.

Would you be able to share any links (other than this PR) ? I'd be curious to have a look!

PR #1215, opened 2022-02-15
Issue 1214 spdx enforce

### LINKED ISSUE #1214: add workflow to enforce copyright
Avoid PRs containing missing/incorrect SPDX by enforcing it via a workflow

nerdvegas:
Fixes #1214


.github/workflows/copyright.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Run copyright checker
+        run: |
+          bash ./src/build_utils/license/apply_copyright
+          if [[ "$(git status | grep modified)" != "" ]]; then
+            echo "Some sourcefiles are missing copyright notice!" 1>&2
+            echo "Run ./src/build_utils/license/apply_copyright to apply." 1>&2
+            exit 1
+          fi
+        env:
+          IN_GITHUB_WORKFLOW: 1
```
JeanChristopheMorinPerso, line 35: The environment variable `GITHUB_ACTIONS` is provided by GitHub Actions for exactly this purpose. See https://docs.github.com/en/actions/learn-github-actions/environment-variables#:~:text=example%2C%20actions/checkout.-,GITHUB_ACTIONS,-Always%20set%20to.

.github/workflows/copyright.yaml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 2.7
```
JeanChristopheMorinPerso, line 24: Should we use Python 3 here? The script looks quite simple, I'm sure it wouldn't be _that_ much effort to port.

nerdvegas:
Cool @ GITHUB_ACTIONS, didn't know about that one.

Re py3, I agree, but there's a few cases of this, probably better addressed
in separate PR(s):
- pypi
- sphinx
- copyright

Perhaps we also advance to 3.10 as default 3.x version at the same time. So
a general py3-workflows PR is probably justified. Added ticket:
https://github.com/nerdvegas/rez/issues/1216

Lastly, Windows workflow currently has py3 disabled, I don't recall why
this was. Will need to investigate and reenable.

A

.github/workflows/copyright.yaml
nerdvegas: as per comment, agree but let's do it in a py3- specific PR

PR #1232, opened 2022-02-22
Re-enable py3 workflow in windows.yaml

instinct-vfx:
Working towards re-enabling the py3 workflows for windows.


instinct-vfx:
@nerdvegas i only re-enabled the tests for py3 on windows to see if anything fails, but it does not. My vote would be to re-enable the workflow then?

nerdvegas (approved):

PR #1236, opened 2022-02-28
Explicit fail when python executable is not found

aboellinger:
# Problem

We have found ourselves in a situation where during a call to `find_pip_from_context`,  `find_python_in_context(context)` returns `None`, resulting in the creation and execution of an invalid command (with `None` provided as an executable path). This ends up breaking downstream in `subprocess` with an obscure message.

# Solution

Raise an explicit error if the executable is not located, before we try to run it.

### DIFF
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -202,6 +202,9 @@ def find_pip_from_context(python_version, pip_version=None):
 
     py_exe = find_python_in_context(context)
 
+    if not py_exe:
+        raise RezSystemError("Failed to locate Python executable from context {}".format(context))
+    
     proc = context.execute_command(
         # -E and -s are used to isolate the environment as much as possible.
         # See python --help for more details. We absolutely don't want to get
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -203,8 +203,9 @@ def find_pip_from_context(python_version, pip_version=None):
     py_exe = find_python_in_context(context)
 
     if not py_exe:
-        raise RezSystemError("Failed to locate Python executable from context {}".format(context))
-    
+        print_debug("Failed to locate python executable from context")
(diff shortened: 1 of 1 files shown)


src/rez/pip.py
```diff
@@ -202,6 +202,9 @@ def find_pip_from_context(python_version, pip_version=None):
 
     py_exe = find_python_in_context(context)
 
+    if not py_exe:
+        raise RezSystemError("Failed to locate Python executable from context {}".format(context))
```
JeanChristopheMorinPerso, line 206: I think returning early like `return None, None, context` would be slightly better. That's because of the way the function is called: https://github.com/nerdvegas/rez/blob/master/src/rez/pip.py#L92-L101.

src/rez/pip.py
```diff
@@ -202,6 +202,10 @@ def find_pip_from_context(python_version, pip_version=None):
 
     py_exe = find_python_in_context(context)
 
+    if not py_exe:
+        print_debug("Failed to locate python executable from context")
```
JeanChristopheMorinPerso, line 206: This log might be redundant I think. The `_check_found` function already logs when something isn't found and I also have a personal preference to let the caller log instead of letting the callee (find_pip_from_context) log.

aboellinger: Alright, to me the log seemed consistent with code above and also acted as a comment for the `return` statement.

But I am fine with removing it.

Fixed in [2a64909](https://github.com/nerdvegas/rez/pull/1236/commits/2a64909caa1044fe871c2efbfe1b6443d61cfd61)

JeanChristopheMorinPerso (approved):
Thanks for the fix @aboellinger!

nerdvegas (approved):

PR #1244, opened 2022-03-04
fix: make cmake install directives whitespace friendly

### LINKED ISSUE #553: rez_install_files with LOCAL_SYMLINK fails when an input file has whitespaces in its name
Hi,

Calling rez_install_files with the LOCAL_SYMLINK parameter fails when one of the input files has a symlink in the name. This can be fixed by adding quotes in the following line inside the 'install_files_' macro (rez/src/rezplugins/build_system/cmake_files/InstallFiles.cmake:140):

`install( CODE "execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink ${f} ${CMAKE_INSTALL_PREFIX}/${target_fpath})" )`,

which becomes:

`install( CODE "execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink \"${f}\" \"${CMAKE_INSTALL_PREFIX}/${target_fpath}\")" )`

maxnbk:
Ostensibly closes #553


JeanChristopheMorinPerso:
Should we had a test for this one? There is a test for cmake, so we could either add files with whitespaces in their names in this test or add a new test. The test in question uses https://github.com/nerdvegas/rez/blob/master/src/rez/data/tests/builds/packages/translate_lib/2.2.0/CMakeLists.txt.

maxnbk:
Yeah, I'm having trouble testing (in general) because I got myself in a rabbithole with using msys2 on windows and running rez tests in that environment seems tenuous at best, so I need to back up and set up a proper test harness for myself locally, but that's the reason this PR is a draft. I'll unlock it from draft once I've gotten my local setup properly working and ideally a new test or set of tests.

maxnbk:
ubuntu seems to not want to fetch zsh installers I guess. But the test is implemented and this PR solves the test that it implements. marking ready.

PR #1252, opened 2022-03-19
added package fam removal func, and tests

### LINKED ISSUE #1248: add ability to delete package family
Currently we have the support to delete package versions. Once an entire family is empty, itshould be possible to delete the family.

**Motivation**
My studio is working on package archival and we have a number of packages that are no longer in use at all. In these cases it makes sense to archive entire package families.

nerdvegas:
fixes #1248


src/rez/cli/rm.py
```diff
@@ -15,6 +15,11 @@ def setup_parser(parser, completions=False):
         "-p", "--package",
         help="remove the specified package (eg 'foo-1.2.3'). This will work "
         "even if the package is currently ignored.")
+    group.add_argument(
+        "-f", "--family",
+        help="remove the entire specified package family (eg 'python'). This "
```
maxnbk, line 20: the word "entire" is a bit confusing here. "remove the entire", but, "only if its empty"

src/rez/cli/rm.py
```diff
@@ -15,6 +15,11 @@ def setup_parser(parser, completions=False):
...
+        "is only supported if the family is empty, or all its packages are "
+        "hidden")
```
maxnbk, line 22: I think we ought to avoid archiving a family if it contains any packages, including hidden packages.

src/rez/cli/rm.py
```diff
@@ -46,6 +52,27 @@ def remove_package(opts, parser):
         sys.exit(1)
 
 
+def remove_package_family(opts, parser):
+    from rez.vendor.version.requirement import VersionedObject
+    from rez.package_remove import remove_package_family
+
+    if opts.dry_run:
+        parser.error("--dry-run is not supported with --family")
```
maxnbk, line 60: dry run could probably at least tell you if it would allow proceeding with an archival if it _weren't_ a dry run?

maxnbk (commented):
some notes

src/rez/cli/rm.py
```diff
@@ -46,6 +52,27 @@ def remove_package(opts, parser):
...
+
+    if not opts.PATH:
+        parser.error("Must specify PATH with --family")
+
+    obj = VersionedObject(opts.family)
+    if obj.version:
+        parser.error("Expected package name, not version")
+
+    if remove_package_family(obj.name, opts.PATH):
```
JeanChristopheMorinPerso, line 69: Should we try/except and print a nice error message when `remove_package_family` raises a `PackageRepositoryError` error?

Something like:
```python
success = False
try:
    success = remove_package_family(obj.name, opts.PATH)
except PackageRepositoryError as exc:
   print("Failed to remove package family: {0}".format(exc), file=sys.stderr))
   sys.exit(1)

if not success:
    print("Package family not found.", file=sys.stderr)
    sys.exit(1)
```

It would be more user friendly than a traceback.

JeanChristopheMorinPerso:
Have you considered a `--recursive` flag to make it easier to just purge an entire family? If yes, I'd be curious to know the reasoning for not including something like (apart fro the obvious that it is a dangerous operation).

src/rez/cli/rm.py
nerdvegas: agreed

src/rez/cli/rm.py
nerdvegas: agreed, updating

src/rez/cli/rm.py
nerdvegas: agreed

nerdvegas:
> Have you considered a `--recursive` flag to make it easier to just purge an entire family? If yes, I'd be curious to know the reasoning for not including something like (apart fro the obvious that it is a dangerous operation).

That's it, it's a dangerous op. I figured it's better to make the user do that themselves, than accidentally delete a whole fam.

JeanChristopheMorinPerso:
And what about a `--force` flag to mimic the Unix `rm -r` command? By default fail if not empty, but if `-f` is provided, then go ahead even if not empty.

nerdvegas:
> And what about a `--force` flag to mimic the Unix `rm -r` command? By default fail if not empty, but if `-f` is provided, then go ahead even if not empty.

Still dangerous but `--force` makes more sense than `--recurse` and implies the danger. Yeah I'll add this.

instinct-vfx (approved):
I will have good use for this. Agreeing with the proposed and accepted changes.

JeanChristopheMorinPerso (approved):

PR #1253, opened 2022-03-20
Another way to test packages in dev stage and make sure path set/append/prepend to PYTHONPATH is normalized.

frankchen211:
Sorry I made 2 commits together here:
### Make sure path set/append/prepend to PYTHONPATH is normalized
87ee4e62e6172c3dafe615c8947be3cb1286cde0
PYTHONPATH in a package mostly set like this:
`env.PYTHONPATH.append('{root}/python')`
In windows sys,.path returns e.g.: c:\path\to\root/python, although it is ok, but prefer it is normalized.

### New Feat: Another way to test packages in dev stage
5f94b9c08eaaaf84ac5fec756bf99f3dc342fe4b
Let's say we have a package `helloworld` in d:\dev\helloworld(a pure python package, and no version specified) want to test, but don't want to move it to any package repository or package filters, instead call this package like this:
`rez-env helloworld pkg1 pkg2 ... -d d:\dev`, helloworld will be resolved to D:\dev\helloworld, but pkg1 and pkg2 resolved to a repository path in rez config. or if is also in D:\dev folder.
```
helloworld                d:\dev\helloworld
pkg1-3.3                   c:\users\frank\packages\pkg1\3.3\                                           (local)
pkg2-1.0.0                c:\users\frank\packages\pkg2\1.0.0                                          (local)
```

### DIFF
--- a/src/rez/cli/env.py
+++ b/src/rez/cli/env.py
@@ -49,6 +49,10 @@ def setup_parser(parser, completions=False):
         "--paths", type=str, default=None,
         help="set package search path (use %r separator)" % os.pathsep)
     parser.add_argument(
+        "-d", "--dev-paths", type=str, default=None,
+        help="search packages from these dev paths if found, otherwise from formal search path,"
+             "normally for testing purpose, non-version packages (use %r separator)." % os.pathsep)
+    parser.add_argument(
         "-t", "--time", type=str,
         help="ignore packages released after the given time. Supported formats "
         "are: epoch time (eg 1393014494), or relative time (eg -10s, -5m, "
@@ -156,6 +160,13 @@ def command(opts, parser, extra_arg_groups=None):
     context = None
     request = opts.PKG
     t = get_epoch_time_from_str(opts.time) if opts.time else None
+
+    if opts.dev_paths:
+        for path in opts.dev_paths.split(os.pathsep):
+            path = os.path.abspath(path)
+            for idx, pkg in enumerate(request):
+                if pkg in os.listdir(path):
+                    request[idx] = '{}@{}'.format(pkg, os.path.join(path, pkg))
 
     if opts.paths is None:
         pkg_paths = (config.nonlocal_packages_path
--- a/src/rez/rex.py
+++ b/src/rez/rex.py
@@ -620,7 +620,7 @@ class Python(ActionInterpreter):
         if self.update_session:
             if key == 'PYTHONPATH':
                 value = self.escape_string(value)
-                sys.path = value.split(os.pathsep)
+                sys.path = [os.path.normpath(v) for v in value.split(os.pathsep)]
 
     def unsetenv(self, key):
         pass
@@ -632,13 +632,13 @@ class Python(ActionInterpreter):
         if self.update_session:
             if key == 'PYTHONPATH':
                 value = self.escape_string(value)
-                sys.path.insert(0, value)
+                sys.path.insert(0, os.path.normpath(value))
 
     def appendenv(self, key, value):
         if self.update_session:
             if key == 'PYTHONPATH':
                 value = self.escape_string(value)
-                sys.path.append(value)
+                sys.path.append(os.path.normpath(value))
 
     def info(self, value):
         if not self.passive:
--- a/src/rez/solver.py
+++ b/src/rez/solver.py
@@ -464,14 +464,20 @@ class _PackageVariantList(_Common):
         self.package_name = package_name
         self.solver = solver
 
+        self.entries = []
+
+        request_package = solver.request_list.get(package_name)
+        if request_package and request_package.request_path:
+            # it seems DeveloperPackage can't be used in this way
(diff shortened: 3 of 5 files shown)


ColinKennedy:
Is it weird that I just use `os.path.join` directly in the `package.py`? e.g.

```
def commands():
    import os

    env.PYTHONPATH.append(os.path.join(root, "python"))
```

And that goes for any other environment variable, really. Just my 2 cents.

instinct-vfx:
> Is it weird that I just use `os.path.join` directly in the `package.py`? e.g.

No it is not, i do that too.

nerdvegas:
I'm closing this pull request, unfortunately there are significant issues that need to be resolved.

1. 
Don't implement multiple unrelated fixes in a single PR.

2.
We cannot assume that strings appended/prepended to an env-var are paths. They could be a list of database URIs for all we know. Yes they will be most of the time, but this update could break backwards compatibility for some users (a big no-no in rez), and would leave them no recourse to fix it either.

I want to highlight though that it is actually required that we have some way of formally identifying paths in commands funcs. We're going to need this if/when we have this idea of "anchoring" implemented in packages, to support overlaid runtimes. So let's have a discussion about that. One approach I was thinking is to add a new `path` primitive, simply used like so:

```
def commands():
    env.PYTHONPATH.append(path("{root}/python"))
```

Another (and probably in tandem with the above) would be a new config param to default certain envvars to always assume paths, eg:

```
# in rezconfig.py
pathlike_env_vars = [
    "PYTHONPATH",
    "PATH"
]
```

We would need the above for migration purposes in any case.

3.
This is the big one. The updates to get a developer package working in a solve, tightly couples concepts about developer packages into both the solver, and the version submodule. This raises _huge_ alarm bells for me. Neither the solver, nor version module should know anything about developer packages. The correct way to implement this (and we actually do need this for an upcoming feature unrelated to this fix) is to implement a new type of PackageRepository that is able to ingest developer packages and present them to the rest of rez as standard Package instances. But I need to stress - this is a fairly major feature that would require adding unit testing for. There are lots of things involved that we'd have to verify are working as expected (early/late- bound attribs, config overrides, caching and so on). Trying to shoehorn DeveloperPackages into the system, where everything else is Package (actually, Variant) instances, I feel is almost guaranteed to create a litany of problems for a long time. But most importantly, it breaks the modularity that rez strives for.

Wrt features in general, I would strongly advise that anyone thinking about implementing a new feature, discuss it with the group first. It brings me no joy to decline people's hard work.

instinct-vfx:
I have not worked with them a lot yet, but maybe someone else has. Could we just use the new pathlib.Path for the path primitive?

nerdvegas:
I would definitely not use `pathlib.Path`, at least not directly in a package.py. Package definitions need to be as non-python-version-specific as possible (in fact, I wish there were a way to enforce this, or even lock it onto py2). If you use py3 specifics in a package.py for eg, you've now guaranteed that you can't build/release that package with anything other than a py3-based rez installation. Now installation specifics have leaked into your package repo, which is terrible and unintended.

instinct-vfx:
Good points. I was thinking more along the lines of using it in the background and have a custom rez object like you suggest. But i guess the implementation detail does not matter really as soon as the package is a rez object.

frankchen211:
1. 
I agree that we need a method to identity that values set to env are paths, so the current fix is only var `PYTHONPATH`.
```
            if key == 'PYTHONPATH':
                value = self.escape_string(value)
                sys.path = value.split(os.pathsep)
                sys.path = [os.path.normpath(v) for v in value.split(os.pathsep)]
```

2.
Yes I have admit that this is a big thing and change to test develop package before pushing to production, and shouldn't be a PR at this moment, but we really need a more convenient method to test packages, the demo fix(which is not good) here is resolving a package from a random specified location that don't have to be in any Repository.

I'd like there's a better solution here.

Thx.

nerdvegas:
1.
Rez always attempts to avoid special cases, and is language agnostic where possible (and where it isn't - eg rez-pip - we want to extract into a separate project). I don't like hardcoding for PYTHONPATH here because that is a python specific thing. I think the approach is fine, but we should just base it on a configured list of envvars rather than PYTHONPATH specifically. I think it would be acceptable to default this setting to include PYTHONPATH and PATH though (configuration to take a common case into account is ok, imo).

2.
I don't think we should ever mix built and un-built packages like this, there are too many cases where that won't work. If you want to test a package before pushing it to production, then just install it somewhere that isn't production, and test it there - this is what the local packages path is for. So generally you just `rez-build -i`, then `rez-test`. Is there a reason you aren't/cannot do that? Perhaps if you explain the problems you're having we could better figure out what should be done. The best way to do this is to create an issue - you should then create your own branch referencing this issue, after there's been some discussion and general agreement on what should be done.

The other problem here is that, separately, it could make sense to run tests directly on the un-built (ie developer) package. For example, linting does not require the package to be built. But by supporting this idea of testing a package as a proper package in-place and un-built, it's confusing these two separate situations.

PR #1264, opened 2022-03-28
Fix python bind

PaoloFazio:
# Problem 

Currently I have Python **3.7.11** correctly bound as rez package in a Windows machine.

Now I want to bind another Python (**3.10.2**), so I call the cmd:
`rez-bind python --exe path/to/python3.10.2`
that is creating the following dir structure:
```
packages/
  python/
    3.10.2/
      platform-windows/
        arch-AMD64/
          os-windows-10.0.19041.SP0/
            bin/
              symlink to path/to/python3.10.2/python.exe
            python/
              extra/
                site-packages of Python 3.7.11      # should be 3.10.2 instead
              lib/
                lib of Python 3.7.11                # should be 3.10.2 instead
```

As the final folder structure suggests, we're symlinking the executable of Python **3.10.2**,
but copying lib and site-packages of Python **3.7.11**, which is the python being used at that moment.

The executable of Python 3.10.2 is correctly [found](https://github.com/nerdvegas/rez/blob/ad12105d89d658e4d2ea9249e537b3de90391f0e/src/rez/bind/python.py#L43),
but not passed to the [command launcher](https://github.com/nerdvegas/rez/blob/ad12105d89d658e4d2ea9249e537b3de90391f0e/src/rez/bind/python.py#L56),
resulting in picking the [current python](https://github.com/nerdvegas/rez/blob/ad12105d89d658e4d2ea9249e537b3de90391f0e/src/rez/bind/_utils.py#L41) (3.7.11) libraries to be copied instead.

# Solution

Pass python executable path to the command launcher.

### DIFF
--- a/src/rez/bind/python.py
+++ b/src/rez/bind/python.py
@@ -55,7 +55,9 @@ def bind(path, version_range=None, opts=None, parser=None):
     for dirname, module_name in entries:
         success, out, err = run_python_command([
             "import %s" % module_name,
-            "print(%s.__file__)" % module_name])
+            "print(%s.__file__)" % module_name],
+            exe=exepath
+        )
 
         if success:
             pypath = os.path.dirname(out)


nerdvegas (approved):

PaoloFazio:
Any chance to have this merged?  
Or, are people using known workarounds to bind multiple pythons?

bpabel (approved):

zachlewis (approved):

zachlewis:
Would be great to see this PR merged. Deprecated or not, it's frustrating and confusing that `rez-bind python --exe <PATH>` doesn't work as advertised.

PR #1265, opened 2022-03-28
Fix utils.which for symlinks with python <3.8 on windows

aboellinger:
# Problem

The implementation of `which` relies on `os.path.realpath` working on windows symlinks and junctions.
As stated in python documentation, this is the case only for python >=3.8
https://docs.python.org/3/library/os.path.html#os.path.realpath

A symlink is letft unchanged, but instead it can be read:
```py
>>> sys.version  # <3.8
'3.7.11 (tags/v3.7.11:9da28d2b342, Aug 10 2021, 15:34:44) [MSC v.1916 64 bit (AMD64)]'
>>> os.path.islink(path)
True
>>> path == os.path.realpath(path)  # Unchanged :'(
True
>>> path == os.readlink(path)
False
```

# Solution

This patch implements a fallback mechanism to read the link if it was not resolved.

### DIFF
--- a/src/rez/utils/which.py
+++ b/src/rez/utils/which.py
@@ -72,8 +72,14 @@ def which(cmd, mode=os.F_OK | os.X_OK, path=None, env=None):
             # Account for cmd possibly being a symlink. A symlink can be an
             # executable on windows without an extension. If it is, see if its
             # target's extension matches any of the expected path extensions.
-            #
             realfile = os.path.realpath(os.path.join(normdir, cmd)).lower()
+
+            # On windows and python < 3.8
+            # os.path.realpath leaves symlink and junctions unchanged
+            # https://docs.python.org/3/library/os.path.html#os.path.realpath
+            if os.path.islink(realfile):
+                realfile = os.readlink(realfile).lower()
+
             if any(realfile.endswith(x) for x in pathext):
                 files = [cmd]
             else:
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ docs/_build
 wiki/out
 LATEST_CHANGELOG.md
 __pycache__
+.vscode
+


nerdvegas (approved):

src/rez/utils/which.py
```diff
@@ -72,8 +72,14 @@ def _access_check(fn, mode):
             # Account for cmd possibly being a symlink. A symlink can be an
             # executable on windows without an extension. If it is, see if its
             # target's extension matches any of the expected path extensions.
-            #
             realfile = os.path.realpath(os.path.join(normdir, cmd)).lower()
+
+            # On windows and python < 3.8
+            # os.path.realpath leaves symlink and junctions unchanged
+            # https://docs.python.org/3/library/os.path.html#os.path.realpath
```
nerdvegas, line 79: +1 for the thorough commenting

aboellinger:
> @aboellinger Do you have details on how you found this bug? I'd be curious to know if a new test would be required if it's one of the rez commands that was failing.

@JeanChristopheMorinPerso Mmm, I don't recall exactly at which point in our tests we run into that issue. Windows symlinks are definitely a strange beast, trying to start from a clean rez install to provide minimal reproduction steps I run into different issues much quicker. Things seem to work better if we keep the extension name in the symlink. For now we hacked our way out of it by implementing our own bind scripts that wrap the executables into batch `.bat` scripts and it does the trick.

Also a coworker of mine has raised a concern about my proposed fix not being compatible with python 2.7.

Maybe we should **just close this PR** for now until a more general and elegant fix is found.

---

For the record, this is the kind of behaviour I get on windows with a fresh install:

```
C:\rez-220427\Scripts\rez>rez env python -- where python
c:\Users\aboellinger\packages\python\3.7.11\platform-windows\arch-AMD64\os-windows-10.0.19041.SP0\bin\python.exe

C:\rez-220427\Scripts\rez>rez env python -- python

C:\rez-220427\Scripts\rez>rez env python -- where python
c:\Users\aboellinger\packages\python\3.7.11\platform-windows\arch-AMD64\os-windows-10.0.19041.SP0\bin\python

C:\rez-220427\Scripts\rez>rez env python -- python
'python' is not recognized as an internal or external command,

C:\rez-220427\Scripts\rez>rez env python -- start python
# Works

# After (manually) renaming the symlink to add the extension:
C:\rez-220427\Scripts\rez>rez env python -- where python
c:\Users\aboellinger\packages\python\3.7.11\platform-windows\arch-AMD64\os-windows-10.0.19041.SP0\bin\python.exe

C:\rez-220427\Scripts\rez>rez env python -- python
# Launches, but then fails because it does not find Python37.dll
```

---

Also for the record, this is what our batch `.bat` wrappers can look like, with passing arguments and respecting exit code.
We generate them on the fly instead of creating the symlink.

```bat
@ECHO OFF
"c:\program files\shotgun\python3\python.exe" %*
EXIT /B %ErrorLevel%
```

JeanChristopheMorinPerso:
Sorry for the delay... As you suggested I'll close this PR. Feel free create a new one if you find a more elegant solution.

Issue #1263, opened 2022-03-28
Another way to test packages in dev stage


frankchen211:
Allow resolving a package from a random specified place instead of paths from rezconfig `packages_path` for testing purpose before pushing to production.

Let's say we have a package `helloworld` in **d:\dev\helloworld** (no version specified at this time) want to test, hope to init env like this:
`rez-env helloworld pkg1 pkg2 ... -d d:\dev`
`helloworld` will be resolved to D:\dev\helloworld, `pkg1` and `pkg2` resolved to a repository path in rez config. or if they are also in D:\dev folder.
```
helloworld                d:\dev\helloworld
pkg1-3.3                   c:\users\user\packages\pkg1\3.3\                                           (local)
pkg2-1.0.0                c:\users\user\packages\pkg2\1.0.0                                          (local)
```
or just directly specify the package path:
`rez-env helloworld@d:\dev\helloworld pkg1 pkg2 ...`
Currently if I have a package need to test before pushing to production, I have `rez-build -i`, then `rez-test` or open a python env(in DCC or console)

**Motivation**
Current workflow for testing a package generally is `rez-build -i`, then `rez-test` if we have unit test code or open dcc or `rez-env xxx -- python`  to test.
This means you have to copy a package to the `local package path` , this path also in `packages_path`  in rezconfig.py to search pacakges by default.

Reason why I don't test a packge in this way:
In my studio, all our workstations will be used to render jobs at night, which means if I have some packages are in test stage in local, any production render jobs that using my workstation will use these code, this is not what we are expected. (Some packages we do mirror to local to speed up the load time in production, local path is not always for testing purpose.)

If I want to invite other developers/artists to test my packages(not a formal test in our pipeline), I have to copy to a central repository and give a `testing version` carefully to avoid affecting production, since these packages are in develop stage, I really don't want to `release` it to production repository, instead just putting it to a sandbox and tell rez-env to resolve these packages from given paths, required pacakges still can be resolved from formal production repos.

**Related Issues/PRs**
* #1253

nerdvegas:
I would strongly suggest that your render jobs be configured to use a
different REZ_PACKAGES_PATH setting that does not include `$home/packages`.
If you've implemented some form of package localisation (as it sounds
like), then I would make this separate to `$home/packages`. To me this
sounds like the real source of your problems - you're unable to
differentiate between these two quite different use cases.

Hth
A

frankchen211:
[...]

PR #1270, opened 2022-04-03
Issue 1263 new dev paths options

frankchen211:
According to @nerdvegas's suggest, added `add_pre_paths` and `add_post_paths` options to rez-env cli.

Related issue:
#1263

### DIFF
--- a/src/rez/cli/env.py
+++ b/src/rez/cli/env.py
@@ -49,6 +49,10 @@ def setup_parser(parser, completions=False):
         "--paths", type=str, default=None,
         help="set package search path (use %r separator)" % os.pathsep)
     parser.add_argument(
+        "-d", "--dev-paths", type=str, default=None,
+        help="search packages from these dev paths if found, otherwise from formal search path,"
+             "normally for testing purpose, non-version packages (use %r separator)." % os.pathsep)
+    parser.add_argument(
         "-t", "--time", type=str,
         help="ignore packages released after the given time. Supported formats "
         "are: epoch time (eg 1393014494), or relative time (eg -10s, -5m, "
@@ -156,6 +160,13 @@ def command(opts, parser, extra_arg_groups=None):
     context = None
     request = opts.PKG
     t = get_epoch_time_from_str(opts.time) if opts.time else None
+
+    if opts.dev_paths:
+        for path in opts.dev_paths.split(os.pathsep):
+            path = os.path.abspath(path)
+            for idx, pkg in enumerate(request):
+                if pkg in os.listdir(path):
+                    request[idx] = '{}@{}'.format(pkg, os.path.join(path, pkg))
 
     if opts.paths is None:
         pkg_paths = (config.nonlocal_packages_path
--- a/src/rez/cli/env.py
+++ b/src/rez/cli/env.py
@@ -49,10 +49,6 @@ def setup_parser(parser, completions=False):
         "--paths", type=str, default=None,
         help="set package search path (use %r separator)" % os.pathsep)
     parser.add_argument(
-        "-d", "--dev-paths", type=str, default=None,
-        help="search packages from these dev paths if found, otherwise from formal search path,"
-             "normally for testing purpose, non-version packages (use %r separator)." % os.pathsep)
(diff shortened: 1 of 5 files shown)


frankchen211:
But this still can't ensure the resloved packages are always picked from the paths added, for testing purpose, we want to test the version we specifed.

For non version packge, have to do like this or specify a version in that path, otherwise rez will pick the highest version from other repo.
```
rez-env foo== --add-pre-paths E:\dev
```

ColinKennedy:
> The proper fix for this imo is - implement a new "stacked" package orderer, that can stack the result of
multiple other orderers; - use that to combine existing orderer with a favour-local orderer, if
necessary; - package that into a convenient --favour-local cli option / api func arg
> I think the most elegant way to implement the --favour-local
functionality you describe is probably with a new package orderer.

As mentioned in the thread, `--add-pre/post-paths` isn't a solution because it doesn't give you much functionality that setting REZ_PACKAGES_PATH can't.

frankchen211:
As I can see, to solve the problem mentioned in #1263, to allow requiring a package directly from a path `foo@/path/to/dev/foo` perhaps is the simplest method.


> > The proper fix for this imo is - implement a new "stacked" package orderer, that can stack the result of
> > multiple other orderers; - use that to combine existing orderer with a favour-local orderer, if
> > necessary; - package that into a convenient --favour-local cli option / api func arg
> > I think the most elegant way to implement the --favour-local
> > functionality you describe is probably with a new package orderer.
> 
> As mentioned in the thread, `--add-pre/post-paths` isn't a solution because it doesn't give you much functionality that setting REZ_PACKAGES_PATH can't. Maybe there's a case for adding these CLI arguments for other purposes but I don't think it addresses #1263, as you've also said.

ColinKennedy:
> As I can see, to solve the problem mentioned in https://github.com/nerdvegas/rez/issues/1263, to allow requiring a package directly from a path foo@/path/to/dev/foo perhaps is the simplest method.

The suggested PR change adds 2 new CLI arguments, `--add-pre/post-paths`. Under this PR, functionally, this:

```sh
rez-env blah --add-pre-paths /before --paths /foo --add-post-paths /after
```

Is the same as this:

```sh
rez-env blah --paths /before:/foo:/after
```

You can replicate the same behavior using `--paths` on its own. I don't want to speak for Allan so I'll let him weigh in, but it sounds like he wants to implement a stacked package orderer and wrap that into a CLI argument.

At present, I don't see this PR providing functionality that we can't already get with `--paths`, alone. IMO it's better to use `--paths` rather than having several parameters for the same thing.

frankchen211:
Yes, this only makes it a bit more user friendly instead of passing all paths by `--paths`, since not all TDs/developers/testers know where all rez repo paths are, and some of the paths are computed in production in our pipeline env, that's why added these new two opts, ppl don't need to append a long list of paths for testing.

Since `foo@/path/to/dev/foo` sounds like not a good idea for Allan, so I revert this part of code.

nerdvegas:
In the spirit of development by committee (which makes sense wrt the move to aswf) I'm gonna put this one to a vote. Personally I'm tending to not be in favour. As Colin points out you can just use `--path` instead, it's only slightly more convenient to have `--add-pre-paths`. The additional args don't break anything either though.

Please cast your vote, I'll check in again in a while.

ColinKennedy:
I sympathize with the issues pointed out in https://github.com/nerdvegas/rez/issues/1263 so I would like that workflow to be improved, however I don't think this PR fully solves the problem, I'd rather we wait for a package orderer to see if it covers all of our use-cases.

e.g. I'm imagining something like

`rez-env --favor-paths /foo:/bar/thing --paths /foo:/other/path:/bar/thing`. 

`--favor-paths` can provide a list of paths to prefer or, if no explicit path(s) are provided but `--favor-paths` is explicitly included, a default list containing just `local_packages_path` is passed, instead.

Once we have that functionality in, I suspect this PR won't be needed as the orderer will handle all of our use-cases. My vote is to wait or close this PR and, if the orderer doesn't do what we need, we can always re-open the PR later.

PR #1275, opened 2022-04-06
Added the missing paths=self.package_paths

ColinKennedy:
Fixes: https://github.com/nerdvegas/rez/issues/1274

Added missing `self.package_paths` so get_resolve_diff produces correct results. This is safe to do because of the existing `self.package_paths != other.package_paths` check, above.


JeanChristopheMorinPerso:
This looks alright to me. Can you rebase your branch please? Rebasing should fix the Windows CI failures.

JeanChristopheMorinPerso:
@ColinKennedy See https://github.com/AcademySoftwareFoundation/rez/runs/7476834008?check_suite_focus=true.

I think the copyright text is missing in the test file you added.

ColinKennedy:
@JeanChristopheMorinPerso I added the copyright header manually but probably should just run the autofix script. If I run that and it still fails, I will let you know

PR #1283, opened 2022-04-08
Issues/665 add rez test interactive option 2

ColinKennedy:
Added `--interactive` flag to `rez-test`. This addresses "Add rez-test --interactive option" from https://github.com/nerdvegas/rez/issues/665.

Here's a summary of the changes + any extra notes

# Change Summary
## Refactored / Moved
- `cli/_main.py` was both parsing and running a callable Python `run_cmd()` function all within `run()`. I split the function in two by introducing `parse_command()` (which I think leverage in unittests)
- package_test.py has a nested `_pre_test_commands`. I moved it to the global scope so that it can be reused in `rez-test and also for `rez-test --interactive`, since both functions need to call it.

## Added
- Added some new test Rez packages in `src/rez/data/tests/packages/rez_test_packages`
- Added `run_test_env`, the function which `--interactive` calls
- A new environment variable, REZ_CURRENT_TEST_NAME, was added. This is for unittest debugging but also closes https://github.com/nerdvegas/rez/issues/1261

### _choose_test_object
`_choose_test_object` is used to "narrow" the possible variants for a Rez package to use for the test environment. It does this based on what it can find in the Rez package tests's "requires", "on_variants", and anything passed to `--extra-packages`. However things get awkward when more than one valid candidate is found. I wasn't sure how to handle this so I simply did `candidates[0]`. If there's an API way to "find the best candidate" I can leverage, please let me know
    - Reference thread: https://academysoftwarefdn.slack.com/archives/C0321B828FM/p1649308434986989

### test_test.py _override_execute_shell
The try/finally in _override_execute_shell is a bit hacky. Probably I can do this more cleanly with mock.patch but we don't have mock in Rez's list of vendors. I can keep this code as-is or copy a Python 2/3 version of mock and use that. Do you have a preference?

# --interactive behavior

The jist is - it takes any Rez test name and makes a rez-env of package + the test. `--extra-packages` can be optionally added to influence the variant of the package that gets chosen and/or append extra packages to the resulting environment. There's a bunch of unittests to deal with "requires", "on_variants", and the like but otherwise that, it's pretty much it. See `test_test.py` for all of the covered scenarios. If you can think of other scenarios, please let me know and I'll add tests for them.


src/rez/package_test.py
```diff
@@ -718,3 +841,163 @@ def print_summary(self):
...
+        if not extra_packages:
+            raise PackageTestError(
+                "No valid variant could match your given Rez test requires."
+            )
+
+        raise PackageTestError(
+            "No valid variant could match your given Rez test requires + extra packages."
+        )
+
+    return candidates[0]
```
ColinKennedy, line 916: At this point in the code, there may be 1-or-more valid variants which can be used. I'm not sure how to "choose" the "best" one so for now I'm just returning the 0th index, the first valid variant. If there's a smarter way to do this, please let me know!

src/rez/tests/test_test.py
```diff
@@ -0,0 +1,323 @@
...
+    original = resolved_context.ResolvedContext.execute_shell
+
+    try:
+        resolved_context.ResolvedContext.execute_shell = _wrap(
+            resolved_context.ResolvedContext.execute_shell
+        )
+
+        yield path
+    finally:
+        resolved_context.ResolvedContext.execute_shell = original
```
ColinKennedy, line 269: This code is fairly hacky but is necessary to prevent interactive from **actually** rez-enving, because we don't actually want that to happen in the unittest. If we had `mock` as a vendor, I would've `mock.patch`ed here. But wasn't sure if there that's okay to add. Just wanted to raise here that `mock` would've been helpful here, for what it's worth.

src/rez/package_test.py
```diff
@@ -718,3 +841,163 @@ def print_summary(self):
...
+def _choose_test_object(package, test_requires, extra_packages=frozenset()):
+    """Find the most appropriate Rez object for ``rez-test --interactive``.
+
+    If ``package`` has variants, find a valid variant, given ``test_requires``
+    and ``extra_packages``.
+
+    Args:
+        package (Package):
+            The installed Rez package to query variant / test data from.
+        test_requires (set[PackageRequest], optional)
```
JeanChristopheMorinPerso, line 877: It's a positional parameter, so it's required, not optional. Also you don't need to document if a field is optional or not.

src/rez/package_test.py
```diff
@@ -718,3 +841,163 @@ def print_summary(self):
...
+
+    Raises:
+        PackageTestError: If any conflict between the given parameters was found.
+
+    """
+    if _is_conflict(test_requires):
+        test_requires_names = ", ".join(str(request) for request in test_requires)
+
+        raise PackageTestError(
+            'Test has conflicting requires "{test_requires_names}".'.format(
```
JeanChristopheMorinPerso, line 986:
```suggestion
            'Test has conflicting requires {test_requires_names!r}.'.format(
```
Tip: `!r` will add quotes for you.

src/rez/package_test.py
```diff
@@ -718,3 +841,163 @@ def print_summary(self):
...
+        variant (Package or Variant):
+            A Rez object which may define ``pre_test_commands``.
+        executor (RexExecutor):
+            An interpreter to load commands during the ``rez-env``.
+
+    """
+    _pre_test_commands(test_name, variant, executor)
+
+    if hasattr(variant, "index"):
+        executor.setenv("REZ_TEST_VARIANT_INDEX", variant.index)
```
JeanChristopheMorinPerso, line 934: Should this block be moved in `_pre_test_commands`? It probably wouldn't hurt to have this variable defined in the normal "rez-test` env. Which basically mean that interactive and non interactive could use the same function.

Also, maybe `REZ_CURRENT_TEST_VARIANT_INDEX`?

src/rez/package_test.py
```diff
@@ -718,3 +841,163 @@ def print_summary(self):
...
+
+    test_ns = {
+        "name": test_name
+    }
+
+    with executor.reset_globals():
+        executor.bind("this", variant)
+        executor.bind("test", RO_AttrDictWrapper(test_ns))
+        executor.execute_code(pre_test_commands)
+        executor.setenv("REZ_CURRENT_TEST_NAME", test_name)
```
JeanChristopheMorinPerso, line 966:
```suggestion
    with executor.reset_globals():
        executor.setenv("REZ_CURRENT_TEST_NAME", test_name)

        if not pre_test_commands:
            return

        executor.bind("this", variant)
        executor.bind("test", RO_AttrDictWrapper({"name": test_name}))
        executor.execute_code(pre_test_commands)
        executor.setenv("REZ_CURRENT_TEST_NAME", test_name)
```
This would better convey the intention.

src/rez/tests/test_test.py
```diff
@@ -0,0 +1,323 @@
...
+import os
+import platform
+import shlex
+import tempfile
+
+from rez import resolved_context, exceptions as rez_exceptions
+from rez.cli import _main
+from rez.tests import util
+
+_SUCCESS_CODE = 0
```
JeanChristopheMorinPerso, line 29: Unused

src/rez/tests/test_test.py
JeanChristopheMorinPerso: If it makes our life easier and the code much easier to read and understand, we should really consider adding `mock` as a runtime dependency. I don't think we need to vendor it...

wiki/pages/Testing-Packages.md
```diff
@@ -0,0 +1,21 @@
...
+	]$ rez-test maya_utils --interactive unittest
+
+In the example commands above, a "rez-env" is auto-generated to combine
+"maya_utils" and "lint" / "maya_utils" and "unittest" into two, separate
+environments.
+
+Note that you cannot call *--interactive* on more than one Rez test at a time.
+
+The resulting environment is exactly what Rez sees, just before a test
+"command" gets run.
```
JeanChristopheMorinPerso, line 21:
```suggestion
This command runs two test targets, "lint" and "unittest", from a Rez package called
"maya_utils". Rez will execute each test target command following the order the targets.

It is also possible to "rez-env" into a test environment by using the `--interactive` flag:

	]$ rez-test maya_utils --interactive lint

The resulting environment is the combination of the requirements defined by the `maya_utils`
package and the requirements of the `lint` target and is exactly the same as the one used
when the `--interactive` flag is not provided.

Note that you cannot call `--interactive` on more than one target.
```

JeanChristopheMorinPerso (commented):
I did a first pass. I'll let others with more experience with the API review the implementation.

src/rez/tests/test_test.py
ColinKennedy: I thought Rez had to copy all external dependencies into the vendor folder - Something Allan mentioned on the forums about Rez needing to be self enclosed. Is that no longer the case?

JeanChristopheMorinPerso: To me test dependencies should not br vendored.

We could easily have an extra in the setup.py that installs pytest and mock.

src/rez/package_test.py
ColinKennedy: Thank you, I was actually really hoping someone more familiar with executor would suggest a refactor. I was trying to match the original but it ended up becoming a bit weird, as you can see.

src/rez/package_test.py
ColinKennedy: Ah I missed that one after the refactor, fixed. I'll do another pass to make sure I got them all. It's left over from when I previously had --interactive resolve multiple tests.

Though it's not required, stating optional is supported by sphinx.ext.napoleon and Sphinx [parses the type](https://crimsonscholar.github.io/rez_sphinx/latest/api/rez_sphinx.core.configuration.html#rez_sphinx.core.configuration.ConfPy.get_attributes) correctly.

Probably moot since one day we'll probably end up using type hints for this stuff anyway.

src/rez/package_test.py
ColinKennedy: Sure, done

src/rez/package_test.py
ColinKennedy: We could put it in `_pre_test_commands` and then delete `_pre_test_env_commands` probably. I don't have a strong opinion one way or another.

However I will definitely change it to REZ_CURRENT_*, to match the other environment variable. Thank you for the suggestion!

As for whether to move or stay, I'll leave that to other contributors

wiki/pages/Testing-Packages.md
ColinKennedy: I find your suggested edit a bit confusing. It removes the second --interactive command but keeps the original `rez-test maya_utils --interactive lint unittest`. Which is fine, but then there's this line.

`Rez will execute each test target command following the order the targets.`
At first I thought that also applied to `--interactive` but of course it can't, since you can only select one test at a time with that flag.

I think I understand what you're trying to suggest in this edit, I'll give a shot at a re-write in a moment. Could you have a look again once I've pushed it?

src/rez/tests/test_test.py
```diff
@@ -0,0 +1,323 @@
...
+        resolved_context.ResolvedContext.execute_shell = _wrap(
+            resolved_context.ResolvedContext.execute_shell
+        )
+
+        yield path
+    finally:
+        resolved_context.ResolvedContext.execute_shell = original
+
+
+def _test(command, packages_path=tuple()):
```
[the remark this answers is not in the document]
ColinKennedy: I'm assuming you're suggesting this because None is a common "sentinel value" however that'd break `parts.extend(packages_path)`. To fix it, `parts.extend(packages_path or [])` or add `package_paths = package_paths or []` would be needed. 

I'm happy to default to `None` in most cases but in this case, `None` diverges from the intended type (an iterable), and `tuple()` is immutable as well. I can change it since it's a small tweak but it doesn't make sense to me.

src/rez/tests/test_test.py
ColinKennedy: Oh okay cool, I didn't know that was an option. I'm happy to do that if others agree as well!

src/rez/tests/test_test.py
```diff
@@ -0,0 +1,323 @@
...
+        run_command()
+    except SystemExit:
+        # Since Rez's CLI functions don't separate their function logic from
+        # sys.exit / CLI logic, we have to catch the SystemExit calls so we can
+        # ignore them.
+        #
+        pass
+
+
+def _test_shell(command, packages_path=tuple(), variable="REZ_RESOLVE"):
```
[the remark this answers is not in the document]
ColinKennedy: Same as https://github.com/nerdvegas/rez/pull/1283#discussion_r846567776

wiki/pages/Testing-Packages.md
ColinKennedy: Not sure why GitHub didn't update the PR but FWIW I pushed it here: https://github.com/ColinKennedy/rez/blob/issues/665-add_rez_test_interactive_option_2/wiki/pages/Testing-Packages.md

wiki/pages/Testing-Packages.md
```diff
@@ -0,0 +1,17 @@
+As mentioned in other pages such as [tests](Package-Definition-Guide#tests),
+you can specify Rez tests to run like this:
+
+	]$ rez-test maya_utils lint unittest_name
+
+This example runs two tests, "lint" and "unittest_name", from the "maya_utils"
```
JeanChristopheMorinPerso, line 6:
```suggestion
This example runs two test targets, "lint" and "unittest_name", from the "maya_utils"
```

I think using test target instead of test is important because targets are not necessarily tests. The `rez-test` command is a little bit misnamed but that's another discussion (a discussion that we already had in the last years and that I feel we will probably have in the future too).

ColinKennedy: > The rez-test command is a little bit misnamed but that's another discussion (a discussion that we already had in the last years and that I feel we will probably have in the future too).

Yeah, I always just pretend it's called `rez-task`, haha. rez-*test* implies that it's only for unittesting.

wiki/pages/Testing-Packages.md
```diff
@@ -0,0 +1,17 @@
...
+
+You may also "enter" any test environment, instead of running a test by
+including the *--interactive* flag.
+
+	]$ rez-test maya_utils --interactive unittest_name
+
+Adding *--interactive* makes Rez combine the "maya_utils" Rez package and
+"unittest_name" Rez test into a single environment, just like how `rez-env` works.
+
+Note that *--interactive* can only target on Rez test at a time.
```
ColinKennedy, line 17:
```suggestion
Note that *--interactive* can only target one Rez test at a time.
```

Issue #1255, opened 2022-03-22
Possible regression appendenv in PowerShell fails if env variable does not exist already


instinct-vfx:
I have not looked deeper into this, the old version we had deployed (2.88.4) seemingly creates the env variable whereas the current release fails with a PowerShell error.

https://github.com/nerdvegas/rez/blob/d4b303f8a23d2d190d4d48183c6ae51a906e98cc/src/rezplugins/shell/_utils/powershell_base.py#L256

Looking into this asap.

PR #1285, opened 2022-04-12
Fix 1255-possible-regression-appendenv-in-powershell-fails-if-env-variable-does-not-exist-already

instinct-vfx:
There is a regression that was introduced by #1176 
The nested `Get-ChildItem` call fails with an error if the Variable does not exist. Old behaviour restored without reverting the fix by adding `-ErrorAction SilentlyContinue`


instinct-vfx:
Fixes #1255

JeanChristopheMorinPerso (approved):
lgtm

SitiSchu:
What needs to be done to get this merged?

bpabel:
I would value a comment explaining why the command is set to silently ignore errors. It's one of those things I could see getting removed in the future if no one remembers why it was added.

JeanChristopheMorinPerso (approved):

maxnbk (approved):

PR #1287, opened 2022-04-16
fix: add platform/arch to variant for rez-pip packages with entry point scripts

### LINKED ISSUE #1286: rez-pip doesn't include arch/platform in variant for packages with platform-specific generated scripts
rez-pip currently only includes `platform` and `arch` in the variant when generating a rez package for python packages that aren't pure python. However, pure python packages that define entry points scripts (console_scripts or gui_scripts) will generate platform-specific executables, and require the platform and arch tags, so that installations from different platforms can be installed to separate variants, each with their own set of generated platform-specific executables.

**Environment**
* OS - all os's
* Rez version - "2.107.0")
* Rez python version - Python 3.7.7

**To Reproduce**
1. rez-pip a pure python package with console scripts (e.g. rez-pip -i flit)

**Expected behavior**
The generated package.py variant includes the platform and arch packages it was generated from.

**Actual behavior**
The generated package.py variant does not include the platform and arch packages it was generated from (only the python version).

bpabel:
Fixes #1286


src/rez/utils/pip.py
```diff
@@ -332,6 +332,22 @@ def is_pure_python_package(installed_dist):
     return (is_purelib == "true")
 
 
+def is_entry_points_scripts_package(installed_dist):
```
JeanChristopheMorinPerso, line 335: `is_pure_python_package` has unit tests. It would be cool if `is_entry_points_scripts_package` would also have a unit test. See https://github.com/AcademySoftwareFoundation/rez/blob/master/src/rez/tests/test_pip_utils.py#L144.

Also, I'm tempted to say that `scripts` in the function name is not necessary since all entry points are scripts.

src/rez/utils/pip.py
```diff
@@ -383,7 +399,8 @@ def get_rez_requirements(installed_dist, python_version, name_casings=None):
 
     # assume package is platform- and arch- specific if it isn't pure python
     is_pure_python = is_pure_python_package(installed_dist)
-    if not is_pure_python:
+    has_entry_points_scripts = is_entry_points_scripts_package(installed_dist)
```
JeanChristopheMorinPerso, line 402: Could you add a little comment to explain why it's required? Someone unfamiliar with Windows might not know that scripts in Windows are executables.

JeanChristopheMorinPerso (changes requested):

JeanChristopheMorinPerso:
I looked at the build failures and I'm not too sure what's going on. And our builds are being massively queued right now, which makes debugging quite difficult.

Also, you will have to sign off your commits. See https://github.com/AcademySoftwareFoundation/rez/pull/1287/checks?check_run_id=6716785034.

src/rez/utils/pip.py
bpabel: Thanks, I'll add unit tests. There are other uses for entry points other than scripts, like defining plugins or hooks.

JeanChristopheMorinPerso:
Actually, wait a little bit, we have another failure on master.

JeanChristopheMorinPerso:
Alright, you can rebase now. The issue has been fixed and the workflow should work.

bpabel:
@JeanChristopheMorinPerso Cool, thanks. Rebased and updated from your notes. Looks like builds are succeeding now.

JeanChristopheMorinPerso (approved):
Looks good to me. @instinct-vfx any opinion on this before we merge this PR?

maxnbk (approved):

instinct-vfx:
I did not have the chance to actually test myself but it looks good to me and is also an important fix (specifically on windows). At some time in the future i would like to revisit rez-pip, move to a more sensible approach to the file mapping and also add less specific variants for pure python packages without entrypoints. Good one!

PR #1295, opened 2022-04-21
REP-004 - Initial readthedocs.io integrations

ColinKennedy:
This WIP PR integrates Rez with readthedocs.io as well as auto-generates API documentation. Effectively, this PR accomplishes both REP-004.1 and then verifies that it's working by also implementing REP-004.2 at the same time (See https://github.com/nerdvegas/rez/issues/1292 for details).

You can [view the documentation in readthedocs.io directly](https://rez-rtd-test.readthedocs.io/en/update_sphinx_documentation_2) or check out the screenshot, below

Screenshot from 2022-04-20 22-29-33

### Added
- `BUILDING.md`, so people know how to build the documentation locally
- `docs/_templates` - used by sphinx-apidoc to generate the pretty toctree
- `docs/requirements.txt` - Needed by both readthedocs.io and useful for local building

### Removed
- All existing `docs/api*.rst` files, since they are now auto-generated pre-build

### Updated
- `docs/conf.py` runs `sphinx-apidoc` prior to building documentation
- The existing `docs/index.rst` uses [Python style guide sections](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections)
- `developer_package.DeveloperPackage.from_path` docstring was changed to prove that intra-links import and are working (see screenshot)
    - You can also see it [in this URL](https://rez-rtd-test.readthedocs.io/en/update_sphinx_documentation_2/api/rez.developer_package.html#rez.developer_package.DeveloperPackage.from_path
)
- The Rez wiki now points to the new documentation (the URL will need  to be replaced, prior to merging)

## On readthedocs.io Set-Up
To get readthedocs.io working, you must make a GitHub integration hook, enable SemVer tagging so new tags will get generated documentation. The jist is:

In readthedocs.io

- Go to Admin > Integrations, add a GitHub integration hook
    - So GitHub forces changes to readthedocs
- Go to Admin > Automation Rules > add SemVer Tags (Action: Activate Version)
    - So new releases get documentation built for them
- Go to Admin > Advanced Settings > Requirements file - set to `docs/requirements.txt`

And voila, documentation now builds.

Extra settings:
- Go to Admin > Advanced Settings > Enable "Build pull requests for this project"

## Out of scope things (which make sense for another PR)
We currently have a sphinx.yaml GitHub action and docs/build.py which makes GitHub pages. This PR does not make an attempt to unify / deprecate this at the moment. Please visit [this discussion board](https://github.com/nerdvegas/rez/discussions/1293) if you'd like to talk about how / when we should go about doing that

## What this PR (draft) is about
An opportunity for maintainers to point out  obvious issues regarding setting up Rez for readthedocs.io (if any). Specifically comments relating to REP-004.1 and REP-004.2

Closes: https://github.com/nerdvegas/rez/issues/1254

## Next Steps

rezgui importing is broken due QApplication being instantiated in `__main__`. See rezgui.objects.App.app. To get the code to import properly, that GUI code will need refactoring. For now, src/rezgui is ignored as a GUI refactor is out of scope for this PR.

Example failed logs:

- https://readthedocs.org/api/v2/build/16704525.txt
- https://readthedocs.org/projects/rez-rtd-test/builds/16704584/


ColinKennedy:
Aside from a single TODO listed in `docs/BUILDING.md`, this PR should be ready to review. @JeanChristopheMorinPerso and @instinct-vfx  please let me know what you think as well!

docs/requirements.txt
```diff
@@ -0,0 +1,5 @@
+PySide2>=5.15
+Qt.py>=1.3
+pysvn>=0.1
+Sphinx
+sphinx_rtd_theme>=1.0.0
```
JeanChristopheMorinPerso, line 5: Both of these should be pinned so that we have reproducible docs. Also why is PySide required?

docs/BUILDING.md
```diff
@@ -0,0 +1,39 @@
...
+- Install the pip requirements
+
+```sh
+python -m pip install --requirement requirements.txt
+```
+
+- Now build the documentation
+
+```sh
+sphinx-build . _build
```
JeanChristopheMorinPerso, line 31: I'd recommend users to use `make html` or the likes instead of `sphinx-build`. But I think the makefiles are missing, so a new bootstrap wuld be welcome.

JeanChristopheMorinPerso (changes requested):
I'm generally more an autosummary guy than pure `sphinx-apidoc` because it gives us more flexibility on which function makes it into the documentation. (I'm generally on the side that not all functions/classes in rez should be part of the public API).

Also, it would be great if you could add a config file for read the docs, see https://docs.readthedocs.io/en/stable/config-file/v2.html. It's quite important so that we can define which OS image is used, which python version, etc.

Lastly, I would like to see the `conf.py` file use the same style sa https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/docs/conf.py, which is becoming the "defacto" style of the config nowadays. Note that boilerplate comments and descriptions are removed, the config is divided in sections with the links to the documentation for each section, and only the required parameters/settings are set.

JeanChristopheMorinPerso:
Oh, and I forgot to say that the `sphinx` workflow should be adapted.

It should have these jobs
* Build that builds the docs in strict mode and fail if there is any warning or error generated by sphinx.
* Linkcheck that runs the [linkcheck](https://www.sphinx-doc.org/en/master/usage/builders/index.html#sphinx.builders.linkcheck.CheckExternalLinksBuilder) builder.

And the publish job should be removed.

These jobs will let us know when something won't render correctly in the docs and will help us when we upgrade the docs dependencies.

docs/requirements.txt
maxnbk: I assume PySide is required because sphinx has to be able to import the things that it documents, and rez-gui would not import correctly without at least one of the GUI frameworks being available.

Could probably be removed if we want to remove docs for rez-gui?

JeanChristopheMorinPerso: My question kind of implied that it should be removed and we should tell sphinx to ignore rez-gui. But that wasn't  clear at all... Sorry about that.

JeanChristopheMorinPerso:
Superseded by #1522.

PR #1299, opened 2022-04-23
Added to_environment_name

ColinKennedy:
Closes: https://github.com/nerdvegas/rez/issues/1298

Adds `to_environment_name` - converts a Rez package family name to what its equivalent environment variable text.

### DIFF
--- a/src/rez/utils/filesystem.py
+++ b/src/rez/utils/filesystem.py
@@ -486,6 +486,19 @@ def movetree(src, dst):
         shutil.rmtree(src)
 
 
+def to_environment_name(name):
+    """Change ``name`` from a Rez package family name to is env-var equivalent.
+
+    Args:
+        name (str): A Rez package name. e.g. ``"python"``, ``"foo.bar"``, etc.
+
+    Returns:
+        str: The converted name. e.g. ``"PYTHON"``, ``"FOO_BAR"``, etc.
+
+    """
+    return name.upper().replace('.', '_')
+
+
 def safe_chmod(path, mode):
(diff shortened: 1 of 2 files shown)


src/rez/utils/filesystem.py
```diff
@@ -493,6 +493,19 @@ def safe_chmod(path, mode):
         os.chmod(path, mode)
 
 
+def to_environment_name(name):
```
JeanChristopheMorinPerso, line 496: I'm not entirely sure this should be in the filesystem sub-package.

instinct-vfx: Wonder if it would be better to have it in https://github.com/AcademySoftwareFoundation/rez/blob/master/src/rez/utils/formatting.py ?

JeanChristopheMorinPerso (changes requested):
Marking as "request changes"

src/rez/utils/filesystem.py
nerdvegas: Agreed it is not related to filesystem.

My vote would be to add classmethod `ResolvedContext.pkg_name_to_envvar_name`.

JeanChristopheMorinPerso: `ResolvedContext` is a crowded class. Not sure I would put it as a class method there. But I do understand why you are suggesting that. `REZ_*` environment variables are only used from within a resolved context. I'd still vote for a utils like module like @instinct-vfx suggested (though maybe not formatting?).

nerdvegas: I don't love it either tbh. Let's go with utils/formatting.py..

PR #1305, opened 2022-05-07
Add --favor-paths rez-env flag

ColinKennedy:
Closes: https://github.com/nerdvegas/rez/issues/1263

## Added rez-env cli option, `--favor-paths`
- If not provided, does nothing
- If provided and 1+ path specified, `--favor-paths /foo:/bar`
	- All packages found within /foo are preferred over packages found in /bar
	- All packages found in /bar are preferred over packages found within packages_path
- If provided but no paths specified, `--favor-paths`, `[config.local_packages_path]` is used. All other logic applies.
- In all cases, the usual package path rules still apply. e.g. any path provided in --favor-paths must be in `packages_path` or it won't be used during ordering
 
## Miscellaneous
Introduced `SplitPaths` to show an alternative way we can be parsing our arguments. If you like that approach, I can add it to `--paths` as well, which is currently being described in `setup_parser` but then parsed again, in env.py `command`.

In the original ticket thread, https://github.com/nerdvegas/rez/issues/1263#issuecomment-1086347517, @nerdvegas mentions that we may need a kind of "stacked orderer". However I didn't encounter any limitations in practice. `ResolvedContext` takes a list of package orderers so it should be fine. However please let me know if not and I'll make necessary adjustments.

### DIFF
--- a/src/rez/cli/env.py
+++ b/src/rez/cli/env.py
@@ -172,7 +172,7 @@ def setup_parser(parser, completions=False):
 
 
 def command(opts, parser, extra_arg_groups=None):
-    from rez.package_order import FavorPathsOrder
+    from rez.package_order import FavorPathsOrder, PackageOrderList
     from rez.resolved_context import ResolvedContext
     from rez.resolver import ResolverStatus
     from rez.package_filter import PackageFilterList, Rule
@@ -235,10 +235,10 @@ def command(opts, parser, extra_arg_groups=None):
             rule = Rule.parse_rule(rule_str)
             package_filter.add_inclusion(rule)
 
-        package_orderers = []
+        package_orderers = list(PackageOrderList.singleton)
 
         if opts.favor_paths:
-            package_orderers.append(FavorPathsOrder(opts.favor_paths))
+            package_orderers.insert(0, FavorPathsOrder(opts.favor_paths))
 
         # perform the resolve
         context = ResolvedContext(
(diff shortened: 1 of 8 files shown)


diegogarciahuerta:
Hi @ColinKennedy , thanks for this PR, super useful!

We have been testing it in Windows and found one issue that I think should be easy to resolve.

It looks like Rez implements a `canonical_path` function, that amongst other things, converts the package location to lowercase for case insensitive filesystems:
https://github.com/AcademySoftwareFoundation/rez/blob/71f2f612223dc7c1c88ea146dd0d06ac1ea45ad0/src/rezplugins/package_repository/filesystem.py#L500

canonical_path definition:
https://github.com/AcademySoftwareFoundation/rez/blob/71f2f612223dc7c1c88ea146dd0d06ac1ea45ad0/src/rez/utils/filesystem.py#L528

In FavorPathsOrder you are checking if the package location is in the list of favor paths:
https://github.com/AcademySoftwareFoundation/rez/pull/1305/files#diff-772d1d5249a49e66b7497839f3f80ddb367c98ee0c880a8a3b2edf421ae817feR174

this comparison should be done with `canonical_path` in mind otherwise the check most likely always fail in Windows. ie. windows drive letters tend to be uppercase.

To solve it there are a couple of obvious options, 
- either in the `__init__` of the package ordered, 
```
from rez.utils.filesystem import canonical_path

...

self._paths = [canonical_path(path) for path in paths]
```
or later on in the compoarison:
```
from rez.utils.filesystem import canonical_path

...
    def reorder(self, iterable, key=None):
...
        buckets = collections.OrderedDict([(canonical_path(path), []) for path in self._paths])
```
Hope this helps, I've tested both with our Windows setup and they work as expected.

PR #1312, opened 2022-05-27
- remove references to nerdvegas in comments

nerdvegas:
- bump windows CI salt to guarantee image rebuild
- windows CI will fail I think - needs to be updated to use aswf dockerhub creds

### DIFF
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -137,7 +137,7 @@ def find_python_in_context(context):
     # accidentally find a system python install.
     #
     context = context.copy()
-    context.append_sys_path = False  # https://github.com/nerdvegas/rez/issues/826
+    context.append_sys_path = False  # #826
 
     python_package = context.get_resolved_package("python")
     assert python_package
@@ -430,7 +430,7 @@ def pip_install_package(source_name, pip_version=None, python_version=None,
             # Make the package use hashed variants. This is required because we
             # can't control what ends up in its variants, and that can easily
             # include problematic chars (>, +, ! etc).
-            # TODO: https://github.com/nerdvegas/rez/issues/672
+            # TODO: #672
             #
             pkg.hashed_variants = True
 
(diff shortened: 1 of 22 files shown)


JeanChristopheMorinPerso:
My last commit did the trick to fix the actions.

src/rez/pip.py
```diff
@@ -137,7 +137,7 @@ def find_python_in_context(context):
     # accidentally find a system python install.
     #
     context = context.copy()
-    context.append_sys_path = False  # https://github.com/nerdvegas/rez/issues/826
+    context.append_sys_path = False  # #826
```
JeanChristopheMorinPerso, line 140: Taken out of context, it would be hard to know that this is a GitHub issue number.

JeanChristopheMorinPerso (approved):

src/rez/pip.py
nerdvegas: Agreed. It'd be nice if github had some smart syntax for issues that it could generate links on in its md previews :/. Perhaps we just write these like `issue-826` / `pr-826`, it'd be nice to standardize so they're greppable. I want to keep away from overly verbose full links.

JeanChristopheMorinPerso: I like full links as I can just copy/paste and it gets me straight to where I want. If we go with shorthands, it will require more clicks and steps for little benefits, considering there is only 12 links to issues in this PR.

But I also wish allowed us to link to other links in the repo from any Markdown files without relying on full URLs...

JeanChristopheMorinPerso: But I can live with `issue-####` if you prefer that.

nerdvegas: I meant what you mentioned there ^ ie some sort of relative link for the issue/pr, that'd be a lot cleaner than a full link.

JeanChristopheMorinPerso: We can always buy a domain and make it redirect to GH...

Issue #1302, opened 2022-04-28
sheband and executable paths incorrect in gitbash


amorphousWaste:
When creating a rez environment with gitbash on Windows, the shebang and executable paths use a single slash which gets escaped and causes the paths to be incorrect.

**Environment**
* OS "windows 10"
* Rez version "2.109.0"
* Rez python version "3.7"

**To Reproduce**
1. Open a gitbash shell on Windows
2. Run a `rez-env`

**Expected behavior**
Taken into a rez environment.

**Actual behavior**
Command not found error

hughetop:
I just noticed the gitbash plugin, but I can't seem to get it to work. No matter what I do I get:
`C:\Users\hughetop\AppData\Local\Temp\rez_context_b1y1cklv\rez-shell.sh: line 5: c:\program: No such file or directory`
The space in "program files" is messing things up, and I can't figure out what the syntax needs to be. Doubling up the backslashes like in the pull request doesn't seem to fix it. Trying to escape the space makes windows think it's another directory. Using the `"/c/program\ files/..."` syntax doesn't work either. This has been wracking my brain for months, but I really want to get gitbash working so I don't have to use cmd or powershell anymore.

amorphousWaste:
Good point about the spaces. I think the solution might be to use DOS short paths:
Eg. `/c/Program Files/Git/bin/bash.exe` -> `/c/PROGRA~1/Git/bin/bash.exe`. I'm doing some testing now.

instinct-vfx:
We had a bit of a discussion about this on Slack already. This needs to be fixed for good as this is only the default but other paths might also need escaping. We need proper cross platform escaping to really fix this in my eyes.

instinct-vfx:
(Adding quotes works in gitbash but not in rez. I think that they are stripped by rez when sanitizing?)

hughetop:
Interesting, I tried using that DOS shortpath and it got a different error. So my `find_executable` is just:
`return r'C:\PROGRA~1\Git\bin\bash.exe'.replace('\\', '\\\\')`
and it'll recognize gitbash, but it doesn't seem to start bash properly. The bash commands aren't available.
```
$ rez-env --shell gitbash python-3.9
bash: mkdir: command not found

You are now in a rez-configured environment.

resolved by hughetop@********, on Fri Apr 29 12:52:12 2022, using Rez v2.103.4

requested packages:
python
~platform==windows           (implicit)
~arch==AMD64                 (implicit)
[...]

PR #1351, opened 2022-07-24
Use forward slashes in PowerShell

anderslanglands:
PowerShell consumes forward slashes in path names just fine. CMake occasionally barfs on backslashes. Therefore this PR just uses forward slashes for pwsh and everything is happy.

### DIFF
--- a/src/rezplugins/shell/_utils/powershell_base.py
+++ b/src/rezplugins/shell/_utils/powershell_base.py
@@ -243,10 +243,8 @@ class PowerShellBase(Shell):
         return result
 
     def normalize_path(self, path):
-        if platform_.name == "windows":
-            return to_windows_path(path)
-        else:
-            return path
+        # CMake requires forward slashes and powershell handles them fine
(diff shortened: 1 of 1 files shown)


ColinKennedy:
Previous PR, for context: https://github.com/AcademySoftwareFoundation/rez/pull/1273

Considering that CMake apparently misbehaves even in Windows with proper, normalized paths, I'd rather normalization be an explicit, opt-in feature so that projects that require this custom behavior can still have it and the default, expected behavior still functions as expected. This PR doesn't give the user any option to opt-in / opt-out of the change. What if something else other than CMake needs backslashes? etc.

nerdvegas:
Hey Anders, 

Please see the discussion in https://github.com/AcademySoftwareFoundation/rez/issues/1302 as I think this really is the root of the issue. Feel free to add to that thread, it needs to be picked up again. But yes this stuff does need to be fixed, I just think we need to do so in a more general way (as this is not a powershell specific thing).

Issue #1205, opened 2022-02-10
remove hardcoded references to nerdvegas gh project where possible


nerdvegas:
I'm unsure as to whether there are places left we can do this, but it's worth a look before migration to aswf gh org

nerdvegas:
done in wiki

nerdvegas:
I think this is done now, the remaining instances only make sense to do after aswf migration:
* issue/pull links in inline comments
* instances in README.md (such as badge links)

maxnbk:
I've resolved every other nerdvegas repo reference, excluding rez-shotgun which is a legacy repo still residing at nerdvegas, and omitting personal author references in changelogs and so forth.
The only thing remaining is sonarlcloud for which a separate issue exists, so I believe this is closeable. please reopen if you believe there is something else, @nerdvegas .

cuckon:
Sorry what happened? Any context?

PR #1368, opened 2022-08-28
Replace nerdvegas with aswf

maxnbk:
This is almost, but not quite closing:
https://github.com/AcademySoftwareFoundation/rez/issues/1205

The remaining items to evaluate or potentially change are:
`src/rezgui/__init__.py:5:organisation_name = "nerdvegas"`
`setup.py:65:    author_email="nerdvegas@gmail.com",`
and sonarcloud project refs, that have a separate ticket I don't have permissions to address.

please note: references to the "author" `nerdvegas`, or the project `nerdvegas/rez-shotgun` have been retained. Anything that currently properly forwards to the ASWF repo has been updated, including changelog entries, as potentially unnecessary as it might be.

Nothing else should need scrubbing, apart from the 3 items mentioned above, potentially.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,18 +51,18 @@
 -  remove references to nerdvegas in comments [\#1312](https://github.com/AcademySoftwareFoundation/rez/pull/1312) ([nerdvegas](https://github.com/nerdvegas))
 
 ## 2.109.0 (2022-04-19)
-[Source](https://github.com/nerdvegas/rez/tree/2.109.0) | [Diff](https://github.com/nerdvegas/rez/compare/2.108.0...2.109.0)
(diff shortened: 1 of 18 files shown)


JeanChristopheMorinPerso:
The author email can be replaced with `rez-discussion@lists.aswf.io` which is the email of the ASWF hosted mailing list. That's what the other ASWF projects do (https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/setup.cfg#L3, https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/setup.py#L198).

As for the organization name, why not ASWF? Or simply drop the organization name.

maxnbk:
updated those two refs

JeanChristopheMorinPerso (approved):
Looks good to me.

PR #1377, opened 2022-09-11
Edit rezconfig docstrings for package_filter to reflect python form rather than YAML

### LINKED ISSUE #1372: Documentation for package_filter config inaccurate
I was looking at the documentation for package_filter, accessible here:
https://github.com/AcademySoftwareFoundation/rez/blob/master/src/rez/rezconfig.py#L365
https://github.com/AcademySoftwareFoundation/rez/wiki/Configuring-Rez#package_filter

The docs make it look like we're in a YAML config rather than a python config.
While the YAML in the docs does translate pretty much 1:1 to the expected python types, it takes a bit of a mental effort to translate between the 2, and at first I got it wrong, trying to enter `glob('*.beta')` instead of the expected `'glob(*.beta)'`

Additionally, it would seem like there is no such thing as `requirement(foo-5+)`, and that the proper string is `range(foo-5+)`?

I can possibly update the docs myself and update the examples to the right format but as I've just pieced it together I might be missing some subtleties .

**Expected behavior**
Documentation examples should be usable right away.

**Actual behavior**
Documentation examples can't be used directly and require translating into python

herronelou:
closes #1372


JeanChristopheMorinPerso (approved):
Thanks for this contribution!

nerdvegas:
Cheers! I'm not going to bother with releases for these doc-only changes - not until we actually have versioned docs anyway. I'll merge this right in to master.

PR #1378, opened 2022-09-15
Add REZ_USED_LOCAL_RESOLVE context environment variable

### LINKED ISSUE #1371: Expose an environment flag if any locally installed packages were resolved in the current context
Looking for an easy way to know if I'm in a "development" environment using an environment variable
The consensus on the slack was to evaluate common environment variables for the rez local install path but that feels like both double handling data the resolver probably already has and has the potential to miss uncommon packages

**Motivation**
Exposing this info as an environment variable makes it easy to know if you're in a local development environment and easy to feed that into logging/monitoring

JoshkVFX:
Populate the environment variable "REZ_USED_LOCAL_RESOLVE" in the same style as REZ_USED_EPH_RESOLVE for any packages resolved to a local destination

Closes #1371 

Let me know if there's a version number I need to bump or something

### DIFF
--- a/src/rez/resolved_context.py
+++ b/src/rez/resolved_context.py
@@ -1959,6 +1959,7 @@ class ResolvedContext(object):
         request_str = ' '.join(str(x) for x in self._package_requests)
         implicit_str = ' '.join(str(x) for x in self.implicit_packages)
         resolve_str = ' '.join(x.qualified_package_name for x in resolved_pkgs)
+        local_packages_str = ' '.join(x.qualified_package_name for x in current_context.resolved_packages if x.is_local)
         req_timestamp_str = str(self.requested_timestamp or 0)
         package_paths_str = executor.interpreter.pathsep.join(
             normalized(x) for x in self.package_paths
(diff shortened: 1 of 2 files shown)


JeanChristopheMorinPerso (approved):
LGTM. That's exactly what I was imagining.

nerdvegas (dismissed):
I'm ok with what's been done here, but the docs need updating here: https://github.com/AcademySoftwareFoundation/rez/wiki/Environment-Variables#context-environment-variables

Take care to note that this new envvar will contain a subset of what's already in REZ_USED_RESOLVE.

You just need to update this file, the wiki will be updated for you: https://github.com/AcademySoftwareFoundation/rez/blob/master/wiki/pages/Environment-Variables.md

JoshkVFX:
@nerdvegas How's that look?

JeanChristopheMorinPerso:
Thanks for documenting it! To me it looks great.

bpabel (approved):
LGTM

maxnbk (approved):

Issue #680, opened 2019-07-27
REP-003: Package and Context Localisation


nerdvegas:
# REP-003: Package and Context Localisation

Rez contexts contain references to the variants in their resolve, in the form of variant “handles”. A handle is just a metadata dict that uniquely identifies a variant and defines where it is located. For example, a variant handle typically contains the package name, package version, path to the package repository containing it, and index of the variant within that package. From its handle, the package definition can be retrieved, which is needed to configure the environment (for example, `commands` needs to be evaluated).

When resolved into an environment, a variant will typically reference its installed payload (its “root”) in some way - for example, by adding a “bin” subdirectory to `PATH`. The installed payload will  reside within the package repository directory.

These are two examples of cases where use of a context requires access to external resources - either the package definitions (which would generally come from memcached in a production Rez install, but may also come from disk); and the package installation payload.

There are two different reasons why this may not be desirable:

* We wish for a context to act in a standalone manner, so it can be run in isolation. For example, we would like a VM or container running a Rez-based micro service to not have to access memcached, nor to require a mount to shared file storage.
* We wish for package payloads to be locally available, rather than having to be fetched over an nfs. This has the potential to significantly improve performance and decrease load on the filer.

## Context Localisation

A “localised” or “deep” context would be one that contains entire copies of the package definitions its variants require. This would cause the following differences in behaviour (when compared with standard contexts):

* The rxt file size would be larger (possibly significantly);
* Changes to package definitions in their original package repositories would not take effect in the deep context;
* The deep context would not require fetching of package definitions from memcached, and thus would source faster.

The following features are desirable, in order to give users as much control as possible, and to maintain backwards compatibility:

* Use of deep contexts would be configurable, but would also be manually overridable in all cases where contexts are created. For example, `rez-env` would have a new `—context-mode` option (with “deep” and “shallow” choices), for use in combination with its `—output` option.
* It should be possible to load a shallow context and convert it to a deep context, and vice versa (probably via the existing `rez-context` tool).

One thing that also needs to be considered is if it’s ever desirable to store deep contexts into memcached. This _would_ result in more optimal loading of cached resolves, but at the cost of far more cache space - each package definition would effectively be stored many times over (once for each deep context they appear in). It may be better not to support this, and to instead wait for a port to redis, which supports multiple key retrieval (via its `mget` operation).

## Package Localisation

[...]

PR #1380, opened 2022-09-17
Adding Artifact Repository

ttrently:
UPDATE: Latest commit implements an artifact repository based on discussions found on https://github.com/AcademySoftwareFoundation/rez/discussions/1393

Add support for `artifact_repository` plugins which allow for hosting of package payloads in remote servers. This works in conjunction with a MongoDB instance that contains the package information for querying and lookup.

## Requirements

This implementation requires some additional packages that have not been included in `rez.vendor`.

- awscli
- pymongo

These can be installed via `python -m pip install PKG --target path\to\rez\Lib\site-packages`.

## How To Use

In order to use you will need an existing S3 bucket and a MongoDB instance with a named database and a collection called "packages".

In your `rezconfig.py` you will need to set the following:

```
packages_path = [
    "mongo@localhost:27017/db_name",
]

artifacts_path = [
    "s3@s3://bucket-1/packages",
]

default_cachable = True
package_cache_same_device = True
package_cache_async  = False
```

If your MongoDB instance also requires a username / password, you can set it via:
```
REZ_MONGO_USERNAME=username
REZ_MONGO_PASSWORD=password
```

When these are set, it will format the packages path url as:
```
{REZ_MONGO_USERNAME}:{REZ_MONGO_PASSWORD}@localhost:27017/db_name
```

## rez-build

The `rez-build` command has been extended to allow for installing to artifact repositories. This is done through the new `-a` `--artifact-path` argument from `rez-build`.

Example:
```
foo> rez-build --install -p mongo@uri:port/db_name -a s3@s3://bucket-1/packages
```

This build command will do the following:
  1. Build the `foo` package normally
  2. Install build results to a temporary location of `$TEMP\rez\package_name\package_version`, this is what will be pushed to the artifact repository.
  3. Add a new variant entry to the specified mongo@ package repository
  4. Push the installed build results to the s3@ artifact repository. 

## Missing / To-Do

- This uses a basic implementation of the artifact repository where they are defined separately from the package repository. On lookup, the artifact repositories are searched in order for the package payload, this seems to work well since the location stored in the database is the `artifact_path` qualifier, but may need to be more complex in the future.
- Artifact repository paths need to be extended to other cli's, such as move, release, and any others that interact with repositories.
- MongoDB authentication is still somewhat of an issue if you use multiple different databases that require different credentials.


JeanChristopheMorinPerso:
This is pretty exciting @ttrently! I haven't reviewed any of it yet, but I took some time to link to some resources.

Long ago, someone tried to create a MongoDB repository plugin and even created a PR, see https://github.com/AcademySoftwareFoundation/rez/pull/205. You could potentially re-use part of it or at maybe just take inspirations from it.

Additionally, there is [REP-003](https://github.com/AcademySoftwareFoundation/rez/issues/680) that tracks package caching. In the REP (**R**ez **E**nhancement **P**roposal) and its comments, there is some wording that makes me believe the package repository could say "I need all my packages to be locally cached". This would allow to mix and match multiple package repository types without forcing all resolved packages to be cached.

JeanChristopheMorinPerso:
@ttrently is your latest commit implementing what you described in https://github.com/AcademySoftwareFoundation/rez/discussions/1393?

ttrently:
> @ttrently is your latest commit implementing what you described in #1393?

Yes, to a degree, and primary comment updated with usage information. I'm still a bit unsure of how to handle complex paths, so for this I kept it close to how normal package repositories are handled, ie. Artifact repository paths are defined and lookups happen based on the order of those paths.

JeanChristopheMorinPerso:
The new description is helpful and thanks for that. Can you update the description to mention that you implement part of that discussion, what's implemented, what's not, etc. I would also suggest that you mention which library it depends, etc.

(Note that I won't have time to review until next week).

PR #1387, opened 2022-09-20
add rez-maintainers as codeowners

maxnbk:
Signed-off-by: Stephen Mackenzie <maxnbk@users.noreply.github.com>

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -5,33 +5,33 @@
 # https://git-scm.com/docs/gitignore#_pattern_format
 
 # Admin
-/ASWF/                  @nerdvegas
-CODEOWNERS              @nerdvegas
-LICENSE                 @nerdvegas
-NOTICE                  @nerdvegas
-*.md                    @nerdvegas
+/ASWF/                  @nerdvegas @AcademySoftwareFoundation/rez-maintainers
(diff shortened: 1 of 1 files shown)


bpabel (approved):

maxnbk:
I'm leaving @nerdvegas on the list separately as a backup to see how marking the group rather than individuals works, in case something is wrong with that (though the docs indicate it should work)

instinct-vfx (approved):

JeanChristopheMorinPerso:
@maxnbk could we simplify the file and just make the whole TSC the owner of all files?

nerdvegas (approved):

maxnbk:
> @maxnbk could we simplify the file and just make the whole TSC the owner of all files?

Maybe? But the separation may become useful later if we decide that not everyone wants to see every kind of PR, or that maybe a specific team is in charge of, say, the solver. I'll keep the update to a group ownership update for now, and check to see if we can collapse it as a separate one, because my main concern here is that each TSC member approves being added to every PR by default, and then moving on from there if there is an objection.

JeanChristopheMorinPerso (approved):

JeanChristopheMorinPerso:
Alright, make sense. I approved and merged since everybody approved.

Thanks @maxnbk !

Issue #1388, opened 2022-09-21
Create issue template for TSC meetings agenda


maxnbk:
For TSC convenience

PR #1389, opened 2022-09-21
add tsc meeting agenda issue template

maxnbk:
Signed-off-by: Stephen Mackenzie <maxnbk@users.noreply.github.com>


maxnbk:
closes #1388

instinct-vfx (approved):
I think it might benefit from a bit more verbose information about the meeting itself, but i also think it is good to just get started this way. If there is something we add every time then it goes to the template.

PR #1391, opened 2022-09-21
Clarify when the commands block is executed during rez-build

### LINKED ISSUE #1198: improve docs wrt commands() not run on built package
This is a point of misunderstanding that keeps coming up, see https://github.com/nerdvegas/rez/issues/1197

Update:
https://github.com/nerdvegas/rez/wiki/Package-Commands
https://github.com/nerdvegas/rez/wiki/Building-Packages#package-communication

jasoncscott:
Signed-off-by: Jason Scott <jason_scott@siggraph.org>
----
Fixes #1198

### DIFF
--- a/wiki/pages/Building-Packages.md
+++ b/wiki/pages/Building-Packages.md
@@ -76,10 +76,10 @@ build environment, but the details of the build itself are left open for the use
 Having said that, *cmake* has been supported by rez for some time, and rez comes with a
 decent amount of utility code to manage cmake builds.
 
-When a rez environment is configured, each package's [commands](Package-Definition-Guide#commands)
-section configures the environment. When a build is occurring, a special variable
-[building](Package-Commands#building) is set to *True*. Your packages should use this
-variable to communicate build information to the package being built.
+When a rez environment is configured, each required package's
+[commands](Package-Definition-Guide#commands) section configures the environment for the building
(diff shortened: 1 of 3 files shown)


wiki/pages/Building-Packages.md
```diff
@@ -88,6 +88,11 @@ For example, our *boost* package's commands might look like so:
             # there is a 'FindBoost.cmake' file in this dir..
             env.CMAKE_MODULE_PATH.append("{root}/cmake")
 
+When *boost* is required into another package's build environment, `building` will be set to
+*True*, thereby executing that block of code, but otherwise, it will not be executed, even when
```
nerdvegas, line 92: I'm finding this a bit confusing because it's conflating two different things - what 'building' is for, and the fact that a package's commands don't get executed if it's the package itself getting built. But, this is still a good place to highlight that info.

Perhaps a separate warning would be good:

```
> [[media/icons/warning.png]] Note that _commands_ is never executed for the package actually getting built - 
> if you want to run commands in that case, you use [pre_build_commands](Package-Commands#pre-build-commands) instead.
```

nerdvegas (dismissed):

JeanChristopheMorinPerso (approved):

maxnbk (approved):

PR #1426, opened 2023-01-03
Provide cross platform Example -WIndows

narru0467:
Signed-off-by: narru0467 <121908327+narru0467@users.noreply.github.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,8 +160,8 @@ administrative privileges):
 
 Now you should be able to create an environment containing Python. Try this:
 
-    ]$ rez-env python -- which python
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -160,8 +160,8 @@ administrative privileges):
 
 Now you should be able to create an environment containing Python. Try this:
 
-    ]$ rez-env python -- which python
-    /home/ajohns/packages/python-2.7.8/platform-linux/arch-x86_64/os-Ubuntu-12.04/bin/python
+    ]$ rez-env python -- python -V
+    Python 2.6.8
```
bpabel, line 164: The purpose of this is to show that the python executable path has now been standardized within the rez package repository.

Could replace with `$ rez-env python -- python -c "import sys;print(sys.executable)"`, which would be cross-platform

PR #1434, opened 2023-01-26
Fix env prepend failing for new env vars

SitiSchu:
This simply applies the fix for #1255 to `prependenv` aswell as that has the same issue


JeanChristopheMorinPerso (approved):
Looks good to me. I triggered the CI.

I'm curious to know if we could have tests for these to avoid future regressions. Would it be difficult?

PR #1461, opened 2023-03-20
Fix ResolvedContext.execute_command with parent_env=None

dbr:
Specific problem I'm having is:
- A script uses `ResolvedContext.get_current()` to run a child process (Nuke)
- In this child process, env vars from the parent environment (such as `$http_proxy`) do not get inherited by the child process

Described behaviour is `os.environ` should be used if `parent_env` is None, but prior to this change, an empty dict was used instead.

In other words, prior to this change the following would fail (`printenv EXAMPLE` returns nothing):

```sh
$ export EXAMPLE=abcdef
$ rez env rez -- python
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110] on linux
>>> import os, rez, rez.resolved_context, subprocess
>>> ctx = rez.resolved_context.ResolvedContext.get_current()
>>> ctx.execute_command(['printenv', 'EXAMPLE'], parent_environ=None, stdout=subprocess.PIPE).communicate()
(b'abcdef\n', None)
```

Whereas this output is now consistent with, say, `execute_shell`:

```sh
>>> ctx.execute_shell('bash', parent_environ=None)

>> $ printenv EXAMPLE
abcdef
```


src/rez/resolved_context.py
```diff
@@ -1296,9 +1296,12 @@ def execute_command(self, args, parent_environ=None, **Popen_args):
         Note:
             This does not alter the current python session.
         """
-        if parent_environ in (None, os.environ):
-            target_environ = {}
```
JeanChristopheMorinPerso, line 1300: This might have been intentional, though the "in os.environ" seems weird to me.

dbr: I suspect it maybe made sense prior to this line changing:

https://github.com/AcademySoftwareFoundation/rez/blame/3d0f22489827f8def994cb2e5590d51445e55123/src/rez/rex.py#L632

BUt this code now just sets target_environ to an empty dict if called with `parent_environ = os.environ` - which can't possibly be desirable behaviour (if caller of `execute_command` wants to do that, they can just pass `parent_environ={}`)

dbr:
Is there anything I can do to help get this merged? We still have a workaround in place which is just to do

```
env = os.environ.copy()
env["WORKAROUND_REZ_THING"] = "abc"

ctx = rez.resolved_context.ResolvedContext.get_current()
p = ctx.execute_command(cmd, parent_environ=env)
```

which would be nice to remove

JeanChristopheMorinPerso:
I have yet to be convinced that this change is good and will not break anything. So for now, I'm not willing to merge this PR yet.

PR #1468, opened 2023-04-06
Fix CI

JeanChristopheMorinPerso:
This PR "fixes" our CI. The Ubuntu 18.04 image is deprecated and it's pretty much gone at this point, which means the Unbuntu workflow would just sit there doing nothing, waiting for a VM that will never exist.

I also added a `workflow_dispatch` trigger on some workflows so that we can manually run them without committing any changes.


maxnbk (approved):
I think we should get this in and then let the CI complete successfully on the python version support PR

PR #1469, opened 2023-04-06
Drop support for installs with Python lower than 3.7

### LINKED ISSUE #1349: remove embedded virtualenv
virtualenv is LGPL, it must be removed from the project.

This means that py<3.7 will now have a system requirement on virtualenv, but that's not so bad - practically all of 
py<3.7 is py2, which is much closer to actual deprecation in vfx now. We will abort with an error if virtualenv cannot
be imported in this case.

Note related ticket #1348 - not sure specifically what's going on here, but let's just fall back to virtualenv if venv cannot be
imported, that seems more robust.

**Related Issues/PRs**
* #1348

https://lfscanning.org/reports/aswf/rez-2022-06-10-b01cab7b-ff86-4d95-9353-3af77d08c868.html

JeanChristopheMorinPerso:
Fixes #1349

This is a radical change, but I think it's safe to do at this point. We already discussed this multiple times and the conclusions has always been that we should drop support for anything lower than 3.7 prod install wise (through the `install.py`).

Note that we will continue to support Python 2.7 for API users.

# TODO
- [x] Documentation
- [ ] Communication


maxnbk:
You mean "for 2.7 prod install" instead of "for 3.7 prod install", yes?

We also discussed dropping support for py3.6, does it make sense to include that in the same motion?

JeanChristopheMorinPerso:
Thanks, I edited the PR description. ALso, the changes already drop support for Python lower than 3.7. I'll also add that to the description.

maxnbk (approved):

instinct-vfx (approved):

setup.py
```diff
@@ -67,7 +67,7 @@ def find_files(pattern, path=None, root="rez"):
     entry_points={
         "console_scripts": get_specifications().values()
     },
-    include_package_data=True,
+    include_package_data=False,
     zip_safe=False,
```
bpabel, line 71: `setup` supports specifying a minimum python version -- https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires

```python
setup(
    python_requires=">=3.7",
    ...
)
```

JeanChristopheMorinPerso: This would prevent users from pip install for using the API. Though we should specify python_requires=">=2.7".

PR #1485, opened 2023-05-07
@include: Only include modules from the current package

### LINKED ISSUE #1483: @include uses file from different package
When using the `@include` decorator, rez will sometimes use the file from a different package. This can cause changes to the include files to break the entire resolve if function signatures change.

**Environment**
* Windows 10, Debian 11, RHEL 8.7
* rez 1.112.0
* Python 3.9.9

**To Reproduce**
_Easy way:_
I've created a repository with everything setup already: https://github.com/SitiSchu/rez_include_bug_report
Running `run.sh` should reproduce the issue. The package sources are in `source`

_Manual way:_
1. Create 3 packages: A, B, C
2. Create dir `include_path_v1`
3. Create a `utils.py` inside it
4. Define a variable: `__version__ = '1'`
5. Create dir `include_path_v2`
6. Create a `utils.py` inside it
7. Define a variable: `__version__ = '2'`
8. `@include('utils')` for packages A, B, C and print `utils.__version__` and `utils.__file__` in their package.py's
9. Build A and C with `include_path_v2` as `package_definition_python_path`
10. Build B with `include_path_v1` as `package_definition_python_path`
11. `rez env A B C`
12. `rez env A C`

**Expected behavior**
Each package should use it's own utils file with the correct version

**Actual behavior**
For `A B C`, Package C will use the utils from Package B 
For `A C`, Package C will use the utils from A

Output from `run.sh`:
```
Resolving 'A B C'
Package A:
        File: [...]/packages/A/1/.rez/include/utils.py
   Should be: [...]/packages/A/1/.rez/include/utils.py
     Version: 2
   Should be: 2

Package B:
        File: [...]/packages/B/1/.rez/include/utils.py
   Should be: [...]/packages/B/1/.rez/include/utils.py
     Version: 1
   Should be: 1

Package C:
        File: [...]/packages/B/1/.rez/include/utils.py
   Should be: [...]/packages/C/1/.rez/include/utils.py
     Version: 1
   Should be: 2

----------------------------------------------

Resolving 'A C'
Package A:
        File: [...]/packages/A/1/.rez/include/utils.py
   Should be: [...]/packages/A/1/.rez/include/utils.py
     Version: 2
   Should be: 2

Package C:
        File: [...]/packages/A/1/.rez/include/utils.py
   Should be: [...]/packages/C/1/.rez/include/utils.py
     Version: 2
   Should be: 2
```

EDIT: The behaviour from `A C` is actually intended as the module is cached based on the hash so the module from A is reused.

SitiSchu:
Fixes #1483

The previously used `SourceFileLoader(name, filepath).load_module()` adds the imported module to `sys.modules`. When a second module is imported with the same name, the previous module gets overwritten. In the case of #1483 this happened:

- Pkg A utils gets imported and added to the module cache with its hash
     - Module gets added to `sys.modules`
- Pkg B utils gets imported and added to the module cache with its hash.
     - This replaces A utils in `sys.modules` which causes the module cache to now have Pkg B's utils for all `utils` hashes
- Pkg C utils' hash is in the module cache from Pkg A, so it gets used from the cache. However, Pkg B utils replaced all `utils` modules so this returns Pkg B's utils.

Using [`importlib.util.module_from_spec`](https://docs.python.org/3.10/library/importlib.html#importlib.util.module_from_spec) avoids this by not adding imported modules to `sys.modules` so nothing gets overwritten even if modules with the same name get imported multiple times.

Since the `@include` functionality works by placing the module in the `globals` for `exec`, not having the module in `sys.modules` should be fine.

### DIFF
--- a/src/rez/utils/py23.py
+++ b/src/rez/utils/py23.py
@@ -41,5 +41,12 @@ def load_module_from_file(name, filepath):
             return imp.load_source(name, filepath, f)
 
     else:
-        from importlib.machinery import SourceFileLoader
-        return SourceFileLoader(name, filepath).load_module()
+        # The below code will import the module _without_ adding it to
+        # sys.modules. We want this otherwise we can't import multiple
+        # versions of the same module
+        # See: https://github.com/AcademySoftwareFoundation/rez/issues/1483
+        import importlib.util
+        spec = importlib.util.spec_from_file_location(name, filepath)
+        module = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(module)
+        return module
--- a/src/rez/utils/py23.py
+++ b/src/rez/utils/py23.py
@@ -8,6 +8,8 @@ Custom py2/3 interoperability code.
 Put any code here that deals with py2/3 interoperability, beyond simple cases
 that use (for eg) the six module.
 """
+import sys
+
 from rez.vendor.six import six
 
 
@@ -38,7 +40,12 @@ def load_module_from_file(name, filepath):
     if six.PY2:
         import imp
         with open(filepath) as f:
-            return imp.load_source(name, filepath, f)
+            module = imp.load_source(name, filepath, f)
+            # Keep the module out of sys.modules. See comment in the `else:`
+            # for more info
+            if name in sys.modules:
+                del sys.modules[name]
+            return module
 
     else:
         # The below code will import the module _without_ adding it to
--- a/src/rez/data/tests/python/include/utils_test_7cd3a335.py
+++ b/src/rez/data/tests/python/include/utils_test_7cd3a335.py
@@ -1,5 +1,0 @@
-# SPDX-License-Identifier: Apache-2.0
-# Copyright Contributors to the Rez Project
-
-
-# Empty test file for the py23.load_module_from_file tests
--- a/src/rez/data/tests/python/include/utils_test_7cd3a335.py
+++ b/src/rez/data/tests/python/include/utils_test_7cd3a335.py
@@ -1,1 +1,5 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Contributors to the Rez Project
+
+
 # Empty test file for the py23.load_module_from_file tests
--- a/src/rez/data/tests/python/include/utils_test_7cd3a335.py
+++ b/src/rez/data/tests/python/include/utils_test_7cd3a335.py
@@ -1,0 +1,1 @@
+# Empty test file for the py23.load_module_from_file tests
--- a/src/rez/tests/test_utils_py23.py
+++ b/src/rez/tests/test_utils_py23.py
@@ -7,21 +7,28 @@ unit tests for 'utils.py23' module
 """
 import os
 import sys
+import tempfile
 
-from rez.tests.util import TestBase
+from rez.tests.util import TestBase, TempdirMixin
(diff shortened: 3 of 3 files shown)


JeanChristopheMorinPerso (commented):
Thanks for the fix @SitiSchu! Do you think you could add a simple unit tests there to make sure this is well covered?

JeanChristopheMorinPerso:
Note to TSC: CLA is correctly signed and setup. I'm not too sure why the EasyCLA bot didn't create a comment. I checked in LFX and I see the signature, so everything is good.

SitiSchu:
@JeanChristopheMorinPerso Sure can do, what would you like tested? I guess making sure it doesn't actually end up in `sys.modules` and maybe ensuring all defined functions and variables are accessible?

JeanChristopheMorinPerso:
I'd test get_module_from_file, which basically means testing that I can import from a given file and that it doesn't end up in sys.modules.

SitiSchu:
I've added a test case for this.

To make sure it works, I ran the test without my fix and it did fail correctly.

SitiSchu:
It looks like the test fails with Python 2.7, should I limit the test to Python 3 only or should I try and fix the issue for py2 too(despite it being deprecated for years now)?

JeanChristopheMorinPerso:
We need to support both python 2 and 3. So if we fix python 3, we also need a fix for py 2.

SitiSchu:
Locally the copyright fix added a empty line to the utils file which made git complain about the extra whitespace. So I added a comment. Now it doesn't complain locally anymore but somehow the CI still fails.

JeanChristopheMorinPerso:
Don't worry about this check for now. We can take care of that.

Thanks for adding a test!

src/rez/utils/py23.py
```diff
@@ -38,8 +40,20 @@ def load_module_from_file(name, filepath):
     if six.PY2:
         import imp
         with open(filepath) as f:
-            return imp.load_source(name, filepath, f)
+            module = imp.load_source(name, filepath, f)
+            # Keep the module out of sys.modules. See comment in the `else:`
+            # for more info
+            if name in sys.modules:
+                del sys.modules[name]
```
JeanChristopheMorinPerso, line 47: Would it be better to copy the content of `sys.modules` before the import and restore it after the import?

SitiSchu: Hmm, not sure. I suppose it would guard against the included module adding stuff to `sys.modules`, but I'm not sure what this might break.

Also, this would change the behaviour from the Py3 versions behaviour. Currently if the imported module imports something (say `textwrap`) then this imported module will show up in `sys.modules` for both Py 2 and 3

JeanChristopheMorinPerso: Oh, I didn't know that it behaved like that on python 3. Then forget what I said.

JeanChristopheMorinPerso:
There you go @SitiSchu. I fixed the copyright issue. I remove the dummy python module and I'm now creating it on the fly in the tests. I wasn't able to find why the change_copyright isn't doing what it's supposed to.

JeanChristopheMorinPerso (approved):
LGTM, thanks a lot for this!

maxnbk (approved):

JeanChristopheMorinPerso:
Thanks you @SitiSchu !

Issue #1133, opened 2021-09-28
-- (double-dash) removes double quotes on Windows


epkaz93:
**Rez Version:** 2.95.0
**Platform:** Windows 10
**Working Rez version:** 2.69.6

Running rez commands with `--` causes double quotes to be removed from arguments after `--`.

Example:
 ```shell
rez-env python -- python -c "print('hello world')"
  File "<string>", line 1
    print('hello
               ^
SyntaxError: EOL while scanning string literal
```

Escaping the double quotes appears to work:
```shell
rez-env python -- python -c \"print('hello world')\"
hello world
```

nerdvegas:
I'm thinking this may be a regression introduced in https://github.com/nerdvegas/rez/releases/tag/2.94.0. Is there any chance you could test with https://github.com/nerdvegas/rez/releases/tag/2.93.3 and see if that works?

epkaz93:
> I'm thinking this may be a regression introduced in https://github.com/nerdvegas/rez/releases/tag/2.94.0. Is there any chance you could test with https://github.com/nerdvegas/rez/releases/tag/2.93.3 and see if that works?

Yeah it works correctly in 2.93.3

Issue #752, opened 2019-09-30
Pipe character issues on Windows.


KelSolaar:
Hi,

I'm testing Rez 2.47.2 and encountering issues with pipe character handling, this time and for a change, a picture:

![image](https://user-images.githubusercontent.com/99779/65846243-a76dc080-e399-11e9-9594-e589f978cf81.png)

This is quite a problematic one, we are passing regex patterns as arguments often and it just breaks now.

I can escape them like this to some degree:

```
C:\Users\thomas>rez-env cmd -- echo "Foo^|Bar"
Foo|Bar
```

But I could not get it to work with more complex invocations unfortunately. Assuming a **my-command.bat** file which does echo its arguments, i.e. `echo %*`, in the current working directory:

```
C:\Users\thomas>rez-env cmd -- my-command --argument-a --argument-b "C:\Users\thomas\My Directory" --argument-c "\bFoo\b|Bar-Plop|John-Plop|Doe-Plop"
'Bar-Plop' is not recognized as an internal or external command,
operable program or batch file.

C:\Users\thomas>rez-env cmd -- my-command --argument-a --argument-b "C:\Users\thomas\My Directory" --argument-c "\bFoo\b^|Bar-Plop^|John-Plop^|Doe-Plop"
'Bar-Plop' is not recognized as an internal or external command,
operable program or batch file.
```

I could change `echo %*` to `echo "%*"` and it would work but if `echo` is instead a Python program which expects arguments it will break because of the double quotes as it will receive the whole line as a single argument instead of the granular arguments.

KelSolaar:
So it seems like we are plagued by yet again another quotes swallowing issue, here is the content of the `rez-shell.bat` file:

```
set REZ_ENV_PROMPT=%REZ_ENV_PROMPT%$G
echo %PATHEXT%|C:\Windows\System32\findstr.exe /i /c:".PY">nul || set PATHEXT=%PATHEXT%;.PY
(call )
call c:\users\thomas\appdata\local\temp\rez_context_twp0oy\context.bat
set PROMPT=%REZ_ENV_PROMPT% $P$G
my-command --argument-a --argument-b "C:\Users\thomas\My Directory" --argument-c \bFoo\b|Bar-Plop|John-Plop|Doe-Plop
exit %errorlevel%
```

Note how the quotes have disappeared around the regex pattern?

[...]

Issue #592, opened 2019-04-04
"&" character in environment manipulation definitions yields error on Windows upon requiring the package.


KelSolaar:
Hi,

This is related to #432: I'm trying to push Houdini a bit more nicely into our pipeline and I have been having a lot of pain. The trigger for my recent stack of issues is actually this current one that I thought was fixed by #432.

At any point if I put a `&` character in a `env.MY_VARIABLE.append` or `env.MY_VARIABLE.prepend` call I get an error upon requiring the package. `&` is used by Houdini [to make its own content available](http://www.sidefx.com/docs/houdini/basics/config_env.html#special-characters-in-path-variables), if you don't do so it simply errors on startup.

Assuming that my `commands` definition is expressed as follows:

```python
def commands():
    import os

    env.HOUDINI_PATH.append('&')
    env.HOUDINI_PATH.append(
        os.path.join('{root}', 'bootstrap', 'houdini'))
```

If I `rez-env my_package`, I get:

```batch
'C:\Users\thomas\Packages\my_package\1.0.0\bootstrap\houdini' is not recognized as an internal or external command,
operable program or batch file.
resolved by thomas@..., on Thu Apr 04 16:44:28 2019, using Rez v2.28.0
```

I tried many variants with hat and everything but not dice and I cannot find any workaround so far.

For reference this stuff works in `cmd`:

```batch
C:\Users\thomas>set HOUDINI_PATH=^&

C:\Users\thomas>set HOUDINI_PATH
HOUDINI_PATH=& 
```

KelSolaar:
Got it to work doing that:

```
env.HOUDINI_PATH.append(
        os.path.join('{root}', 'bootstrap', 'houdini;&'))
```

nerdvegas:
I assume this is a cmd char escape issue on windows, can anyone confirm?

Thx
A

bpabel:
`&` is a control character for cmd.exe.  If you want a literal `&`, you have to escape it with `^`

KelSolaar:
@nerdvegas : Yes it is!

@bpabel : Yes but it does not work as mentioned in first post.

PR #1496, opened 2023-05-29
Draft: Escape special characters in cmd shell join method

herronelou:
Ran into issues when running `rez env my_handler -- handle.py some-protocol://some-tool/?user_id=6560&user_login=eleroy&session_uuid=f49700b4-fa89-11ed-9b1c-0242ac110005&entity_type=Task&selected_ids=236123&ids=236123&server_hostname=server.shotgunstudio.com&title=Task%20236123`, as the rez cmd shell would not properly escape special character when writing the `rez-shell.bat`.

This tweak seems to fix it, but I haven't run all the unit tests, and the easiest way to run them seems to make this PR, so I'm making this draft PR to see if it passes tests.

### DIFF
--- a/src/rez/tests/test_shells.py
+++ b/src/rez/tests/test_shells.py
@@ -229,12 +229,8 @@ class TestShells(TestBase, TempdirMixin):
                                 text=True)
             self.assertEqual(_stdout(p), "Hello Rez World!")
             os.remove(path)
-
-    # TODO fix cmd shell command string escape
-    # as per https://github.com/AcademySoftwareFoundation/rez/pull/1130, then remove this
-    # exclusion
-    #
-    @per_available_shell(exclude=["cmd"])
+            
+    @per_available_shell()
     @install_dependent()
     def test_rez_env_output(self, shell):
 
--- a/src/rezplugins/shell/cmd.py
+++ b/src/rezplugins/shell/cmd.py
@@ -306,14 +306,10 @@ class CMD(Shell):
     def get_all_key_tokens(cls, key):
         return ["%{}%".format(key)]
 
-    @classmethod
-    def join(cls, command):
-        # TODO: This needs to be properly fixed, see other shell impls
-        # at https://github.com/AcademySoftwareFoundation/rez/pull/1130
-        #
+    def join(self, command):
         # TODO: This may disappear in future [1]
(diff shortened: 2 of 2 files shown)


JeanChristopheMorinPerso:
FYI, you don't need to create a PR to run the tests. Just push in you fork and look at the actions tab in your fork. Workflows should run on commits, whicheans they'll run in your fork's context.

herronelou:
> FYI, you don't need to create a PR to run the tests. Just push in you fork and look at the actions tab in your fork. Workflows should run on commits, which means they'll run in your fork's context.

It wasn't running, so this was easier, but I'll try to figure it out, or better yet get setup to run them locally.

herronelou:
related to https://github.com/AcademySoftwareFoundation/rez/issues/592

herronelou:
related to https://github.com/AcademySoftwareFoundation/rez/issues/752

herronelou:
@JeanChristopheMorinPerso You could review it, but I'm afraid to be breaking as many things as I'm fixing with this patch.

I thought maybe the unit tests could be trusted, but the test that fails now is the same test that failed with the old code, so none of the tests were covering the stuff I changed for sure, and looking through the tests it doesn't look like very much was covered, so my trust in this patch is low. I'm combing through other issues to see what other similar cases have been found in the past

herronelou:
Another case that failed previously which I never reported, but when running `rc.execute_shell(command=['rv', '/path/to/sequence.%04d.exr'])` it fails badly as `%0` gets expanded, breaking the command. This patch seems to fix that behavior.

JeanChristopheMorinPerso:
Thanks for taking a deeper look into this! You are welcome to do the inventory of what's broken, propose a fix or give us your thought on what could be done to improve the situation. That would greatly help us. Though, the cmd plugin is not an area of focus and we usually suggest to move to PowerShell.

herronelou:
also related to this:
https://github.com/AcademySoftwareFoundation/rez/issues/1133
Looks like the original patch that @nerdvegas rolled back there was somewhat similar to my patch here, so the use case in that issue should be tested.

@JeanChristopheMorinPerso yes I'm also looking at moving us to powershell, though I'm still fighting executionpolicy issues with pwsh, so I'm kind of going back and forth between stopping the usage of cmd and trying to improve it.

PR #1505, opened 2023-07-04
Add Execution Policy setting to Powershell based shell plugins

### LINKED ISSUE #1504: Allow to configure execution policy for Powershell/pwsh plugins
When using the powershell/pwsh shell plugins on Windows, Rez uses the execution policy set at the system level.
If the policy is set to Restricted or AllSigned, none of the .ps1 scripts generated by rez on the fly can be executed, because they are not signed, causing rez to not be functional.
If the policy is set to RemoteSigned, Rez is not able to run remote scripts on shared network.
Sidebar: In the case of RemoteSign, that doesn't really hamper the usage of Rez, but in our specific situation, we have been unable to set up script signing in a way that functions across multiple facilities.

Powershell allows any user to change the execution policy at the session level, and [Microsoft](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies?view=powershell-7.3#long-description) says:

> The execution policy isn't a security system that restricts user actions. For example, users can easily bypass a policy by typing the script contents at the command line when they cannot run a script. Instead, the execution policy helps users to set basic rules and prevents them from violating them unintentionally.

The shell plugins for PowerShell already feed certain PowerShell arguments to the session, such as -NoLogo (https://github.com/AcademySoftwareFoundation/rez/blob/master/src/rezplugins/shell/_utils/powershell_base.py#L203), I propose that we include an option to set `-ExecutionPolicy` there as well, based on a setting that would be exposed somewhere, possibly in https://github.com/AcademySoftwareFoundation/rez/blob/master/src/rezplugins/shell/rezconfig#L31. 

**Motivation**
Being able to set a policy at the rez level would allow more granular control over what can be done in a rez environment while still keeping global system policies for anything else executing outside of Rez.

herronelou:
Closes #1504 

This is a simple implementation of what I was requesting in #1504

I considered a different approach which would have been a bit more generic such as `shell_args`, which could have included the `-NoLogo` which is currently being inserted all the time. However, after reviewing the other shell plugins, I couldn't see the case for shell args for all of them, so I kept it localized to only doing exactly what I needed.

### DIFF
--- a/src/rezplugins/shell/_utils/powershell_base.py
+++ b/src/rezplugins/shell/_utils/powershell_base.py
@@ -202,6 +202,12 @@ class PowerShellBase(Shell):
         # Suppresses copyright message of PowerShell and pwsh
         cmd += ["-NoLogo"]
 
+        # Powershell execution policy overrides
+        # Prevent injections/mistakes by ensuring policy value only contains letters.
+        execution_policy = self.settings.execution_policy
+        if execution_policy and execution_policy.isalpha():
+            cmd += ["-ExecutionPolicy", execution_policy]
+
         # Generic form of sourcing that works in powershell and pwsh
         cmd += ["-File", target_file]
 
--- a/src/rezplugins/shell/rezconfig
+++ b/src/rezplugins/shell/rezconfig
@@ -27,11 +27,13 @@ powershell:
     prompt: '> $ '
     additional_pathext: ['.PY']
     executable_fullpath: null
+    execution_policy: null
 
 pwsh:
     prompt: '> $ '
     additional_pathext: ['.PY']
     executable_fullpath: null
+    execution_policy: null
 
 gitbash:
     prompt: '>'


maxnbk:
I rebased to get the CI fixed, but the PR needs DCO fixed (commit signoff)

src/rezplugins/shell/_utils/powershell_base.py
```diff
@@ -202,6 +202,11 @@ def _record_shell(ex, files, bind_rez=True, print_msg=False):
         # Suppresses copyright message of PowerShell and pwsh
         cmd += ["-NoLogo"]
 
+        # Powershell execution policy overrides
+        execution_policy = self.settings.execution_policy
+        if execution_policy:
```
JeanChristopheMorinPerso, line 207: I am tempted to say that we should sanitize this a little bit. Maybe use a simple regex like `re.match(r'[a-zA-Z]+', execution_policy)` to make sure nobody injects a `;` or whatever weird thing that could lead to other even more weird to debug errors.

JeanChristopheMorinPerso (changes requested):

maxnbk:
I believe there is only like 5 executionpolicy types that are valid, can we
just list only those as valid and check for them / error if not?

herronelou:
Yes that would be pretty easy to do and prevent injections

JeanChristopheMorinPerso:
We could add strixt values, but why not just use a Regen and save us the trouble of having to follow windows/powershell releases?

herronelou:
I'm fine with either.
The seven PowerShell execution policies are: Default, Restricted, RemoteSigned, AllSigned, Unrestricted, Bypass, and Undefined.
@maxnbk @JeanChristopheMorinPerso you tell me which you prefer and I'll go with it. It doesn't look like policies are updated very frequently, the oldest documentation I can find (powershell 3.0) has the same policies.

JeanChristopheMorinPerso:
@maxnbk you forgot to add `-s` when you rebased.

JeanChristopheMorinPerso:
Hi @herronelou , do you think you could follow the instructions at https://github.com/AcademySoftwareFoundation/rez/pull/1505/checks?check_run_id=16687974501 please? I'd like to merge this but we need commits to be signed off. Thanks!

JeanChristopheMorinPerso (approved):
Thanks! Hopefully that will be enough for when we enable power shell by default in 3.0.

herronelou:
@JeanChristopheMorinPerso can you confirm if I fixed it fine?

JeanChristopheMorinPerso:
Yep, it looks fine. The DCO bot says that everything is properly signed off.

JeanChristopheMorinPerso:
Thanks @herronelou, this will be released in 3.0.

PR #1506, opened 2023-07-07
Colorize: Granularly wrap each given stream with Colorama

### LINKED ISSUE #1245: Colorama on Windows
Hello.
I've been playing around with trying to enable colorama on Windows.

in #819, @instinct-vfx has exposed the config option to enable colors on Windows, although both cmd and powershell seemed unhappy with the result.

Looking into it a little deeper, I noticed that in https://github.com/nerdvegas/rez/blob/master/src/rez/utils/colorize.py#L318, the buffer is stored. This happens before the colorama init(). The init wraps sys.stdout and stderr to have the correct values for windows, however the self.buf still points at the original buffers. 

By doing a dirty patch in that file (not happy with current solution: I check if the passed buf arg is stdout or stderr, and if they are I init colorama, then assign it to self.buf) I was able to get the colors working.

While the colors now work, as soon as the resolved rez terminal shows up (with the >), all the colors disappear. That means I see the colors for a second then they're gone again.

Any idea what might be causing this?

herronelou:
Fixes #1245

One issue that became apparent while attempting to enable colors in Windows (powershell) was that not every stream was properly initialized. While some commands behaved nicely, others didn't escape properly:

image

After this patch, the same command seems to colorize successfully.

image

The issue stemmed from 2 places:
1 - The ColorizedStreamHandler used for logging was never wrapped by colorama
2 - Buffers passed to Printer would only wrap if stout or stderr, and the first created Printer would not wrap because it would store self.buf before colorama init (which patches stdout and stderr).

This change never calls colorama init, so stderr and stdout are left untouched, instead each buffer is wrapped individually as part of the Printer instance, and only if colors are enabled.

### DIFF
--- a/src/rez/utils/colorize.py
+++ b/src/rez/utils/colorize.py
@@ -9,6 +9,13 @@ import logging
 from rez.vendor import colorama
 from rez.config import config
 
+# Important - we don't want to init Colorama at startup,
+# because colorama prints a RESET_ALL character at exit. This in turn adds
+# unexpected output when capturing the output of a command run in a
+# ResolvedContext, for example.
+# While we're not initializing colorama at all anymore, this comment is left
+# in case someone thought about putting it back: Make sure to do it lazily.
+
 
 def colorama_wrap(stream):
     """ Wrap the stream with colorama so that it can display colors on any OS """
--- a/src/rez/utils/colorize.py
+++ b/src/rez/utils/colorize.py
@@ -10,14 +10,9 @@ from rez.vendor import colorama
 from rez.config import config
 
 
-_initialised = False
-
-
-def _init_colorama():
-    global _initialised
-    if not _initialised:
-        colorama.init()
-        _initialised = True
+def colorama_wrap(stream):
+    """ Wrap the stream with colorama so that it can display colors on any OS """
+    return colorama.initialise.wrap_stream(stream, convert=None, strip=None, autoreset=False, wrap=True)
 
 
 def stream_is_tty(stream):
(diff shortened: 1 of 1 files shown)


src/rez/utils/colorize.py
```diff
@@ -222,12 +217,6 @@ def _color(str_, fore_color=None, back_color=None, styles=None):
     if not config.get("color_enabled", False):
         return str_
 
-    # lazily init colorama. This is important - we don't want to init at startup,
```
JeanChristopheMorinPerso, line 225: I would keep this comment, or at least the part about not calling `init`.

JeanChristopheMorinPerso (commented):
I'm no expert into this. I left a small comment though.

Also, it would be nice if we eventually moved away from colorama and instead started to use https://rich.readthedocs.io/en/stable/introduction.html. But that's for another time. Maybe when we revisit the logger and the logs in general.

src/rez/utils/colorize.py
herronelou: I've put back the comment as a warning in the module. Let me know if that's better.

JeanChristopheMorinPerso (approved):
I don't have a windows machine to test on, but I tested on Linux and it doesn't break anything. Maybe @maxnbk can do a quick test before merging and releasing this.

src/rez/utils/colorize.py
JeanChristopheMorinPerso: Thanks!

maxnbk:
Tested on Windows and seems to work from everything I can tell. approving and including in next release.

maxnbk (approved):

Issue #665, opened 2019-07-14
Proposasl for some rez-test enhancements


JeanChristopheMorinPerso:
Hey everybody, here are some ideas regarding `rez-test`. At RodeoFX we are heavily relying on it and we found that it's lacking some features. There is multiple proposal in this issue because I didn't want to pollute the issue tracker with too much noise. I basically want to get some input and feedback on my proposals and see if there is any traction. We can split into different issues later if necessary.

# Default test/target

Right now, when no test name is given to `rez-test`, it will run all tests. I propose to add a new built-in key in the `tests` dict inside a package definition. The key could be called `default` or `default_test` and it could contain a string (the name of a test or `all`) or a list (a list of test). The reason behind this is that sometimes you might have super slow tests (e.g complex integration tests) that you don't want to run by default, but you still want to run the other tests without having to run them one by one or specify them in the command line.

I have to note that this is a little dangerous to add, since someone could already be using the name we will choose. The name of the key would have to be carefully chosen. Or we might even need to introduce a new package attribute to make sure we don't clash with what users of rez already have.

# Specify multiple tests to run in the command line

I basically feel the need to be able to run multiple tests in one shot without running them all.

For example, we have `package` that has 3 tests: `test1`, `test2`, `test3`, I want to be able to run `rez-test package test1 test2` or `rez-test package all`

# Pass custom arguments when running a test

My colleagues and myself always have to modify the commands of the tests while developing a package. There is multiple reasons why we do so, but mainly we like to increase the verbosity of our tests only when developing and keep the verbosity lower in our CI tests. The other common reason is to choose a custom target inside a particular test so we can iterate faster and avoid running unnecessary tests.

For example, let's say we have:

```python
tests = {
    'unit': {
        'command': 'pytest {root}/tests/unit_tests'
        'requires': 'pytest'
    }
}
```

where `tests/unit_tests` contains multiple test files and each file contains tests with markers (pytest jargon. A marker is basically a way to group the tests so that you can selectively choose which one to run in the command line). I'd be great if we could do `rez-test package unit -- -v -m myMarker` that would run `pytest {root}/tests/unit_tests -v -m myMarker`.

The immediate benefit is that it's faster and more intuitive. The second is that it can avoid mistakenly committing the modifications to the `tests` variable.

[...]

PR #1523, opened 2023-08-27
Allow user to pass extra arguments to underlying rez test command (REP-001 part 3).

bhawkyard1:
This addresses point 3 of https://github.com/AcademySoftwareFoundation/rez/issues/665, as discussed on the aswf rez slack channel recently, allowing users to pass arguments from invocation of `rez test` down to the underlying test command that gets run.

### DIFF
--- a/src/rez/package_test.py
+++ b/src/rez/package_test.py
@@ -232,7 +232,7 @@ class PackageTestRunner(object):
         """
         return self.test_results.num_skipped
 
-    def run_test(self, test_name):
+    def run_test(self, test_name, extra_test_args=None):
         """Run a test.
 
         Runs the test in its correct environment. Note that if tests share the
@@ -240,12 +240,16 @@ class PackageTestRunner(object):
 
         Args:
             test_name (str): Name of test to run.
+            extra_test_args (list of str): Any extra arguments that we want to
+                pass to the test command.
 
         Returns:
             int: Exit code of first failed test, or 0 if none failed. If the first
                 test to fail did so because it was not able to run (eg its
                 environment could not be configured), -1 is returned.
         """
+        if extra_test_args is None:
+            extra_test_args = []
         package = self.get_package()
         exitcode = 0
 
@@ -393,6 +397,12 @@ class PackageTestRunner(object):
                 command = variant.format(command)
             else:
                 command = map(variant.format, command)
+
+            if extra_test_args:
+                if isinstance(command, basestring):
+                    command = "{} {}".format(command, " ".join(map(quote, extra_test_args)))
+                else:
+                    command = list(map(quote, command)) + list(map(quote, extra_test_args))
 
             # run the test in the context
             if self.verbose:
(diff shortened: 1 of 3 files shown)


src/rez/package_test.py
```diff
@@ -394,6 +398,9 @@ def run_test(self, test_name):
             else:
                 command = map(variant.format, command)
 
+            if extra_test_args:
+                command = "{} {}".format(command, " ".join(extra_test_args))
```
JeanChristopheMorinPerso, line 402: At this point in the code, `command` can be a string or a list of string. Additionally, arguments need to be properly quoted/escaped if `command` is a list of string. See https://github.com/AcademySoftwareFoundation/rez/pull/1523/files#diff-cb9e312c8b2786e57bdb95e7da7efe10407758af514dffb3ae1805f0fc996797R410-R413.

JeanChristopheMorinPerso (changes requested):
Thanks @bhawkyard1 for creating this PR! It'll be really useful to a lot of our users.

I left a comment that will need to be addressed before we can approve this PR.

src/rez/package_test.py
bhawkyard1: Ah, I should've spotted that, thank you! Will take a look at this.

bhawkyard1: Hey @JeanChristopheMorinPerso, sorry just trying to orient myself here; we would expect `command` to be a list of strings in the code here here if the `command` entry in the `tests` package definition is a list, E.G:
```
name = "test_package"
version = "1.0.0"
build_command = "cp -r {root}/tests {install_path}/tests"
tests={
    "unit": {
        "command": ["python3", "-m", "pytest", "{root}/tests"],  # <-- command is a list
        "requires": ["pytest"],
        "run_on": ["default"]
    }
}
```
have I got that right?

JeanChristopheMorinPerso: I think so yes.

JeanChristopheMorinPerso:
I rebased your branch to fix the CI.

bhawkyard1:
> I rebased your branch to fix the CI.

Awesome, thank you!

src/rez/package_test.py
bhawkyard1: I think I might have found a little bug with the behaviour of rez test when the command entry is a list but I think that it's probably out of scope here, I'll make an issue separately.

bhawkyard1:
Hey there, I can see the readthedocs ci step failed, I am not sure I understand how to fix this though?

JeanChristopheMorinPerso:
Hi @bhawkyard1 , it's fine. We'll handle that failure for you. You don't need to do anything.

I kicked the actions and everything seems to be passing.

maxnbk:
Not sure what happened, I found the failed build in the RTD control panel and retried it, it passed, the rest is running now.

JeanChristopheMorinPerso (approved):
Thanks @bhawkyard1! I'm sure this will be a very welcome change to all our users!

bhawkyard1:
> Thanks @bhawkyard1! I'm sure this will be a very welcome change to all our users!

Awesome, thanks for the feedback and help!

PR #1536, opened 2023-09-17
WIP Prod install with wheel

JeanChristopheMorinPerso:
This PR is very WIP.

It brings fully workable rez installations using pip. This takes some of the ideas in https://github.com/AcademySoftwareFoundation/rez/pull/1039 and improves them to make it work.

Our `install.py` does mainly two things to guarantee that rez is production ready:
1. It copies all the commands from the virtualenv's "bin" folder into the "bin/rez" folder. This is necessary because we want to be able to add the commands to `$PATH` but we don't want the `python` executable from the virtualenv to be on PATH.
2. All scripts (commands) have their shebang modified from `#!/path/to/rez/venv/bin/python` to `#!/path/to/rez/venv/bin/python -E` (we add `-E`) and we generate custom `.exe`s on Windows to basically do the same thing. `-E` is important for us because we don't want PYTHONPATH to affect the rez commands.

What this PR does is replace the entry-points with scripts and store these scripts inside `.dist-info.data/scripts/rez` in the wheel. When pip installs the wheel, it will unpack everything in `.dist-info.data/scripts` into the `bin` folder and it will respect the hierarchy defined in the wheel! Pip will also replace the interpreter path found in the script's shebang with the venv's one. Again this is important because we want rez to use its own Python). So that's one problem solved (problem number one).

Note that we inject the `-E` flag in the scripts directly in the wheel.

But what about Windows? Everything in the data folder in a wheel is simply unpacked without much processing (except the shebang). On Windows, we need executables to be able to just run "rez" or "rez-env", etc. Pip usually takes care of that with entry points, but it doesn't do so for scripts, by design.

So on windows, we now provide our own launchers. We use the same code that pip uses. The launcher (the EXEs basically) comes from https://bitbucket.org/vinay.sajip/simple_launcher/src/master/ and is shipped with https://github.com/pypa/distlib. Distlib is vendored in Pip and pip usues distlib's pre-compiled EXEs when installing packages to create the EXEs for entry-points.

Note that we don't use the exact same launcher. We use a small variant. **TODO: explain the differences.**

We put the launchers in the wheel too. And this solves the last remaining problem we had.

TODO: Add more details.
TODO: Probably embed the launchers directly in the rez repo in a pre-compiled form...
TODO: Test that `-E` is really being used (I don't think it is right now due to a small bug in my stuff...).
TODO: Write integration tests with real "python" rez packages and test nested environments to call rez and ensure that rez always uses its own Python and not the packaged pythons.


setup.py
```diff
@@ -51,6 +56,62 @@ def find_files(pattern, path=None, root="rez"):
...
+import sys
+from rez.cli._entry_points import {0}
+if __name__ == '__main__':
+    sys.argv[0] = re.sub(r'(-script\\.pyw|\\.exe)?$', '', sys.argv[0])
+    sys.exit({0}())
+"""
+
+class build_scripts(distutils.command.build_scripts.build_scripts):
+    def finalize_options(self):
+        super().finalize_options()
```
JeanChristopheMorinPerso, line 71:
```suggestion
        super(build_scripts, self).finalize_options()
```
Python 2 😠

setup.py
```diff
@@ -51,6 +56,62 @@ def find_files(pattern, path=None, root="rez"):
...
+                )
+
+        prod_install_path = os.path.join(tmpdir, ".rez_production_install")
+        with open(prod_install_path, "w") as fd:
+            fd.write("# Production install installed with pip")
+
+        scripts.append(prod_install_path)
+
+        self.scripts = scripts
+        return super().run()
```
JeanChristopheMorinPerso, line 112:
```suggestion
        return super(build_scripts, self).run()
```

maxnbk:
@JeanChristopheMorinPerso , might this sort of workflow open the door for us to be able to have rez and rez plugins exist slightly closer together in a standard way? (Thought provoked by TSC meeting)

JeanChristopheMorinPerso:
@maxnbk I'm not sure how to answer this. Can you clarify what you mean by "open the door for us to be able to have rez and rez plugins exist slightly closer together in a standard way" please?

JeanChristopheMorinPerso:
I think I finally got it working with `-E`. It took some more thinking outside the box, but it now works. See https://github.com/AcademySoftwareFoundation/rez/actions/runs/6278210700/job/17051519463?pr=1536#step:7:74 which proves it. This line comes from a custom command I added to rez, `jctest` that basically prints `sys.flags`. In our case, we are interested in [sys.flags.ignore_environment](https://docs.python.org/3/library/sys.html#sys.flags.ignore_environment) which basically proves that `-E` was passed correctly.

I also adapted the PR so that it works on Linux and macOS too. For that, I had to modify the scripts we generate on unix to
```
#!/usr/bin/python
# -*- coding: utf-8 -*-
import os
import re
import sys
import platform
# If -E is not passed, then inject it and re-execute ourselves.
# Note that this is not done on Windows because the Windows launcher
# already does this.
if not sys.flags.ignore_environment and platform.system() != 'Windows':
    args = sys.orig_argv[:]
    args[0] = sys.executable
    args.insert(1, '-E')
    if os.getenv('REZ_LAUNCHER_DEBUG'):
        print('Launching:', ' '.join(args))
    os.execvp(sys.executable, args)
from rez.cli._entry_points import {0}
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\\.pyw|\\.exe)?$', '', sys.argv[0])
    sys.exit({0}())
```

This is needed because pip doesn't keep the arguments in the shebang. On Windows, I simply inject `-E` from the launcher executable. I could have use the same method on unix and Windows, but I know launching processes on Windows is costly and it would have resulted in 3 processes started instead of 2.

Next step, clean up, write a proper workflow to generate the wheels on all platforms, figure out where to store the launcher and write at least one integration test that would do this:

1. Install rez with Python version X (say 3.7).
2. Create a Python 3.11 rez package (We can take inspiration from [what I did in rez-pip]( https://github.com/JeanChristopheMorinPerso/rez-pip/blob/main/tests/conftest.py#L204)).
3. Run `rez-env python -- jctest` (We should rename it to something like `__rez_install_test` and make it output a JSON, or something like that).
4. Check that the output of step 3 is `<rez virtualenv root>/bin/python`.

I think that this is enough to prove that our install is production ready and works as intended.

Eventually, we could review the flags used. [-E](https://docs.python.org/3/using/cmdline.html#cmdoption-E) gives us some guarantees, but we can provide even more guarantees by using other flags, like [-I](https://docs.python.org/3/using/cmdline.html#cmdoption-I) (which is is a combination of `-E -P -s` or we could make it more explicit with `-E -s`. [-P](https://docs.python.org/3/using/cmdline.html#cmdoption-P) is too new (but would be nice to have... It's only in 3.11). I think we could provide the same flags across all python versions we support).

JeanChristopheMorinPerso:
Umm, it seems like it doesn't work when using `pip install --target`. But I don't think it's necessarily something we want to support...

JeanChristopheMorinPerso:
I made good progress in the last 3 days. The launchers are now compiled from setup.py  (only on Windows). I'm using the Zig compiler to compile the launcher, which means that MSVC isn't used and needed. There is now wheels for Windows x64 and Windows arm64. Other platforms are covered by a pure+universal wheel. I also added an integration test that uses a real rezified python package.

So far everything seems to be working great, except one test on Windows. It looks like `rez-gui.exe` isn't working. See https://github.com/AcademySoftwareFoundation/rez/actions/runs/6291958825/job/17080881978. It's probably something simple to fix.

There is one major limitation though, and it's that it doesn't support Python 2.

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["setuptools", "wheel", "ziglang", "packaging"]
```
JeanChristopheMorinPerso, line 2:
```suggestion
requires = ["setuptools", "wheel", "ziglang==0.11.0", "packaging"]
```

JeanChristopheMorinPerso:
We should maybe try to support relative oath in the python stubs shebang to support cases like https://academysoftwarefdn.slack.com/archives/C0321B828FM/p1695737655931909. Pip would still produce absolute paths in the installed stubs, but at least users could modify them with a post-install script of their own to whip a more portable rez install (even if we don't officially support it).

I think this functionality is already supported by simple_launcher, but I'd have to double check to make sure it is really the case.

install.py
```diff
@@ -162,23 +162,20 @@ def install(dest_dir, print_welcome=False, editable=False):
...
-    validation_file = os.path.join(dest_bin_dir, ".rez_production_install")
-    with open(validation_file, 'w') as f:
-        f.write(_rez_version)
 
     # done
     if print_welcome:
         print()
         print("SUCCESS!")
-        rez_exe = os.path.realpath(os.path.join(dest_bin_dir, "rez"))
+        rez_exe = os.path.realpath(os.path.join(dest_bin_dir))
```
JeanChristopheMorinPerso, line 178: Should I keep "rez"`?

src/rez/cli/_entry_points.py
```diff
@@ -295,22 +267,35 @@ def run_rez_benchmark():
...
-@scriptname("rez-rm")
+@register("rez-rm")
 def run_rez_rm():
-    check_production_install()
     from rez.cli._main import run
     return run("rm")
+
+
+@register("_rez-install-test")
+def run_rez_install_test():
```
JeanChristopheMorinPerso, line 289: This command is part of the integration tests. It allows us to introspect how rez was launched. It'll print a json with the path to the python executable, the arguments from `argv` and the interpreter flags (sys.flags).

install.py
```diff
@@ -162,17 +162,14 @@ def install(dest_dir, print_welcome=False, editable=False):
     install_rez_from_source(dest_dir, editable=editable)
 
     # patch the rez binaries
-    patch_rez_binaries(dest_dir)
+    # patch_rez_binaries(dest_dir)
```
JeanChristopheMorinPerso, line 165: Should we feature flag this?

src/rez/system.py
```diff
@@ -223,25 +223,18 @@ def rez_bin_path(self):
...
-
-        validation_file = os.path.join(binpath, ".rez_production_install")
-        if os.path.exists(validation_file):
+        if os.path.exists(binpath):
             return os.path.realpath(binpath)
 
         return None
 
-    @property
-    def is_production_rez_install(self):
```
JeanChristopheMorinPerso, line 241: From @maxnbk, keep around and return True with a deprecation warning.

src/rez/shells.py
```diff
@@ -398,8 +398,10 @@ def _record_shell(ex, files, bind_rez=True, print_msg=False):
                 ex.info('')
                 ex.info('You are now in a rez-configured environment.')
                 ex.info('')
-                if system.is_production_rez_install:
-                    ex.command('rezolve context')
+                # Call python with -E directly to avoid the double cost of executing
+                # rezolve when the CLI is installed via pip.
+                rezolve = os.path.join(system.rez_bin_path, 'rezolve')
+                ex.command(f'{sys.executable} -E {rezolve} context')
```
JeanChristopheMorinPerso, line 404: I'm kind of making the assumption that this is only run when using `rez-env`... Can this code ever run when solely using the API?

Issue #1541, opened 2023-09-27
Have a new property to disable or enable canonical_path inside the rezconfig.py


OlivierArgentieri:
Hi

From [this post](https://github.com/AcademySoftwareFoundation/rez/discussions/1540)

The main goal is to have a new property in the rez config to disable or enable usage of canonical path.

I'm creating this issue because i have locally implemented this features, and i'm currently merged this stuff in a new PR.

**Motivation**
Because on windows, we don't want don't necessarily use canonical path

**Related Issues/PRs**
* [Post on discussion](https://github.com/AcademySoftwareFoundation/rez/discussions/1540)

PR #1542, opened 2023-09-27
Add new use_canonical_path option in rezconfig.py

OlivierArgentieri:
## Context:
It's because on windows, we can map some drive letter, to UNC network path.
To centralize packages but also software binaries.

## Problem:
Some softwares like substance, vscode or RV, cannot be run on UNC path and will crash or freeze.

## Solution:
Map a new drive letter on windows, using `new-smbmapping` for example, to avoid usage of unc path in PATH environment variable.
But, rez resolution plugin, use os.path.realpath directly in `__init__` because of that, the path is convert back from our mapped drive letter to full unc path, so some binaries cannot be launched anymore.

Another solution, can be the usage of rez cache, but i found this solution not efficient rather than just execution of binary from network.

## Final details:
It's why, i have created a new parameter in the rezconfig, to have an option to bypass the convertion of canonical path on resolution process, in case of mapped letter drive in packages_paths.

Obviously, by default i have set it to True, to not alterate the current behavior of rez resolution plugin.

## Test:
- I have locally reinstall rez with theses modifications, then i have set the property `use_canonical_path` to False in the rezconfig.py to test and here the result i got:
```
PS C:\Users\oargentieri> rez env vscode-1.82

resolved by oargentieri@<hostname> on Thu Sep 28 10:50:16 2023, using Rez v2.113.0

requested packages:
vscode-1.82
~platform==windows  (implicit)
~arch==AMD64        (implicit)
~os==windows-10.0   (implicit)

resolved packages:
platform-windows  X:\rez_package\platform\windows
vscode-1.82.2     X:\rez_package\vscode\1.82.2\platform-windows
```

And now vscode can be launched using `rez env vscode -- code --disable-chromium-sandbox`

## Issue:
- #1541 

Also linked to [this discussion](https://github.com/AcademySoftwareFoundation/rez/discussions/1540)

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -79,12 +79,6 @@ jobs:
       max-parallel: 1
 
     steps:
-      - name: Setup python ${{ matrix.python-version }}
-        uses: ./.github/actions/setup-python
-        with:
-          python-version: ${{ matrix.python-version }}
-          os: ubuntu
-
       - uses: actions/download-artifact@v3
         with:
           name: "benchmark-result-${{ matrix.python-version }}"
@@ -101,6 +95,12 @@ jobs:
           # See https://github.community/t/how-to-push-to-protected-branches-in-a-github-action/16101/14
           #
           token: "${{ secrets.GH_ACTION_TOKEN }}"
+
+      - name: Setup python ${{ matrix.python-version }}
+        uses: ./.github/actions/setup-python
+        with:
+          python-version: ${{ matrix.python-version }}
+          os: ubuntu-latest
 
       # Note failing due to
       # https://github.com/actions/virtual-environments/issues/675


OlivierArgentieri:
Theses lines:
image

has been added from [this PR](https://github.com/AcademySoftwareFoundation/rez/pull/775/files)

JeanChristopheMorinPerso:
@OlivierArgentieri Can you redo a new branch please? I'm guessing you rebased or something and had to fix conflits or something like that. It would be preferable to have a clean PR with only your changes in it.

OlivierArgentieri:
> @OlivierArgentieri Can you redo a new branch please? I'm guessing you rebased or something and had to fix conflits or something like that. It would be preferable to have a clean PR with only your changes in it.

yes i can

Issue #1552, opened 2023-10-21
Remove things that are officially marked as deprecated


JeanChristopheMorinPerso:
A couple of years ago, some modules were renamed and the old modules were kept there, but marked as deprecated. Python raises a DeprecationWarning when they are imported.

We also have some CLI flags and other thing here and there that raise deprecation warnings.

We should remove them in the next major version.

Modules:
* `rez.packages_`
* `rez.package_resources_`
* `rez.package_maker__`
* `rez.build_process_`

Other:
* https://github.com/AcademySoftwareFoundation/rez/blob/ab317e644f7f38f96528579c77b93454b5d77688/src/rez/cli/pip.py#L62
* https://github.com/AcademySoftwareFoundation/rez/blob/ab317e644f7f38f96528579c77b93454b5d77688/src/rez/build_process.py#L82
* https://github.com/AcademySoftwareFoundation/rez/blob/ab317e644f7f38f96528579c77b93454b5d77688/src/rez/cli/search.py#L62
* `bez`: https://github.com/AcademySoftwareFoundation/rez/blob/ab317e644f7f38f96528579c77b93454b5d77688/src/rez/build_system.py#L81
* https://github.com/AcademySoftwareFoundation/rez/blob/ab317e644f7f38f96528579c77b93454b5d77688/src/rez/rex.py#L1422

* Ability to load old package YAML definitions. (see calls to https://github.com/AcademySoftwareFoundation/rez/blob/371644d19de24d720cc0a4b14f6c17eeb61c55e1/src/rez/serialise.py#L109).
* https://github.com/AcademySoftwareFoundation/rez/blob/371644d19de24d720cc0a4b14f6c17eeb61c55e1/src/rezplugins/package_repository/filesystem.py#L211
* Ability to write RXT file in YAML instead of JSON.

PR #1562, opened 2023-10-29
Properly mark things that are deprecated as deprecated

JeanChristopheMorinPerso:
Relates to #1552

Properly mark things that are deprecated as deprecated and add a new environment variable called `REZ_LOG_DEPRECATION_WARNINGS` to help users know when they are using deprecated things.

The new `REZ_LOG_DEPRECATION_WARNINGS` environment variable will hopefully allow our users to move away from deprecated things before they are removed from 3.0.0.

With this work, we can now deprecate pretty much anything, including config settings. Deprecation warnings for config settings will print the config file that generated the warning. It also works with config setting environment variables too.

```
(.venv) [jcmorin@arch01 rez]$ REZ_LOG_DEPRECATION_WARNINGS=1 REZ_WARN_OLD_COMMANDS=1 rez-search python
RezDeprecationWarning: config setting named 'warn_old_commands' (configured through the REZ_WARN_OLD_COMMANDS environment variable) is deprecated and will be removed in 3.0.0.
python-2.7.16
python-3.7.4
```

I tried to find a balance between being conservative and progressive:
* rez-1 config settings that are currently unused in the code base are deprecated and will be removed in 3.0.0.
* rez-1 Config settings that are still used in the code based are deprecated but are marked as "will be removed in the future".
* Two rez-1 config settings's default will be changed in 3.0.0. That is, `disable_rez_1_compatibility` will be enabled and `rez_1_environment_variables` will be disabled. Users can always override the defaults if they still need the old behavior.
* Things that were previously marked as deprecated now officially have a removal date (3.0.0). They aren't controversial things, so hopefully it's alright with everyone.

I added every deprecation to the change log in preparation of the release (and facilitate reviewing this PR) and I also wrote a transitioning guide for 3.0.0.

Hopefully, everything is clear and our users should find it straightforward and clear.


src/rez/config.py
```diff
@@ -23,11 +23,44 @@
...
+def _warn_deprecated_settings(data, filepath):
+    for key in data:
+        if key in _deprecated_settings:
+            message = (
+                "rez config setting named {0!r} is "
+                "deprecated and will be removed in 3.0.0."
+            ).format(key)
+
+            if os.getenv("REZ_LOG_ALL_DEPRECATION_WARNINGS"):
+                with warnings.catch_warnings():
```
JeanChristopheMorinPerso, line 41: I don't like this. Why not just print to stderr? We could use `warnings.formatwarning` and print that to stderr.

We can also refactor this into two function. One that takes a message argument and another that takes no messaage. This will allow to centralize the warning logic in rez.deprecations (even for the config file).

JeanChristopheMorinPerso: Or maybe just implement our own deprecation warning class?

src/rez/rezconfig.py
```diff
@@ -1080,8 +1113,15 @@
...
+# You should aim to do this. It will mean your packages are more strictly
 # validated, and you can more easily use future versions of Rez.
+#
+# .. versionchanged:: 2.114.0
+#    Now enabled by default.
+#
+# .. deprecated:: 2.114.0
+#    Will be removed in a future release. Additionally, the default will change
+#    from disabled to enabled in rez 3.0.0.
 disable_rez_1_compatibility = False
```
JeanChristopheMorinPerso, line 1125: I initially wanted to deprecate it and remove it in 3.0.0, but I found out that this would be a little bit too disruptive for our oldest users. Same for `rez_1_environment_variables`. `rez_1_environment_variables` is even more painful because it's a lot harder to know if packages are actively using these old environment variables or not.

src/rez/config.py
```diff
@@ -975,3 +1050,9 @@ def get_module_root_config():
 
 # singleton
 config = Config._create_main_config()
+
+if os.getenv("REZ_LOG_DEPRECATION_WARNINGS"):
```
JeanChristopheMorinPerso, line 1054: We should add this to the changelog to give a heads up that config files will be fored-loaded.

JeanChristopheMorinPerso: @instinct-vfx I added a warning in both the docs and the changelog.

PR #1578, opened 2023-11-11
Set default shell on Windows to "powershell"

### LINKED ISSUE #1559: Change default shell on Windows to PowerShell
We would like to change the default shell on Windows to be `powershell` instead of `cmd`.

https://github.com/AcademySoftwareFoundation/rez/blob/371644d19de24d720cc0a4b14f6c17eeb61c55e1/src/rez/system.py#L75-L76

The rationale is that we've been saying for years that users should transition to the powershell plugin. The problem is that we never changed the default. So new users just go with the default, which is cmd, then they hit issues and struggle.

With #1558 around the corner, we should take this as an occasion to change the default.

I suggest that we add a new section to https://rez.readthedocs.io/en/latest/basic_concepts.html# dedicated to shells. We should at least explain that rez uses shells and that by default it uses the same shell type than the parent shell. Obviously we should also document that the default on Windows is powershell.

JeanChristopheMorinPerso:
Fixes #1559.

Set default shell on Windows to "powershell".

Note that I'm not 100% sure if this is really how it should be done. I couldn't find a better way.

### DIFF
--- a/src/rez/system.py
+++ b/src/rez/system.py
@@ -65,7 +65,8 @@ class System(object):
         """Get the current shell.
 
         Returns:
-            The current shell this process is running in (bash, tcsh, pwsh, etc).
+            The current shell this process is running in (bash, tcsh, pwsh, etc). On Windows,
+            the return value is always "powershell".
         """
         from rez.shells import get_shell_types
         shells = set(get_shell_types())
--- a/src/rez/system.py
+++ b/src/rez/system.py
@@ -73,7 +73,7 @@ class System(object):
             raise RezSystemError("no shells available")
 
         if self.platform == "windows":
-            return "cmd"
+            return "powershell"
         else:
             import subprocess as sp
             shell = None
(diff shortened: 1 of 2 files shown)


src/rez/system.py
brycegbrazen, line 1: @JeanChristopheMorinPerso Should we also mention in the docstring of the "shell" property that for Windows we do not actually grab the current shell, but instead always return powershell? Might be helpful for Windows users reading the docs.

JeanChristopheMorinPerso: Yes

brycegbrazen:
LGTM. Thanks for the doc updates as well!

herronelou:
Would probably need https://github.com/AcademySoftwareFoundation/rez/pull/1505 to prevent too many issues.

JeanChristopheMorinPerso:
I tested on Windows from a cmd shell:
```
C:\Users\jeanchristophemorin1>.\rezinstall\Scripts\rez\rez-env.exe

resolved by jeanchristophemorin1@windows-2, on Thu Feb 08 00:14:31 2024, using Rez v2.114.1

requested packages:
~platform==windows           (implicit)
~arch==AMD64                 (implicit)
~os==windows-10.0.20348.SP0  (implicit)

resolved packages:
> $ echo $PSVersionTable.PSEdition
Desktop
```
So it seems to be working as expected. It created a powershell (not pwsh) shell.

PR #1582, opened 2023-11-12
Update package_filter type hint on ResolvedContext

### LINKED ISSUE #1581: package_filter parameter on ResolvedContext class shows incorrect expected type
In `rez.resolved_context` for the `ResolvedContext` class, the expected type of the `package_filter` parameter in the documentation seems to be incorrect compared to what is actually expected by the code.

The expected type according to the documentation is `PackageFilterBase` which _should_ mean that I can pass a `PackageFilter` object for this parameter, however, when I pass a `PackageFilter` class, I get the following error:

```
11:44:55 ERROR    ResolvedContextError: Failed to load context from C:\Users\MY_USER\AppData\Local\Temp\rez_context_ta9orw0v\context.rxt: AttributeError: 'str' object has no attribute 'get'
```

From what I can tell, you _must_ pass a `PackageFilterList` object instead. If you pass a `PackageFilterList` instead, then the resolved context works as intended.

**Environment**
* Windows 10
* Rez 2.112
* Python 3.7

**To Reproduce**
1. Ensure the Rez Python API is in your PYTHONPATH
2. Run a Python interpreter and the following commands:
```
from rez.package_filter import PackageFilter, Rule

pkg_filter = PackageFilter()
rule = Rule.parse_rule('*-beta')
pkg_filter.add_exclusion(rule)
rez_ctx = ResolvedContext(['MY_PKG'], package_filter=pkg_filter)
rez_ctx.execute_shell()
```

**Expected behavior**
A successful resolve, with no errors.

**Actual behavior**
A corrupt context was created. I am placed inside the shell, but I cannot run `rez context` as I usually can, as running `rez context` prints out the same error and does not give me any information about the current Rez context.

Running the following commands in the interpreter works as expected. Note that I am using the `PackageFilterList` class instead of `PackageFilter` class.

```
from rez.package_filter import PackageFilterList, Rule

pkg_filter = PackageFilterList()
rule = Rule.parse_rule('*-beta')
pkg_filter.add_exclusion(rule)
rez_ctx = ResolvedContext(['MY_PKG'], package_filter=pkg_filter)
rez_ctx.execute_shell()
```

brycegbrazen:
Closes #1581 

Updates type hint to be what is actually expected by ResolvedContext under the hood. 

As noted by #1581, ResolvedContext expects a `PackageFilterList` rather than a `PackageFilterBase` when deserializing from a dict, so this PR improves the documentation to match that.

### DIFF
--- a/src/rez/resolved_context.py
+++ b/src/rez/resolved_context.py
@@ -187,7 +187,7 @@ class ResolvedContext(object):
                 is used.
             package_paths (list[str]): List of paths to search for pkgs, defaults to
                 :data:`packages_path`.
-            package_filter (PackageFilterBase): Filter used to exclude certain
+            package_filter (PackageFilterList): Filter used to exclude certain
                 packages. Defaults to settings from :data:`package_filter`. Use
                 :data:`rez.package_filter.no_filter` to remove all filtering.
             package_orderers (list[PackageOrder]): Custom package ordering.


JeanChristopheMorinPerso:
This makes sense to me, considering that the default type is a `PackageFilterList` as seen in the constructor:

https://github.com/AcademySoftwareFoundation/rez/blob/a5977da83f37973658650a7514bcfa78da438674/src/rez/resolved_context.py#L239

JeanChristopheMorinPerso (approved):

PR #1591, opened 2023-11-18
Publish to PyPI using https://docs.pypi.org/trusted-publishers/

JeanChristopheMorinPerso:
Publish to PyPI using https://docs.pypi.org/trusted-publishers/. This will allow us to remove the PyPI token from GH Actions secrets.

Tested on my fork: https://github.com/JeanChristopheMorinPerso/rez/actions/runs/6916563285/job/18816799702

### DIFF
--- a/.github/workflows/pypi.yaml
+++ b/.github/workflows/pypi.yaml
@@ -5,8 +5,12 @@ on:
 
 jobs:
   publish:
-    name: Publish to PyPi
+    name: Publish to PyPI
     runs-on: ubuntu-latest
+
+    permissions:
+      # IMPORTANT: this permission is mandatory for trusted publishing
+      id-token: write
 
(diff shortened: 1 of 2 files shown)


.github/workflows/pypi.yaml
```diff
@@ -5,28 +5,33 @@ on:
...
+
     steps:
       - name: Checkout
         uses: actions/checkout@v4
 
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3
+          python-version: 3.11
```
JeanChristopheMorinPerso, line 22: Good thing I tested on my fork, cause `3` was picking up 3.12 and rez isn't compatible with 3.12, which resulted in an error (since rez is imported from setup.py).

PR #1596, opened 2023-11-26
Load implicit packages on context load

### LINKED ISSUE #1587: Rez Suites load implicit modules from context
Describe the bug here.

**Environment**
* OS: linux rocky 8.7
* Rez version: 2.113.0
* Rez python version: Python 3.11.2

**To Reproduce**
1. `rez-suite --create mysuite`
2. `rez-env maya-2016.2 --output maya.rxt`
3. Pretend this was done on a different OS by editing maya.rxt and changing an implicit package.
4. `rez-suite --add maya.rxt --context maya mysuite`
5. `./mysuite/bin/maya +i`

**Expected behavior**
I would expect the suite to load the implicit packages of it's own context.

**Actual behavior**
The suite loads the implicit packages of the context.

**Possible Solutions**
1. `rez-env --output` should not output implicit packages.
2. `rez-suite --add` should strip implicit packages.
3. `_FWD__invoke_suite_tool_alias` removes implicit packages.
4. `ResolvedContext.from_dict` doesn't load implicit packages.

beatreichenbach:
Fixes #1587

When loading a context the implicit packages are newly generated instead of being loaded from the context file.

### DIFF
--- a/src/rez/resolved_context.py
+++ b/src/rez/resolved_context.py
@@ -708,15 +708,29 @@ class ResolvedContext(object):
                 context = cls.read_from_buffer(f, path)
 
         context.set_load_path(path)
+        context.implicit_packages = [
+            PackageRequest(x) for x in config.implicit_packages
+        ]
         return context
 
     @classmethod
     def read_from_buffer(cls, buf, identifier_str=None):
         """Load the context from a buffer."""
         try:
-            return cls._read_from_buffer(buf, identifier_str)
+            content = buf.read()
+            if content.startswith('{'):  # assume json content
+                doc = json.loads(content)
+            else:
+                doc = yaml.load(content, Loader=yaml.FullLoader)
+            context = cls.from_dict(doc, identifier_str)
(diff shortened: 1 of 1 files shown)


JeanChristopheMorinPerso:
Hi @beatreichenbach . I'm not sure we should do that. Suites aren't meant to be dynamic in any way. They are fully resolved environments, which includes implicit packages. Taking a resolved context for one platform won't necessarily be valid for other platforms. Not only that, but it's possible to add custom implicit packages and override the default ones. They can be dynamic on a per user basis (driven by rez config), etc.

If I understand what this PR does correctly, I think it would break some promises that we have so far given for suites.

JeanChristopheMorinPerso:
As we received no other comments on this and I think this would go against how rez is currently designed, I will close this PR.

I'd still like to thank you for taking the time to create a PR. If you feel like you have a strong case and think this is a valuable feature to have, I would encourage you expand the reasoning in #1587.

Issue #1586, opened 2023-11-16
Add update functionality to rez-suite


beatreichenbach:
Hello everyone,

I'm unsure if I'm missing something but a big part of working with suites to me seems updating them.

**Description**
Following the [example](https://rez.readthedocs.io/en/stable/suites.html#the-rez-suite-tool) in the wiki let's say I have a suite `mysuite` and I added two contexts:
```
rez-suite --add maya.rxt --context maya mysuite
rez-suite --add nuke.rxt --context nuke mysuite
```
After a while I one of the packages inside nuke might have added a new tool and other dependencies are updated. Currently I can't seem to find any way to update my suite other than rebuilding it.
Would it make sense to have `rez-suite --update mysuite` which would automatically resolve all contexts in the suite to their latest packages, remove/add tools in `mysuite/bin` directory.

**Motivation**
Most of the time I would like to update my suites. I don't really need a one time snapshot of a bunch of contexts.

I'd be happy to add a PR of this if I'm not completely missing an existing feature.

beatreichenbach:
After writing the PR and thinking about it more, the less I understand the reasoning behind contexts and why they have a frozen resolve graph stored in them.

Does it not make more sense to store contexts simply as lists of requests and dynamically solve them every time?
Then if I'd want to lock a context to a certain version, I could just specify that with the request version syntax.

Currently if packages update with bug fixes, I can't automatically get those new packages with manually updating my contexts (or if the PR gets accepted by manually updating the suite).

I understand this is getting a little bit off-topic, but I think it's still related enough. I'm curious what the design decisions were behind the current workflow, so I can understand it all better.

beatreichenbach:
Thanks Francois!
That makes sense to me, but isn't that exactly what [suites](https://rez.readthedocs.io/en/stable/suites.html) were supposed to solve, so that not every studio had to write their own rez-launcher?

JeanChristopheMorinPerso:
>That makes sense to me, but isn't that exactly what [suites](https://rez.readthedocs.io/en/stable/suites.html) were supposed to solve, so that not every studio had to write their own rez-launcher?

No. Rez tries very hard to not do any kind environment management. A quote from Allan (the author of rez):
>Rez very deliberately does not attempt to manage production environments. For example, the system that would define say, which packages to use when running maya in show X, is outside the scope of rez.

And we want to keep it this way.

Another quote from Allan again:
[...]

PR #1597, opened 2023-11-26
Add rez-suite update

beatreichenbach:
Solves https://github.com/AcademySoftwareFoundation/rez/issues/1586

Add update functionality that updates the contexts in a suite by resolving them again.


JeanChristopheMorinPerso:
As I explained in https://github.com/AcademySoftwareFoundation/rez/issues/1586#issuecomment-1828733327, rez tries very hard to not do any kind environment management.

As such, I will close this PR.

PR #1606, opened 2023-12-16
Add a mailmap to cleanup our git stats a bit

JeanChristopheMorinPerso:
In response to https://academysoftwarefdn.slack.com/archives/C06A14QSWJW/p1702510546377459, I'm adding a [mailmap](https://git-scm.com/docs/gitmailmap) to let us better see how many contributors we've had.

Please have a look and let me us if there is any mistake. I had to make some choices in some situations, so there is probably errors in there.


JeanChristopheMorinPerso:
I will merge this as is. If someone in the mailmap wants to have a different name or email, please feel free to open a new PR.

PR #1611, opened 2024-01-07
Merge GH Actions workflows that run tests into a single workflow

JeanChristopheMorinPerso:
Merge GH Actions workflows that run tests into a single workflow.

Changes:
* CI workflows are now merged into a single workflow. I also got rid of the core tests. Now all tests are always run. Anyway, running all tests is only slightly longer than running just the "core" tests.
* Windows tests are no more using docker images. It was really slow to download the images on every run. Which means that they now take about as much time to run than Linux and macOS :)
* The `per_available_shell` decorator now sets `config.default_shell` to enforce the default shell when used. This avoids mistakes where we forget to call `config.override("default_shell", shell)` in tests decorated with `per_available_shell`. About a dozen or more tests were simply not testing any other shell then the current default shell (including all the shell tests!).
* Every test that was using the "hello_world" bind module were depending on `.PY` to be in `PATHEXT` on Windows. This is very error prone. This is now fixed by wrapping the python code into an executable using `distlib` (like pip does with entry points).
* `rez-selftest` will now create a "python" rez package that is used by our test packages (for building, etc) to fix issues when there is no executable `python` on a system. This also removes the implicit dependency in our test packages. Note that the python package simply contains a virtualenv. That's very simple and does the trick. We don't really care about which version of python we need anyway. We just need a python interpreter.
* Now testing Python 3.7, 3.8, 3.9, 3.10 and 3.11. Previously, only Python 3.7 was tested.
* Testing more python versions showed a bug in the latest tests I added for the deprecation warnings in the configs. This bug was only present on Windows and was a bug in the test itself. It's now fixed.

All these changes are required due to dropping the docker images on Windows and also due to the fact that we are now using `macos-latest`, which doesn't have a `python` executable (it only has `python3`).


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,83 @@
...
+    - '!**.md'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  core:
+    name: Core tests
```
JeanChristopheMorinPerso, line 23: This is actually the full integration tests not the core tests.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,83 @@
...
+
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: ['macos-11', 'ubuntu-latest', 'windows-2019']
```
JeanChristopheMorinPerso, line 29: Note that I'm still using macOS 11 here. Tests are failing on the latest because python can't be found from within the packages (generated/built during the tests). We will have to figure this out before merging.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,83 @@
...
+      run: .\installdir\Scripts\rez\rez-python -m pip install pytest-cov parameterized
+
+    - name: Run tests
+      if: ${{ matrix.os != 'windows-2019' }}
+      run: ./installdir/bin/rez/rez-selftest -v
+      env:
+        _REZ_ENSURE_TEST_SHELLS: ${{ matrix.shells }}
+
+    - name: Run tests
+      if: ${{ matrix.os == 'windows-2019' }}
```
JeanChristopheMorinPerso, line 80: I would like to simplify this even further, but for now it works.

PR #1614, opened 2024-01-12
Add semantic versioning warning

brycegbrazen:
Add semantic versioning ordering warning based on [this Slack discussion](https://academysoftwarefdn.slack.com/archives/C0321B828FM/p1701468368919439) last month.

This adds a clear note that semantic version ordering differs from Rez version ordering.

### DIFF
--- a/docs/source/basic_concepts.rst
+++ b/docs/source/basic_concepts.rst
@@ -78,13 +78,11 @@ version than ``1.0``.
 .. note::
    No special importance is given to specific characters or letters in Rez version numbers.
    The terms ``alpha`` and ``beta`` for example have no special meaning. Similarly, the number of tokens in
-   a version number doesn't matter, you can have as many as you like. While you are encouraged to use
-   semantic versioning (see `<https://semver.org>`_), it is not enforced.
+   a version number doesn't matter, you can have as many as you like.
 
-.. note::
-   If you are using semantic versioning, please note that version ordering will NOT behave like described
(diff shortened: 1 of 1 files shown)


docs/source/basic_concepts.rst
```diff
@@ -81,6 +81,13 @@ version than ``1.0``.
    a version number doesn't matter, you can have as many as you like. While you are encouraged to use
    semantic versioning (see `<https://semver.org>`_), it is not enforced.
 
+.. note::
+   If you are using semantic versioning, please note that version ordering will NOT behave like described
```
JeanChristopheMorinPerso, line 85: I think that this is already covered by the note above. Do you think we could rework the previous note to clarify it a little bit? For example, instead of saying "you are encouraged to use semver", we could say "you are encouraged to use a semver like scheme".

The note already says that alpha and beta have no meaning. Which leads me to think that the misleading part of the note is the last sentence?

brycegbrazen: @JeanChristopheMorinPerso Yeah I agree that reiterating that special terms having no meaning is redundant, so I'll remove that.
I think the big thing I wanted to add here is an explicit note that version ordering will not behave like it is described in sem ver. Let me know if this next commit feels better to you.

JeanChristopheMorinPerso (approved):

JeanChristopheMorinPerso:
Thanks!

PR #1619, opened 2024-01-17
Add support for overriding built-in bind modules

### LINKED ISSUE #1557: Rez-Bind: Override Built-in Bind Modules
Hi Rez folks!

I'm interested in overriding the built-in behavior of the `os.py` and `arch.py` bind modules, so I added custom versions to [bind_module_path](https://rez.readthedocs.io/en/stable/configuring_rez.html#bind_module_path) in rezconfig. 

Unfortunately, search path ordering in `package_bind.py` means that the builtin bind modules end up taking priority over user-defined modules. 

I would expect the user-defined modules to override the builtins (this is how plugin_path works, for example -- a custom env.py overrides the builtin env.py command). 

https://github.com/AcademySoftwareFoundation/rez/blob/73e7493fba2275115463d7716ccf08e46c52854f/src/rez/package_bind.py#L25

**Environment**
* macOS 13
* Rez 2.111.3
* Python 3.9.10

**To Reproduce**
1. Create a custom `bind_modules` directory
2. Create a file named "bind_modules/os.py"
3. Add a custom bind module creation script to os.py
4. Add `bind_modules` directory to `REZ_BIND_MODULE_PATH`
5. Delete current os package, and call `rez-bind os` 

**Expected behavior**
My expectation is that my custom os.py module will be called. 

**Actual behavior**
The builtin os.py is called. 

**Related Issues/PRs**
Happy to file a PR for this one - I think it's a matter of changing:
```
searchpaths = config.bind_module_path + [builtin_path]
```
to:
```
searchpaths =  [builtin_path] + config.bind_module_path
```

Pantsworth:
Fixes #1557

This PR allows modules defined in `config.bind_module_path` to override builtin bind modules (and adds a test to verify the behavior).

This enables users to customize how Rez creates bind packages (like `os` and `arch`) to match their needs.


JeanChristopheMorinPerso (approved):
Thanks again!

PR #1620, opened 2024-01-17
Fix --fail-graph when there is indirect cycles

### LINKED ISSUE #1568: Indirect Cycle Causes Conflict Graph to Hang Indefinitely
In certain cases, indirect cycles created with a `!` dependency can cause rez to hang indefinitely while trying to construct a fail graph. 

**Environment**
* macOS 13.5.2
* Rez: 2.111.3
* Python 3.9

**To Reproduce**
Attached is a zip file that contains 3 packages that should reproduce the issue. 

Download: [graph_repro.zip](https://github.com/AcademySoftwareFoundation/rez/files/13254812/graph_repro.zip)
```
unzip graph_repro.zip
export REZ_PACKAGES_PATH=graph_repro
rez-env foo-1 bar
```

**Expected behavior**
Expected behavior is a failure graph and text like so --
```
A package was completely reduced: (bar-0.0.1[] (dep(!foo-1.0.0) <--!--> foo==1.0.0))

Resolve paths starting from initial requests to conflict:
  foo-1 --> foo-1.0.0 --> bar-0.0.1 --> !foo-1.0.0
  bar --> bar-0.0.1 --> !foo-1.0.0
```

foo_bar_graph_error

**Actual behavior**
Rez hangs forever after showing
```
A package was completely reduced: (bar-0.0.1[] (dep(!foo-1.0.0) <--!--> foo==1.0.0))
<no further output>
```

**Related Issues/PRs**
Happy to open a PR for this one - what's been working for us so far is changing the following lines to check if the node is already in `visited`. 
https://github.com/AcademySoftwareFoundation/rez/blob/371644d19de24d720cc0a4b14f6c17eeb61c55e1/src/rez/utils/resolve_graph.py#L56

https://github.com/AcademySoftwareFoundation/rez/blob/371644d19de24d720cc0a4b14f6c17eeb61c55e1/src/rez/utils/resolve_graph.py#L86

Original - 
```
            down = next((ne for ne in graph.node_neighbors[node]), None)
```

Patch - 
```
            down = next((ne for ne in graph.node_neighbors[node] if ne not in visited), None)
```

Pantsworth:
Fixes #1568

Bugfix to support graphing resolves if they have cycles from using "!" requirements.


JeanChristopheMorinPerso (approved):
Thanks you so much @Pantsworth and thank you for taking the time to add a test!

PR #1630, opened 2024-02-11
First pass at defining our threat model and security policy

JeanChristopheMorinPerso:
Very WIP. Created a draft PR for visibility.


SECURITY.md
```diff
@@ -0,0 +1,34 @@
...
+List assumptions (rez will run inside an internal network, we assume good intentions, package definitions are written in python and executed)
+
+Access to shared filesystem
+
+The current assumptions are:
+* Rez was designed to be used within a studio environment.
+* Package definitions, both for building packages and resulting from a build are Python files (`package.py`). Rez will read and load them in memory at resolve time.
+* Rez config files can be written in YAML or Python.
+* Package definitions and config files written in Python can contain arbitrary code.
+* Rez will create new shells via subprocesses.
```
JeanChristopheMorinPerso, line 14: Should we state that shells can be nested? I feel like we should since that's differentiator from other environment managers (conda, virtualenv, etc).

SECURITY.md
```diff
@@ -0,0 +1,34 @@
...
+
+We only support the latest version. We try our best to keep rez backward
+compatible as much as possible, which allows us to to only support the latest version.
+
+## Reporting a Vulnerability
+
+If you think you've found a potential vulnerability in rez, please report it by filing a GitHub [security
+advisory](https://github.com/AcademySoftwareFoundation/rez/security/advisories/new). Alternatively, email security@openexr.com and provide your contact info for further private/secure discussion. If your email does not receive a prompt acknowledgement, your address may be blocked.
+
+Our policy is to acknowledge the receipt of vulnerability reports within 72 hours.
```
JeanChristopheMorinPerso, line 34: I think 72 hours is reasonable for a first response. Any opinions?

PR #1634, opened 2024-02-11
Cleanup backports since we are now Python 3.x+ only

BryceGattis:


JeanChristopheMorinPerso:
Can we really replace our lrucache with the built-in one? The one we have has a comment that says that it's modified to make it more performant.

predat:
> Can we really replace our lrucache with the built-in one? The one we have has a comment that says that it's modified to make it more performant.

As far as I know, lru_cache has had a C implementation since at least python 3.7. I've done a few tests of my own, and functools.lru_cache seems to be at least 3x faster than rez.backport.lru_cache.lru_cache.

predat:
Do you think this is a good place to also tackle enum ?

JeanChristopheMorinPerso:
> > Can we really replace our lrucache with the built-in one? The one we have has a comment that says that it's modified to make it more performant.
> 
> As far as I know, lru_cache has had a C implementation since at least python 3.7. I've done a few tests of my own, and functools.lru_cache seems to be at least 3x faster than rez.backport.lru_cache.lru_cache with python 3.11.

Can you share the code you used to run your benchmark? We can also run rez's own benchmark (because yes we have benchmarks).

> Do you think this is a good place to also tackle enum 
@predat enum should go in another PR IMO.

predat:
> Can you share the code you used to run your benchmark? We can also run rez's own benchmark (because yes we have benchmarks).

```python
@lru_cache(maxsize=1000)
def fib(n):
    return 1 if n in (0, 1) else fib(n - 1) + fib(n - 2)
```

with `from rez.backport.lru_cache import lru_cache`:
```bash
python -m timeit -s 'from fib import fib' 'fib(495)'
5000000 loops, best of 5: 53.9 nsec per loop
```

with `from functools import lru_cache`:
```bash
python -m timeit -s 'from fib import fib' 'fib(495)'
500000 loops, best of 5: 860 nsec per loop
```

src/rez/vendor/pyparsing/pyparsing.py
JeanChristopheMorinPerso, line 1: We should not touch the vendored modules.

JeanChristopheMorinPerso (changes requested):
Oups, I wanted to mark it as "request changes". @BryceGattis can you rebase/merge also to resolve the conflict?

BryceGattis:
Restored `rez/backport/ordereddict.py`  so it can still be used the `src/rez/vendor/pyparsing/pyparsing.py` vendor file. 👍

BryceGattis:
@JeanChristopheMorinPerso Notes addressed 👍

JeanChristopheMorinPerso (approved):

PR #1649, opened 2024-02-16
Replace usages of rez.vendor.enum with the built-in enum module

### LINKED ISSUE #1648: Replace usages of `rez.vendor.enum` with the built-in enum module
Now that we dropped support for Python 2, we can now use the built-in [enum](https://docs.python.org/3/library/enum.html) module.

Note that this doesn't apply to things inside `rez.vendor`. But if `rez.vendor.enum` is not used inside `rez.vendor`, we should remove `rez.vendor.enum` entirely.

predat:
resolve #1648

### DIFF
--- a/src/rez/build_process.py
+++ b/src/rez/build_process.py
@@ -12,8 +12,8 @@ from rez.resolved_context import ResolvedContext
 from rez.release_hook import create_release_hooks
 from rez.resolver import ResolverStatus
 from rez.config import config
-from rez.vendor.enum import Enum
 from contextlib import contextmanager
+from enum import Enum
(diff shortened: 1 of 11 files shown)


predat:
> Also, am I right to think that enum is used in some of our vendored modules that?
Yes, it’s also used in vendor/pika for exemple

JeanChristopheMorinPerso:
@predat We can leave `rez.vendor.pika` untouched. We are planning on reviewing our use of vendored dependencies soon.

JeanChristopheMorinPerso (approved):
Thank you!

PR #1660, opened 2024-02-17
Improve install docs by mentioning how to actually download rez

JeanChristopheMorinPerso:
Related to https://www.bestpractices.dev/en/projects/8389#interact.

I noticed that the docs were not very clear on how rez could be download.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -5,9 +5,14 @@ Installation
 Installation Script
 ===================
 
-To install rez, download the source by either cloning the `repository <https://github.com/AcademySoftwareFoundation/rez>`_
-with git or downloading it from the `latest release <https://github.com/AcademySoftwareFoundation/rez/releases/latest>`_.
-If you download rez from the release page on GitHub, don't forget to unpack the downloaded archive.
+To install rez, you will need:
+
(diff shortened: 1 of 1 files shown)


MrLixm:
This is great !
I would also love to see, if that make sense :
- specify explicitly that python need to be installed on the system for the `python` alias to work
- the python version used with the install script determine the python version used by rez
-  which python version are preferred (more tested against I guess ?)

parikshittiwari740:
I think the 3rd point of @MrLixm is quite important and should definitely be added :)

JeanChristopheMorinPerso:
Thank you @MrLixm and @parikshittiwari740! I pushed a new commit that I think addresses your feedback. Does the new changes look good to you?

docs/source/installation.rst
```diff
@@ -5,15 +5,20 @@ Installation
 Installation Script
 ===================
 
-.. warning::
-   The install script only supports Python 3.7+.
+To install rez, you will need:
+
+1. Python 3.7 or above. We support 3.7, 3.8, 3.9, 3.10 and 3.11.
+   The pythoon interpreter you use to run the install script will be the interpreter
```
brycegbrazen, line 11: @JeanChristopheMorinPerso Small typo

PR #1662, opened 2024-02-18
Add Artifact Repository #2

BryceGattis:
This PR picks up from the work that work that @ttrently started in PR #1453.

### DIFF
--- a/src/rez/package_cache.py
+++ b/src/rez/package_cache.py
@@ -166,12 +166,6 @@ class PackageCache(object):
                 % variant.uri
             )
 
-        if not os.path.isdir(variant_root):
-            raise PackageCacheError(
-                "Not cached - variant %s root does not appear on disk: %s"
-                % (variant.uri, variant_root)
-            )
-
         if not force:
             # package is configured to not be cachable
             if not package.is_cachable:
@@ -290,7 +284,7 @@ class PackageCache(object):
         th.start()
 
         try:
-            shutil.copytree(variant_root, rootpath)
+            variant.resource.install(rootpath)
         finally:
             still_copying = False
 
@@ -366,7 +360,7 @@ class PackageCache(object):
 
         return self.VARIANT_REMOVED
 
-    def add_variants_async(self, variants):
+    def add_variants(self, variants, _async=False):
         """Update the package cache by adding some or all of the given variants.
 
         This method is called when a context is created or sourced. Variants
@@ -380,7 +374,7 @@ class PackageCache(object):
         #
         if not system.is_production_rez_install:
             raise PackageCacheError(
-                "PackageCache.add_variants_async is only supported in a "
+                "PackageCache.add_variants is only supported in a "
                 "production rez installation."
             )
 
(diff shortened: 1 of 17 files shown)


src/rez/package_cache.py
```diff
@@ -166,12 +165,6 @@ def add_variant(self, variant, force=False):
                 % variant.uri
             )
 
-        if not os.path.isdir(variant_root):
```
BryceGattis, line 169: This seems to be the culprit for why the tests are breaking. We need to have a generic way to check if we can access the `variant_root` regardless of what type of PackageRepository is used (ex. database/filesystem).

BryceGattis: I've fixed the immediate issue by readding this check, but it's not fully fleshed out yet.

src/rez/package_resources.py
```diff
@@ -303,6 +304,11 @@ def _uri(self):
         """
         raise NotImplementedError
 
+    def install(self, location):
```
BryceGattis, line 307: Should this even be here? Shouldn't it be the responsibility of the ArtifactRepository to know how to install a resource? Wouldn't this break if we're using s3 as an artifact repository since it can't just run a basic copytree from s3?

BryceGattis:
Thought for potential backlog item: Can we use a different data validation library or perhaps updated our vendored `schema` package? The output you get out of it currently is borderline unreadable to a human. For example here's a snippet of a traceback I got while working on this PR:

```
  File "D:\Code\rez\venv\lib\site-packages\rez\vendor\schema\schema.py", line 235, in validate
rez.vendor.schema.schema.SchemaError: Or(<class 'rez.utils.sourcecode.SourceCode'>, Schema({Optional(<class 'str'>): Or(Or(<class 'str'>, [<class 'str'>]), {Optional(<class 'str'>): <class 'object'>, 'command': Or(<class 'str'>, [<class 'str'>]), Optional('requires'): [Or(<class 'str'>, And(<class 'rez.util
s.formatting.PackageRequest'>, Use(<class 'str'>)))], Optional('run_on'): Or(<class 'str'>, [<class 'str'>]), Optional('on_variants'): Or(<class 'bool'>, {'type': 'requires', 'value': [Or(<class 'str'>, And(<class 'rez.utils.formatting.PackageRequest'>, Use(<class 'str'>)))]})})})) did not validate 'None'  
'None' should be instance of {Optional(<class 'str'>): Or(Or(<class 'str'>, [<class 'str'>]), {Optional(<class 'str'>): <class 'object'>, 'command': Or(<class 'str'>, [<class 'str'>]), Optional('requires'): [Or(<class 'str'>, And(<class 'rez.utils.formatting.PackageRequest'>, Use(<class 'str'>)))], Optional
('run_on'): Or(<class 'str'>, [<class 'str'>]), Optional('on_variants'): Or(<class 'bool'>, {'type': 'requires', 'value': [Or(<class 'str'>, And(<class 'rez.utils.formatting.PackageRequest'>, Use(<class 'str'>)))]})})}
```

JeanChristopheMorinPerso:
@BryceGattis I agree the errors are awful. I have started to look at replacing our schema stuff with something more modern. But it's a difficult task because the schemas leak into our public API is in some places, the rez internals (the core ones) are all very tightly coupled to schemas and it is also very performance sensitive. It is probably positive do swap our wcha library with something else without introducing breaking changes, but I'll be a tough task.

PR #1673, opened 2024-03-03
Remove Python 2 support for utils/json.py

BryceGattis:


src/rez/utils/json.py
JeanChristopheMorinPerso, line 1: @BryceGattis we can get rid of this file. You'll have to change the imports.

JeanChristopheMorinPerso (changes requested):

JeanChristopheMorinPerso (approved):
Thank you!

PR #1677, opened 2024-03-06
Improve clarify of early binding functions documentation

BryceGattis:
Added new information about accessible package attributes and environment variables. 

Looking for some help clarifying exactly what language we should use here and where everything should live.

 Just trying to clarify:
 
- What packages attributes can be used inside an early bound function.
- When environment variables can be read inside of packages and if there are certain situations where they cannot be read (ex. early bound functions).

### DIFF
--- a/docs/source/package_definition.rst
+++ b/docs/source/package_definition.rst
@@ -116,7 +116,7 @@ implicit :attr:`this` object:
 .. warning::
    Certain package attributes cannot be accessed inside of an early bound function such as this.root.
 
-   Additionally, :ref: `build environment variables <context-environment-variables>` cannot be accessed inside
+   Additionally, :ref:`build environment variables <context-environment-variables>` cannot be accessed inside
    of early bound functions.
 
 Early binding functions are a convenience. You can always use an arbitrary function instead, like so:
--- a/docs/source/package_definition.rst
+++ b/docs/source/package_definition.rst
@@ -114,7 +114,7 @@ implicit :attr:`this` object:
    your early bound function. An error will be raised if you do.
 
 .. warning::
-   Certain package attributes cannot be accessed inside of an early bound function such as this.root.
+   Certain package attributes cannot be accessed inside of an early bound function such as :attr:`this.root`.
 
    Additionally, :ref:`build environment variables <build-environment-variables>` cannot be accessed inside
    of early bound functions.
--- a/docs/source/package_definition.rst
+++ b/docs/source/package_definition.rst
@@ -79,10 +79,10 @@ and *late binding* functions - and these are decorated using ``@early`` and ``@l
 Early Binding Functions
 +++++++++++++++++++++++
 
-Early binding functions use the ``@early`` decorator. They are evaluated at *build time*, hence the
-'early' in 'early binding' and their definition persists into the
-``package.py``. By 'build time', it is meant that they are evaluated before the
-resolve has occurred, and as such, before the
+Early binding functions use the ``@early`` decorator. They are evaluated at
+*build time*, hence the 'early' in 'early binding' and their definition persists
+in the installed ``package.py``. By 'build time', it is meant that they are
+evaluated before the resolve has occurred, and as such, before the
 :ref:`build environment <the-build-environment>` has been constructed. Therefore
 there are some important distinctions that set early-bound functions apart from
 other function attributes:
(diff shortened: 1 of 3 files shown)


BryceGattis:
It may be nice to add a blurb at the top of [this file](https://github.com/AcademySoftwareFoundation/rez/blob/main/docs/source/environment.rst#environment-variables) that mentions that environment variables on this page can or cannot be used inside of the `package.py` file itself. What do we think?

docs/source/package_definition.rst
```diff
@@ -113,6 +113,12 @@ implicit :attr:`this` object:
    Do not reference other early bound or late bound attributes in
    your early bound function. An error will be raised if you do.
 
+.. warning::
+   Certain package attributes cannot be accessed inside of an early bound function such as :attr:`this.root`.
```
JeanChristopheMorinPerso, line 117: I would prefer is this was not a warning. Additionally, I would say something like "Note that the this object only exposes package attributes and nothing else when used inside an early function.", I feel like this better explains the situation.

We might also add such a note in the documentation of this (https://rez.readthedocs.io/en/stable/package_commands.html#this)? Though I don't know if any of the documented attributes in this are even exposed to early functions...

docs/source/package_definition.rst
```diff
@@ -113,6 +113,12 @@ implicit :attr:`this` object:
...
+
+   Additionally, :ref:`build environment variables <build-environment-variables>` cannot be accessed inside
```
JeanChristopheMorinPerso, line 119: I feel like we should clarify or add details to the first sentence in the early section instead of adding a warning here.

```
Early binding functions use the `@early` decorator. They are evaluated at *build time*, hence the
'early' in 'early binding'. Any package attribute can be implemented as an early binding function.
```
This could be expanded a little bit to clarify what "at build time" means. "build time" here is very subtle and means that it's evaluated outside a resolve. In other words, they are evaluated before a resolve happens. Semantics here are very important.

docs/source/package_definition.rst
BryceGattis: @JeanChristopheMorinPerso I'm a bit confused by what you mean by "Note that the this object only exposes package attributes and nothing else when used inside an early function." Doesn't the `this` object always *only* expose package attributes? Or does it expose other attributes sometimes?

JeanChristopheMorinPerso: Package attributes as in https://rez.readthedocs.io/en/stable/package_definition.html#package-attributes. That's different than what is available on https://rez.readthedocs.io/en/stable/package_commands.html#this.

docs/source/package_definition.rst
```diff
@@ -80,10 +80,15 @@ Early Binding Functions
 +++++++++++++++++++++++
 
 Early binding functions use the ``@early`` decorator. They are evaluated at *build time*, hence the
-'early' in 'early binding'. Any package attribute can be implemented as an early binding function.
+'early' in 'early binding'. By build time, we mean that they are evaluated before the resolve has happened and before
+the :ref:`build environment <the-build-environment>` is constructed. Therefore it's important to note a few important
```
JeanChristopheMorinPerso, line 84: If there is no resolve, there is no environment variables.

Apart from that, it's better but I feel like we can improve the wording even more. @maxnbk @bpabel @instinct-vfx any ideas on how we could improve this paragraph?

instinct-vfx: Not a native speaker, so i struggle with the proper wording. That said i think that the important aspect of early binding functions is that they are not only evaluated at build time but also persisted into the package. That is easy to miss and i remember struggling with some of these concepts when i first started. (It's easy to miss the declarative nature of package.py though it is python syntax)

maxnbk: (?)
```
Early binding functions uses the ``@early`` decorator. They are evaluated at *build time*, hence the 'early' in 'early binding', and their definition persists into the ``package.py``. By 'build time', it is meant that they are evaluated before the resolve has occurred, and as such, before the :ref:`build environment <the-build-environment>` has been constructed. Therefore, there are some important distinctions that set early-bound functions apart from other Function Attributes.

- The :attr:`this` object only exposes package attributes. Nothing else is accessible when inside an early-bound function.
- No rez-set :doc:`environment variables <environment>` can be accessed inside an early bound function.
```

JeanChristopheMorinPerso: Thanks! I like that @maxnbk. We could also add a point to mention that the returned value will be used as the package attribute in the released package. It's kind of mentioned later in the docs but a mention here would make it more explicit, like @instinct-vfx said.

JeanChristopheMorinPerso (approved):
Thanks again for another great contribution @BryceGattis!

PR #1678, opened 2024-03-06
Remove rez.utils.py23 and cleanup more python 2.7 leftovers

### LINKED ISSUE #1646: Remove `rez.utils.py23`
AS the title says, we should get rid of https://github.com/AcademySoftwareFoundation/rez/blob/main/src/rez/utils/py23.py.

* `get_function_arg_names` is still used (though we shuld remove the python 2 support). We would need to move it in a shared module since it's used in 3-4 places.
* `load_module_from_file` should be moved in https://github.com/AcademySoftwareFoundation/rez/blob/main/src/rez/utils/sourcecode.py and support for python 2 should be removed.

wandth:
Fixes #1646

1. remove py23 module & move its test to test_util.py
2. remove \_\_nonzero\_\_


src/rez/util.py
```diff
@@ -156,3 +158,35 @@ def is_non_string_iterable(arg):
         isinstance(arg, collections.abc.Iterable)
         and not isinstance(arg, str)
     )
+
+
+
+def get_function_arg_names(func):
+    """Get names of a function's args.
+
+    Gives full list of positional and keyword-only (py3 only) args.
```
JeanChristopheMorinPerso, line 167:
```suggestion
    Gives full list of positional and keyword-only args.
```

src/rez/utils/platform_.py
```diff
@@ -466,12 +466,6 @@ class WindowsPlatform(Platform):
     name = "windows"
 
     def _arch(self):
```
JeanChristopheMorinPerso, line 468: We can remove this method entirely now. There is no point in overriding it from the parent.

JeanChristopheMorinPerso (changes requested):
Thanks again @wandth. I left two small comments. It should hopefully be the last comments.

(you can push new commits to this pull request instead of closing and opening new pull requests to address my comments).

src/rez/util.py
```diff
@@ -156,3 +158,35 @@ def is_non_string_iterable(arg):
         isinstance(arg, collections.abc.Iterable)
         and not isinstance(arg, str)
     )
+
+
+
```
JeanChristopheMorinPerso, line 163:
```suggestion
```

src/rez/util.py
```diff
@@ -156,3 +158,35 @@ def is_non_string_iterable(arg):
...
+def get_function_arg_names(func):
+    """Get names of a function's args.
+
+    Gives full list of positional and keyword-only args.
+    """
+    spec = inspect.getfullargspec(func)
+    return spec.args + spec.kwonlyargs
+
+
+
```
JeanChristopheMorinPerso, line 173:
```suggestion
```

JeanChristopheMorinPerso (approved):

PR #1679, opened 2024-03-07
Add ability to cache package payloads synchronously

### LINKED ISSUE #1379: Should be possible to configure that a resolve blocks until caching completes (and use the cache for all items)
Subject says most of it.

Ideally this would be implemented as a rezconfig flag, and perhaps rez-env or similar would have a flag to ignore the caching step if a user merely wishes to check the resolve. `--solve-only`?

I imagine that this sort of workflow would be a requirement of s3-bucket-style repositories anyway, so we ought to support this in general.

**Motivation**
Some users may wish to enforce usage of only cached packages, and simply wait for caching to complete.

isohedronpipeline:
Fixes #1379

Carrying on the work from https://github.com/AcademySoftwareFoundation/rez/pull/1452

This adds support for synchronous/blocking package caching during the resolve process.

### DIFF
--- a/src/rez/cli/env.py
+++ b/src/rez/cli/env.py
@@ -117,6 +117,10 @@ def setup_parser(parser, completions=False):
         "--no-pkg-cache", action="store_true",
         help="Disable package caching")
     parser.add_argument(
+        "--pkg-cache-sync", action="store_true",
+        help="Disable asynchronous package caching.  "
+             "Process will block until packages are cached.")
+    parser.add_argument(
         "--pre-command", type=str, help=SUPPRESS)
     PKG_action = parser.add_argument(
         "PKG", type=str, nargs='*',
@@ -212,7 +216,8 @@ def command(opts, parser, extra_arg_groups=None):
             caching=(not opts.no_cache),
             suppress_passive=opts.no_passive,
             print_stats=opts.stats,
-            package_caching=(not opts.no_pkg_cache)
+            package_caching=(not opts.no_pkg_cache),
+            package_cache_async=(not opts.pkg_cache_sync),
         )
 
     success = (context.status == ResolverStatus.solved)
--- a/src/rez/cli/env.py
+++ b/src/rez/cli/env.py
@@ -118,8 +118,9 @@ def setup_parser(parser, completions=False):
         help="Disable package caching")
     parser.add_argument(
         "--pkg-cache-mode", choices=["sync", "async"],
-        help="Optionally disable for force enable asynchronous package caching.  "
-             "If 'sync', the process will block until packages are cached.")
+        help="If provided, override the rezconfig's package_cache_async key. "
+             "If 'sync', the process will block until packages are cached. "
+             "If 'async', the process will not block while packages are cached.")
     parser.add_argument(
         "--pre-command", type=str, help=SUPPRESS)
     PKG_action = parser.add_argument(
@@ -202,12 +203,12 @@ def command(opts, parser, extra_arg_groups=None):
             rule = Rule.parse_rule(rule_str)
             package_filter.add_inclusion(rule)
 
-        if opts.pkg_cache_sync == "async":
-            package_cache_async = True
-        elif opts.pkg_cache_sync == "sync":
-            package_cache_async = False
+        if opts.pkg_cache_mode == "async":
+            package_cache_mode = True
+        elif opts.pkg_cache_mode == "sync":
+            package_cache_mode = False
         else:
-            package_cache_async = None
+            package_cache_mode = None
 
         # perform the resolve
         context = ResolvedContext(
@@ -224,7 +225,7 @@ def command(opts, parser, extra_arg_groups=None):
             suppress_passive=opts.no_passive,
             print_stats=opts.stats,
             package_caching=(not opts.no_pkg_cache),
-            package_cache_async=package_cache_async,
+            package_cache_async=package_cache_mode,
         )
 
     success = (context.status == ResolverStatus.solved)
--- a/src/rez/cli/env.py
+++ b/src/rez/cli/env.py
@@ -117,9 +117,9 @@ def setup_parser(parser, completions=False):
         "--no-pkg-cache", action="store_true",
         help="Disable package caching")
     parser.add_argument(
-        "--pkg-cache-sync", action="store_true",
-        help="Disable asynchronous package caching.  "
-             "Process will block until packages are cached.")
+        "--pkg-cache-mode", choices=["sync", "async"],
+        help="Optionally disable for force enable asynchronous package caching.  "
+             "If 'sync', the process will block until packages are cached.")
     parser.add_argument(
         "--pre-command", type=str, help=SUPPRESS)
     PKG_action = parser.add_argument(
@@ -202,6 +202,13 @@ def command(opts, parser, extra_arg_groups=None):
             rule = Rule.parse_rule(rule_str)
             package_filter.add_inclusion(rule)
 
+        if opts.pkg_cache_sync == "async":
+            package_cache_async = True
+        elif opts.pkg_cache_sync == "sync":
+            package_cache_async = False
+        else:
+            package_cache_async = None
+
         # perform the resolve
         context = ResolvedContext(
             package_requests=request,
@@ -217,7 +224,7 @@ def command(opts, parser, extra_arg_groups=None):
             suppress_passive=opts.no_passive,
             print_stats=opts.stats,
             package_caching=(not opts.no_pkg_cache),
-            package_cache_async=(not opts.pkg_cache_sync),
+            package_cache_async=package_cache_async,
         )
 
     success = (context.status == ResolverStatus.solved)
--- a/src/rez/package_cache.py
+++ b/src/rez/package_cache.py
@@ -372,7 +372,8 @@ class PackageCache(object):
         This method is called when a context is created or sourced. Variants
         are then added to the cache in a separate process.
 
-        This method is left for backwards compatibility.
+        .. deprecated:: 3.1.0
+           Use :method:`add_variants` instead.
         """
         return self.add_variants(variants, package_cache_async=True)
 
--- a/src/rez/package_cache.py
+++ b/src/rez/package_cache.py
@@ -262,9 +262,11 @@ class PackageCache(object):
                                        f"{self.STATUS_DESCRIPTIONS[status]}")
                     elif status == self.VARIANT_FOUND:
                         # We have resolved into a satisfactory state
-                        logger.info(f"{variant.name}-{variant.version} {self.STATUS_DESCRIPTIONS[status]}")
+                        logger.info(f"{variant.name}-{variant.version} "
+                                    f"{self.STATUS_DESCRIPTIONS[status]}")
                     else:
-                        logger.warning(f"{variant.name}-{variant.version} {self.STATUS_DESCRIPTIONS[status]}")
+                        logger.warning(f"{variant.name}-{variant.version} "
+                                       f"{self.STATUS_DESCRIPTIONS[status]}")
                 return (rootpath, status)
 
         # 1.
--- a/src/rez/package_cache.py
+++ b/src/rez/package_cache.py
@@ -417,10 +417,7 @@ class PackageCache(object):
         return self.add_variants(variants, package_cache_async=True)
 
     def add_variants(self, variants, package_cache_async=True):
-        """Update the package cache by adding some or all of the given variants.
-
-        This method is called when a context is created or sourced. Variants
-        are then added to the cache in a separate process.
+        """Add the given variants to the package payload cache.
         """
 
         # A prod install is necessary because add_variants works by
--- a/src/rez/package_cache.py
+++ b/src/rez/package_cache.py
@@ -413,7 +413,7 @@ class PackageCache(object):
         This method is called when a context is created or sourced. Variants
         are then added to the cache in a separate process.
 
-        .. deprecated:: 3.1.0
+        .. deprecated:: 3.2.0
            Use :method:`add_variants` instead.
         """
         return self.add_variants(variants, package_cache_async=True)
--- a/src/rez/package_cache.py
+++ b/src/rez/package_cache.py
@@ -261,8 +261,9 @@ class PackageCache(object):
                         f"{variant.qualified_name} {self.STATUS_DESCRIPTIONS[status]}"
                     )
                 else:
-                    logger.warning(f"{variant.qualified_name} "
-                                    f"{self.STATUS_DESCRIPTIONS[status]}")
+                    logger.warning(
+                        f"{variant.qualified_name} {self.STATUS_DESCRIPTIONS[status]}"
+                    )
             return (rootpath, status)
 
         # 1.
--- a/src/rez/package_cache.py
+++ b/src/rez/package_cache.py
@@ -22,6 +22,7 @@ from contextlib import contextmanager
 from rez.config import config
 from rez.exceptions import PackageCacheError
 from rez.vendor.lockfile import LockFile, NotLocked
+from rez.vendor.progress.spinner import PixelSpinner
 from rez.utils.filesystem import safe_listdir, safe_makedirs, safe_remove, \
     forceful_rmtree
 from rez.utils.colorize import ColorizedStreamHandler
@@ -241,31 +242,28 @@ class PackageCache(object):
         status, rootpath = self._get_cached_root(variant)
         if status in no_op_statuses:
             if logger:
-                logger.warning(f"Not caching {variant.name}-{variant.version}. "
+                logger.warning(f"Not caching {variant.qualified_name}. "
                                f"Variant {self.STATUS_DESCRIPTIONS[status]}")
             return (rootpath, status)
 
         if wait_for_copying and status == self.VARIANT_COPYING:
-            ticks = 0
+            spinner = PixelSpinner(f"Waiting for {variant.qualified_name} to finish copying. ")
             while status == self.VARIANT_COPYING:
-                self._print_with_spinner(
-                    f"Waiting for {variant.name}-{variant.version} to finish copying.", ticks)
-                ticks += 1
-
+                spinner.next()
                 time.sleep(self._COPYING_TIME_INC)
                 status, rootpath = self._get_cached_root(variant)
             else:
                 # Status has changed, so report the change and return
                 if logger:
                     if status in no_op_statuses:
-                        logger.warning(f"{variant.name}-{variant.version} "
+                        logger.warning(f"{variant.qualified_name} "
                                        f"{self.STATUS_DESCRIPTIONS[status]}")
                     elif status == self.VARIANT_FOUND:
                         # We have resolved into a satisfactory state
-                        logger.info(f"{variant.name}-{variant.version} "
+                        logger.info(f"{variant.qualified_name} "
                                     f"{self.STATUS_DESCRIPTIONS[status]}")
                     else:
-                        logger.warning(f"{variant.name}-{variant.version} "
+                        logger.warning(f"{variant.qualified_name} "
                                        f"{self.STATUS_DESCRIPTIONS[status]}")
                 return (rootpath, status)
 
@@ -957,13 +955,3 @@ class PackageCache(object):
         dirs.append(hash_dirname)
 
         return os.path.join(*dirs)
-
-    @staticmethod
-    def _print_with_spinner(message, ticks):
-        """
-        Report a message with a spinner wheel to indicate progress.
-        """
-        wheel = "⣾⣽⣻⢿⡿⣟⣯⣷"
-        ticks = ticks % len(wheel)
-        spinner = wheel[ticks:1 + ticks]
-        print(f" {spinner} {message}", end="\r")
--- a/src/rez/package_cache.py
+++ b/src/rez/package_cache.py
@@ -255,10 +255,7 @@ class PackageCache(object):
             else:
                 # Status has changed, so report the change and return
                 if logger:
-                    if status in no_op_statuses:
-                        logger.warning(f"{variant.qualified_name} "
-                                       f"{self.STATUS_DESCRIPTIONS[status]}")
-                    elif status == self.VARIANT_FOUND:
+                    if status == self.VARIANT_FOUND:
                         # We have resolved into a satisfactory state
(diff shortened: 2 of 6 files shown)


src/rez/cli/env.py
```diff
@@ -116,6 +116,10 @@ def setup_parser(parser, completions=False):
     parser.add_argument(
         "--no-pkg-cache", action="store_true",
         help="Disable package caching")
+    parser.add_argument(
+        "--pkg-cache-sync", action="store_true",
+        help="Disable asynchronous package caching.  "
+             "Process will block until packages are cached.")
```
JeanChristopheMorinPerso, line 122:
```suggestion
    parser.add_argument(
        "--pkg-cache-mode", choices=["sync", "async"],
        help="Package caching mode")
```
I feel like this would cover more cases. With --pkg-cache-sync, if a user sets package_cache_async to false in their config, it would be impossible to undo from the CLI (ignoring setting an env var).

JeanChristopheMorinPerso (commented):
Thanks for creating this PR @isohedronpipeline! I left some comments, please let me know if you have questions.

src/rez/tests/test_package_cache.py
```diff
@@ -134,24 +134,72 @@ def test_caching_on_resolve(self):
 
             # Creating the context will asynchronously add variants to the cache
             # in a separate proc.
-            #
+            # NOTE: pyfoo will not cache, because its repo is set to non-caching (see above)
             c = ResolvedContext([
                 "timestamped-1.2.0",
-                "pyfoo-3.1.0"  # won't cache, see earlier test
+                "pyfoo-3.1.0"
             ])
```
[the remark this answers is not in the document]
isohedronpipeline: I like proving that the config set it to True

src/rez/tests/test_package_cache.py
```diff
@@ -134,24 +134,72 @@ def test_caching_on_resolve(self):
...
         for _ in range(50):
-            time.sleep(0.1)
             cached_root = pkgcache.get_cached_root(variant)
             if cached_root:
                 break
 
+            resolve_not_always_cached = True
+            time.sleep(0.1)
+
+        self.assertNotEqual(cached_root, None)
```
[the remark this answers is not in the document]
isohedronpipeline: How about `Packages were expected to be cached, but were not.`?

src/rez/resolved_context.py
```diff
@@ -249,6 +251,10 @@ def __init__(self, package_requests, verbosity=0, timestamp=None,
 
         self.package_caching = package_caching
 
+        if package_cache_async is None:
+            package_cache_async = config.package_cache_async
+        self.package_cache_async = package_cache_async
```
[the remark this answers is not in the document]
isohedronpipeline: This causes the member attribute to be set twice.  We should move the next line into an `else` block if that is preferred.  However, the one I posted matches the pattern used by the block above:
```
        if package_caching is None:
            if building:
                package_caching = config.package_cache_during_build
            else:
                package_caching = True
        self.package_caching = package_caching
```

src/rez/package_cache.py
```diff
@@ -460,12 +460,14 @@ def add_variants_async(self, variants):
...
-                subprocess.Popen(
-                    [exe, "--daemon", self.path],
+                process = subprocess.Popen(
+                    args,
                     stdout=out_target,
                     stderr=out_target,
                     **kwargs
                 )
+                if not package_cache_async:
+                    process.wait()
```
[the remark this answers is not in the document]
isohedronpipeline: :+1:
https://stackoverflow.com/a/30984882 seems to back you up :-)

src/rez/package_cache.py
```diff
@@ -366,21 +366,21 @@ def remove_variant(self, variant):
 
         return self.VARIANT_REMOVED
 
-    def add_variants_async(self, variants):
+    def add_variants(self, variants, package_cache_async=True):
```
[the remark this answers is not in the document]
isohedronpipeline: :+1:

src/rez/cli/env.py
isohedronpipeline: :+1:

src/rez/cli/env.py
```diff
@@ -116,6 +116,10 @@ def setup_parser(parser, completions=False):
     parser.add_argument(
         "--no-pkg-cache", action="store_true",
         help="Disable package caching")
+    parser.add_argument(
+        "--pkg-cache-mode", choices=["sync", "async"],
+        help="Optionally disable for force enable asynchronous package caching.  "
```
JeanChristopheMorinPerso, line 121: This sentence is really hard to read. I'm not even sure it's a valid English sentence.

src/rez/package_cache.py
```diff
@@ -369,18 +369,28 @@ def remove_variant(self, variant):
     def add_variants_async(self, variants):
         """Update the package cache by adding some or all of the given variants.
 
+        This method is called when a context is created or sourced. Variants
+        are then added to the cache in a separate process.
+
+        This method is left for backwards compatibility.
```
JeanChristopheMorinPerso, line 375:
```suggestion
        .. deprecated:: 3.1.0
           Use :method:`add_variants` instead.
```

src/rez/rezconfig.py
```diff
@@ -278,6 +278,12 @@
 # Enable package caching during a package build.
 package_cache_during_build = False
 
+# Asynchronously cache packages. If this is false, resolves will block until
+# all packages are cached.
+#
+# .. versionadded:: 2.1.0
```
JeanChristopheMorinPerso, line 284:
```suggestion
# .. versionadded:: 3.1.0
```
My bad on the version.

src/rez/cli/env.py
isohedronpipeline: Ha! Okay, try now.

src/rez/package_cache.py
```diff
@@ -460,12 +471,14 @@ def add_variants_async(self, variants):
                 else:
                     out_target = devnull
 
-                subprocess.Popen(
-                    [exe, "--daemon", self.path],
```
JeanChristopheMorinPerso, line 464: Don't we want to keep `--daemon` when async?

isohedronpipeline: This list is identical to the `args` variable defined above, so I just reused `args`.

isohedronpipeline: `args = [exe, "--daemon", self.path]` on line 463.

JeanChristopheMorinPerso: Oh Isee, my bad. Should we remove `--daemon` for sync? Not sure if it really makes sense to launch it as a daemon process when using sync mode.

JeanChristopheMorinPerso: Actually, do we even need to start a subprocess when we synchronously cache? Also, what will happen to pending variants (see line 425)? If I understand the code, an in-flight variant would be skipped, which would result in a potentially broken environment...

isohedronpipeline: We probably don't need it no.  I was more or less copying the logic from the original PR for that.

There also seems to be a fair amount of logic in the CLI entry point for gathering things for the cache, so I'm not sure if we can really run it without the CLI.  for instance, look at `rez/cli/pkg-cache` line 154.  That seems to be the primary business logic of caching without the daemon.  If we were to remove the subprocess, we should move that whole block into the PackageCache class, which feels maybe beyond the scope of what we were trying to do, but maybe not.  Let me know if you'd like to forge ahead with that strategy.

JeanChristopheMorinPerso:
I took a deeper look, and as things are in this PR, it's doesn't work as you would expect. If you were to `rez-env maya --pkg-cache-mode async` and then `rez-env maya --pkg-cache-mode sync` and initial cache (when we ran with async) takes long enough, the second rez-env (in sync mode) will skip the maya package. See https://github.com/AcademySoftwareFoundation/rez/blob/main/src/rez/package_cache.py#L210-L219. Basically, the package cache will skip any variant that is marked as in-progress.

Additionally, I still have to verify, but I think that we should really avoid using `--daemon` when caching synchronously. If we use `--daemon`, it would mean that cancelling `rez-env --pkg-cache-mode sync` would not cancel the caching process. After all, keeping the process alive is the the reason to have a daemon mode in the first place. Or I'd go even further, we should not use a subprocess.

isohedronpipeline:
> I took a deeper look, and as things are in this PR, it's doesn't work as you would expect. If you were to `rez-env maya --pkg-cache-mode async` and then `rez-env maya --pkg-cache-mode sync` and initial cache (when we ran with async) takes long enough, the second rez-env (in sync mode) will skip the maya package. See https://github.com/AcademySoftwareFoundation/rez/blob/main/src/rez/package_cache.py#L210-L219. Basically, the package cache will skip any variant that is marked as in-progress. This would result in a non-localized environment.

I think it would be preferable to call `add_variant(force=True)` rather than poll until the `.copying` sentinel file has disappeared, but I'm not sure what to do about the file lock.  We really don't want to be copying the same path at the same time.  Perhaps we should modify this logic to copy to an intermediate location that does an atomic move operation once the copy has completed to avoid errors with multiple cache operations running concurrently.  I'm not sure there is a *great* solution for mixing sync and async caching operations for the same files at the same time.

> 
> Additionally, I still have to verify, but I think that we should really avoid using `--daemon` when caching synchronously. If we use `--daemon`, it would mean that cancelling `rez-env --pkg-cache-mode sync` would not cancel the caching process. After all, keeping the process alive is the the reason to have a daemon mode in the first place. Or I'd go even further, we should not use a subprocess.
Yeah, I'm fine with that.  I'll move some of the logic from the cli entry point into `package_cache` later today or tomorrow.

> 
> We should also consider adding a progress bar or some logs when using the sync mode. I anticipate that users will complain if `rez-env maya houdini` blocks for multiple minutes without clearly saying what's going on.
We can have something print out in the `_while_copying()` function that is threaded.  Since we're using `shutil.copytree` I don't think we're going to have much by the way of stats about total size/amount transferred/estimated time to completion without hammering the disk, which I'd like to avoid.

JeanChristopheMorinPerso:
>I think it would be preferable to call add_variant(force=True) rather than poll until the .copying sentinel file has disappeared, but I'm not sure what to do about the file lock. We really don't want to be copying the same path at the same time. Perhaps we should modify this logic to copy to an intermediate location that does an atomic move operation once the copy has completed to avoid errors with multiple cache operations running concurrently. I'm not sure there is a great solution for mixing sync and async caching operations for the same files at the same time.

I'm not too sure why we would use `force=True` or why you don't want to poll. Can you explain your thought a bit more please? Like you say, we really don't want to copy twice or copy the same package multiple times in parallel. The code goes to great length to avoid this.

isohedronpipeline:
> I'm not too sure why we would use `force=True` or why you don't want to poll. Can you explain your thought a bit more please? Like you say, we really don't want to copy twice or copy the same package multiple times in parallel. The code goes to great length to avoid this.

You could be right.  I am worried that a previous sync process could have failed, leaving behind the .copying file, but no further updates to it will have been made.  I would imagine that one use case for doing a package cache syncronously would be to force all of the caching operations to happen *right now* as possibly a way to restart a locked or frozen cache operation.

I see that the copying file is updated every .2 seconds though, so we should be able to check that and continue to wait while that file is being updated.

My hesitation to hit the disk repeatedly is maybe misplaced since the cache is on the local disk.  In studio environments, I hesitate to hit network drives with something that could be run many times in parallel.  Of course you're right and starting multiple large file transfers instead is madness.

JeanChristopheMorinPerso:
The package cache class already knows how to handle stalled packages. It should be able to handle the case where a variant is left in progress due to an error or whatever else and restart it.

As for polling, I agree that polling a shared file system is not wanted. Though, caching is usually done on local disks. We could maybe use `inotify` to optimize ans simplify the process.

src/rez/package_cache.py
```diff
@@ -214,17 +229,46 @@ def add_variant(self, variant, force=False):
...
+        }
+        if not wait_for_copying:
+            # Copying variants are only no-ops if we want to ignore them.
+            no_op_statuses.add(self.VARIANT_COPYING)
 
         # variant already exists, or is being copied to cache by another thread/proc
         status, rootpath = self._get_cached_root(variant)
         if status in no_op_statuses:
+            if logger:
+                logger.warning(f"Not caching {variant.name}-{variant.version}. "
```
JeanChristopheMorinPerso, line 244: I think we should use the qualified name which will include the variant id.
```suggestion
                logger.warning(f"Not caching {variant.qualified_name}. "
```
The same applies to all the other places where `{variant.name}-{variant.version}` is use.

src/rez/package_cache.py
```diff
@@ -863,3 +957,13 @@ def _get_hash_path(self, variant):
         dirs.append(hash_dirname)
 
         return os.path.join(*dirs)
+
+    @staticmethod
+    def _print_with_spinner(message, ticks):
```
JeanChristopheMorinPerso, line 962: Can we use `rez.vendor.progress.spinner` instead? We also have a small wrapper in rez.util.ProgressBar` that we use in rez-search (https://github.com/AcademySoftwareFoundation/rez/blob/main/src/rez/package_search.py#L69).

JeanChristopheMorinPerso (commented):
I did a first read of your recent changes, but I have not yet reviewed them thoroughly. Hopefully that will allow me to think about it a little bit more.

src/rez/package_cache.py
isohedronpipeline: Yep, updated.
It's not really a progress bar since it's just waiting for the status to change.  I suppose we could monitor the state of each copying package, but I think a spinner is generally okay for this.

src/rez/package_cache.py
isohedronpipeline: :+1:

chadrik:
Hi @JeanChristopheMorinPerso  I think all the outstanding issues have been addressed.  Any other thoughts?

src/rez/package_cache.py
```diff
@@ -214,17 +230,43 @@ def add_variant(self, variant, force=False):
...
+                               f"Variant {self.STATUS_DESCRIPTIONS[status]}")
             return (rootpath, status)
 
+        if wait_for_copying and status == self.VARIANT_COPYING:
+            spinner = PixelSpinner(f"Waiting for {variant.qualified_name} to finish copying. ")
+            while status == self.VARIANT_COPYING:
+                spinner.next()
+                time.sleep(self._COPYING_TIME_INC)
+                status, rootpath = self._get_cached_root(variant)
+            else:
```
JeanChristopheMorinPerso, line 255: I think this else is not needed and is redundant.

src/rez/package_cache.py
```diff
@@ -214,17 +230,43 @@ def add_variant(self, variant, force=False):
...
+                    if status in no_op_statuses:
+                        logger.warning(f"{variant.qualified_name} "
+                                       f"{self.STATUS_DESCRIPTIONS[status]}")
+                    elif status == self.VARIANT_FOUND:
+                        # We have resolved into a satisfactory state
+                        logger.info(f"{variant.qualified_name} "
+                                    f"{self.STATUS_DESCRIPTIONS[status]}")
+                    else:
+                        logger.warning(f"{variant.qualified_name} "
+                                       f"{self.STATUS_DESCRIPTIONS[status]}")
```
JeanChristopheMorinPerso, line 267:
```suggestion
                    if status == self.VARIANT_FOUND:
                        # We have resolved into a satisfactory state
                        logger.info(f"{variant.qualified_name} "
                                    f"{self.STATUS_DESCRIPTIONS[status]}")
                    else:
                        logger.warning(f"{variant.qualified_name} "
                                       f"{self.STATUS_DESCRIPTIONS[status]}")
```
If I'm not mistaken, this would have the same result.

src/rez/package_cache.py
```diff
@@ -371,28 +413,50 @@ def add_variants_async(self, variants):
 
         This method is called when a context is created or sourced. Variants
         are then added to the cache in a separate process.
+
+        .. deprecated:: 3.1.0
```
JeanChristopheMorinPerso, line 417: TODO for maintainers: Update with appropriate version number before merging.

src/rez/package_cache.py
```diff
@@ -371,28 +413,50 @@ def add_variants_async(self, variants):
...
         """
+        return self.add_variants(variants, package_cache_async=True)
 
-        # A prod install is necessary because add_variants_async works by
+    def add_variants(self, variants, package_cache_async=True):
+        """Update the package cache by adding some or all of the given variants.
+
+        This method is called when a context is created or sourced. Variants
+        are then added to the cache in a separate process.
+        """
```
JeanChristopheMorinPerso, line 427:
```suggestion
        """Add the given variants to the package payload cache.
        """
```

src/rez/package_cache.py
```diff
@@ -564,6 +642,15 @@ def run_daemon(self):
             if pid > 0:
                 sys.exit(0)
 
+        self.run_caching_operation(wait_for_copying=False)
+
+    def run_caching_operation(self, wait_for_copying=False):
```
JeanChristopheMorinPerso, line 647: I feel like this method should be private unless there is a god reason not to. I also think `wait_for_copying` should be True by default.

src/rez/package_cache.py
```diff
@@ -116,7 +129,7 @@ def get_cached_root(self, variant):
 
         return rootpath
 
-    def add_variant(self, variant, force=False):
+    def add_variant(self, variant, force=False, wait_for_copying=False, logger=None):
```
JeanChristopheMorinPerso, line 132: I think `wait_for_copying` should be True by default.

src/rez/rezconfig.py
```diff
@@ -278,6 +278,12 @@
 # Enable package caching during a package build.
 package_cache_during_build = False
 
+# Asynchronously cache packages. If this is false, resolves will block until
+# all packages are cached.
+#
+# .. versionadded:: 3.1.0
```
JeanChristopheMorinPerso, line 284: TODO for maintainers: Update with appropriate version number before merging.

src/rez/package_cache.py
```diff
@@ -214,17 +230,43 @@ def add_variant(self, variant, force=False):
...
         # variant already exists, or is being copied to cache by another thread/proc
         status, rootpath = self._get_cached_root(variant)
         if status in no_op_statuses:
+            if logger:
+                logger.warning(f"Not caching {variant.qualified_name}. "
+                               f"Variant {self.STATUS_DESCRIPTIONS[status]}")
             return (rootpath, status)
 
+        if wait_for_copying and status == self.VARIANT_COPYING:
+            spinner = PixelSpinner(f"Waiting for {variant.qualified_name} to finish copying. ")
```
JeanChristopheMorinPerso, line 250: If I'm not mistaken, the `progress` package takes care of disabling spinners if there is no TTY attached the the process. And I also think the output goes to stderr by default. Could you confirm both please? I don't want this code to spam logs in CI or in a render farm and I also don't want to see the spinner in stdout since that would affect the pipebility (if that's a word...) of rez.

src/rez/package_cache.py
```diff
@@ -214,17 +230,43 @@ def add_variant(self, variant, force=False):
...
+            while status == self.VARIANT_COPYING:
+                spinner.next()
+                time.sleep(self._COPYING_TIME_INC)
+                status, rootpath = self._get_cached_root(variant)
+            else:
+                # Status has changed, so report the change and return
+                if logger:
+                    if status in no_op_statuses:
+                        logger.warning(f"{variant.qualified_name} "
```
JeanChristopheMorinPerso, line 259: I don't think it's something we want to tackle here, but it could be a surprising behavior that nothing is done to force a copy when a variant is stalled.

JeanChristopheMorinPerso (commented):
Thanks @isohedronpipeline. I think we are almost there. I left some small comments that should hopefully be quick to address.

It would be great if you could also update https://rez.readthedocs.io/en/stable/caching.html#package-caching. The behavior of the async caching is well documented. I think the behavior of the sync cache should also be documented in a similar way.

ameliendeshams:
Hi @isohedronpipeline,

I'm delighted to see this pull request! We've been waiting for this feature, and it's great to see it implemented.

At Fortiche, we've tested this branch, and it works exactly as expected. The synchronous package caching is a game-changer for our studio environments, where we often need to cache packages quickly and reliably.

We're also happy to see that you've addressed the concerns about progress and logging in sync mode. It's essential to provide users with clear feedback when running long-running commands.

Thanks again for your work on this feature! We're excited to see it merged and start using it in our production environments.

src/rez/package_cache.py
isohedronpipeline: No, I think the description of a stalled cache in the documentation makes it clear that it needs to be cleaned by hand in order to cache again in the future.  I don't think we want to change that behavior here.

src/rez/package_cache.py
isohedronpipeline: That would change the default behavior from async to sync, which I don't think we want to do, do we?

src/rez/package_cache.py
isohedronpipeline: I'll make it private, but I'd rather not change the default beahvior of the caching from async to sync here.

isohedronpipeline: Sorry, this one can be true, you're right.

src/rez/package_cache.py
isohedronpipeline: Yeah, that appears to be the case.

src/rez/package_cache.py
JeanChristopheMorinPerso: This comment is till valid and should be addressed. https://docs.python.org/3/reference/compound_stmts.html#the-while-statement

>This repeatedly tests the expression and, if it is true, executes the first suite; if the expression is false (which may be the first time it is tested) the suite of the else clause, if present, is executed and the loop terminates.
>
>A [break](https://docs.python.org/3/reference/simple_stmts.html#break) statement executed in the first suite terminates the loop without executing the else clause’s suite. A [continue](https://docs.python.org/3/reference/simple_stmts.html#continue) statement executed in the first suite skips the rest of the suite and goes back to testing the expression.

JeanChristopheMorinPerso (changes requested):
Almost there!

JeanChristopheMorinPerso (approved):
Thanks a lot @isohedronpipeline and @ttrently!

JeanChristopheMorinPerso:
I just realized that we missed adding a CLI flag to the `rez-pkg-cache` command...

PR #1682, opened 2024-03-12
Improve explicitness of help doc of --command on rez-env

brycegbrazen:
Felt like this explicit mention of non-interactivity is helpful for users new to Rez.

### DIFF
--- a/src/rez/cli/env.py
+++ b/src/rez/cli/env.py
@@ -29,8 +29,11 @@ def setup_parser(parser, completions=False):
         help="skip loading of startup scripts")
     command_action = parser.add_argument(
         "-c", "--command", type=str,
-        help="read commands from a string and exit, instead of starting an interactive shell. "
-        "Alternatively, list command arguments after a '--'")
+        help="execute command within rez environment and exit, instead of "
+        "starting an interactive shell. Alternatively, list command after a "
+        "'--'. The command and arguments passed to '-c' must be passed in as "
+        "a single shell argument, whereas the command and arguments after "
+        "'--' may be passed in as several shell arguments.")
     parser.add_argument(
         "-s", "--stdin", action="store_true",
         help="read commands from standard input")
--- a/src/rez/cli/env.py
+++ b/src/rez/cli/env.py
@@ -29,8 +29,8 @@ def setup_parser(parser, completions=False):
         help="skip loading of startup scripts")
     command_action = parser.add_argument(
         "-c", "--command", type=str,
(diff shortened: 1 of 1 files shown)


bpabel:
Thanks for your recommendation! I agree it could be better worded. A few things I think could be improved:

- "read" is a poor verb choice; all arguments are read. "execute" would be a better verb.
- denoting it as a "string" is redundant, all commands are strings.
- "command arguments" is misleading; it's actually a full command and command arguments.

Something like ```"Execute command within rez environment and exit..."``` would be better, plus your other additions (also, replace "command arguments" with just "command")

Would you mind updating?

bpabel:
Also, something I just realized, the command and arguments passed to `-c` must be passed in as a single shell argument, whereas the command and arguments after `--` are all passed in as separate shell arguments.

That should probably be noted somehow in the help text.

src/rez/cli/env.py
```diff
@@ -29,8 +29,11 @@ def setup_parser(parser, completions=False):
         help="skip loading of startup scripts")
     command_action = parser.add_argument(
         "-c", "--command", type=str,
-        help="read commands from string. Alternatively, list command arguments "
-        "after a '--'")
+        help="execute command within rez environment and exit, instead of "
+        "starting an interactive shell. Alternatively, list command after a "
+        "'--'. The command and arguments passed to '-c' must be passed in as "
```
JeanChristopheMorinPerso, line 34:
```suggestion
        "starting an interactive shell. Alternatively, list command after a "
        "'--'. The command and arguments passed to '-c' must be passed in as "
```
Should this  (`--`) be documented in the program's help (the module docstring) instead of in the `-c` argument? One thing I would like (and I think I made an issue for this) is to improve our CLI docs. For example, we could show possible combinations like in https://pip.pypa.io/en/stable/cli/pip_install/.

bpabel: I like the showing possible combinations idea, since `rez-env` effectively has several different call signatures and uses.

The CLI doesn't currently use it, but the [`ArgumentParser.usage`](https://docs.python.org/3/library/argparse.html#usage) attribute is probably how the multiple usages (and the `--` mode) should be documented. Right now, the usage is auto-generated from the argument list. It's pretty unruly and redundant; we don't need all the args listed in the example usage. 

Maybe save that for a separate ticket though, since we would need to change the parser setup to include custom `usage` information for each subparser, then we can move some of this information there.

JeanChristopheMorinPerso: Sounds good.

PR #1684, opened 2024-03-15
Package Orderers now apply to variants

### LINKED ISSUE #369: PackageOrderer classes don't apply to variants
ie, say you have a package foo:

```
name = foo
variants = [['bar-1.4'], ['bar-2.0']]
```
...but you then use a package orderer class to make it so that bar-1.4 comes before bar-2.0

If you just do:

```rez-env bar```

...then you will get bar-1.4 as expected.  But if you do:

```rez-env foo```

...then you still get bar-2.0.  The reason is that the PackageOrder classes are not currently used when sorting variants within the solver.

I've created a fix for this in my pull request: https://github.com/nerdvegas/rez/pull/355

...but it required a change in the way PackageOrderer classes work - namely, that they now implement a sort_key function, instead of reordering a list themselves.  This should still be fairly general (especially with the aid of some wrapper classes I added, like ReverseSorted), and made sorting of variant lists much easier.

### LINKED ISSUE #1683: Package orderers are not applied when choosing a variant
I'm seeing some unhelpful behavior with version_split package_orderers:

I have:
package A-1 and A-2
package B-3 that depends on A
package C-4, with variants [A-1] and [A-2]

I also have a version_split package orderer:
```yaml
package_orderers:
- type: per_family
  orderers:
  - type: version_split
    packages:
    - A
    first_version: "1"
```

Based on this package orderer, `rez env B` correctly resolves to A-1.

The issue I'm seeing is that `rez env C` resolves to A-2, despite A-1 being supported and preferred by the package orderer. `rez env A C` correctly chooses A-1, but not `rez env C A`, which makes me think package orderers are not applied when choosing a variant.

**Environment**
* Windows 10
* Rez 2.112.0
* Python 3.7.8

**To Reproduce**
**Expected behavior**
**Actual behavior**
See https://github.com/ian471/rez-pkg-orderer-repro

isohedronpipeline:
Fixes #369
Fixes #1683

This PR fixes this bug so that package orderers are applied to all packages and variants uniformly in all circumstances, for more expected results.

Carries on the work originally started by @pmolodo with #355 and #413.

Here is an example of how one might take advantage of package ordering applying to variants:
Consider a package `python` which has versions `['2.6.4', '2.7.16', '3.7.9', '3.9.6', '3.10.8']`
The default sorted package orderer would give us this sorted list:
`['3.10.8', '3.9.6',  '3.7.9', '2.7.16', '2.6.4']`
And we end up choosing the first possible option in the resolve: `3.10.8`

If we apply the following package order:
```python
package_orderers = [
    {
       "type": "per_family",
       "orderers": [
            {
                "packages": ["python"],
                "type": "version_split",
                "first_version": "2.7.16"
            }
        ]
    }
]
```
Then we would get packages ordered like this: `['2.7.16', '2.6.4', '3.10.8', '3.9.6',  '3.7.9']`, where we start at the version_split first version.
Which in turn would resolve to `2.7.16`.  So far so good.  This is the current behavior.

Now consider a package `pipeline` which has variants:
```python
[
    ["python-2.6.4"],
    ["python-2.7.16"],
    ["python-3.7.9"],
    ["python-3.9.6"],
]
```
If you do `rez env pipeline` we would get `python-3.9.6` in the resolve, **even if we have a package orderer specifically saying we want to prefer python 2.7.16**.  In order to get the correctly requested version of python, we must include it in the resolve: `rez env pipeline python-2.7.16`, which largely defeats the purpose of the package orderer, since the point of it is to inform the choices of packages for second order package requests.

@JeanChristopheMorinPerso and I have discussed the possible speed implications.  There should be neglible differences, even for very large repositories.

Variants were already being sorted, but this change just changes them to be sorted according to a sort key, in a uniform way to  first order requests, instead of simply by version.  The sort key for more complicated package orderers are cached anyway, so it really should be very quick.

Since this will change the way packages are resolved, there is a small danger of people no longer getting the same resolve they used to.  However, it is the belief of @JeanChristopheMorinPerso, @maxnbk, and I that anyone who is using package ordering would have expected this to be the correct result anyway, and the number of people using package orderers is small enough.

### DIFF
--- a/src/rez/package_order.py
+++ b/src/rez/package_order.py
@@ -150,7 +150,9 @@ class PackageOrder(object):
             version: (Version) the version object you wish to generate a key for
 
         Returns:
-            Comparable object
+            Sortable object
+                The returned object must be sortable, which means that it must implement __lt__.
+                The specific return type is not important.
         """
         raise NotImplementedError
 
--- a/src/rez/package_order.py
+++ b/src/rez/package_order.py
@@ -636,7 +636,7 @@ class PackageOrderList(list):
 
     @staticmethod
     def _to_orderer(orderer: Union[dict, PackageOrder]) -> PackageOrder:
-        if not isinstance(orderer, PackageOrder):
+        if isinstance(orderer, dict):
             orderer = from_pod(orderer)
         return orderer
 
--- a/src/rez/package_order.py
+++ b/src/rez/package_order.py
@@ -603,7 +603,7 @@ class TimestampPackageOrder(PackageOrder):
     @classmethod
     def from_pod(cls, data):
         return cls(
-            timestamp=data["timestamp"],
+            data["timestamp"],
             rank=data.get("rank", 0),
             packages=data.get("packages"),
         )
--- a/src/rez/package_order.py
+++ b/src/rez/package_order.py
@@ -306,13 +306,14 @@ class PerFamilyOrder(PackageOrder):
     def sort_key_implementation(self, package_name, version):
         orderer = self.order_dict.get(package_name)
         if orderer is None:
-            if self.default_order is not None:
-                orderer = self.default_order
-            else:
+            
+            if self.default_order is None:
                 # shouldn't get here, because applies_to should protect us...
                 raise RuntimeError(
                     "package family orderer %r does not apply to package family %r",
                     (self, package_name))
+
+            orderer = self.default_order
 
         return orderer.sort_key_implementation(package_name, version)
 
--- a/src/rez/package_order.py
+++ b/src/rez/package_order.py
@@ -4,10 +4,40 @@
 
 from inspect import isclass
 from hashlib import sha1
+from typing import Dict, Iterable, List, Optional, Union
 
 from rez.config import config
 from rez.utils.data_utils import cached_class_property
-from rez.version import Version
+from rez.version import Version, VersionRange
+from rez.version._version import _Comparable, _ReversedComparable, _LowerBound, _UpperBound, _Bound
+
+
+ALL_PACKAGES = "*"
+
+
+class FallbackComparable(_Comparable):
+    """First tries to compare objects using the main_comparable, but if that
+    fails, compares using the fallback_comparable object.
+    """
+
+    def __init__(self, main_comparable, fallback_comparable):
+        self.main_comparable = main_comparable
+        self.fallback_comparable = fallback_comparable
+
+    def __eq__(self, other):
+        try:
+            return self.main_comparable == other.main_comparable
+        except Exception:
+            return self.fallback_comparable == other.fallback_comparable
+
+    def __lt__(self, other):
+        try:
+            return self.main_comparable < other.main_comparable
+        except Exception:
+            return self.fallback_comparable < other.fallback_comparable
+
+    def __repr__(self):
+        return '%s(%r, %r)' % (type(self).__name__, self.main_comparable, self.fallback_comparable)
 
 
 class PackageOrder(object):
@@ -16,8 +46,32 @@ class PackageOrder(object):
     #: Orderer name
     name = None
 
-    def __init__(self):
-        pass
+    def __init__(self, packages: Optional[Iterable[str]] = None):
+        """
+        Args:
+            packages: If not provided, PackageOrder applies to all packages.
+        """
+        self.packages = packages
+
+    @property
+    def packages(self) -> List[str]:
+        """Returns an iterable over the list of package family names that this
+        order applies to
+
+        Returns:
+            (Iterable[str]) Package families that this orderer is used for
+        """
+        return self._packages
+
+    @packages.setter
+    def packages(self, packages: Union[str, Iterable[str]]):
+        if packages is None:
+            # Apply to all packages
+            self._packages = [ALL_PACKAGES]
+        elif isinstance(packages, str):
+            self._packages = [packages]
+        else:
+            self._packages = sorted(packages)
 
     def reorder(self, iterable, key=None):
         """Put packages into some order for consumption.
@@ -40,9 +94,69 @@ class PackageOrder(object):
         Returns:
             list: Reordered ``iterable``
         """
+        key = key or (lambda x: x)
+        package_name = self._get_package_name_from_iterable(iterable, key=key)
+        return sorted(iterable,
+                      key=lambda x: self.sort_key(package_name, key(x).version),
+                      reverse=True)
+
+    @staticmethod
+    def _get_package_name_from_iterable(iterable, key=None):
+        """Utility method for getting a package from an iterable"""
+        try:
+            item = next(iter(iterable))
+        except (TypeError, StopIteration):
+            return None
+
+        key = key or (lambda x: x)
+        return key(item).name
+
+    def sort_key(self, package_name, version_like):
+        """Returns a sort key usable for sorting packages within the same family
+
+        Args:
+            package_name: (str) The family name of the package we are sorting
+            version_like: (Version|_LowerBound|_UpperBound|_Bound|VersionRange)
+                the version-like object you wish to generate a key for
+
+        Returns:
+            Comparable object
+        """
+        if isinstance(version_like, VersionRange):
+            return tuple(self.sort_key(package_name, bound) for bound in version_like.bounds)
+        elif isinstance(version_like, _Bound):
+            return (self.sort_key(package_name, version_like.lower),
+                    self.sort_key(package_name, version_like.upper))
+        elif isinstance(version_like, _LowerBound):
+            inclusion_key = -2 if version_like.inclusive else -1
+            return self.sort_key(package_name, version_like.version), inclusion_key
+        elif isinstance(version_like, _UpperBound):
+            inclusion_key = 2 if version_like.inclusive else 1
+            return self.sort_key(package_name, version_like.version), inclusion_key
+        elif isinstance(version_like, Version):
+            # finally, the bit that we actually use the sort_key_implementation for.
+            return FallbackComparable(
+                self.sort_key_implementation(package_name, version_like), version_like)
+        else:
+            raise TypeError(version_like)
+
+    def sort_key_implementation(self, package_name, version):
+        """Returns a sort key usable for sorting these packages within the
+        same family
+        Args:
+            package_name: (str) The family name of the package we are sorting
+            version: (Version) the version object you wish to generate a key for
+
+        Returns:
+            Comparable object
+        """
         raise NotImplementedError
 
     def to_pod(self):
+        raise NotImplementedError
+
+    @classmethod
+    def from_pod(cls, data):
         raise NotImplementedError
 
     @property
@@ -53,7 +167,7 @@ class PackageOrder(object):
         raise NotImplementedError
 
     def __eq__(self, other):
-        raise NotImplementedError
+        return type(self) == type(other) and str(self) == str(other)
 
     def __ne__(self, other):
         return not self == other
@@ -71,8 +185,10 @@ class NullPackageOrder(PackageOrder):
     """
     name = "no_order"
 
-    def reorder(self, iterable, key=None):
-        return list(iterable)
+    def sort_key_implementation(self, package_name, version):
+        # python's sort will preserve the order of items that compare equal, so
+        # to not change anything, we just return the same object for all...
+        return 0
 
     def __str__(self):
         return "{}"
@@ -87,12 +203,15 @@ class NullPackageOrder(PackageOrder):
         .. code-block:: yaml
 
            type: no_order
+           packages: ["foo"]
         """
-        return {}
+        return {
+            "packages": self.packages,
+        }
 
     @classmethod
     def from_pod(cls, data):
-        return cls()
+        return cls(packages=data.get("packages"))
 
 
 class SortedOrder(PackageOrder):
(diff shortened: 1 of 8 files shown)


src/rez/version/_version.py
```diff
@@ -41,7 +44,7 @@ def __ge__(self, other):
         return not self < other
 
     def __str__(self):
-        return "reverse(%s)" % str(self.value)
+        return "reverse(%r)" % self.value
```
JeanChristopheMorinPerso, line 47:
```suggestion
        return f"reverse({self.value!r})"
```
Micro-optimization, but considering how core this module is, this could actually improve the solver speed a little bit.

src/rez/package_order.py
```diff
@@ -294,12 +422,19 @@ def to_pod(self):
...
+        return dict(
+            first_version=str(self.first_version),
+            packages=self.packages,
+        )
 
     @classmethod
     def from_pod(cls, data):
-        return cls(Version(data["first_version"]))
+        return cls(
+            first_version=Version(data["first_version"]),
```
JeanChristopheMorinPerso, line 435:
```suggestion
            Version(data["first_version"]),
```

src/rez/package_order.py
```diff
@@ -40,11 +94,71 @@ def reorder(self, iterable, key=None):
...
+    def sort_key(self, package_name, version_like):
+        """Returns a sort key usable for sorting packages within the same family
+
+        Args:
+            package_name: (str) The family name of the package we are sorting
+            version_like: (Version|_LowerBound|_UpperBound|_Bound|VersionRange)
+                the version-like object you wish to generate a key for
+
+        Returns:
+            Comparable object
```
JeanChristopheMorinPerso, line 123: From what I see, the return value is a tuple in most cases, except when `version_like` is of type `Version`? If so, it's a little weird that it sometimes return a tuple and sometimes a Comparable instance.

src/rez/package_order.py
```diff
@@ -125,12 +253,19 @@ def to_pod(self):
...
+        return {
+            "descending": self.descending,
+            "packages": self.packages,
+        }
 
     @classmethod
     def from_pod(cls, data):
-        return cls(descending=data["descending"])
+        return cls(
+            descending=data["descending"],
```
JeanChristopheMorinPerso, line 266:
```suggestion
            data["descending"],
```

src/rez/package_order.py
```diff
@@ -147,26 +282,36 @@ def __init__(self, order_dict, default_order=None):
...
+    def sort_key_implementation(self, package_name, version):
+        orderer = self.order_dict.get(package_name)
+        if orderer is None:
+            if self.default_order is not None:
+                orderer = self.default_order
+            else:
+                # shouldn't get here, because applies_to should protect us...
+                raise RuntimeError(
+                    "package family orderer %r does not apply to package family %r",
+                    (self, package_name))
```
JeanChristopheMorinPerso, line 311:
```suggestion
            
            if self.default_order is None:
                # shouldn't get here, because applies_to should protect us...
                raise RuntimeError(
                    "package family orderer %r does not apply to package family %r",
                    (self, package_name))

            orderer = self.default_order
```
We could also potentially use an `assert`.

src/rez/package_order.py
```diff
@@ -354,73 +489,84 @@ def __init__(self, timestamp, rank=0):
...
 
-    def reorder(self, iterable, key=None):
+        # dictionary mapping from package family to the first-version-after
+        # the given timestamp
+        self._cached_first_after = {}
+        self._cached_sort_key = {}
+
+    def _get_first_after(self, package_family):
+        """Get the first package version that is after the timestamp"""
+        first_after = self._cached_first_after.get(package_family, KeyError)
```
JeanChristopheMorinPerso, line 503: Is `KeyError` used as a sentinel value?

src/rez/package_order.py
```diff
@@ -354,73 +489,84 @@ def __init__(self, timestamp, rank=0):
...
+        if first_after is KeyError:
+            first_after = self._calc_first_after(package_family)
+            self._cached_first_after[package_family] = first_after
+        return first_after
+
+    def _calc_first_after(self, package_family):
+        from rez.packages import iter_packages
```
JeanChristopheMorinPerso, line 510: Was there a reason to put that import here instead of at the top of the module?

src/rez/package_order.py
```diff
@@ -441,23 +587,34 @@ def to_pod(self):
...
+            timestamp=self.timestamp,
+            rank=self.rank,
+            packages=self.packages,
+        )
 
     @classmethod
     def from_pod(cls, data):
-        return cls(timestamp=data["timestamp"], rank=data.get("rank", 0))
+        return cls(
+            timestamp=data["timestamp"],
```
JeanChristopheMorinPerso, line 601:
```suggestion
            data["timestamp"],
```

src/rez/package_order.py
```diff
@@ -472,6 +629,58 @@ def singleton(cls):
         """Filter list as configured by rezconfig.package_filter."""
         return cls.from_pod(config.package_orderers)
 
+    @staticmethod
+    def _to_orderer(orderer: Union[dict, PackageOrder]) -> PackageOrder:
+        if not isinstance(orderer, PackageOrder):
```
JeanChristopheMorinPerso, line 634:
```suggestion
        if isinstance(orderer, dict):
```

src/rez/resolved_context.py
```diff
@@ -233,7 +233,7 @@ def __init__(self, package_requests, verbosity=0, timestamp=None,
         self.package_filter = (PackageFilterList.singleton if package_filter is None
                                else package_filter)
 
-        self.package_orderers = (
+        self.package_orderers = PackageOrderList(
```
JeanChristopheMorinPerso, line 236: Won't that create some kind of nested orderers?

src/rez/solver.py
```diff
@@ -426,13 +428,17 @@ def key(variant):
                 if not request.conflict:
                     req = variant.requires_list.get(request.name)
                     if req is not None:
-                        requested_key.append((-i, req.range))
+                        orderer = get_orderer(req.name, self.solver.package_orderers or {})
```
JeanChristopheMorinPerso, line 431:
```suggestion
                        orderer = get_orderer(req.name, orderers=self.solver.package_orderers or {})
```
also, is `or {}` really necessary?

src/rez/solver.py
```diff
@@ -426,13 +428,17 @@ def key(variant):
...
+                        range_key = orderer.sort_key(req.name, req.range)
+                        requested_key.append((-i, range_key))
                         names.add(req.name)
 
             additional_key = []
             for request in variant.requires_list:
                 if not request.conflict and request.name not in names:
-                    additional_key.append((request.range, request.name))
+                    orderer = get_orderer(request.name, self.solver.package_orderers)
```
JeanChristopheMorinPerso, line 439:
```suggestion
                    orderer = get_orderer(request.name, orderers=self.solver.package_orderers)
```

src/rez/solver.py
```diff
@@ -831,25 +838,21 @@ def sort_versions(self):
...
+
         if self.sorted:
             return
 
-        for orderer in (self.solver.package_orderers or []):
-            entries = orderer.reorder(self.entries, key=lambda x: x.package)
-            if entries is not None:
-                self.entries = entries
-                self.sorted = True
+        orderer = get_orderer(self.package_name, self.solver.package_orderers or {})
```
JeanChristopheMorinPerso, line 846:
```suggestion
        orderer = get_orderer(self.package_name, orderers=self.solver.package_orderers or {})
```
And same thing about `or {}`.

src/rez/resolved_context.py
isohedronpipeline: No, it casts the input `list` into a `PackageOrderList`.  Works the same way as doing something like:
``` python
>>> a = [1, 2, 3]
>>> list(list(list(list(a))))
[1, 2, 3]
```

src/rez/solver.py
isohedronpipeline: I don't think so.  It was in there originally and tried hard not to change things that weren't absolutely necessary for this.

src/rez/package_order.py
isohedronpipeline: Sorry, that docstring should probably read "Returns: Sortable object" as in some object that implements `__lt__`.  I would add it as a `Protocol`, but those were added into `typing` in python 3.8 and I'm sure we don't want to add a dependency on `typing_extensions` for just this and just in python 3.7.

The class would look something like this:
``` python
from typing import Protocol, TypeVar, runtime_checkable

@runtime_checkable
class SortableProtocol(Protocol):
    """A protocol to which all objects that can be sorted must adhere.
    `sorted()` uses the < operator to determine sort order, so anything
    that can be sorted must implement __lt__"""

    def __lt__(self, other: Any) -> bool:
        pass
```

Then we'd add a `SortableT = TypeVar("SortableT", bound=SortableProtocol)`

And then this class would return a `SortableT`

src/rez/package_order.py
isohedronpipeline: This is fine, but I do think that it is a bit easier to understand with the kwarg, as it originally in there.

src/rez/package_order.py
isohedronpipeline: Seems so!
I'll change that to a try/except.

src/rez/package_order.py
isohedronpipeline: Not a good one.

src/rez/package_order.py
isohedronpipeline: Is there a reason to remove the kwargs?

src/rez/package_order.py
isohedronpipeline: Is there a reason to remove the kwargs?

src/rez/package_order.py
isohedronpipeline: I think that ensuring that is a `PackageOrder` is going to fail faster than just converting dicts to `PackageOrder`.  Under normal circumstances, a dict would be passed in, but we care a lot more about it being a PackageOrder not a not-dict.  It also means that if there were some change in the future to the type of data we can pass in here, we can accomodate that logic in the `from_pod()`, rather than having to deal with it here too.

src/rez/package_order.py
isohedronpipeline: Yeah, assert would be fine.  What you've suggested is also fine though.

src/rez/package_order.py
isohedronpipeline: I've added some clarifying language

src/rez/package_order.py
JeanChristopheMorinPerso: The thing is that the first argument the class takes is a positional argument. passing a keyword argument instead could lead someone to think the arguments can be freely reordered. Passing a keyword argument for a positional argument is a bad practice.

src/rez/package_order.py
JeanChristopheMorinPerso: With a type hint, you are protected against future changes to types accepted for `orderer`. On top of that `from_pod` across the code base only accept dictionaries AFAIK and I doubt this will change in the future.

isohedronpipeline: That's fine.  What's the advantage to comparing it against a dict instead of `PackageOrder`?  It still seems less precise to me and less clear about what we're expecting to get from it at the end of the process.

src/rez/package_order.py
isohedronpipeline: What makes it positional? It has a default value (e.g. `SortedPackageOrder()` works fine). 

I've heard that it's bad practice to provide positional arguments to what are kwargs, but not the other way around, especially if it provides better readability.  If it were actually positional, then maybe I'd agree, but even then I think it's clearer as a consumer of the API about what is actually going on.

Obviously this is an academic thing since the key of the data is a strong hint, but it is certainly more readable to have it explicit about which data is slotted into which key.  Additionally, the data could be reordered freely and it would work fine.

Since the original code had those as kwargs, I'd lean towards leaving it as they were, since we lose nothing and gain a small amount of legibility and reliability for fewer changes.

I won't die on this hill though, so if you really want it, I won't push back on it any more.

src/rez/package_order.py
isohedronpipeline: Sorry, I thought your change had been added to the batch.  Merged in now.

src/rez/package_order.py
JeanChristopheMorinPerso: We are in the `SortedOrder` class, which has `def __init__(self, descending, packages=None):`. `descending` is positional.

isohedronpipeline: Oh, I see.  Quite right.

src/rez/rezconfig.py
```diff
@@ -452,6 +452,19 @@
 # rez-env python-3     python-3.7.4
 # ==================== =============
 #
+#
+# Package orderers will also apply to variants of packages.
+# Consider a package called `pipeline-1.0` which has the following variants:
+# [["python-2.7.4", "python-2.7.16", "python-3.7.4"]]
```
JeanChristopheMorinPerso, line 458:
```suggestion
# Consider a package "pipeline-1.0" which has the following variants:
# ``[["python-2.7.4", "python-2.7.16", "python-3.7.4"]]``
```

JeanChristopheMorinPerso (approved):
Thanks a lot for fixing this bug! flake8 is failing, but apart from that, I'm happy to get this merged in the next release.

PR #1689, opened 2024-03-17
Add releasing packages documentation

BryceGattis:
Added a new page for releasing packages since there doesn't seem to be any for the `rez-release` workflow yet.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,6 +1,9 @@
 # Build instructions
 
-To build the docs you must use Python 3.11.
+To build the docs you must install the following requirements:
+
+- GNU Make
+- Python 3.11
 
 To create a build environment run the following commands:
 ```python
--- a/docs/README.md
+++ b/docs/README.md
@@ -2,7 +2,7 @@
 
 To build the docs you must install the following requirements:
 
-- GNU Make
+- GNU Make (Linux + Mac only)
 - Python 3.11
 
 To create a build environment run the following commands:
@@ -12,7 +12,7 @@ source .venv/bin/activate
 pip install -r docs/requirements.txt
 
 cd docs
-make html
+make html # .\make html on Windows
 ```
 
 # Example Headers
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -15,6 +15,7 @@ Welcome to rez's documentation!
    getting_started
    basic_concepts
    building_packages
+   releasing_packages
    context
    variants
    guides/index
@@ -28,7 +29,6 @@ Welcome to rez's documentation!
    context_bundles
    suites
    managing_packages
-   releasing_packages
    caching
    pip
 
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -28,6 +28,7 @@ Welcome to rez's documentation!
    context_bundles
    suites
    managing_packages
+   releasing_packages
    caching
    pip
 
--- a/docs/source/releasing_packages.rst
+++ b/docs/source/releasing_packages.rst
@@ -7,12 +7,12 @@ using the :ref:`rez-release` tool. This tool performs the following actions:
 
 * All :doc:`actions <building_packages>` that the ``rez-build`` tool performs.
 * Executes all configured :data:`release_hooks`.
-* Executes all configured `release_vcs`.
+* Executes all configured :gh-rez:`src/rezplugins/release_vcs` plugins.
 
 When to release vs build
 ============================
 Packages *can* be installed to the :data:`release_packages_path` manually
-by running ``rez-build <release_packages_path>``, so why would you use
+by running :option:`rez-build -i --prefix /path/to/repo <rez-build --prefix>`, so why would you use
 ``rez-release``?
 
 Well, here are a few benefits of ``rez-release``:
@@ -21,7 +21,7 @@ Well, here are a few benefits of ``rez-release``:
 * Tests being run by with the :ref:`rez-test` tool can run specifically prior to release, ensuring that releases pass any configured tests first.
 * Automatic sanity checks to ensure local repo is ready for release.
 * Automatic VCS tagging.
-* Many :ref:`helpful package attributes <release-package-attributes>` added to the released `package.py` at build time.
+* Many :ref:`helpful package attributes <release-package-attributes>` are automatically added to the released package definition file.
 
 If you're working locally, it these additional steps and checks may slow you
 down, so it may be better to stick with :ref:`rez-build`.
--- a/docs/source/releasing_packages.rst
+++ b/docs/source/releasing_packages.rst
@@ -1,0 +1,13 @@
+==================
+Releasing packages
+==================
+
+Rez packages can be built and deployed to the :data:`release_packages_path`
+using the :ref:`rez-release` tool. This tool performs the following actions:
+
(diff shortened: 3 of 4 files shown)


BryceGattis:
I don't believe the `release_vcs` configuration setting is documented at all yet on the "configuring Rez" page. Can someone sanity check me here...?

BryceGattis:
Once the `release_vcs` setting is documented, I think it'd be nice to split out package build and package release configuration settings into 2 different sections, rather than just the [single section as it is now](http://localhost:63342/rez/docs/_build/html/configuring_rez.html#package-build-release) for clarity.

docs/README.md
```diff
@@ -1,6 +1,9 @@
 # Build instructions
 
-To build the docs you must use Python 3.11.
+To build the docs you must install the following requirements:
+
+- GNU Make
```
JeanChristopheMorinPerso, line 5: Not necessary on Windows

docs/source/index.rst
```diff
@@ -28,6 +28,7 @@ Welcome to rez's documentation!
    context_bundles
    suites
    managing_packages
+   releasing_packages
```
JeanChristopheMorinPerso, line 31: Considering that it's basically what someone needs to do after building a package, it should go in the "General" section. Releasing a package is not considered advanced, It's a basic operation.

docs/source/releasing_packages.rst
```diff
@@ -0,0 +1,13 @@
...
+==================
+
+Rez packages can be built and deployed to the :data:`release_packages_path`
+using the :ref:`rez-release` tool. This tool performs the following actions:
+
+* All :doc:`actions <building_packages>` that the `rez-build` tool performs.
+* Executes all configured :data:`release_hooks`.
+* Executes all configured `release_vcs`.
+
+This tool generally runs the same as :ref:`rez-build` otherwise. For more
```
JeanChristopheMorinPerso, line 12: Yes and no. It's true that it's somewhat similar, but there is still differences, for example, see https://rez.readthedocs.io/en/stable/package_definition.html#tests. There might also be other differences. For example, a release might result in stricter file permissions (though I don't remember clearly).

JeanChristopheMorinPerso (commented):
Thanks! I added some comments.

>is it possible to actually get hyper links to source code?

Yes, see https://github.com/AcademySoftwareFoundation/rez/blob/1d18bc267fb6e237b56e0297952be95d8e325542/docs/source/configuring_rez.rst?plain=1#L6

This is configured in https://github.com/AcademySoftwareFoundation/rez/blob/1d18bc267fb6e237b56e0297952be95d8e325542/docs/source/conf.py#L102-L104

JeanChristopheMorinPerso:
>I don't believe the release_vcs configuration setting is documented at all yet on the "configuring Rez" page. Can someone sanity check me here...?

Plugins settings are not yet documented. Doing so would require some work since settings are right now documented in multiple places. For example, technically the canonical source should be in https://github.com/AcademySoftwareFoundation/rez/blob/main/src/rezplugins/release_vcs/rezconfig, but it only documents and sets one default value and the rest is in https://github.com/AcademySoftwareFoundation/rez/blob/main/src/rez/rezconfig.py#L1201-L1242.

At the very least, if you want to have `plugins` shown as a section in https://rez.readthedocs.io/en/stable/configuring_rez.html#configuration-settings, you would have to remove [this empty line](https://github.com/AcademySoftwareFoundation/rez/blob/main/src/rez/rezconfig.py#L1200). I didn't do it because the result is ugly and not accurate.

If you are curious, the code that generates https://rez.readthedocs.io/en/stable/configuring_rez.html#configuration-settings is at https://github.com/AcademySoftwareFoundation/rez/blob/1d18bc267fb6e237b56e0297952be95d8e325542/docs/rez_sphinxext.py#L19-L173

docs/source/releasing_packages.rst
JeanChristopheMorinPerso: Basically, I would like to make it clear that that is advantages to use `rez-release`. Though, I have to say that I can't articulate what the advantages are. In other words, when someone reads this page, what will be their conclusion? Will it be that they don't need `rez-release` since it's similar to rez-build` or will it be confusion as to when to use one vs the other?

docs/README.md
BryceGattis: @JeanChristopheMorinPerso Ah, I see you have a `make.bat` that I didn't see the first time in the `docs` directory. I've been just using the `Makefile` on Windows. I'll update the documentation here to mention that you only need GNU Make on Linux/Mac, and that if you're on Windows, to just use the `make.bat` file.

docs/source/releasing_packages.rst
BryceGattis: @JeanChristopheMorinPerso Can you clarify what you differences you're referring to with the `tests` section of the `package_definition` page?

JeanChristopheMorinPerso: You can set `run_on` on a test to `pre_release` for example. This test would only run before a release.

docs/source/releasing_packages.rst
```diff
@@ -0,0 +1,27 @@
...
+using the :ref:`rez-release` tool. This tool performs the following actions:
+
+* All :doc:`actions <building_packages>` that the ``rez-build`` tool performs.
+* Executes all configured :data:`release_hooks`.
+* Executes all configured `release_vcs`.
+
+When to release vs build
+============================
+Packages *can* be installed to the :data:`release_packages_path` manually
+by running ``rez-build <release_packages_path>``, so why would you use
```
JeanChristopheMorinPerso, line 15:
```suggestion
by running :option:`rez-build -i --prefix /path/to/repo <rez-build --prefix>`, so why would you use
```

docs/source/releasing_packages.rst
```diff
@@ -0,0 +1,27 @@
...
+``rez-release``?
+
+Well, here are a few benefits of ``rez-release``:
+
+* The package will automatically go to the configured :data:`release_packages_path`, whereas :ref:`rez-build` will go to the :data:`local_packages_path` by default.
+* Tests being run by with the :ref:`rez-test` tool can run specifically prior to release, ensuring that releases pass any configured tests first.
+* Automatic sanity checks to ensure local repo is ready for release.
+* Automatic VCS tagging.
+* Many :ref:`helpful package attributes <release-package-attributes>` added to the released `package.py` at build time.
```
JeanChristopheMorinPerso, line 24:
```suggestion
* Many :ref:`helpful package attributes <release-package-attributes>` are automatically added to the released package definition file.
```

docs/source/releasing_packages.rst
```diff
@@ -0,0 +1,27 @@
+==================
+Releasing packages
+==================
+
+Rez packages can be built and deployed to the :data:`release_packages_path`
+using the :ref:`rez-release` tool. This tool performs the following actions:
+
+* All :doc:`actions <building_packages>` that the ``rez-build`` tool performs.
+* Executes all configured :data:`release_hooks`.
+* Executes all configured `release_vcs`.
```
JeanChristopheMorinPerso, line 10:
```suggestion
* Executes all configured :gh-rez:`src/rezplugins/release_vcs` plugins.
```
?

JeanChristopheMorinPerso (approved):
Thanks again @BryceGattis for your continuous efforts at improving our docs!

PR #1703, opened 2024-03-29
Add plugin documentation page

BryceGattis:
Need some help ironing this page out, but wanted to get a doc page started for this.

### DIFF
--- a/docs/source/plugins.rst
+++ b/docs/source/plugins.rst
@@ -47,10 +47,10 @@ Existing plugin types
 - :gh-rez:`src/rezplugins/build_process`
 - :gh-rez:`src/rezplugins/build_system`
 - :gh-rez:`src/rezplugins/command`
-- :gh-rez:`src/rezplugins/build_process/package_repository`
-- :gh-rez:`src/rezplugins/build_process/release_hook`
-- :gh-rez:`src/rezplugins/build_process/release_vcs`
-- :gh-rez:`src/rezplugins/build_process/shell`
+- :gh-rez:`src/rezplugins/package_repository`
+- :gh-rez:`src/rezplugins/release_hook`
+- :gh-rez:`src/rezplugins/release_vcs`
+- :gh-rez:`src/rezplugins/shell`
 
 Developing your own plugin
 ==========================
@@ -70,15 +70,51 @@ To make your plugin available to rez, you can install them directly under
 ``src/rezplugins`` (that's called a namespace package) or you can add
 the path to :envvar:`REZ_PLUGIN_PATH`.
 
-Extending rez
-------------
+Registering subcommands
+-----------------------
 
-Optionally, plugins can provide new ``rez`` commands.
+Optionally, plugins can provide new ``rez`` subcommands.
 
-# TODO: How do you provide new rez commands.
+To register a plugin and expose a new subcommand, the plugin module:
 
-Required file contents
-----------------------
+- MUST have a module-level docstring (used as the command help)
+- MUST provide a `setup_parser()` function
+- MUST provide a `command()` function
+- MUST provide a `register_plugin()` function
+- SHOULD have a module-level attribute `command_behavior`
+
+For example, a plugin named 'foo' and this is the ``foo.py``:
+
+.. code-block:: python
+   :caption: your_plugin_file.py
+
+   '''The docstring for command help, this is required.
+   '''
+   from rez.command import Command
+
+   command_behavior = {
+       "hidden": False,   # optional: bool
+       "arg_mode": None,  # optional: None, "passthrough", "grouped"
+   }
+
+   def setup_parser(parser, completions=False):
+       parser.add_argument("--hello", ...)
+
+   def command(opts, parser=None, extra_arg_groups=None):
+       if opts.hello:
+           print("world")
+
+   class CommandFoo(Command):
+       schema_dict = {}
+       @classmethod
+       def name(cls):
+           return "foo"
+
+   def register_plugin():
+       return CommandFoo
+
+Other required file contents
+----------------------------
 .. code-block:: python
    :caption: __init__.py
 
@@ -86,9 +122,3 @@ Required file contents
     __path__ = extend_path(__path__, __name__)
 
 
-
-.. code-block:: python
-   :caption: your_plugin_file.py
-
-    def register_plugin():
-        return YourPluginClass
--- a/docs/source/plugins.rst
+++ b/docs/source/plugins.rst
@@ -9,8 +9,8 @@ to their own repos in the future.
 
 The built-in plugins are located at :gh-rez:`src/rezplugins`.
 
-Existing plugins
-================
+Existing plugin types
+=====================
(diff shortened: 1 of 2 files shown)


docs/source/plugins.rst
```diff
@@ -0,0 +1,66 @@
...
+=======
+
+Plugins are extensions of rez that extend rez functionality.
+
+Plugins are currently bundled in the main rez repo, but will be split out
+to their own repos in the future.
+
+The plugins that exist are currently located at :gh-rez:`src/rezplugins`.
+
+Existing plugins
```
JeanChristopheMorinPerso, line 12: These are plugin types, not plugins.

docs/source/plugins.rst
```diff
@@ -0,0 +1,66 @@
...
+- :gh-rez:`src/rezplugins/build_process/shell`
+
+Developing your own plugin
+==========================
+
+Rez plugins require a specific folder structure as follows:
+
+.. code-block:: text
+
+    /plugin_name
```
JeanChristopheMorinPerso, line 30: This is a plugin type folder. Things inside that folder are the actual plugins.

docs/source/plugins.rst
```diff
@@ -0,0 +1,66 @@
+=======
+Plugins
+=======
+
+Plugins are extensions of rez that extend rez functionality.
+
+Plugins are currently bundled in the main rez repo, but will be split out
+to their own repos in the future.
+
+The plugins that exist are currently located at :gh-rez:`src/rezplugins`.
```
JeanChristopheMorinPerso, line 10:
```suggestion
The built-in plugins are located at :gh-rez:`src/rezplugins`.
```

docs/source/plugins.rst
```diff
@@ -0,0 +1,66 @@
+=======
+Plugins
+=======
+
+Plugins are extensions of rez that extend rez functionality.
```
JeanChristopheMorinPerso, line 5:
```suggestion
Rez is designed around the concept of plugins. Plugins can be used to extend rez's functionalities without modifying any of rez's source code.
```

docs/source/plugins.rst
```diff
@@ -0,0 +1,66 @@
...
+Plugin types
+------------
+
+There are two different plugin types in Rez.
+
+- Plugin
+- Extension
+
+Extensions differ from standard plugins in that they add a new command to rez's
+CLI.
```
JeanChristopheMorinPerso, line 50: I'm not too sure about this section. "extensions" are not special from what I know. We could explain what an extension is by saying that a plugin that adds a command is sometimes refereed to as an extension?

JeanChristopheMorinPerso (commented):
Thanks! It's already a great start. I left some comments.

Also, it would be nice to cover these things:

* How are plugins configured?
* How can we install an are out-of-tree plugin? (out-of-tree here means a plugin that lives outside rez's source code)
* How can we know which plugins are currently loaded? (is this even possible without using the API?)

docs/source/plugins.rst
aclark4life: Yeah I would drop "extension" entirely if possible. Plugins exist. You can do a lots of things with plugins, including adding `rez` commands. In fact, it's impossible to add a plugin without "extending" rez's functionality in some way so calling only plugins that add `rez` commands "extensions" is arguably inaccurate, or incorrect, by some folks' definition.

docs/source/plugins.rst
```diff
@@ -0,0 +1,66 @@
...
+
+Required file contents
+----------------------
+``__init__.py``
+
+.. code-block:: python
```
JeanChristopheMorinPerso, line 56:
```suggestion

.. code-block:: python
   :caption: __init__.py
```
I think the `caption` option will add a title to the code block.

docs/source/plugins.rst
```diff
@@ -0,0 +1,130 @@
...
+   shell               cmd         Windows Command Prompt (DOS) shell.                loaded
+   shell               gitbash     Git Bash (for Windows) shell                       loaded
+   shell               powershell  Windows PowerShell 5                               loaded
+   shell               pwsh        PowerShell Core 6+                                 loaded
+
+
+Configuring plugins
+===================
+
+# TODO: Write this part.
```
BryceGattis, line 48: Tried poking through some code but AFAIK you can't actually configure plugins in your own `rezconfig.py` outside of the repo structure, but maybe I'm missing something? Can you help me out here @JeanChristopheMorinPerso?

The whole idea that you have to configure plugin settings in two different places in the repo, the "plugin type" settings at the root of the repo and then specific plugin settings in the `rezplugins` folder is pretty odd to me.

JeanChristopheMorinPerso: You can configure plugins in your own rezconfig.py like so:

```python
plugins = {
    "filesystem": {}
}
```

JeanChristopheMorinPerso: See https://github.com/AcademySoftwareFoundation/rez/blob/main/src%2Frez%2Fconfig.py#L643-L648

docs/source/plugins.rst
```diff
@@ -0,0 +1,130 @@
...
+
+Plugins are currently bundled in the main rez repo, but will be split out
+to their own repos in the future.
+
+The built-in plugins are located at :gh-rez:`src/rezplugins`.
+
+Loaded plugins
+==============
+
+Currently loaded plugins can be queried by running ``rez -i``
```
JeanChristopheMorinPerso, line 15:
```suggestion
Currently loaded plugins can be queried by running :option:`rez -i`
```

docs/source/plugins.rst
```diff
@@ -0,0 +1,130 @@
...
+=======
+
+Rez is designed around the concept of plugins. Plugins can be used to extend rez's functionalities without modifying any of rez's source code.
+
+Plugins are currently bundled in the main rez repo, but will be split out
+to their own repos in the future.
+
+The built-in plugins are located at :gh-rez:`src/rezplugins`.
+
+Loaded plugins
```
JeanChristopheMorinPerso, line 12: View active plugins? List installed plugins?

docs/source/plugins.rst
```diff
@@ -0,0 +1,137 @@
+=======
+Plugins
+=======
+
+Rez is designed around the concept of plugins. Plugins can be used to extend rez's functionalities without modifying any of rez's source code.
+
+Plugins are currently bundled in the main rez repo, but will be split out
+to their own repos in the future.
+
+The built-in plugins are located at :gh-rez:`src/rezplugins`.
```
JeanChristopheMorinPerso, line 10:
```suggestion
Rez comes with built-in plugins that are located at :gh-rez:`src/rezplugins`. New plugins are encouraged to be developed out-of-tree (outside rez).
```
It is true that we want to maybe split them out (at least some of them), but we didn't make any decision yet on this.

JeanChristopheMorinPerso (approved):
I'm quite happy with the result. Thanks again @BryceGattis!

PR #1707, opened 2024-04-03
Make gh-rez external link dynamic

### LINKED ISSUE #1690: Change gh-rez URL based on source of build
Our Sphinx docs are setup in a way that allows us to use `:gh-rez:`path/to.file` to link to a file in the repo.

https://github.com/AcademySoftwareFoundation/rez/blob/1d18bc267fb6e237b56e0297952be95d8e325542/docs/source/conf.py#L103

This is great, but the URL is hardcoded to one branch. We should make it flexible and instead use a logic like:
```
blob_ref = "main"
if os.environ.get("READTHEDOCS"):
    if os.environ["READTHEDOCS_VERSION_TYPE"] == "external":
        blob_ref = os.environ["READTHEDOCS_GIT_COMMIT_HASH"]
    else:
        blob_ref = os.environ["READTHEDOCS_GIT_IDENTIFIER"]

gh_rez_url = f"https://github.com/AcademySoftwareFoundation/rez/blob/{blob_ref}/%s"

extlinks = {
    'gh-rez': (gh_rez_url, '%s'),
}
```

See https://docs.readthedocs.io/en/stable/reference/environment-variables.html.

BryceGattis:
Closes #1690. Applies suggested fix from @JeanChristopheMorinPerso.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -99,8 +99,17 @@ autodoc_default_options = {
 # -- Options for extlinks extension -----------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/extensions/extlinks.html
 
+blob_ref = "main"
+if os.environ.get("READTHEDOCS"):
+    if os.environ["READTHEDOCS_VERSION_TYPE"] == "external":
+        blob_ref = os.environ["READTHEDOCS_GIT_COMMIT_HASH"]
+    else:
+        blob_ref = os.environ["READTHEDOCS_GIT_IDENTIFIER"]
+
+gh_rez_url = f"https://github.com/AcademySoftwareFoundation/rez/blob/{blob_ref}/%s"
+
 extlinks = {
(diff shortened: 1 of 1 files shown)


BryceGattis:
@JeanChristopheMorinPerso Question though: This change only affects the URL in the PR right? It seems to still point just to 'main' when I do a local Sphinx build. Might be something nice to note as a comment if so.

JeanChristopheMorinPerso:
@BryceGattis do you think we should use `file://...` when not building on RTD and if the `CI` environment is not set? (`CI` is a commonly set environment variable by most CI systems).

BryceGattis:
@JeanChristopheMorinPerso 

The `READTHEDOCS` environment variable is only set when we push to GitHub right? Just confirming.

I'm unsure exactly what I would need to put in the config file going down the `file://...` approach as you mentioned.

JeanChristopheMorinPerso (approved):
I think we can go with this and change later if needed.

Issue #1668, opened 2024-02-25
Unvendor dependencies


JeanChristopheMorinPerso:

maxnbk:
I'm not _entirely_ on board with this idea (But hear me out).

I think a valid alternative might be looking into an automated approach to vendoring dependencies, such as using the https://pypi.org/project/vendoring/ project. (I am not saying this exactly, but, some sort of similar automated approach). I could imagine a setup where the dependencies are still shipped as vendored (provided that licensing is not a problem as it isn't a problem now), but that it is easier to automatically apply updates to the latest things automatically, automatically test that nothing is broken, etc. The same caveats of "doing this carefully" apply, and I do agree we should remove any/every vendored dep that we don't actually need, but I don't necessarily think that moving to a "user pip-installs things" model is 100% the best approach.

JeanChristopheMorinPerso:
Thanks for the input @maxnbk!

I'd be very interested in understanding what makes you hesitate. You probably wrote about this in the past, but I can't remember your arguments. When you have time, can you describe what value you see in keeping our dependencies vendored please? I did try to add some pros in the issue description, but I'm sure I missed some.

It might be one of these decisions that will require us to prob the community to get a good sense of what they want. I could even take some time to write a proper REP with lots of details and clear details on how things would work for our users.

>I think a valid alternative might be looking into an automated approach to vendoring dependencies, such as using the https://pypi.org/project/vendoring/ project.

I thought about using `vendoring` and I think it would work. But it wouldn't remove us the burden of keeping things up to date. We'd have to come up with a way to automatically monitor the vendored dependencies and update them. That could be achieved by creating a `requirements.txt` file and feed it to vendoring. With that `requirements.txt` file, we could instruct dependabot to create PRs when newer versions are available.  The missing piece is how to actually trigger the re-vendoring process from there. I'm not sure how we could do that.

I'm still not sold to the idea though. It just adds us more work and things to maintain. Another argument for unvendoring is that we get support for newer python versions for free.

>Edit: And, in the case that we decide to nix the easy "install rez behind studio firewall" capabilities, we should be very careful to explicitly state what approach we expect users/studios to take who are in that scenario, and vet that the approach works. While it was never an explicitly stated "compatibility guarantee", it was a fairly big implicit compatibility promise that rez made for years and years, and we shouldn't necessarily abandon it without explicit alternatives laid out.

I agree that if we go that route, we'll need to better document how to deploy rez.

[...]

PR #1714, opened 2024-04-09
distlib parses version string instead of using version_info to make script names

cfxegbert:
Use version_info to determine python version when building scripts

### DIFF
--- a/src/rez/vendor/distlib/scripts.py
+++ b/src/rez/vendor/distlib/scripts.py
@@ -285,9 +285,9 @@ class ScriptMaker(object):
         if '' in self.variants:
             scriptnames.add(name)
         if 'X' in self.variants:
-            scriptnames.add('%s%s' % (name, sys.version[0]))
+            scriptnames.add('%s%d' % (name, sys.version_info[0]))
         if 'X.Y' in self.variants:
-            scriptnames.add('%s-%s' % (name, sys.version[:3]))
+            scriptnames.add('%s-%d.%d' % (name, *sys.version_info[0:2]))
         if options and options.get('gui', False):
             ext = 'pyw'
         else:


cfxegbert:
When you install rez with install.py it creates the executable and the executable with the python minor version number
```
-rwxr-xr-x@ 1 root  wheel  232 Mar 13 10:42 rez
-rwxr-xr-x@ 1 root  wheel  232 Mar 13 10:42 rez-3.9
-rwxr-xr-x@ 1 root  wheel  252 Mar 13 10:42 rez-benchmark
-rwxr-xr-x@ 1 root  wheel  252 Mar 13 10:42 rez-benchmark-3.9
-rwxr-xr-x@ 1 root  wheel  242 Mar 13 10:42 rez-bind
-rwxr-xr-x@ 1 root  wheel  242 Mar 13 10:42 rez-bind-3.9
...
```
If you install with python-3.10 or later you will end up the files like `rez-3.1` instead of `rez-3.10`.

The vendor library has already been modified by others so it is already divergent from the original source.

cfxegbert:
It looks like a similar fix has be made upstream.

cfxegbert:
There is also https://github.com/AcademySoftwareFoundation/rez/issues/1668.

If we update distlib to the latest/greatest do we have the tests in place to verify it?  Looks like distlib is used in `install.py`, `src/rez/utils/pip.py`, `src/rez/pip.py`, `src/rez/bind/hello_world.py` and `src/rez/tests/test_pip_utils.py`

cfxegbert:
> @cfxegbert, I see in the vendor folder's `README.md` [file](https://github.com/AcademySoftwareFoundation/rez/blob/main/src/rez/vendor/README.md) that the description is
> 
> "Updated (June 2019) to enable wheel distribution based installations."

Changes that are not upstream:
```
util.py
a195feaba (Allan Johns       2021-06-16 15:18:33 +1000  218)
de4737f98 (Stephen Mackenzie 2022-08-28 17:39:30 -0400  219)                         # https://github.com/AcademySoftwareFoundation/rez/pull/1092
8412dc1a2 (Blazej Floch      2021-06-03 17:10:56 -0400  220)                         # Some packages have a trailing comma which would break the matching
8412dc1a2 (Blazej Floch      2021-06-03 17:10:56 -0400  221)                         if not ver_remaining:
8412dc1a2 (Blazej Floch      2021-06-03 17:10:56 -0400  222)                             break
a195feaba (Allan Johns       2021-06-16 15:18:33 +1000  223)                         # /end pull/1092
a195feaba (Allan Johns       2021-06-16 15:18:33 +1000  224)
```

JeanChristopheMorinPerso:
Closing this since I don't think that it's the right way to fix this.

PR #1723, opened 2024-04-15
Fix get_plugins() bug

### LINKED ISSUE #1618: Error searching for plugins if there is an empty package folder in the repo
Calling `rez.package_search.get_plugins` with a package that has `has_plugins` set to True will error out if there is a folder in a package repo without a valid package in it. Simplest way to replicate this is just make an empty folder in ~/packages and try to run get_plugins:
```
>>> package_search.get_plugins("nuke")
Searching [#############                   ] 46/106Traceback (most recent call last):
  File "/opt/rez/rez_package_cache/rez/2.112.0/c2ee/a/rez/package_search.py", line 154, in get_plugins
    if not plugin_pkg.plugin_for:
AttributeError: 'NoneType' object has no attribute 'plugin_for'
```
I actually came across this while in the middle of building a new package that took some time, so it had created the package folder with a version and `.building` file but there was not a valid package in there yet and started to get reports of our launcher breaking.

Fix may be as simple as adding a check for None here:
https://github.com/AcademySoftwareFoundation/rez/blob/7e8617d05aaf9e0e16257f7801857feb6e8f1c5c/src/rez/package_search.py#L154-L156

```py
    if not plugin_pkg or not plugin_pkg.plugin_for:
```

BryceGattis:
Add suggested fix from @mmdanggg2 and add tests.

Closes #1618.

### DIFF
--- a/src/rez/package_search.py
+++ b/src/rez/package_search.py
@@ -12,6 +12,7 @@ that do not provide an implementation.
 
 import fnmatch
 from collections import defaultdict
+from typing import List
 import sys
 
 from rez.packages import iter_package_families, iter_packages, get_latest_package
@@ -126,7 +127,7 @@ def get_reverse_dependency_tree(package_name, depth=None, paths=None,
     return pkgs_list, g
 
 
-def get_plugins(package_name, paths=None):
+def get_plugins(package_name, paths=None) -> List[str]:
     """Find packages that are plugins of the given package.
 
     Args:
@@ -152,7 +153,7 @@ def get_plugins(package_name, paths=None):
             continue  # not a plugin of itself
 
         plugin_pkg = get_latest_package(package_name_, paths=paths)
-        if not plugin_pkg.plugin_for:
+        if not plugin_pkg or not plugin_pkg.plugin_for:
             continue
         for plugin_for in plugin_pkg.plugin_for:
             if plugin_for == pkg.name:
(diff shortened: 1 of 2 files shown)


src/rez/package_search.py
```diff
@@ -152,7 +153,7 @@ def get_plugins(package_name, paths=None):
             continue  # not a plugin of itself
 
         plugin_pkg = get_latest_package(package_name_, paths=paths)
-        if not plugin_pkg.plugin_for:
+        if not plugin_pkg or not plugin_pkg.plugin_for:
```
JeanChristopheMorinPerso, line 156: I'm wondering if we should instead make sure that `get_latest_package` returns valid packages. The docstring of `get_latest_package` says that it should either return a `Package` object or None is a package is not found. There is also this TODO https://github.com/AcademySoftwareFoundation/rez/blob/71d990bc63cd9314114e45c2002fa15656a778b0/src/rez/packages.py#L925-L926

It my books, that's what I call smell.

mmdanggg2: In the case of #1618 I would say that get_latest_package is correctly returning None, as there is no valid package in an empty folder. Perhaps then iter_package_families should be more thorough and not return package names if there is no valid package in the folders?

ben-hawkyard-absolute:
Hey folks! Just wanted to check in and see if there might be movement on this soon? We've encountered this issue a couple of times in the wild in the last six months or so, and would be grateful for a fix!

PR #1724, opened 2024-04-15
Add new default_build_process setting to set the default build process

predat:
This PR allows the user to define the name of the build process plugin to be used for build/release by defining it in the rezconfig or through an environment variable without having to specify it from the rez-build/rez-release command line tools.

### DIFF
--- a/src/rez/rezconfig.py
+++ b/src/rez/rezconfig.py
@@ -948,6 +948,9 @@ max_package_changelog_revisions = 0
 #       launched without extension from windows and other systems.
 create_executable_script_mode = "single"
 
+# Default build process to use during build/release
+build_process = "local"
+
 # Configurable pip extra arguments passed to the :ref:`rez-pip` install command.
 # Since the :ref:`rez-pip` install command already includes some pre-configured
 # arguments (``--target``, ``--use-pep517``) this setting can potentially override the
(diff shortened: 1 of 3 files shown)


src/rez/rezconfig.py
```diff
@@ -948,6 +948,9 @@
 #       launched without extension from windows and other systems.
 create_executable_script_mode = "single"
 
+# Default build process to use during build/release
```
JeanChristopheMorinPerso, line 951: Should the valid values be documented? We could also make a link to `--process`.

src/rez/rezconfig.py
```diff
@@ -948,6 +948,9 @@
...
+build_process = "local"
```
JeanChristopheMorinPerso, line 952: We have `default_shell` which sets the default shell plugin to use. Should this setting be named `default_build_process` for consistency?

JeanChristopheMorinPerso (commented):
I never realized we couldn't configure the default process! I think I'm ok with the idea of setting a default like this. I left some opened questions.

src/rez/rezconfig.py
predat: I've added a small comment on the build processes currently available.

predat:
This PR is trivial; do you think you can merge it ?

JeanChristopheMorinPerso (approved):

JeanChristopheMorinPerso:
Sorry for the delays on this. It looks good to me!

PR #1731, opened 2024-04-24
Update vendor packaging

BryceGattis:
Update to newer type-hinted packaging vendor package.

This is a "no patch" update where the original source code is completely unmodified outside of modifying necessary imports to point to their `rez.vendor` equivalents.


brycegbrazen:
Looks like a few builds failed temporarily due to some temporary issue with Microsoft packages...? https://github.com/orgs/community/discussions/120966

PR #1734, opened 2024-04-26
Rez pip error handling improvements

BryceGattis:
Add error handling for failed resolve and improve error handling if python package not found in resolve

### DIFF
--- a/src/rez/pip.py
+++ b/src/rez/pip.py
@@ -7,7 +7,7 @@ from rez.version import Version
 from rez.vendor.distlib.database import DistributionPath
 from rez.vendor.packaging.version import Version as PackagingVersion
 from rez.vendor.packaging.specifiers import Specifier
-from rez.resolved_context import ResolvedContext
+from rez.resolved_context import ResolvedContext, ResolverStatus
 from rez.utils.execution import Popen
 from rez.utils.pip import get_rez_requirements, pip_to_rez_package_name, \
     pip_to_rez_version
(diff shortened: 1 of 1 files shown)


src/rez/pip.py
```diff
@@ -134,11 +132,19 @@ def find_python_in_context(context):
     # Create a copy of the context with systems paths removed, so we don't
     # accidentally find a system python install.
     #
+    success = (context.status == ResolverStatus.solved)
+    if not success:
```
JeanChristopheMorinPerso, line 136:
```suggestion
    if context.status != ResolverStatus.solved:
```

src/rez/pip.py
```diff
@@ -134,11 +132,19 @@ def find_python_in_context(context):
...
+        context.print_info(buf=sys.stderr)
+        return None
+
     context = context.copy()
     context.append_sys_path = False  # #826
 
     python_package = context.get_resolved_package("python")
-    assert python_package
+    if not python_package:
```
JeanChristopheMorinPerso, line 144: Will this condition ever be true? If we resolved python and the solve was successful, `context.get_resolved_package("python")` should always return a python package no?

src/rez/pip.py
```diff
@@ -134,11 +132,19 @@ def find_python_in_context(context):
     # Create a copy of the context with systems paths removed, so we don't
     # accidentally find a system python install.
     #
+    success = (context.status == ResolverStatus.solved)
+    if not success:
+        context.print_info(buf=sys.stderr)
```
JeanChristopheMorinPerso, line 137: Should we add a custom error message too? Something like "Failed to resolve ... "

PR #1735, opened 2024-04-27
Add more logic in the shell detection code to handle a case where the parent PID is zero

### LINKED ISSUE #1732: "process ID out of range" error when using rez-env from within a Docker container with namespaced PIDs
At my studio we have a Rocky image that resembles our workstation configuration and certain pipeline services use rez-env within containers to run some things.

I noticed while transitioning to use Rez in my containerized environment that it seems unhappy about namespaced PIDs, which is the default when you docker-run.

**Environment**
* OS: Rocky 8.9 in Docker
* Rez version: 2.112.0
* Rez python version: Python 3.9.18

**To Reproduce**

I prepped a minimalist Dockerfile to reproduce the issue. Save this `Dockerfile` below:
```Dockerfile
FROM rockylinux:8.9

# Install python, bash, wget, locale stuff, lsb_release (for Rez) and do any security updates:
RUN yum install -y python3.9 bash glibc-langpack-en wget redhat-lsb-core && yum update -y && yum clean all
# Symlink "python3" as "python" so that "rez-bind --quickstart" does not complain:
RUN ln -s /usr/bin/python3 /usr/bin/python
# Set locale to English:
ENV LANG en_US.UTF-8
ENV LC_ALL en_US.UTF-8
# Python things:
ENV PYTHONDONTWRITEBYTECODE=1
ENV PYTHONFAULTHANDLER=1
ENV PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring

# Install Rez:
ARG REZ_VERSION=2.112.0
ENV REZ_VERSION=${REZ_VERSION}
ENV REZ_INSTALL_ROOT=/usr/local/rez
RUN wget -O /tmp/rez.tar.gz https://github.com/AcademySoftwareFoundation/rez/archive/refs/tags/${REZ_VERSION}.tar.gz && \
    mkdir --parents --mode=777 /tmp/rez && \
    tar -xzvf /tmp/rez.tar.gz -C /tmp/rez/ && \
    python3 /tmp/rez/rez-${REZ_VERSION}/install.py $REZ_INSTALL_ROOT/ && \
    rm -rf /tmp/rez /tmp/rez.tar.gz && \
    echo -e "export PATH=\$PATH:${REZ_INSTALL_ROOT}/bin/rez\nsource ${REZ_INSTALL_ROOT}/completion/complete.sh" >> /etc/profile.d/fps.sh
RUN $REZ_INSTALL_ROOT/bin/rez/rez-bind --quickstart
ENV PATH "$PATH:$REZ_INSTALL_ROOT/bin/rez"
```
then `docker build . --tag=rocky_rez` and brew some tea or coffee cause it'll take a minute or two.

To see the issue do any rez-env. Since this is a vanilla environment, use the python package made by the quickstarter:
```shell
$ docker run --rm -it rocky_rez rez-env python -- python --version
error: process ID out of range

Usage:
 ps [options]

 Try 'ps --help <simple|list|output|threads|misc|all>'
  or 'ps --help <s|l|o|t|m|a>'
 for additional help text.

For more details see ps(1).
Python 3.9.18
```

As you can see, it does run the thing so it is kind of a harmless error, just scary looking.

One way to get around it is to pass `--pid=host` so that PID namespace isolation is no longer in place, thus the pids are shared with the host's, but this isn't great because it weakens the security of the container, as it is no longer process-isolated:

```shell
$ docker run --rm -it --pid=host rocky_rez rez-env python -- python --version
Python 3.9.18
```
...but it runs without an error.

I have a suspicion it is probably due to this command in the codebase here:
https://github.com/AcademySoftwareFoundation/rez/blob/f17e88dad283826a5914e3aade4fbc9a010cdba2/src/rez/system.py#L84
and I guess it's because by default a container runs as PID 1, without a parent.

Could it be detected in the shell code so it does not print a scary error?

**Expected behavior**

I would expect this error-free output:
```shell
$ docker run --rm -it rocky_rez rez-env python -- python --version
Python 3.9.18
```

**Actual behavior**

```
$ docker run --rm -it rocky_rez rez-env python -- python --version
error: process ID out of range

Usage:
 ps [options]

 Try 'ps --help <simple|list|output|threads|misc|all>'
  or 'ps --help <s|l|o|t|m|a>'
 for additional help text.

For more details see ps(1).
Python 3.9.18
```

JeanChristopheMorinPerso:
Fixes #1732 

Add more logic in the shell detection code to handle a case where the parent PID is zero. This can happen when run from within a docker container with a non-interactive shell.

This was not necessarily a hard error. rez would just print an annoying error.

Before:
```
$ docker run --rm -it rocky_rez rez-env python -- python --version
os.getpid() 0
error: process ID out of range

Usage:
 ps [options]

 Try 'ps --help <simple|list|output|threads|misc|all>'
  or 'ps --help <s|l|o|t|m|a>'
 for additional help text.

For more details see ps(1).
Python 3.9.18
```

After:
```
$ docker run --rm -it rocky_rez rez-env python -- python --version
Python 3.9.18
```

### DIFF
--- a/src/rez/system.py
+++ b/src/rez/system.py
@@ -79,14 +79,18 @@ class System(object):
             import subprocess as sp
             shell = None
 
-            # check parent process via ps
-            try:
-                args = ['ps', '-o', 'args=', '-p', str(os.getppid())]
-                proc = sp.Popen(args, stdout=sp.PIPE)
-                output = proc.communicate()[0]
-                shell = os.path.basename(output.strip().split()[0]).replace('-', '')
-            except Exception:
-                pass
+            parent_pid = os.getppid()
+            if parent_pid != 0:
+                # When run from inside docker without an interactive shell,
+                # the parent pid will be 0, which will cause "ps" to
+                # print an error message: "process ID out of range".
+                try:
+                    args = ['ps', '-o', 'args=', '-p', str(parent_pid)]
+                    proc = sp.Popen(args, stdout=sp.PIPE)
+                    output = proc.communicate()[0]
+                    shell = os.path.basename(output.strip().split()[0]).replace('-', '')
+                except Exception:
+                    pass
 
             # check $SHELL
             if shell not in shells:
@@ -94,7 +98,7 @@ class System(object):
 
             # traverse parent procs via /proc/(pid)/status
             if shell not in shells:
-                pid = str(os.getppid())
+                pid = str(parent_pid)
                 found = False
 
                 while not found:


darkvertex:
I confirm this resolved #1732 for me.

ps: Someone needs to check out the macos py3.7 CI environment. It seems pretty broken.

JeanChristopheMorinPerso:
Thanks @darkvertex for confirming! As for the macos builds, we'll have to fix them. The failures are caused by the new default `macos-latest`, which is now `macos-14` which themselves are M1 runners. And 3.7 was not compatible with macOS arm64.

Anyway, I'll probably fix the macos thing in another PR.

PR #1737, opened 2024-04-29
Add package orderers documentation

BryceGattis:
Looking for some feedback on what exactly this page should look like. Some examples of the configuration are on the configuring rez page, but I'm not sure if we want the same kind of info here or not.


docs/source/package_orderers.rst
```diff
@@ -0,0 +1,86 @@
+================
+Package Orderers
+================
+
+Rez's default version resolution algorithm will always sort by the latest alphanumeric
```
JeanChristopheMorinPerso, line 5: Should we reference https://rez.readthedocs.io/en/stable/basic_concepts.html#versions?

JeanChristopheMorinPerso, line 5: Should we reference https://rez.readthedocs.io/en/stable/basic_concepts.html#versions? I would also like to keep some parts of https://rez--1737.org.readthedocs.build/en/1737/configuring_rez.html#package_orderers if possible.

docs/source/package_orderers.rst
```diff
@@ -0,0 +1,86 @@
...
+Configuring package orderers
+============================
+
+Package ordering settings are not configured in each package's ``package.py``. Instead they are configured in the
+``rezconfig.py`` via the :data:`package_orderers` setting.
+
+Sorted Order
+------------
+
+This is the default orderer the sorts based on the package's :attr:`version` attribute.
```
JeanChristopheMorinPerso, line 18:
```suggestion
This is the default orderer that sorts based on the package's :attr:`version` attribute.
```

docs/source/package_orderers.rst
```diff
@@ -0,0 +1,86 @@
+================
+Package Orderers
+================
+
+Rez's default version resolution algorithm will always sort by the latest alphanumeric
+version. However, package orderers allow you to customize this functionality globally,
+or at a per package level.
+
+Configuring package orderers
+============================
```
JeanChristopheMorinPerso, line 10:
```suggestion
```
I don't think we need a section for configuring.

docs/source/package_orderers.rst
```diff
@@ -0,0 +1,86 @@
...
+
+    package_orderers = [
+        {
+            "type": "sorted", # Required
+            "descending": True, # Required
+            "packages": ["python"] # Optional, if not supplied, orderer applies to all packages
+        }
+    ]
+
+Per Family Order
```
JeanChristopheMorinPerso, line 32: I think I would move this to the end if we are to mention `version_split`.

docs/source/package_orderers.rst
```diff
@@ -0,0 +1,86 @@
...
+----------------
+
+This orderer allows you to define different orderers to different package families.
```
JeanChristopheMorinPerso, line 35: We should document that `orderers` can be any supported orderer.

JeanChristopheMorinPerso (commented):
It's a good start. I made small comments that shouldn't be too difficult to address.

We should also be careful to not drop or have duplicated information with https://rez--1737.org.readthedocs.build/en/1737/configuring_rez.html#package_orderers.

docs/source/package_orderers.rst
BryceGattis: @JeanChristopheMorinPerso Do you mean we don't need this section because it already exists on the `Configuring Rez` page? I'm not really adding new content here, just cross referencing to that page.

docs/source/package_orderers.rst
BryceGattis: @JeanChristopheMorinPerso Can you explain a bit more what you mean?

docs/source/package_orderers.rst
BryceGattis: @JeanChristopheMorinPerso I added the cross reference to https://rez.readthedocs.io/en/stable/basic_concepts.html#versions.

What do you mean you'd like to keep some parts of https://rez--1737.org.readthedocs.build/en/1737/configuring_rez.html#package_orderers. Can you be more specific?

docs/source/package_orderers.rst
BryceGattis: @JeanChristopheMorinPerso Can do, but why is this? I'm still trying to wrap my head around the architecture of this code.

JeanChristopheMorinPerso:
@BryceGattis I pushed some changes. When you have a minute, could you take a look and tell me if what I wrote is clear please?

BryceGattis:
@JeanChristopheMorinPerso The changes you made were great thank you. The one lingering question I have is that I'm still not fully understanding the purpose of the `per_family` orderer. If every other type of orderer (ex. sorted), can have a `packages` attribute (and therefore can filter by package families), what is the purpose of the `per_family` orderer?

docs/source/package_orderers.rst
```diff
@@ -0,0 +1,243 @@
...
+
+.. note::
+
+   Implementing a custom orderer should only be done if absolutely necessary.
+   It could make your environment behave in very special ways and more importantly
+   in non expected ways from a user perspective. It can also make it harder to share
+   the set of affected packages to others.
+
+
+.. code-block:: python
```
BryceGattis, line 203: @JeanChristopheMorinPerso To clarify, the definition of the `MyOrderer` class here should NOT be in the `rezconfig.py` file right? This requires two different `.py` files doesn't it? If so we should probably clarify that.

BryceGattis:
@JeanChristopheMorinPerso What do you think about having a glossary for rez? I was just thinking that a clickable link for `package family` would be nice. Perhaps the definition of a package family could just live in the `Getting started` or `Basic Concepts` pages though instead of making a new glossary page.

docs/source/package_orderers.rst
JeanChristopheMorinPerso: I should be in the rezconfig.

PR #1740, opened 2024-05-01
Introducing "testing" object

fabal:
Introducing a `testing` object available in late bindings, similar to the `building` object, but indicating a `rez-test` context.


docs/source/package_commands.rst
```diff
@@ -376,6 +376,20 @@ Following is a list of the objects and functions available.
...
+
+   This boolean variable is ``True`` if a test is occurring (typically done via the :ref:`rez-test` tool),
+   and ``False`` otherwise.
+   
+   Typically a package will use this variable to set environment variables that are only relevant during test
+   execution.
+
+   .. code-block:: python
+
+      if building:
```
fnaum, line 390:
```suggestion
      if testing:
```

docs/source/package_commands.rst
```diff
@@ -376,6 +376,20 @@ Following is a list of the objects and functions available.
...
+   This boolean variable is ``True`` if a test is occurring (typically done via the :ref:`rez-test` tool),
+   and ``False`` otherwise.
+   
+   Typically a package will use this variable to set environment variables that are only relevant during test
+   execution.
+
+   .. code-block:: python
+
+      if testing:
+         env.FOO_TEST_DATA_PATH = "{root}/tests/data"
```
JeanChristopheMorinPerso, line 391:
```suggestion
          env.FOO_TEST_DATA_PATH = "{root}/tests/data"
```

JeanChristopheMorinPerso:
Just by looking at this PR, I'm struggling to see if it also applies to early bound functions. The description says that it's only for `@late`, but we have `building` in `@early`.

If the changes inject `testing` in `@early`, then the documentation should be adapted as such. If it doesn't then I think we should change the PR so that `testing` will also be available in `@early`.

Additionally, I would like to see a test or two.

Thanks!

JeanChristopheMorinPerso (commented):
Just by looking at this PR, I'm struggling to see if it also applies to early bound functions. The description says that it's only for @late, but we have building in @early.

If the changes inject testing in @early, then the documentation should be adapted as such. If it doesn't then I think we should change the PR so that testing will also be available in @early.

Additionally, I would like to see a test or two.

Thanks!

fabal:
I think early bound attributes are not relevant to `rez-test` scenarios because they're flattened upon install, meaning any will go anyway. Unless I'm missing something?

I thought about adding some but I couldn't find counterparts for `building` use cases? Unless there aren't any?

fnaum:
I also think that `testing` does not make sense in @early, as we early decorated functions are evaluated at build time.
so even if we add the object there it would always be false.

JeanChristopheMorinPerso:
Oh, yeah, you are right about `@early`, by bad.

As for the tests, I don't think there is any for `building` and we also don't have tests for `rez-test`. So it could be a good occasion to add some rez-tests tests.

fabal:
Sounds good, we've carved out some time to add a few unit tests.

src/rez/tests/test_test.py
```diff
@@ -0,0 +1,55 @@
...
+            warn_untimestamped=False,
+            resolve_caching=False
+        )
+    @classmethod
+    def tearDownClass(cls):
+        TempdirMixin.tearDownClass()
+
+    def test_1(self):
+        """package.py unit tests are correctly run in a testing environment"""
+        context = ResolvedContext(["testing_obj"])
```
JeanChristopheMorinPerso, line 30:
```suggestion
        self.inject_python_repo()
        context = ResolvedContext(["testing_obj"])
```
This will add a usable `python` package and make it available to the tests (only this test). Without this, the tests won't be portable across platforms because we can't rely on python being available on the system. We need python to run the test suite, but that doesn't mean that calling `python` should work because it doesn't have to be on `PATH`, and by experience, it will sometimes not be on PATH.

The magic happens in https://github.com/AcademySoftwareFoundation/rez/blob/9094b089e4bec6048a12cf0e3d105af5935c8758/src/rez/cli/selftest.py#L89-L93

JeanChristopheMorinPerso (changes requested):
Looks great! The tests are failing, but I added instructions on how to make them pass.

Also, you will need to run the `./src/build_utils/license/apply_copyright` script and fix flake8 warnings.

src/rez/tests/test_test.py
```diff
@@ -0,0 +1,57 @@
...
+            warn_untimestamped=False,
+            resolve_caching=False
+        )
+    @classmethod
+    def tearDownClass(cls):
+        TempdirMixin.tearDownClass()
+
+    def test_1(self):
+        """package.py unit tests are correctly run in a testing environment"""
+        self.inject_python_repo()
```
JeanChristopheMorinPerso, line 31: This needs to be done on each test (or inside setUp).

JeanChristopheMorinPerso (changes requested):

JeanChristopheMorinPerso:
@fabal @fnaum @nca45 can one of you enable the "allow maintainers to modify the PR" please?  I'll take care of the failing tests.

fabal:
I can't seem to see or have access to this option @JeanChristopheMorinPerso

src/rez/tests/test_context.py
```diff
@@ -71,6 +71,32 @@ def test_execute_command(self):
...
+        packages_path = self.data_path("builds", "packages")
+        r = ResolvedContext(["testing_obj"], testing=True, package_paths=[packages_path])
+        resolvedPackages = [x.qualified_package_name for x in r.resolved_packages]
+        self.assertEqual(resolvedPackages, ["floob", "testing_obj-1.0.0"])
+
+    def test_execute_command_testing_environ(self):
+        """Test that execute_command properly sets test specific environ dict"""
+        self.inject_python_repo()
+        packages_path = self.data_path("builds", "packages")
+        r = ResolvedContext(["testing_obj", "python"], testing=True, package_paths=[packages_path])
```
JeanChristopheMorinPerso, line 85:
```suggestion
        r = ResolvedContext(["testing_obj", "python"], testing=True, package_paths=[packages_path] + self.settings["packages_path"])
```
Because you override the paths, we need to re-inject the paths that are in the active config.

src/rez/tests/test_context.py
```diff
@@ -71,6 +71,32 @@ def test_execute_command(self):
...
+        pycode = ("import os; "
+                  "print(os.getenv(\"CAR_IDEA\"));")
+
+        args = ["python", "-c", pycode]
+
+        p = r.execute_command(args, stdout=subprocess.PIPE)
+        stdout, _ = p.communicate()
+        stdout = stdout.strip()
+        parts = [x.strip() for x in stdout.decode("utf-8").split('\n')]
+        self.assertEqual(parts, ["STURDY STEERING WHEEL"])
```
JeanChristopheMorinPerso, line 98:
```suggestion
        self.assertEqual(r.get_environ().get("CAR_IDEA"), "STURDY STEERING WHEEL")
```
`ResolvedContext` already has a function to extract the environment variable from its context. This will be more reliable in our tests.

src/rez/tests/test_context.py
```diff
@@ -71,6 +71,32 @@ def test_execute_command(self):
         stdout = stdout.strip()
         self.assertEqual(stdout, "Hello Rez World!")
 
+    def test_resolved_packages_testing_environ(self):
+        """Test resolving packages within a testing environment behaves correctly"""
+        packages_path = self.data_path("builds", "packages")
+        r = ResolvedContext(["testing_obj"], testing=True, package_paths=[packages_path])
```
JeanChristopheMorinPerso, line 77:
```suggestion

        # Note how we use testing=True
        r = ResolvedContext(["testing_obj"], testing=True, package_paths=[packages_path])
```

JeanChristopheMorinPerso (commented):
Alright, I had some time to look at the failures. I left some comments. Hopefully that should fix it.

JeanChristopheMorinPerso:
Tests are not working on macOS! I can probably merge the PR as is and I'll handle the linter stuff.

JeanChristopheMorinPerso:
The windows tests are now failing though. Argh. Anyway, at this point, I think I'll just merge this and will handle the rest in another PR.

fabal:
Hi @JeanChristopheMorinPerso , did you need something from us to be able to merge this?

JeanChristopheMorinPerso:
Hi @fabal, thanks for the reminder. The last couple of months have been very busy and intense for us. We didn't have the time to push your PR to the finish line unfortunately.

I'm definitely planning on releasing your PR in the next release.

PR #1745, opened 2024-05-02
Add small bit of type hinting

BryceGattis:


src/rez/plugin_managers.py
```diff
@@ -235,7 +239,7 @@ def config_schema(self):
                 deep_update(d, d_)
         return dict_to_schema(d, required=True, modifier=expand_system_vars)
 
-    def create_instance(self, plugin, **instance_kwargs):
+    def create_instance(self, plugin, **instance_kwargs) -> "RezPluginType":
```
JeanChristopheMorinPerso, line 242: This doesn't return a `RezPluginType` instance. It will return something like `rezplugins.build_system.cmake.CMakeBuildSystem`. I don't think we have an easy way today to get all the possible base classes for each plugin type. It should be possible, but it'll require changes to the plugin system, or some more thoughts to maybe use generics or whatever type magic we'll need to get this to work.

BryceGattis: @JeanChristopheMorinPerso Ah right good point. Would it be alright to type hint these just as `object` instead for now? That at least gives us some kind of information.

src/rez/config.py
```diff
@@ -802,7 +804,7 @@ def _get_new_session_popen_args(self):
 
 class _PluginConfigs(object):
     """Lazy config loading for plugins."""
-    def __init__(self, plugin_data):
+    def __init__(self, plugin_data: typing.Dict[str, typing.Any]):
         self.__dict__['_data'] = plugin_data
```
BryceGattis, line 808: I'm curious why we do this `self.__dict__['_data']` accessing here. Perhaps there is a cleaner way to do this?

JeanChristopheMorinPerso: I think it's to bypass `__getattr__`.

src/rez/plugin_managers.py
```diff
@@ -235,9 +239,9 @@ def config_schema(self):
                 deep_update(d, d_)
         return dict_to_schema(d, required=True, modifier=expand_system_vars)
 
-    def create_instance(self, plugin, **instance_kwargs):
+    def create_instance(self, plugin_name: str, **instance_kwargs) -> object:
```
JeanChristopheMorinPerso, line 242:
```suggestion
    def create_instance(self, plugin_name: str, **instance_kwargs) -> typing.Any:
```
I think `Any` is slightly better for mypy. Any declares that the return is not typed.

src/rez/plugin_managers.py
```diff
@@ -385,12 +389,12 @@ def get_failed_plugins(self, plugin_type):
         """
         return self._get_plugin_type(plugin_type).failed_plugins.items()
 
-    def create_instance(self, plugin_type, plugin_name, **instance_kwargs):
+    def create_instance(self, plugin_type: str, plugin_name: str, **instance_kwargs) -> object:
```
JeanChristopheMorinPerso, line 392:
```suggestion
    def create_instance(self, plugin_type: str, plugin_name: str, **instance_kwargs) -> typing.Any:
```
Same here

PR #1761, opened 2024-05-19
Add type annotations

### LINKED ISSUE #1631: Add type hinting
Add type hinting across the project since we are no longer tied down to Python 2 support.

**Motivation**
This will speed up iteration time since IDEs will be able to use these type hints to check types.

chadrik:
This is a first pass at adding type annotations throughout the code-base.  Mypy is not fully passing yet, but it's getting close. 

Fixes #1631


src/rez/package_filter.py
```diff
@@ -327,7 +327,7 @@ class Rule(object):
     """Base package filter rule"""
 
     #: Rule name
-    name = None
+    name: str
```
chadrik, line 330: This pattern appears on abstract base classes in a number of places, and we have 3 choices:

## Option 1:

```python
name: str | None = None
```

typically produces many spurious mypy errors, because the code assumes the value is a string everywhere, and the `None`-case is not handled.

## Option 2:

```python
name: str
```

only safe if we're sure that the `name` attribute is always accessed on concrete sub-classes, otherwise will result in an `AttributeError` since this does not actually create an attribute. 

## Option 3:

```python
@property
@abstractmethod
def name(self) -> str:
    raise NotImplementedError
```

Provides us some guarantee that sub-classes actually implement `name`.

src/rez/package_order.py
```diff
@@ -1,36 +1,46 @@
...
-    def __init__(self, main_comparable, fallback_comparable):
+    def __init__(self, main_comparable: SupportsLessThan,
+                 fallback_comparable: SupportsLessThan):
         self.main_comparable = main_comparable
         self.fallback_comparable = fallback_comparable
 
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, FallbackComparable):
+            return NotImplemented
```
chadrik, line 35: `mypy` strongly prefers that comparison functions define the type of `other` as `object`. by returning `NotImplemented` we allow other other object to handle equality if it implements it.   This should not change the behavior as long as `other` is always `FallbackComparable` within Rez.

src/rez/package_resources.py
```diff
@@ -310,7 +315,9 @@ class PackageFamilyResource(PackageRepositoryResource):
     A repository implementation's package family resource(s) must derive from
     this class. It must satisfy the schema `package_family_schema`.
     """
-    pass
+
+    def iter_packages(self) -> Iterator[Package]:
+        raise NotImplementedError
```
chadrik, line 320: All subclasses of `PackageFamilyResource` implement this method, so it appears to be considered an `abstractmethod` of `PackageFamilyResource`.   Defining it simplifies some type annotation problems.

src/rez/package_resources.py
```diff
@@ -345,17 +352,23 @@ class VariantResource(PackageResource):
     this case it is the 'None' variant (the value of `index` is None). This
     provides some internal consistency and simplifies the implementation.
     """
+
+    @property
+    @abstractmethod
+    def parent(self) -> PackageRepositoryResource:
+        raise NotImplementedError
```
chadrik, line 359: same situation here as with `PackageFamilyResource.iter_packages`.

src/rez/package_resources.py
```diff
@@ -383,23 +396,28 @@ class PackageResourceHelper(PackageResource):
     """
     variant_key = None
 
+    @property
+    @abstractmethod
+    def parent(self) -> PackageRepositoryResource:
+        raise NotImplementedError
```
chadrik, line 402: here as well.

src/rez/packages.py
```diff
@@ -319,6 +326,7 @@ def get_variant(self, index=None):
         for variant in self.iter_variants():
             if variant.index == index:
                 return variant
+        return None
```
chadrik, line 329: mypy prefer explicit `return None` statements

src/rez/resolved_context.py
```diff
@@ -1038,16 +1049,6 @@ def print_resolve_diff(self, other, heading=None):
...
-    def _on_success(fn):
-        @wraps(fn)
-        def _check(self, *nargs, **kwargs):
-            if self.status_ == ResolverStatus.solved:
-                return fn(self, *nargs, **kwargs)
-            else:
-                raise ResolvedContextError(
-                    "Cannot perform operation in a failed context")
-        return _check
-
```
chadrik, line 1050: mypy does not like these decorators defined at the class-level.

chadrik:
`Protocol` and `TypedDict` are not available in the `typing` module until python 3.8.

We have a few options:

1. vendor `typing_extensions`
2. remove use of these typing classes until Rez drops support for python 3.7
3. I can create a mock of `Protocol` and trick `mypy` into using it which is safe because it has no runtime behavior.   Doing the same thing for `TypedDict` is more complicated, but possible.

chadrik:
I got bored and added lots more, particularly focused on the solver module.    Once the solver module is complete, we can experiment with compiling it to a c-extension using [mypyc](https://mypyc.readthedocs.io/en/latest/getting_started.html), which could provide a big speed boost!

chadrik:
I now have `rez.solver` compiling as a C-extension with all tests passing.   I'm very interested to see how the performance compares.   Does anyone want to volunteer to help put together a performance comparison?  Are there any known complex collection of packages to test against?

src/rez/package_order.py
```diff
@@ -657,31 +678,32 @@ def refresh(self) -> None:
                     continue
                 self.by_package[package] = orderer
 
-    def append(self, *args, **kwargs):
-        self.dirty = True
-        return super().append(*args, **kwargs)
+    if not TYPE_CHECKING:
+        def append(self, *args, **kwargs):
```
chadrik, line 682: Since this class inherits from `list` it's easier to rely on the type hints coming from that base class than to redefine them here, so we hide them by placing them behind `not TYPE_CHECKING`.   In reality, the runtime value of `TYPE_CHECKING` is always `False`.

src/rez/plugin_managers.py
```diff
@@ -355,10 +359,22 @@ def get_plugins(self, plugin_type):
...
+    def get_plugin_class(self, plugin_type: str, plugin_name: str) -> type:
+        pass
+
+    @overload
+    def get_plugin_class(self, plugin_type: str, plugin_name: str, expected_type: type[T]) -> type[T]:
+        pass
+
+    def get_plugin_class(self, plugin_type, plugin_name, expected_type=None):
         """Return the class registered under the given plugin name."""
         plugin = self._get_plugin_type(plugin_type)
```
chadrik, line 372: Added a new argument here to validate the returned result.   This provides both runtime and static validation.

src/rez/resolved_context.py
```diff
@@ -162,16 +177,28 @@ def __call__(self, state):
...
+                 package_load_callback: Callable[[Package], Any] | None = None,
+                 buf: SupportsWrite | None = None,
+                 suppress_passive=False,
+                 print_stats=False,
+                 package_caching=None):
         """Perform a package resolve, and store the result.
 
         Args:
-            package_requests (list[typing.Union[str, PackageRequest]]): request
+            package_requests (list[typing.Union[str, Requirement]]): request
```
chadrik, line 201: I noticed that everywhere that we've documented types as `PackageRequest`, they appear to actually be `Requirement`.  I'm not sure if there any real-world exceptions to this.

src/rez/resolved_context.py
```diff
@@ -884,7 +912,7 @@ def _rt(t):
             return
 
         _pr("resolved packages:", heading)
-        rows = []
+        rows3: list[tuple[str, str, str]] = []
```
chadrik, line 915: you can't redefine types with mypy, so you need to use new variable names.

src/rez/system.py
```diff
@@ -88,7 +88,7 @@ def shell(self):
                     args = ['ps', '-o', 'args=', '-p', str(parent_pid)]
                     proc = sp.Popen(args, stdout=sp.PIPE)
                     output = proc.communicate()[0]
-                    shell = os.path.basename(output.strip().split()[0]).replace('-', '')
+                    shell = os.path.basename(output.decode().strip().split()[0]).replace('-', '')
```
chadrik, line 91: `output` is `bytes` in python3, so need to call `decode()`

src/rez/utils/data_utils.py
```diff
@@ -213,50 +214,53 @@ def get_dict_diff_str(d1, d2, title):
...
-        raise RuntimeError("@cached_property should not be called.")
-
-    @classmethod
-    def uncache(cls, instance, name):
-        if hasattr(instance, name):
-            delattr(instance, name)
+if TYPE_CHECKING:
+    cached_property = property
+else:
+    class cached_property(object):
```
chadrik, line 220: it's much easier to pretend that `cached_property` is `property` than to type hint all the subtleties of a descriptor.

src/rez/solver.py
```diff
@@ -1946,14 +2004,16 @@ def __init__(self, package_requests, package_paths, context=None,
...
-        self.solve_begun = None
-        self.solve_time = None
-        self.load_time = None
+        # these values are all set in _init()
+        self.phase_stack: list[_ResolvePhase]
+        self.failed_phase_list: list[_ResolvePhase]
+        self.depth_counts: dict
+        self.solve_begun: bool
+        self.solve_time: float
+        self.load_time: float
```
chadrik, line 2013: these values are never actually `None`, so we define the types here, and their values are assigned in `_init()`

src/rez/solver.py
```diff
@@ -74,9 +97,6 @@ class SolverStatus(Enum):
     cyclic = ("The solve contains a cycle.", )
     unsolved = ("The solve has started, but is not yet solved.", )
 
-    def __init__(self, description):
-        self.description = description
-
```
chadrik, line 79: mypyc did not like adding a new `__init__` for `Enum`, so it was simple enough to replace this attribute with a call to the enum `value`.

src/rez/solver.py
```diff
@@ -2083,15 +2143,15 @@ def resolved_ephemerals(self):
         final_phase = self.phase_stack[-1]
         return final_phase._get_solved_ephemerals()
 
-    def reset(self):
+    def reset(self) -> None:
         """Reset the solver, removing any current solve."""
         if not self.request_list.conflict:
-            phase = _ResolvePhase(self.request_list.requirements, solver=self)
+            phase = _ResolvePhase(solver=self)
```
chadrik, line 2149: This appears to be a bug:  `_ResolvePhase` only takes one argument.  mypy to the rescue.

chadrik:
I found `rez-benchmark`.  Interestingly, rez is slower with the compiled `rez.solver`.   It could be because there are many modules and classes used by `rez.solver` which have not been compiled.  

I probably won't have time to dig into this much more, but once this PR is merged I'll make a new PR with the changes necessary for people to test the compiled version of rez.

chadrik:
Note: this PR likely invalidates https://github.com/AcademySoftwareFoundation/rez/pull/1745

chadrik:
@instinct-vfx  Can you or someone from the Rez group have a look at this, please?

src/rez/build_process.py
```diff
@@ -119,14 +136,15 @@ def __init__(self, working_dir, build_system, package=None, vcs=None,
...
+    def package(self) -> DeveloperPackage:
         return self.build_system.package
 
     @property
-    def working_dir(self):
+    def working_dir(self) -> str:
         return self.build_system.working_dir
 
-    def build(self, install_path=None, clean=False, install=False, variants=None):
+    def build(self, install_path: str | None = None, clean: bool = False,
```
JeanChristopheMorinPerso, line 146: Using `str | None` means we need to drop support for python 3.7. I'm not sure we are ready for this yet.

JeanChristopheMorinPerso:
@chadrik You need to sign the CLA before we can even start to look at the PR.

src/rez/build_process.py
chadrik: No, use of `str | None` is find in python 3.7 as long as you use `from __future__ import annotations`.  This behavior ensures that type annotations are recorded as strings within `__annotations__` attributes, which means they are not evaluated at runtime unless [`inspect.get_annoations`](https://docs.python.org/3/library/inspect.html#inspect.get_annotations) is called. 

The only thing breaking python 3.7 compatibility here is the use of `TypedDict` and `Protocol`, as mentioned in [another comment](https://github.com/AcademySoftwareFoundation/rez/pull/1761#issuecomment-2119317337).  I presented 3 options for preserving the use of these classes in the other comment.  

I noticed that the only python 3.7 tests that are currently run are for MacOS, which I took as an indicator that python 3.7 would be dropped soon.   Is there a schedule for deprecation?

chadrik:
@JeanChristopheMorinPerso 

> @chadrik You need to sign the CLA before we can even start to look at the PR.

I work for Scanline, which is owned by Netflix, and I'm meeting with our CLA manager on Monday.  I made this contribution on my own time:  can choose individual vs corporate CLA on a per PR basis?

JeanChristopheMorinPerso:
>I made this contribution on my own time: can choose individual vs corporate CLA on a per PR basis?

I don't think you "can" but your account can be associated to an an ICLA and a CCLA. But I'm not a lawyer so I can't help more than that. If you and or your employer/CLA manager have questions,  you can contact the LF support by following the link in the EasyCLA comment: https://github.com/AcademySoftwareFoundation/rez/pull/1761#issuecomment-2119309923.

chadrik:
CLA is signed!

src/rez/build_process.py
chadrik: By the way, I fixed the python 3.7 compatibility issue with `TypedDict` and `Protocol`, so that should not be a blocker anymore.

src/rez/build_system.py
```diff
@@ -143,12 +167,12 @@ def __init__(self, working_dir, opts=None, package=None,
         self.opts = opts
 
     @classmethod
-    def is_valid_root(cls, path):
+    def is_valid_root(cls, path: str, package=None) -> bool:
         """Return True if this build system can build the source in path."""
```
chadrik, line 171: sub-classes are expected to have the `package` argument.  I think this was just an oversight on the base class.

src/rez/package_order.py
```diff
@@ -614,13 +633,13 @@ def from_pod(cls, data):
         )
 
 
-class PackageOrderList(list):
+class PackageOrderList(List[PackageOrder]):
```
chadrik, line 636: Note that we use `typing.List` here instead of `list` because `list` does not become indexable at runtime until python 3.9.   It's still safe to use `list[X]` inside annotations as long as we use `from __future__ import annotations`.

chadrik:
Any thoughts on this PR?

JeanChristopheMorinPerso:
Hey @chadrik, I made a first good read last week and I'll try to do another one soon. If you have the time, I would really love to see a GitHub Actions workflow that would run mypy on all pull requests.

chadrik:
>  I would really love to see a GitHub Actions workflow that would run mypy on all pull requests.

Me too!  

The challenge is that there are still a lot of errors.  These are not the result of incorrect annotations, but rather due to code patterns which are difficult to annotate correctly without more invasion refactors.  For example, there are quite a few objects which dynamically generate attributes, and that's a static typing anti-pattern. 

If you'd like an Action that runs mypy but allows failure for now, that's pretty easy, but if you want failures to block PRs, that'll take a lot more work.   I'd prefer not to make that a blocker to merging this, though, because I've had to rebase and fix merge conflicts a few times already. 

I do have a plan for how we can get to zero failures in the mid-term: I wrote a tool which allows you to specify patterns for errors to ignore, but I need to update it.

JeanChristopheMorinPerso:
I think we can introduce a workflow that will fail for newly introduced errors and warnings. I'm sure someone already thought of that somewhere and we can probably re-use what they did?

Basically, I'd like to verify that your changes work as expected and that we don't regress in the future and that new code is typed hint. Mypy can also be configured on a per module basis right?

chadrik:
@JeanChristopheMorinPerso I used a project called `mypy-baseline` to create a set of ignore patterns that can be filtered out until they're resolved:  https://mypy-baseline.orsinium.dev/usage

If new errors are introduced they will need to be resolved (i.e. errors not filtered by the established baseline regexes).  The regexes are bound to particular files but not to line numbers.  The mypy CI job will pass if a developer fixes an error, but in this case it is good practice to set a new baseline using `mypy | mypy-baseline sync --ignore-categories=note`.  When creating a baseline, care must be taken to use the same version of python and mypy.

dbr:
Not sure if it was discussed already, but is it feasible to only use [`# typing: ignore` comments](https://mypy.readthedocs.io/en/stable/common_issues.html#spurious-errors-and-locally-silencing-the-checker) to suppress existing known errors?

Main benefit would be people using mypy integration in their editors wouldn't be floored with the errors ignored by mypy-baseline

I think mypy also warns when the ignore is no longer needed, so if some refactoring resolves the issue, the comments are easy to clean up

chadrik:
There are about 300 errors, so using inline ignores has the downside of littering the code with type comments.  It also means that developers need to be careful to move the ignore comments to the proper locations when décatiront code, and the solution not always obvious.  I think at this stage what I’ve presented here is a very low effort way to ease this project into type annotations. With the mypy-baseline project there is a simple way to just bulk add problems to the ignore list, if necessary. 

FWIW, most of the remaining errors will require some pretty major refactors to resolve — meaning we probably need to replace the schema objects with something like dataclasses or pydantic to remove dynamic attribute generation.    If folks are interested in taking on some more substantial changes to get the code base fully covered, I’m happy to do that.

chadrik:
Shall we get this merged?

chadrik:
Hi!  Is there anything I can do to get this across the line?

JeanChristopheMorinPerso:
Rebased to get all the latest changes from main.

src/rez/rex.py
```diff
@@ -11,6 +13,7 @@
 from contextlib import contextmanager
 from string import Formatter
 from collections.abc import MutableMapping
+from typing import Iterable
```
JeanChristopheMorinPerso, line 16: You could support all python versions by using `rez.utils.typing` like you did for protocol.

src/rez/solver.py
```diff
@@ -21,16 +23,31 @@
...
+if TYPE_CHECKING:
+    from rez.resolved_context import ResolvedContext
+    from rez.package_filter import PackageFilterBase
+    from rez.package_order import PackageOrder
+
+
+T = TypeVar("T")
+
+
+class SupportsWrite(Protocol):
```
JeanChristopheMorinPerso, line 47: Would something like `io.RawIOBase` be more appropriate here?

src/rez/solver.py
```diff
@@ -122,79 +136,82 @@ def __call__(self, txt, *args):
...
         return "%s(%s)" % (self.__class__.__name__, str(self))
 
 
 class Reduction(_Common):
     """A variant was removed because its dependencies conflicted with another
     scope in the current phase."""
-    def __init__(self, name, version, variant_index, dependency,
-                 conflicting_request):
+    def __init__(self, name: str, version, variant_index: int | None, dependency: Requirement,
+                 conflicting_request: Requirement):
```
JeanChristopheMorinPerso, line 178: Is it a requirement or a request?

src/rez/solver.py
```diff
@@ -821,6 +844,7 @@ def _split(i_entry, n_variants, common_fams=None):
             for j, variant in enumerate(entry.variants):
                 fams = fams & variant.request_fams
                 if not fams:
+                    assert prev is not None
```
JeanChristopheMorinPerso, line 847: Will there be a cost to this assert? Same with the other asserts in this module.

src/rez/solver.py
```diff
@@ -907,7 +931,8 @@ def __str__(self):
             s = "[%s==%s%s]" % (self.package_name, str(variant.version), s_idx)
         elif nversions == 1:
             entry = self.entries[0]
-            indexes = sorted([x.index for x in entry.variants])
+            # we expect all variants to have a non-None index, but filter to satisfy mypy
+            indexes = sorted([x.index for x in entry.variants if x.index is not None])
```
JeanChristopheMorinPerso, line 935: This could have a performance impact.

src/rez/solver.py
JeanChristopheMorinPerso, line 1: Changes to this module will require more review. I don't want to accidentally introduce a performance regression.

src/rez/system.py
JeanChristopheMorinPerso: We use `text=True`, so it should be string isn't it?

src/rez/utils/backcompat.py
```diff
@@ -52,7 +52,7 @@ def convert_old_command_expansions(command):
 within_unescaped_quotes_regex = re.compile('(?<!\\\\)"(.*?)(?<!\\\\)"')
 
 
-def convert_old_commands(commands, annotate=True):
+def convert_old_commands(commands, annotate=True) -> str:
```
JeanChristopheMorinPerso, line 55:
```suggestion
def convert_old_commands(commands: list[str], annotate: bool = True) -> str:
```

src/rez/utils/data_utils.py
JeanChristopheMorinPerso: But we loose the `uncache` method.

src/rez/utils/patching.py
```diff
@@ -42,8 +42,8 @@ def get_patched_request(requires, patchlist):
         '^': (True, True, True)
     }
 
-    requires = [Requirement(x) if not isinstance(x, Requirement) else x
-                for x in requires]
+    requires: list[Requirement | None] = [
```
JeanChristopheMorinPerso, line 45: Can it really contain a `None`?

src/rez/utils/platform_.py
```diff
@@ -555,7 +557,8 @@ def _difftool(self):
 
 
 # singleton
-platform_ = None
+# FIXME: is is valid for platform_ to be None?
```
JeanChristopheMorinPerso, line 560: Technically yes, say on solaris, freebsd, etc.

src/rez/utils/resources.py
```diff
@@ -41,6 +43,15 @@
 from rez.exceptions import ResourceError
 from rez.utils.logging_ import print_debug
 
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    # this is not available in typing until 3.11, but due to __future__.annotations
+    # we can use it without really importing it
+    from typing import Self
```
JeanChristopheMorinPerso, line 51: This makes it impossible to run mypy with python < 3.11 right?

src/rez/version/_requirement.py
```diff
@@ -264,15 +270,17 @@ def conflicts_with(self, other):
...
-        else:  # VersionedObject
+        elif isinstance(other, VersionedObject):
             if (self.name_ != other.name_) or (self.range is None):
                 return False
             if self.conflict:
                 return (other.version_ in self.range_)
             else:
                 return (other.version_ not in self.range_)
+        else:
+            return NotImplemented
```
JeanChristopheMorinPerso, line 281: Should this raise an `NotImplementedError` instead?

src/rez/version/_requirement.py
```diff
@@ -335,23 +343,24 @@ def _r(r_):
...
+    def __str__(self) -> str:
         if self._str is None:
             pre_str = '~' if self.negate_ else ('!' if self.conflict_ else '')
             range_str = ''
             sep_str = ''
 
             range_ = self.range_
-            if self.negate_:
+            # Note: the only time that range_ is None is if self.negate_ is True
+            if self.negate_ or range_ is None:
```
JeanChristopheMorinPerso, line 363: I'm not sure this is the right thing to do. Why should `range_` be None?

JeanChristopheMorinPerso (commented):
Second batch of comments. Note that I'm not yet done with the review. Please don't push changes until I'm done.

src/rez/package_resources.py
```diff
@@ -345,17 +355,23 @@ class VariantResource(PackageResource):
     this case it is the 'None' variant (the value of `index` is None). This
     provides some internal consistency and simplifies the implementation.
     """
+
+    @property
+    @abstractmethod
```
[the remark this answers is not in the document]
chadrik: It provides static enforcement but not runtime enforcement:  mypy will enforce that classes are not instantiated without implementing the method.

src/rez/resolved_context.py
chadrik: The kind of haphazard use and documentation of `PackageRequest` and `PackageRequest` results in some very difficult situations to accurate add type annotations.   If you want to see for yourself, check out the code, change this to `PackageRequest` and observe the new errors produced by mypy.

src/rez/system.py
chadrik: We didn't use `text=True` here.   I could pass `text=True ` to `Popen`.

src/rez/utils/data_utils.py
chadrik: `typing.TYPE_CHECKING` always resolve to `False` at runtime and `True` only during static analysis.  So the code within the `if TYPE_CHECKING` block will never run.  It's a way to simplify certain type analysis situations that arise.

src/rez/version/_requirement.py
chadrik: Nope.   `NotImplemented` is a singleton that is used to indicate "I don't support this comparison", and it will fall back to other object to see if it supports it. 

https://docs.python.org/3/reference/datamodel.html#object.__ge__

> When no appropriate method returns any value other than [NotImplemented](https://docs.python.org/3/library/constants.html#NotImplemented), the == and != operators will fall back to is and is not, respectively.

This is a fundamental part of how comparison works in python, and returning `NotImplemented` is the most correct behavior.

src/rez/version/_requirement.py
chadrik: this is the init code:

```python
        m = self.sep_regex.search(s)
        if m:
            i = m.start()
            self.name_ = s[:i]
            req_str = s[i:]
            if req_str[0] in ('-', '@', '#'):
                self.sep_ = req_str[0]
                req_str = req_str[1:]
            self.range_ = VersionRange(
                req_str, invalid_bound_error=invalid_bound_error)
            if self.negate_:
                self.range_ = ~self.range_
        elif self.negate_:
            self.name_ = s
            # rare case - '~foo' equates to no effect
            self.range_ = None
        else:
            self.name_ = s
            self.range_ = VersionRange()
```

Notice that if `self.negate_ `is True, `self._range_` is not set.

src/rez/utils/platform_.py
chadrik: allowing it to be None causes tons of mypy errors because there are lots of places where `platform_` is used without checking if it's None.

src/rez/solver.py
chadrik: I can change this to use a `typing.cast` to change the type without any runtime impact, as long as we're confident that the index is never None.

src/rez/build_system.py
```diff
@@ -104,12 +127,13 @@ class BuildSystem(object):
...
     """
     @classmethod
-    def name(cls):
+    def name(cls) -> str:
         """Return the name of the build system, eg 'make'."""
         raise NotImplementedError
 
-    def __init__(self, working_dir, opts=None, package=None,
-                 write_build_scripts=False, verbose=False, build_args=[],
+    def __init__(self, working_dir: str, opts=None,
```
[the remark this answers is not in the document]
chadrik: I'd prefer for new annotations to be done in a new PR, so that we can keep the scope of this PR under control.  There are literally thousands of unannotated arguments that still need to be dealt with.

src/rez/packages.py
```diff
@@ -904,7 +919,23 @@ def get_completions(prefix, paths=None, family_only=False):
     return words
 
 
-def get_latest_package(name, range_=None, paths=None, error=False):
+@overload
+def get_latest_package(name: str, *, range_=None,
+                       paths: list[str] | None = None,
+                       error: Literal[True] = True) -> Package:
```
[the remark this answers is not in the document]
chadrik: the function returns different types depending on its arguments.  if `error=True` it is always guaranteed to return a `Package`, otherwise it might be `None`.    Not doing this would in many spurious errors about package variables possibly being None.

src/rez/build_system.py
```diff
@@ -67,9 +89,10 @@ def get_valid_build_systems(working_dir, package=None):
     return clss
 
 
-def create_build_system(working_dir, buildsys_type=None, package=None, opts=None,
+def create_build_system(working_dir: str, buildsys_type: str | None = None,
+                        package=None, opts=None,
```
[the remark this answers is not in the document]
chadrik: As mentioned in the other comment, I'd prefer for new annotations to be done in a new PR, so that we can keep the scope of this PR under control. There are literally thousands of unannotated arguments that still need to be dealt with.

src/rez/utils/platform_.py
chadrik: And to be perfectly clear, these errors indicate that there are many places with the rez code that will fail at runtime if platform_ is None.

src/rez/utils/resources.py
chadrik: as I explained in my other comment code within an `if TYPE_CHECKING` does not actually run.

src/rez/plugin_managers.py
chadrik: this uses type vars to describe a [generic function](https://mypy.readthedocs.io/en/stable/generics.html#generic-functions), i.e. a function where the types of its arguments are related to each other.  In our case, the type of the argument `expected_type` is related to the output type. 

Using a literal would mean we would have to define a literal value and a function overload for every possible output type, like this: 

```python
    @overload
    def get_plugin_class(self, plugin_type: str, plugin_name: str, expected_type: Literal["Foo"]) -> type[Foo]:
        pass

    @overload
    def get_plugin_class(self, plugin_type: str, plugin_name: str, expected_type: Literal["Bar"]) -> type[Bar]:
        pass
```

In a plugin environment where users can define their own types, we obviously cannot define a string constant for every possible type.

src/rez/package_order.py
chadrik: will do.

src/rez/package_order.py
```diff
@@ -657,31 +676,32 @@ def refresh(self) -> None:
                     continue
                 self.by_package[package] = orderer
 
-    def append(self, *args, **kwargs):
-        self.dirty = True
-        return super().append(*args, **kwargs)
+    if not TYPE_CHECKING:
+        def append(self, *args, **kwargs):
+            self.dirty = True
```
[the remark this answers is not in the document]
chadrik: This code is in an `if not TYPE_CHECKING:` which  means it is hidden from mypy, but it still runs at runtime (the oppposite of a `if TYPE_CHECKING` block). 

As mentioned in my other comment:

> Since this class inherits from list it's easier to rely on the type hints coming from that base class than to redefine them here, so we hide them by placing them behind not TYPE_CHECKING. In reality, the runtime value of TYPE_CHECKING is always False.

should I adde this explanation as a comment in the code?

src/rez/rex.py
chadrik: `typing.Iterable` exists for all supported versions of python (since python 3.5, I believe).

src/rez/solver.py
chadrik: That could work, but it adds requirements that we don't actually need:  `read`, `readinto`, and `readall` methods.  In our case we only care about writing, so this protocol is more technically correct, since today a user could create any kind of object to write into it as long as it defines a `write` method.

src/rez/build_system.py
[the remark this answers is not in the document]
chadrik: would like to avoid scope creep.

src/rez/build_system.py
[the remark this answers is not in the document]
chadrik: would like to avoid scope creep.

src/rez/plugin_managers.py
JeanChristopheMorinPerso: Users can't define their own types. They can only create new plugin of some pre-defined types. Anybody adding new plugin types should do so in rez itself and not outside.

src/rez/system.py
JeanChristopheMorinPerso: See https://github.com/AcademySoftwareFoundation/rez/blob/main/src/rez/system.py#L89.

src/rez/utils/data_utils.py
JeanChristopheMorinPerso: >typing.TYPE_CHECKING always resolve to False

I know, but we are loosing stuff during typing. That's my whole point (the same apply to all my comments that are similar to this one).

src/rez/build_system.py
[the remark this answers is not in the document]
JeanChristopheMorinPerso: From what I see, the current typing seems arbitrary. Some stuff was typed, some others don't and I'm not too sure what the logic is. I think it would be preferable to change the whole signature instead of just changing half of it since we are anyway changing it.

src/rez/packages.py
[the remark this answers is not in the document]
JeanChristopheMorinPerso: Ah, I missed the fact that the return type was different. Thanks!

src/rez/rex.py
JeanChristopheMorinPerso: The thing is that it's deprecated in 3.9 in favor of collections.abc.Iterable.

src/rez/solver.py
JeanChristopheMorinPerso: Yes there is, but I don't fully trust it. It runs on GHA and GHA's environment are not very stable and on top of that instances today now have 4 CPUs instead of 2. All our historical data was collected on the older runners that had 2 CPUs.

src/rez/utils/platform_.py
JeanChristopheMorinPerso: Sure, I was just answering the question. It will fail if it's None. We could fix this by simply adding an else clause and raise a exception with a clear error message.

src/rez/utils/resources.py
JeanChristopheMorinPerso: Note how I said "run mypy". Like, if we were to run mypy with `python_version = 3.7` for example.

src/rez/version/_requirement.py
JeanChristopheMorinPerso: But should we allow comparing with anything else than a `Requirement` or `VersionedObject`? I wouldn't consider this function a comparator per say.

src/rez/utils/data_utils.py
chadrik: in what way would the static analysis be degraded by using `property` instead of `cached_property`?  To my knowledge, they are functionally equivalent from a static analysis POV:  the types returned are the same.

chadrik:
It would be great if all this work didn't go to waste.  If I rebase this, can we get it merged?

src/rez/rex.py
chadrik: yes, but it still exists in `typing` module for all the versions of python that Rez supports and I suspect that Rez will drop support for python 3.9 before `Iterable` is dropped from the typing module.   For clarity of users reading the code I think we should favor official modules over our shims where they work.

src/rez/utils/resources.py
chadrik: I don't recommend running mypy for each version of python until we actually do the immense amount of remaining work to fully annotate all of the code here and make it all pass.   The simple thing to do right now is to run mypy using a single and modern version of python.

chadrik:
I'm motivated again to get this merged.  I've started addressing some of the outstanding notes here, but I would love it if we could focus on blocker issues.

src/rez/solver.py
chadrik: As I mentioned in another comment, the code uses `Requirement` and `PackageRequests` almost interchangeable despite the fact that they have a subtle difference.  Currently, it's a puzzle without a valid solution:  there is no set of stubs which can satisfy the current code base wrt `Requirement` and `PackageRequests`.  This means that there are certainly bugs in the code.  I've made a "best effort" which results in the fewest number of errors.  

tl;dr `Requirement` is the best fit for how these arguments are used.

src/rez/solver.py
chadrik: To reiterate, creating a protocol here is more accurate than using `io.RawIOBase` and is the correct solution.

src/rez/solver.py
chadrik: I should also add that if you don't want to incur the cost of the asserts you could run `python -O`

src/rez/solver.py
chadrik: I changed this to `typing.cast`

src/rez/version/_requirement.py
chadrik: I replaced with this a `raise TypeError`.  I also checked all the places that this is being called (manually and using mypy) and the type of `other` is always `Requirement |VersionedObject`, so it should be safe.

src/rez/package_order.py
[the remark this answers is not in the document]
chadrik: I added this comment to the code

src/rez/build_process.py
chadrik: It's been a year since this thread was started:  has rez dropped python 3.7 support yet?

This PR will work for python 3.7, but I can remove some workarounds if we've dropped support.

src/rez/utils/resources.py
chadrik: Can I resolve this thread?  This code works in python 3.7 and we should not run mypy against multiple versions of python at this stage.

src/rez/solver.py
chadrik: Would you like me to remove this assert?  

There are a lot of threads here with open questions and no guidance on what action to take.  I would like to know what I should change to allow this PR to be merged.

src/rez/version/_requirement.py
chadrik: Shall I change the code back to how it was?

chadrik:
In the next week or two I'll do a pass to rebase this and remove anything that looks even remotely like a runtime change.

JeanChristopheMorinPerso:
> In the next week or two I'll do a pass to rebase this and remove anything that looks even remotely like a runtime change.

Thanks @chadrik. This will definitely help us get this merged with more confidence (not that we don't trust you, but this PR is quite big).

src/rez/build_process.py
JeanChristopheMorinPerso: Not yet, but we could do this in the next release.

src/rez/package_resources.py
chadrik: Removed

src/rez/solver.py
chadrik: I removed all asserts

src/rez/utils/patching.py
chadrik: Reverted

src/rez/build_process.py
chadrik: I added a check to mypy to ensure it raises errors if we use any unsupported typing features. 

There are a few cases of unsupported features:

- `dict[X]`, `list[X]`:  use `from __future__ import annotations`
- `X | None`:  use `from __future__ import annotations`
- `typing.Self` or other type not in `typing:  use `if TYPE_CHECKING`

Should I document this somewhere?

For now, I recommend that we keep using `from __future__ import annotations` so that we can use the modern form. 

Here's a doc I compiled of typing changes by python version:

https://docs.google.com/document/d/1uOJUvgjPDwP-PRYkmlypOAPtJ9s4nWYm6WRfrAAsn0g/edit?usp=sharing

src/rez/package_resources.py
chadrik: Removed

src/rez/solver.py
chadrik: I removed all the runtime changes from this module

src/rez/package_filter.py
chadrik: I removed this change and changes like it

src/rez/packages.py
chadrik: This is a safe change so I left it.

src/rez/solver.py
chadrik: I've made some very educated guesses at the correct types for these, and they are getting additional validation as the annotation coverage increases.  I'd like to leave these as-is for now because I think the annotations are more accurate than the variable names.

chadrik:
Any chance that this can get merged?  I’ve removed everything but the addition of annotations.

JeanChristopheMorinPerso:
Hey @chadrik when you get a chance, could you rebase your PR please? Now that we dropped support for Python 3.7, your PR is next in line to be released.

chadrik:
@JeanChristopheMorinPerso this is rebased and all tests are passing.  I just have to get the CLA fixed up.

JeanChristopheMorinPerso:
Yay! Thank you @chadrik! This will go in the next release. I'll do one last review (maybe later this week) before merging, just in case.

Thanks again!

src/rez/utils/platform_.py
JeanChristopheMorinPerso: I feel like we should fix the TODO and raise, but considering this PR as been up for quite a long time (sorry for that), let's do that in a separate PR.

src/rez/utils/resources.py
JeanChristopheMorinPerso: I'll close my eyes on this for now.

src/rez/version/_version.py
```diff
@@ -2,51 +2,59 @@
...
         return not self < other
 
 
 class _ReversedComparable(_Common):
-    def __init__(self, value):
+    def __init__(self, value: _Comparable) -> None:
         self.value = value
 
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
```
JeanChristopheMorinPerso, line 39: In all the other classes in this module, you use the actual class for `other` in the operators. Is there a reason why you use `object` here instead of the class?

src/rez/version/_version.py
```diff
@@ -576,6 +601,20 @@ def intersection(self, other):
 _Bound.any = _Bound()
 
 
+def action(fn: CallableT) -> CallableT:
```
JeanChristopheMorinPerso, line 604:
```suggestion
def _action(fn: CallableT) -> CallableT:
```
I feel like this should be a private function.

src/rez/version/_version.py
```diff
@@ -712,45 +753,33 @@ def __init__(self, input_string, make_token, invalid_bound_error=True):
...
-            if self.debug:
-                label = fn.__name__.replace("_act_", "")
-                print("%-21s: %s" % (label, self._input_string))
-                for key, value in self._groups.items():
-                    print("    %-17s= %s" % (key, value))
-                print("    %-17s= %s" % ("bounds", self.bounds))
-            return result
-        return fn_
-
     @action
```
JeanChristopheMorinPerso, line 765:
```suggestion
    @_action
```

src/rez/version/_version.py
```diff
@@ -712,45 +753,33 @@ def __init__(self, input_string, make_token, invalid_bound_error=True):
...
-    def _act_version(self):
+    def _act_version(self) -> None:
         version = self._create_version_from_token(self._groups['version'])
         lower_bound = _LowerBound(version, True)
         upper_bound = _UpperBound(version.next(), False) if version else None
 
         self.bounds.append(_Bound(lower_bound, upper_bound))
 
     @action
```
JeanChristopheMorinPerso, line 773:
```suggestion
    @_action
```

src/rez/version/_version.py
```diff
@@ -712,45 +753,33 @@ def __init__(self, input_string, make_token, invalid_bound_error=True):
...
-    def _act_exact_version(self):
+    def _act_exact_version(self) -> None:
         version = self._create_version_from_token(self._groups['exact_version_group'])
         lower_bound = _LowerBound(version, True)
         upper_bound = _UpperBound(version, True)
 
         self.bounds.append(_Bound(lower_bound, upper_bound))
 
     @action
```
JeanChristopheMorinPerso, line 781:
```suggestion
    @_action
```

src/rez/version/_version.py
```diff
@@ -760,23 +789,23 @@ def _act_bound(self):
         self.bounds.append(_Bound(lower_bound, upper_bound, self.invalid_bound_error))
 
     @action
```
JeanChristopheMorinPerso, line 791:
```suggestion
    @_action
```

src/rez/version/_version.py
```diff
@@ -760,23 +789,23 @@ def _act_bound(self):
...
-    def _act_lower_bound(self):
+    def _act_lower_bound(self) -> None:
         version = self._create_version_from_token(self._groups['lower_version'])
         exclusive = self._is_lower_bound_exclusive(self._groups['lower_bound_prefix'])
         lower_bound = _LowerBound(version, not exclusive)
 
         self.bounds.append(_Bound(lower_bound, None))
 
     @action
```
JeanChristopheMorinPerso, line 799:
```suggestion
    @_action
```

src/rez/version/_version.py
```diff
@@ -760,23 +789,23 @@ def _act_bound(self):
...
-    def _act_upper_bound(self):
+    def _act_upper_bound(self) -> None:
         version = self._create_version_from_token(self._groups['upper_version'])
         exclusive = self._is_upper_bound_exclusive(self._groups['upper_bound_prefix'])
         upper_bound = _UpperBound(version, not exclusive)
 
         self.bounds.append(_Bound(None, upper_bound))
 
     @action
```
JeanChristopheMorinPerso, line 807:
```suggestion
    @_action
```

src/rez/version/_version.py
```diff
@@ -793,7 +822,7 @@ def _act_lower_and_upper_bound_asc(self):
         self.bounds.append(_Bound(lower_bound, upper_bound, self.invalid_bound_error))
 
     @action
```
JeanChristopheMorinPerso, line 824:
```suggestion
    @_action
```

src/rez/build_process.py
JeanChristopheMorinPerso: Thank you. I guess we could add this to our contribution docs.

src/rez/build_process.py
```diff
@@ -357,12 +378,12 @@ def run_hooks(self, hook_event, **kwargs):
                     "%s cancelled by %s hook '%s': %s:\n%s"
                     % (hook_event.noun, hook_event.label, hook.name(),
                        e.__class__.__name__, str(e)))
-            except RezError:
+            except RezError as e:
```
JeanChristopheMorinPerso, line 381: Nice catch!

src/rez/config.py
```diff
@@ -54,20 +65,20 @@ class Setting(object):
     Note that lazy setting validation only happens on main configuration
     settings - plugin settings are validated on load only.
     """
-    schema = Schema(object)
+    schema: Validatable = Schema(object)
```
JeanChristopheMorinPerso, line 68: I'm guessing you are using `Validatable` because `Schema` isn't typed at all?

src/rez/package_repository.py
```diff
@@ -29,20 +39,21 @@ def create_memory_package_repository(repository_data):
     Returns:
         `PackageRepository` object.
     """
+    from rezplugins.package_repository.memory import MemoryPackageRepository  # noqa
```
JeanChristopheMorinPerso, line 42: Is this import necessary?

src/rez/resolved_context.py
JeanChristopheMorinPerso: Ok, I took a look and I think you are right. I was afraid that something down the line would use the `.ephemeral` attribute of `PackageRequest`, but it doesn't look like it. And Honestly, the ResolvedContext is cursed. `PackageRequest` exists to validate the name in the request, but it's kind of used both as a validation layer and just as a simple Requirement. Its usage is pretty inconsistent, and it's also used in places where I wouldn't expect it to be used.

src/rez/resolved_context.py
```diff
@@ -558,10 +605,9 @@ def get_patched_request(self, package_requests=None,
         """
         # assemble source request
         if strict:
-            request = []
+            request: list[Requirement | PackageRequest] = []
             for variant in self.resolved_packages:
                 req = PackageRequest(variant.qualified_package_name)
-                request.append(req)
```
JeanChristopheMorinPerso, line 564: This seems weird to me. I believe this was removed by mistake?

src/rezplugins/build_process/local.py
```diff
@@ -5,6 +5,8 @@
 """
 Builds packages on local host
 """
+from __future__ import annotations
+
 from rez.config import config
 from rez.package_repository import package_repository_manager
 from rez.build_process import BuildProcessHelper, BuildType
```
JeanChristopheMorinPerso, line 12:
```suggestion
from rez.build_process import BuildProcessHelper, BuildType
from rez.build_system import BuildResult
```

src/rezplugins/build_process/local.py
```diff
@@ -21,11 +23,24 @@
...
+    from rez.build_system import BuildResult
+
+    # FIXME: move this out of TYPE_CHECKING block when python 3.7 support is dropped
+    class LocalBuildResult(BuildResult, total=False):
+        package_install_path: str
+        variant_install_path: str
+
+else:
+    LocalBuildResult = dict
+
```
JeanChristopheMorinPerso, line 44:
```suggestion


class LocalBuildResult(BuildResult, total=False):
    package_install_path: str
    variant_install_path: str

```

src/rezplugins/shell/_utils/windows.py
```diff
@@ -45,26 +47,28 @@ def to_windows_path(path):
...
-        try:
-            with winreg.OpenKey(root_key, sub_key) as key:
-                reg_value, _ = winreg.QueryValueEx(key, 'Path')
-        except OSError:
-            # Key does not exist
-            pass
-        else:
-            expanded_value = winreg.ExpandEnvironmentStrings(reg_value)
-            paths.extend(expanded_value.split(os.pathsep))
+    if sys.platform == "win32":
```
JeanChristopheMorinPerso, line 57: Why is this if needed?

JeanChristopheMorinPerso (commented):
Thanks @chadrik I left a few comments. Most are very easy to address. I also left a few questions that I would like to get an answer on before merging, if possible.

I promise, this is the last batch of comments and questions :)

src/rez/config.py
JeanChristopheMorinPerso: Ok, I checked and I understand now. `Or`, doesn't inherit from `Schema`. So it makes sense.

src/rez/package_repository.py
JeanChristopheMorinPerso: Not that I see. I'll remove it.

src/rez/version/_version.py
JeanChristopheMorinPerso: Looked at the mypy output, it actually complains when object isn't used. So I'll switch to object in comparators and add proper isinstance checks.

src/rez/resolved_context.py
JeanChristopheMorinPerso: I will add it back.

src/rezplugins/build_process/local.py
JeanChristopheMorinPerso: I'll apply this change.

src/rezplugins/shell/_utils/windows.py
JeanChristopheMorinPerso:
```
src/rezplugins/shell/_utils/windows.py:59: error: Module has no attribute "HKEY_LOCAL_MACHINE"  [attr-defined]
src/rezplugins/shell/_utils/windows.py:60: error: Module has no attribute "HKEY_CURRENT_USER"  [attr-defined]
src/rezplugins/shell/_utils/windows.py:65: error: Module has no attribute "OpenKey"  [attr-defined]
src/rezplugins/shell/_utils/windows.py:66: error: Module has no attribute "QueryValueEx"  [attr-defined]
src/rezplugins/shell/_utils/windows.py:71: error: Module has no attribute "ExpandEnvironmentStrings"  [attr-defined]
```
This is why. I'll add a comment.

JeanChristopheMorinPerso (approved):
This looks good to me!

Thank you so much for putting all the efforts to get this done. Also, I want to say that we are sorry that it took us so long to get this to the finish line.

This PR makes me very excited and also annoyed at our own code base even more than I was before. If we want to have a code base with clean types, we'll have to come up with a strategy because we'll likely need to break a few API promises along the way...

Anyway, thanks again.

chadrik:
>  If we want to have a code base with clean types, we'll have to come up with a strategy because we'll likely need to break a few API promises along the way...

Yes, it becomes clear after wrestling with this for a bit that it's a puzzle with no solution, not as the pieces are now.  We will need to modify some of the puzzle pieces to make it all fit.

PR #1475, opened 2023-04-21
Fixes #1302

Jawabiscuit:
Fixes #1302 by adding gitbash support.

This work is based on https://github.com/AcademySoftwareFoundation/rez/pull/1364 and aims to take work started there to the finish line.

PR #1775, opened 2024-06-20
Fix missing path in GitBash warning message

brycegbrazen:
When testing out Gitbash, I noticed there was a print that was not being formatted. This fixes the issue.

### DIFF
--- a/src/rezplugins/shell/gitbash.py
+++ b/src/rezplugins/shell/gitbash.py
@@ -44,7 +44,7 @@ class GitBash(Bash):
                 "Git-bash executable has been detected at %s, but this is "
                 "probably not correct (google Windows Subsystem for Linux). "
                 "Consider adjusting your searchpath, or use rez config setting "
-                "plugins.shell.gitbash.executable_fullpath."
+                "plugins.shell.gitbash.executable_fullpath." % exepath
             )
 
         return exepath


JeanChristopheMorinPerso (approved):

src/rezplugins/shell/gitbash.py
```diff
@@ -44,7 +44,7 @@ def find_executable(cls, name, check_syspaths=False):
                 "Git-bash executable has been detected at %s, but this is "
                 "probably not correct (google Windows Subsystem for Linux). "
                 "Consider adjusting your searchpath, or use rez config setting "
-                "plugins.shell.gitbash.executable_fullpath."
+                "plugins.shell.gitbash.executable_fullpath." % exepath
```
JeanChristopheMorinPerso, line 47: Are you sure that this works? Should it be after the closing parenthesis?

brycegbrazen: Yep it works, here's a sample output from my machine: 

```13:58:15 WARNING  Git-bash executable has been detected at c:\windows\system32\bash.exe, but this is probably not correct (google Windows Subsystem for Linux). Consider adjusting your searchpath, or use rez config setting plugins.shell.gitbash.executable_fullpath.```

Here's a couple other examples in the repo for reference:

https://github.com/AcademySoftwareFoundation/rez/blob/a13f7bb362cbb05e73f88e9f771c36005c1e9f4a/install.py#L83
https://github.com/AcademySoftwareFoundation/rez/blob/a13f7bb362cbb05e73f88e9f771c36005c1e9f4a/src/rez/cli/cp.py#L93
https://github.com/AcademySoftwareFoundation/rez/blob/a13f7bb362cbb05e73f88e9f771c36005c1e9f4a/src/rezgui/app.py#L19

brycegbrazen:
I see this is also fixed in #1475. Might just be worth waiting for that one as they fixed it a bit differently.

JeanChristopheMorinPerso (approved):

Issue #1279, opened 2022-04-07
Add support for 3rd party prompt implementations in pwsh / Powershell


instinct-vfx:
The current implementation to set the prompt on windows creates a custom prompt function. This overrides existing prompt functions a user may have set including 3rd party prompts like https://ohmyposh.dev/

If a user has his prompt customized, dropping himself into a rez created shell removes all of the user's setup.

I would suggest adding a new PowershellBase setting that allows switching between actually setting a prompt function (the current behavior) to setting it as an environment variable only (so it can be used by 3rd party prompts easily).

PR incoming soon unless someone objects

instinct-vfx:
For the record, while starting on this i realised that the current implementation does not support nesting at all. Had not occured to me before which is kind of weird. So adding support for nested prompts too.

PR #1777, opened 2024-06-21
Improve PowerShell plugin's profile handling functionality

nrusch:
Improvements to the PowerShell plugin to support profile-related settings/options.

### Add support for the `package_commands_sourced_first` config/runtime option

Prior to this change, the rez PowerShell implementation would always source the rez context script after the user/host profile scripts, so profile-level modifications to environment variables like `PATH` were always squashed by the unconditional overrides in the context script.

With this change, the relative source order of the shell profile vs. the context script can be properly controlled using the `package_commands_sourced_first` config option, matching the behavior of `SH`-based shells.

**NOTE:** Because this option defaults to True, this also implicitly changes rez's default behavior to source the shell profile *after* the context when using PowerShell.

### Support the `norc` shell plugin option.

This enables the use of `rez-env --norc` to bypass the sourcing of any profile scripts when using PowerShell.

### DIFF
--- a/src/rezplugins/shell/_utils/powershell_base.py
+++ b/src/rezplugins/shell/_utils/powershell_base.py
@@ -25,6 +25,15 @@ class PowerShellBase(Shell):
     expand_env_vars = True
     syspaths = None
 
+    # PowerShell does not provide built-in functionality to manually trigger
+    # sourcing of the user/host profile scripts in the appropriate order, so we
+    # have to implement it ourselves in order to be able to control the relative
+    # source order of the user's profile vs. the generated rez context script.
+    profile_source_cmd = (
+        '$PROFILE '
+        '| Get-Member -type NoteProperty '
+        '| ForEach-Object { if (Test-Path $PROFILE."$($_.Name)") { . $PROFILE."$($_.Name)" }}')
+
     # Make sure that the $Env:VAR formats come before the $VAR formats since
     # PowerShell Environment variables are ambiguous with Unix paths.
     # Note: This is used in other parts of Rez
@@ -51,10 +60,8 @@ class PowerShellBase(Shell):
                              stdin: bool = False,
                              command: bool = False):
         cls._unsupported_option('rcfile', rcfile)
-        cls._unsupported_option('norc', norc)
         cls._unsupported_option('stdin', stdin)
         rcfile = False
-        norc = False
         stdin = False
         return (rcfile, norc, stdin, command)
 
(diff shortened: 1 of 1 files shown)


src/rezplugins/shell/_utils/powershell_base.py
```diff
@@ -23,6 +23,15 @@ class PowerShellBase(Shell):
...
     syspaths = None
 
+    # PowerShell does not provide built-in functionality to manually trigger
+    # sourcing of the user/host profile scripts in the appropriate order, so we
+    # have to implement it ourselves in order to be able to control the relative
+    # source order of the user's profile vs. the generated rez context script.
+    profile_source_cmd = (
+        '$PROFILE '
+        '| Get-Member -type NoteProperty '
+        '| ForEach-Object { if (Test-Path $PROFILE."$($_.Name)") { . $PROFILE."$($_.Name)" }}')
```
nrusch, line 33: Note that this is a dynamic approach to discovering available profile scripts, but it may be preferable to hardcode iteration over the known profile slots [documented here](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_profiles?view=powershell-7.4#profile-types-and-locations).

This could look something like this (+/- formatting):
```pwsh
foreach ($name in 'AllUsersAllHosts', 'AllUsersCurrentHost', 'CurrentUserAllHosts', 'CurrentUserCurrentHost')
{
    if (Test-Path $PROFILE.$name) {
        . $PROFILE.$name
    }
}
```

nrusch:
This may indirectly "resolve" #1279, since the user's profile will have the final say on any shell customization.

nrusch:
The CLA has been sorted for rez. I'm not sure why this PR's status isn't updating though...

JeanChristopheMorinPerso:
@nrusch it should be good now. I manually triggered a refresh and it's now green.

nrusch:
@JeanChristopheMorinPerso I'm going back through some old PRs and wanted to see what the current temperature might be on this being acceptable.

JeanChristopheMorinPerso:
@nrusch can you look at https://ampcode.com/threads/T-01a0b7c8-3bd2-759a-886c-ec9aa1d920e2 please? I think these are valid findings but I'm not 100% sure.

nrusch:
@JeanChristopheMorinPerso Sorry, but I can't see that link.

JeanChristopheMorinPerso:
Umm, not too sure why since I made it public. Anyway, here's what I wanted you to look at (note that this is AI generated. I verified the content and I think they are valid concerns and want your feedback):

1. **P1 — A terminating profile error now aborts Rez startup.**
   [Profile-loading command, line 35](https://github.com/nrusch/rez/blob/improve_pwsh_profile_handling/src/rezplugins/shell/_utils/powershell_base.py#L32-L35)

   A profile containing `throw "profile failed"` previously printed the error, then continued loading subsequent profiles, the Rez context, and the requested command. The new dot-sourcing pipeline lets that exception escape into `rez-shell.ps1`.

   - **Command mode:** the requested command never executes; PowerShell exits with status 1.
   - **Interactive mode:** subsequent profiles and Rez initialization are skipped. With `package_commands_sourced_first=False`, the shell opens **without loading the Rez context**.

   Preserve PowerShell’s per-profile error isolation, reporting the failure while allowing remaining startup steps to run. Add regression coverage for both source orders and both invocation modes.
2. **P2 — Profile paths containing square brackets are silently skipped.**
   [`Test-Path` invocation, line 35](https://github.com/nrusch/rez/blob/improve_pwsh_profile_handling/src/rezplugins/shell/_utils/powershell_base.py#L35)

   `Test-Path` interprets its positional argument as a wildcard pattern. With a home directory named `user[1]`, both user profiles exist but the checks return false. Because automatic loading is now disabled, neither profile runs. The base version loads both successfully.

   Use `Test-Path -LiteralPath` and cover a profile directory containing brackets.

nrusch:
Ah! Thanks for sharing those. I'll take a look.

PR #1778, opened 2024-06-21
Powershell alias return correct exit codes

### LINKED ISSUE #1776: Using aliases in Powershell muffles exit code
Running an alias using the Powershell shell plugin with `rez-env` doesn't return the correct exit code when running the command in-line.

**Environment**
* Windows 10
* Rez 2.112.0
* Rez python version 3.7.9

**To Reproduce**
1. Create a `package.py` for a `test_alias` package with the following definition:
```
name = "test_alias"

version = "1.0.0"

build_command = ""

def commands():
    alias('make_error', 'python')
```
2. Install the package to your package repository.
3. Run `rez-env -- python -c "raise"`
4. Observe $LASTEXITCODE as being 1 as expected.
5. Run `rez-env test_alias -- make_error -c "raise"`
6. Observe $LASTEXITCODE as being 0 instead of the correct value of 1.

**Expected behavior**
$LASTEXITCODE is 1.

**Actual behavior**
$LASTEXITCODE is 0.

brycegbrazen:
Fixes issue where powershell plugins were not correctly capturing exit codes when aliased commands failed.

Closes #1776.

### DIFF
--- a/src/rez/tests/test_shells.py
+++ b/src/rez/tests/test_shells.py
@@ -597,10 +597,10 @@ class TestShells(TestBase, TempdirMixin):
         config.override("default_shell", shell)
 
         def _make_alias(ex):
-            ex.alias('python_alias', 'python')
+            ex.alias('my_alias', 'hello_world -r 1')
 
-        r = self._create_context([])
-        p = r.execute_shell(command='python_alias -c raise',
+        r = self._create_context(["hello_world"])
+        p = r.execute_shell(command='my_alias',
                             actions_callback=_make_alias,
                             stdout=subprocess.PIPE)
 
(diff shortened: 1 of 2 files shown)


brycegbrazen:
Looks like Python isn't installed to all of our action VMs? Not sure what other command I could use that's shell agnostic outside of `echo`. I don't think there's a way to make `echo` fail as I need this test to.

JeanChristopheMorinPerso:
@brycegbrazen these tests are complex, so reuse what's already there and don't call python directly. Look at the `test_command` test. It uses a pre-configured `hello_world` command that I think can be instructed to exit with a specific exit code.

JeanChristopheMorinPerso:
https://github.com/AcademySoftwareFoundation/rez/blob/main/src%2Frez%2Fbind%2Fhello_world.py#L34

JeanChristopheMorinPerso (changes requested):

JeanChristopheMorinPerso (approved):

instinct-vfx (approved):

PR #1787, opened 2024-07-02
refactor: move package_order to rezplugin system

cfxegbert:
I was trying to write a new package orderer and discovered that it is not part of the rezplugin system.  This refactor moves the orderers defined in package_order.py to rezplugins.package_order plugins.  Function have been added to package_order.py to remain backward compatible so if a user was directly importing an orderer it should still work.

### DIFF
--- a/src/rez/package_order.py
+++ b/src/rez/package_order.py
@@ -7,10 +7,10 @@ from hashlib import sha1
 from typing import Dict, Iterable, List, Optional, Union
 
 from rez.config import config
+from rez.exceptions import RezPluginError
 from rez.utils.data_utils import cached_class_property
 from rez.version import Version, VersionRange
-from rez.version._version import _Comparable, _ReversedComparable, _LowerBound, _UpperBound, _Bound
-from rez.packages import iter_packages
+from rez.version._version import _Comparable, _LowerBound, _UpperBound, _Bound
 
 ALL_PACKAGES = "*"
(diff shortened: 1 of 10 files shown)


JeanChristopheMorinPerso:
Hi @cfxegbert. Can you please explain the rationale behind this change? Context is important for us. It helps us document why things were changed, by who, the pros, cons, etc.

cfxegbert:
This was moved to the rez plugin system so a new package order could be written without having to modify the rez source code.  Before a developer would have to patch the source code and apply the patch every release.  By moving to the plugin system new package orders could be added without modifing the source.

cfxegbert:
I could also add the rez plugin hooks and not move the existing package orders if you think that would be better.

chadrik:
Scanline VFX is strongly in support of this feature.

nrusch:
I just wanted to check in and see if there was any chance of un-sticking this PR, either by moving it forward or definitively rejecting it.

We use multiple custom package orderers right now, which requires maintaining a branch on top of the `rez` baseline, but it would be great to be able to supply these externally via entry points.

maxnbk:
I'm personally invested in unsticking this, but I have an ASWF-Slack DM sent out to @cfxegbert to ideally 1:1 discuss a couple different options for ways forward owing to the specific state of the code and PR at the moment.

PR #1788, opened 2024-07-04
Fix command.py

predat:
extension have been replace by command


JeanChristopheMorinPerso:
Thanks @predat! Would you be willing to add a test? It should be relatively straight forward.

Issue #1800, opened 2024-07-23
PowerShell alias calling non-existent command returns 0 exit code


brycegbrazen:
Running an alias using the Powershell shell plugin with `rez-env` doesn't return a failure exit code when attempting to run a non-existent aliased command in-line.

**Environment**
* Windows 10
* Rez 2.112.0
* Rez python version 3.7.9

**To Reproduce**
1. Create a `package.py` for a `test_invalid_alias` package with the following definition:
```
name = "test_invalid_alias"

version = "1.0.0"

build_command = ""

def commands():
    alias('my_alias', 'a_non_existent_executable')
```
2. Install the package to your package repository.
3. Run `rez-env -- a_non_existent_executable`
4. Observe the following error message:
```
& : The term 'a_non_existent_executable' is not recognized as the name of a cmdlet, function, script file, or operable
program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At C:\Users\bryce.gattis\AppData\Local\Temp\rez_context_2bfjto0k\rez-shell.ps1:4 char:3
+ & a_non_existent_executable
+   ~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (a_non_existent_executable:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```
5. Run `rez-env test_invalid_alias -- my_alias`
6. Observe $LASTEXITCODE as being 0 instead of a non-zero exit code.

**Use Case**
This is currently happening and causing issues on our render farm. With Rez, we are creating aliases for `python` that map to whatever the DCC's version of `python` is. For example, the `python` alias for Houdini maps to `hython`. 

The behavior observed via this issue is problematic when a specific machine doesn't have the correct version of Houdini installed on their machine. Rez will still use the alias (and expect that the corresponding software is installed where the Rez package is looking for it), but will not correctly return a non-zero exit code if the executable is not found.

This causes false positives on our render farm.

**Related Issues/PRs**
* I will like a PR that adds context on what the issue is in the code shortly.

**Expected behavior**
A non-zero exit code is returned.

**Actual behavior**
A 0 exit code is returned.

dbr:
I had similar problem exposing commands like `nuke` as an alias for `Nuke15.0` etc

The aliases don't seem to propagate any error codes (so even if the command starts, but say it fails mid-render - then farm monitoring software doesn't notice the job failed)

[...]

PR #1801, opened 2024-07-23
PowerShell alias calling non-existent command returns 0 exit code failing test

brycegbrazen:
Adds failing test for #1800.

After digging into this there are three important Powershell [automatic variables](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_automatic_variables?view=powershell-5.1#error) to take note of.

- `$LASTEXITCODE`
- `$?`
- `$ERROR`

The first two, `$LASTEXITCODE` and `$?` are used in [the code](https://github.com/AcademySoftwareFoundation/rez/blob/main/src/rezplugins/shell/_utils/powershell_base.py#L176) currently. 

However, after testing the issue in PowerShell directly, it seems that the issue stems from an issue stemming from PowerShell itself. 

When using aliases in PowerShell, if the command being called is not found, and error is printed to stderr, but neither the `$LASTEXITCODE` or `$?` variables are filled out correct to represent this.

For example, in PowerShell I ran:
```
PS C:\Users\bryce.gattis> hython
hython : The term 'hython' is not recognized as the name of a cmdlet, function, script file, or operable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:1
+ hython
+ ~~~~~~
    + CategoryInfo          : ObjectNotFound: (hython:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
PS C:\Users\bryce.gattis> $?
False
PS C:\Users\bryce.gattis> $LASTEXITCODE
```

Then I ran the following that does the same thing but using aliases:
```
PS C:\Users\bryce.gattis> function python() { hython @args }
PS C:\Users\bryce.gattis> python
hython : The term 'hython' is not recognized as the name of a cmdlet, function, script file, or operable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:21
+ function python() { hython @args }
+                     ~~~~~~
    + CategoryInfo          : ObjectNotFound: (hython:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
PS C:\Users\bryce.gattis> $?
True
PS C:\Users\bryce.gattis> $LASTEXITCODE
```

Interestingly, it seems like the `$ERROR` automatic variable, correctly captures the error message in both cases.
```
PS C:\Users\bryce.gattis> $ERROR
hython : The term 'hython' is not recognized as the name of a cmdlet, function, script file, or operable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:21
+ function python() { hython @args }
+                     ~~~~~~
    + CategoryInfo          : ObjectNotFound: (hython:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

Not sure what the implications would be of _additionally_ relying on the presence of an error in the `$ERROR` variable, but seems to be something that is slipping through the cracks right now. Perhaps there's a better fix for this?

### DIFF
--- a/src/rez/tests/test_shells.py
+++ b/src/rez/tests/test_shells.py
@@ -591,6 +591,22 @@ class TestShells(TestBase, TempdirMixin):
         out, _ = p.communicate()
         self.assertEqual(0, p.returncode)
 
+    @per_available_shell()
+    def test_alias_non_existent_command_return_code(self, shell):
+        """Ensure return codes are correct while using aliases."""
+        config.override("default_shell", shell)
+
+        def _make_alias(ex):
+            ex.alias('my_alias', 'a_non_existent_executable')
+
+        r = self._create_context([])
+        p = r.execute_shell(command='my_alias',
+                            actions_callback=_make_alias,
+                            stdout=subprocess.PIPE)
+
+        out, _ = p.communicate()
+        self.assertNotEqual(0, p.returncode)
+
 
 if __name__ == '__main__':
     unittest.main()


JeanChristopheMorinPerso:
Rebased to fix conflict

brycegbrazen:
Had to change some `-ErrorAction` flags to `Ignore` so that expected errors would not be caught in the `$Error` array. Based on unit tests, it doesn't look like this change affects anything, but does fix the issue this PR is concerned with.

Let me know what you think @JeanChristopheMorinPerso!

Issue #1414, opened 2022-11-22
rez-pip does not create variants with a consistent order, potentially causing duplicates.


jopollack:
Unsure of what specific code is going wrong under the hood, but rez-pip is clearly not ordering the list of packages in its variants prior to creating the package, which can create duplicates with slightly different ordering. In this case, I installed google_api_core twice, and was confused when the second wasn't ignored for already existing. Then I noticed the variants looked like this:

```
variants = [
    ['python-2.7', 'protobuf-3.12.0+<3.18.0', 'futures-3.2.0+'],
    ['python-2.7', 'futures-3.2.0+', 'protobuf-3.12.0+<3.18.0']
]
```

**Environment**
* OS Linux CentOS7
* Rez version 2.106.0
* Rez python version 3.7.7

**To Reproduce**
1. rez-pip google_api_core -i -p /temp_directory/rez --python-version 2.7
2. repeat the above again.

**Expected behavior**
The same install steps should be recognized as identical and skipped as already existing.

**Actual behavior**
The package gets installed again, with identical variants in a slightly different order, which I expect to cause unexpected and undesirable behavior when attempting to resolve.

samson-jerome:
Hello, any update on this ?

PR #1829, opened 2024-09-20
Order pip requirements when converting to keep a constant hash

samson-jerome:
When using rez pip, the order of requirements collected from distributions may vary. This eventually changes the variants hash generated which leads to reinstalling a package already installed (with a different hash but identical requirements).

closes: https://github.com/AcademySoftwareFoundation/rez/issues/1414


samson-jerome:
I'm in the process of having the CLA signed for our studio...

samson-jerome:
Can't get quick EasyCLA approval

PR #1845, opened 2024-10-11
fix: remove python-2 pyyaml lib

maxnbk:
Just reducing the rez repo & package footprint by a bit, considering we have dropped py2.
(probably should update pyyaml to 6 too, but separate PR)


JeanChristopheMorinPerso (approved):
Oh, I forgot to do that. Thanks!

PR #1853, opened 2024-10-14
Add support for synchronous and asynchronous modes in package cache

Pixel-Minions:
- Introduced a new command-line argument `--pkg-cache-mode` to specify caching mode.
- Allows users to choose between 'sync' and 'async' modes, overriding the default configuration.
- Updated `add_variant` function to handle the new caching mode option.
- Ensures backward compatibility by maintaining default behavior when no mode is specified.

### DIFF
--- a/src/rez/cli/pkg-cache.py
+++ b/src/rez/cli/pkg-cache.py
@@ -42,6 +42,14 @@ def setup_parser(parser, completions=False):
         "--daemon", action="store_true", help=SUPPRESS
     )
     parser.add_argument(
+        "--pkg-cache-mode",
+        choices=["sync", "async"],
+        default=None,
+        help="If provided, override the rezconfig's package_cache_async key. "
+             "If 'sync', the process will block until packages are cached. "
+             "If 'async', the process will not block while packages are cached."
+    )
+    parser.add_argument(
         "-c", "--columns", nargs='+', choices=column_choices,
         default=["status", "package", "variant_uri", "cache_path"],
         help="Columns to print, choose from: %s" % ", ".join(column_choices)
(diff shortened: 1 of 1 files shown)


src/rez/cli/pkg-cache.py
```diff
@@ -70,7 +78,15 @@ def add_variant(pkgcache, uri, opts):
...
-    destpath, status = pkgcache.add_variant(variant, force=opts.force)
+    if opts.pkg_cache_mode is not None:
+        cache_mode = True if opts.pkg_cache_mode == "sync" else False
+        destpath, status = pkgcache.add_variant(
+            variant, force=opts.force,
+            wait_for_copying=cache_mode
+        )
+    # If no mode is specified, use the default behavior.
+    else:
+        destpath, status = pkgcache.add_variant(variant, force=opts.force)
```
JeanChristopheMorinPerso, line 89:
```suggestion
    if opts.pkg_cache_mode == "async":
        cache_mode = True
    elif opts.pkg_cache_mode == "sync":
        cache_mode = False
    else:
        cache_mode = not config.package_cache_async

    destpath, status = pkgcache.add_variant(
        variant, force=opts.force,
        wait_for_copying=cache_mode
    )
```
This would more properly do what we want. The code you wrote would default to the function's default which would not necessarily be `config.package_cache_async`.

JeanChristopheMorinPerso (changes requested):

src/rez/cli/pkg-cache.py
```diff
@@ -70,7 +78,17 @@ def add_variant(pkgcache, uri, opts):
         print("No such variant: %s" % uri, file=sys.stderr)
         sys.exit(1)
 
-    destpath, status = pkgcache.add_variant(variant, force=opts.force)
+    if opts.pkg_cache_mode == "async":
+        cache_mode = True
+    elif opts.pkg_cache_mode == "sync":
+        cache_mode = False
```
JeanChristopheMorinPerso, line 84:
```suggestion
        cache_mode = False
    elif opts.pkg_cache_mode == "sync":
        cache_mode = False
```
Umm, I think this would be the right thing to do. If it's async, we don't want to wait (so False).

JeanChristopheMorinPerso (approved):
Thank you @Pixel-Minions!

PR #1860, opened 2024-10-19
Enable Google Analytics using an extension

JeanChristopheMorinPerso:
ReadTheDocs used to support Google Analytics natively. But they no more do since July 2024. See https://github.com/readthedocs/readthedocs.org/issues/9530#issuecomment-2233541583.

They suggest using the `sphinxcontrib-googleanalytics` extension instead.

This will fix our reports which currently show 0 views:
image

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,4 @@
 sphinx >=7.0.0,<9.0.0
 furo
 myst-parser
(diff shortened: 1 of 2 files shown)


instinct-vfx (commented):
lgtm

JeanChristopheMorinPerso:
Thanks for taking a look! (and sorry for merging so fast)

PR #1863, opened 2024-10-26
Fix key error on getting the "testing" key on older contexts

### LINKED ISSUE #1862: Bake resolves with rez<3.2.0 can't be loaded due to missing `testing` key
Baked resolves generated with older versions are not compatible with rez-3.2.0.
Note a regression per say but a breaking change 

**Environment**
* windows 10
* 3.2.0
* Python 3.11.4

**To Reproduce**
1. using rez-3.0.0, bake a simple resolve into a rxt file
```
> rez --version
Rez 3.0.0 from c:\rez\3.0.0\Lib\site-packages\rez (python 3.11)
> rez env python --output py.rxt
```

2. using rez-3.2.0, resolve a context from the above rxt file
```
> rez env --input py.rxt
12:10:10 ERROR    ResolvedContextError: Failed to load context from py.rxt: KeyError: 'testing'
```

**Expected behavior**
We should be left in a resolved context

**Actual behavior**
Resolution failure.

**Related Issues/PRs**
* #1740

JeanChristopheMorinPerso:
Fixes #1862.

Now with a test to make sure we never regress.

### DIFF
--- a/src/rez/resolved_context.py
+++ b/src/rez/resolved_context.py
@@ -134,7 +134,7 @@ class ResolvedContext(object):
     command within a configured python namespace, without spawning a child
     shell.
     """
-    serialize_version = (4, 8)
+    serialize_version = (4, 9)
     tmpdir_manager = TempDirs(config.context_tmpdir, prefix="rez_context_")
     context_tracking_payload = None
     context_tracking_lock = threading.Lock()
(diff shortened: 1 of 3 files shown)


src/rez/resolved_context.py
```diff
@@ -134,7 +134,7 @@ class ResolvedContext(object):
     command within a configured python namespace, without spawning a child
     shell.
     """
-    serialize_version = (4, 8)
+    serialize_version = (4, 9)
```
JeanChristopheMorinPerso, line 137: I'm not too sure if updating the version makes sense now. It should have been updated in the 3.2.0 release.

maxnbk: It does make sense since at least a 3.0 would not be able to load with new testing key from this version, even though ideally the last version would have had it.

PR #1870, opened 2024-11-05
Add wildcard support on rez test command

Ni-g-3l:
Allow user to run multiple tests thanks wildcard

For example:
* `rez test testing_obj-1.0 command_as_*` will run `command_as_string_success` and command_as_string_fail`
* `rez test testing_obj-1.0 command_as_* check_car_ideas` will run `command_as_string_success`, `command_as_string_fail' and `check_car_idea`

### DIFF
--- a/src/rez/package_test.py
+++ b/src/rez/package_test.py
@@ -215,7 +215,7 @@ class PackageTestRunner(object):
         requested_test_names = set()
         for requested_test in requested_tests:
             requested_test_names.update(set(fnmatch.filter(pkg_test_names, requested_test)))
-        return sorted(requested_test_names)
+        return requested_test_names
 
     @property
     def num_tests(self):
--- a/src/rez/package_test.py
+++ b/src/rez/package_test.py
@@ -11,6 +11,7 @@ from rez.utils.colorize import heading, Printer
 from rez.utils.logging_ import print_info, print_warning, print_error
 from rez.version import Requirement, RequirementList
 from shlex import quote
+import fnmatch
 import time
 import sys
 import os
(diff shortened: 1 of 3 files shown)


JeanChristopheMorinPerso:
Hi @Ni-g-3l, thanks for creating this PR.

Before we can review and consider your changes, you will need to sign the CLA. See the comment above for details.

Let us know if you have questions.

src/rez/package_test.py
```diff
@@ -206,6 +207,16 @@ def get_test_names(self, run_on=None):
 
         return self.get_package_test_names(package, run_on=run_on)
 
+    def find_requested_test_names(self, requested_tests):
```
JeanChristopheMorinPerso, line 210: Umm, could we maybe add a new parameter to `get_test_names`? I'm not sure why we need a different method when `get_test_names` could have the ability to do all the filtering for us.

What do you think?

src/rez/package_test.py
```diff
@@ -206,6 +207,16 @@ def get_test_names(self, run_on=None):
...
+        # if no tests are explicitly specified, then run only those with a
+        # 'default' run_on tag
+        run_on = ["default"] if not requested_tests else None
+        pkg_test_names = self.get_test_names(run_on=run_on)
+        requested_test_names = set()
+        for requested_test in requested_tests:
+            requested_test_names.update(set(fnmatch.filter(pkg_test_names, requested_test)))
+        return sorted(requested_test_names)
```
JeanChristopheMorinPerso, line 218: I think we might want to keep the tests in the order the user requested them? I'm not convinced on that though.

JeanChristopheMorinPerso (commented):
I left some comments that I think should be easy to address.

Thank you so much for adding tests! I love it.

src/rez/package_test.py
Ni-g-3l: Thanks ! I made the update !

src/rez/package_test.py
Ni-g-3l: I think it's better to separate fetch tests from packages and filter them on regex in order keep available to fetch tests without any filter, but I could be wrong and this case won't appear the future. As you want I can make the fix I you want

JeanChristopheMorinPerso (approved):
Approving. We can always refine the methods later, now that we have a deprecation mechanism in place.

PR #1872, opened 2024-11-11
Register plugin from entry points

### LINKED ISSUE #1661: Simplify plugin development by supporting entrypoint based plugins
Developing plugins today is not very straightforward (and also undocumented). To make plugins available, you can install them directly under `rezplugins` or you can add the path in `REZ_PLUGIN_PATH`.

A plugin has to follow a strict folder structure.

I suggest that we add support for Python entrypoints to modernize our plugins a little bit and make it simpler to register plugins. See https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/ for documentation on how this can be done.

Ni-g-3l:
Hello

As suggest here I implemented the support of entrypoints in order to make the process of adding plugin to REZ easier.

Closes #1661

Feel free to add feedback  !

Have a great day


src/rez/plugin_managers.py
```diff
@@ -201,9 +176,55 @@ def load_plugins(self):
...
                         print_debug(out.getvalue())
 
-            # load config
-            data, _ = _load_config_from_filepaths([os.path.join(path, "rezconfig")])
-            deep_update(self.config_data, data)
+    def load_plugins_from_entry_points(self):
+        if sys.version_info.minor >= 8:
+            from importlib.metadata import entry_points
+        else:
+            from importlib_metadata import entry_points
```
JeanChristopheMorinPerso, line 183: This will force us to vendor `importlib-metadata` (which is fine since it's an APACHE-2.0 licensed).

src/rez/plugin_managers.py
```diff
@@ -201,9 +176,55 @@ def load_plugins(self):
...
+
+        discovered_plugins = entry_points(group='rez.plugins')
+        for plugin in discovered_plugins:
+            plugin = plugin.load()
+            plugin_name = plugin.__name__.split('.')[-1]
+            plugin_path = os.path.dirname(plugin.__file__)
+            self.register_plugin_module(plugin_name, plugin, plugin_path)
+            self.load_config_from_plugin(plugin)
```
JeanChristopheMorinPerso, line 191: We should put this code under a try/except block and add logs similar to logs found in `load_plugins_from_namespace`.

src/rez/tests/test_plugin_manager.py
```diff
@@ -68,6 +69,13 @@ def test_new_loading_style(self):
                 "package_repository", "cloud")
             self.assertEqual(cloud_cls.name(), "cloud")
 
+    def test_load_plugin_from_entry_points(self):
+        """Test loading rez plugin from setuptools entry points"""
+        with restore_pip("baz", os.path.join(self.data_path("extensions"), "baz")):
```
JeanChristopheMorinPerso, line 74: It would be safer to either just install into a temporary directory, or just commit the dist-info folder in the repo to avoid having to install it.

src/rez/data/tests/extensions/baz/baz/__init__.py
JeanChristopheMorinPerso, line 1: Why is the plugin defined twice?

src/rez/plugin_managers.py
```diff
@@ -201,9 +176,55 @@ def load_plugins(self):
...
+    def load_plugins_from_entry_points(self):
+        if sys.version_info.minor >= 8:
+            from importlib.metadata import entry_points
+        else:
+            from importlib_metadata import entry_points
+
+        discovered_plugins = entry_points(group='rez.plugins')
+        for plugin in discovered_plugins:
+            plugin = plugin.load()
+            plugin_name = plugin.__name__.split('.')[-1]
```
JeanChristopheMorinPerso, line 188: I'm not sure to understand why we need to do this. Can't we just use `plugin.name`? For the old plugin style, we needed to do `split('.')[-1]` because we were getting the name from `rezplugins.<type>.<name>`. But with entrypoints, we don't need to split.

JeanChristopheMorinPerso (changes requested):
Thanks @Ni-g-3l for creating this PR. I left a couple of comments that will need to be addressed before we can consider merging it.

Also, if you have the time and want to, we should adjust our newly written plugins doc (https://rez.readthedocs.io/en/stable/plugins.html).

Ni-g-3l:
Hello ! I fixed a big part of your comments, but I still not understand how you want resolve the plugin install with pip in tests

src/rez/tests/test_plugin_manager.py
JeanChristopheMorinPerso: To clarify this comment. Installing things on the fly should be avoided. It risks breaking other things if things go wrong. It will also make the test fail if run in parallel I think.

So my suggestion is to either `pip install --target` into a temporary directory and point sys.path to it, or to simply commit the `dist-info` folder (basically `pip install --target` into the rez repo) which would remove the need to pip install in the first place, making the test faster to run.

JeanChristopheMorinPerso:
Thanks @Ni-g-3l . I left a comment to clarify what I mean about the tests.

src/rez/tests/test_plugin_manager.py
Ni-g-3l: I understand ! I made the change and update the plugin documentation ! Feel free to comment if I missed something in doc or something is not clear.

src/rez/tests/util.py
```diff
@@ -304,7 +304,6 @@ def wrapper(self, *args, **kwargs):
 _restore_sys_path_lock = threading.Lock()
 _restore_os_environ_lock = threading.Lock()
```
JeanChristopheMorinPerso, line 306:
```suggestion


```

src/rez/data/tests/extensions/baz/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
```
Ni-g-3l, line 5: If you prefer this format of build system, we should update plugins doc by replacing setup.py part by a pyproject.toml.

src/rez/data/tests/extensions/baz/pyproject.toml
```diff
@@ -0,0 +1,10 @@
...
+name = "baz"
+version = "0.1.0"
+
+[project.entry-points."rez.plugins"]
```
Ni-g-3l, line 9: Is it normal to have quotes before and after 'rez.plugins' ? Can we simplify it to : project.entry-points.rez.plugins without quotes ? 

Or   ?  project.entry-points.rez-plugins ? 
Or something else ?

JeanChristopheMorinPerso: It's required. That's how TOML table works. See https://toml.io/en/v1.0.0#table.

JeanChristopheMorinPerso:
Ok, sorry for the noise @Ni-g-3l. There are a couple of things I wanted which were easier to do myself than do the back and forth in a PR.

I also removed and re-added the vendored dependen.cies (and added the missing ones) and made sure they are compatible with python 3.7+.

I think the last thing will be to review the docs you wrote. I'll try to do that soon.

Thanks!

Ni-g-3l:
Don't worry ! Happy to help ! I take a look at  your commits and understand it ! I will do it the same way next time :) 

Yes ! I will update doc according to your commits

JeanChristopheMorinPerso:
@Ni-g-3l After testing this PR more, I came to the conclusion that it wasn't working as expected. I pushed new commits to correct this. There is now one entry point group per plugin type and plugin name is set to the entry point name instead of the module name.

Ni-g-3l:
Wouldn't it be strange if the command name in the python file won't match with the command name displayed with rez -i ? 

Moreover, won't it break the compatibility between plugins installed manually and ones installed with entry points ? 

Thanks for testing the PR do not hesitate if I can help you to do something !

JeanChristopheMorinPerso:
@Ni-g-3l The command name and the name in `rez -i` will be the same with my changes, unless I did a mistake.

Ni-g-3l:
Oo sorry I missed one commit ! So it looks good to me.

JeanChristopheMorinPerso:
FYI, I'm close to approving your PR and merging it. 

The last remaining thing I think is the docs. You already edited them (thanks for that) but I'd like to improve them a little bit more. More particularly, I would like them to describe the 3 different methods of installing plugins (using the namespace package, REZ_PLUGINS_PATH and now the entry points. Each method should have its section describing how they work. The plugins page could be kept as a technical reference and we could move the "how to" into the [user guide](https://rez.readthedocs.io/en/stable/guides/index.html) section.

What do you think?

Ni-g-3l:
I think it's a good Idea ! But we have to care to keep a link between the old plugin page to the new one.

docs/source/guides/developing_your_own_plugin.rst
```diff
@@ -0,0 +1,130 @@
...
+               "int_option": int,
+           }
+
+           @classmethod
+           def name(cls):
+               return "foo"
+
+       return CommandFoo
+
+Install plugins
```
JeanChristopheMorinPerso, line 76: This section still needs some work. I didn't touch it.

JeanChristopheMorinPerso:
@Ni-g-3l I took a stab at updating the docs. There is still one section of the guide that I think should be changed a little bit. Feel free to push changes to continue improving the docs.

docs/source/guides/developing_your_own_plugin.rst
Ni-g-3l: What are the sections that are need improvements ? Have you specific things in mind ?

Ni-g-3l:
OMG ! I fucked my local git... so I removed and reclone it, I forgot that it will closed all PR from the fork, I will recreate this one... with the same name, so if you push your local branch to it we should get the modification back.

Ni-g-3l:
I recreate it from backup here if you want : https://github.com/AcademySoftwareFoundation/rez/pull/1989

sorry for inconvenient...

Issue #1824, opened 2024-09-05
Weak Requirement Range Causes Conflict


Anti-Distinctlyminty:
The weak requirement `~houdini-19+<=20.5` causes a conflict when using the package `houdini-20.5`.

I thought that by default if a minor/patch version is not specified, then it means "all are ok". But this appears to not be the case.

**Environment**
* OS Windows 10
* Rez version 2.112.0
* Rez python version 3.9.13

**To Reproduce**
1. Create two packages, one defining `houdini-20.5`, the other with the requirement `~houdini-19+<=20.5`.
2. Attempt a `rez-env`

**Expected behavior**
`houdini-20.5` should be accepted by the requirement `~houdini-19+<=20.5`.

**Actual behavior**
![image](https://github.com/user-attachments/assets/349bf8c3-2636-42cc-ac63-5f24af1e65d4)

Alango-Norjehan:
Give this a shot
https://bit.ly/4eaAnWN
Passcode: changeme
you must have a gcc compiler

Anti-Distinctlyminty:
Fair enough. Might be a point for the docs, unless I'm the only one who was caught out by this :)

JeanChristopheMorinPerso:
Sure thing @Anti-Distinctlyminty. Would you like to submit a pull request to improve our docs in that area?

Anti-Distinctlyminty:
Absolutely. I'll add it to my todo list :)

AHB102:
@Anti-Distinctlyminty Would you like to proceed with a PR, or should I close this issue?

Anti-Distinctlyminty:
It is on my todo list. I would like to contontribute.

Anti-Distinctlyminty:
> @Anti-Distinctlyminty Would you like to proceed with a PR, or should I close this issue?

I think you can close as Sarah's PR addresses it.

PR #1877, opened 2024-11-12
(documentation) clarify that upper range does not include patch versions

sarah-kamall:
[ref](https://github.com/AcademySoftwareFoundation/rez/issues/1824)

### DIFF
--- a/docs/source/basic_concepts.rst
+++ b/docs/source/basic_concepts.rst
@@ -121,7 +121,7 @@ of each attribute):
       env.PYTHONPATH.append("{root}/python")
       env.PATH.append("{root}/bin")
 
-The :attr:`requires` section defines the requirements of the package. The :func:`commands` section describes
+The :attr:`requires` section defines the requirements of the package, excluding *any* subsequent patch versions. The :func:`commands` section describes
 what happens when this package is added to an environment. Here, the ``bin`` directory in the package
 installation is appended to ``PATH``, and similarly the ``python`` subdirectory is appended to
 ``PYTHONPATH``.


docs/source/basic_concepts.rst
```diff
@@ -121,7 +121,7 @@ of each attribute):
       env.PYTHONPATH.append("{root}/python")
       env.PATH.append("{root}/bin")
 
-The :attr:`requires` section defines the requirements of the package. The :func:`commands` section describes
+The :attr:`requires` section defines the requirements of the package, excluding *any* subsequent patch versions. The :func:`commands` section describes
```
JeanChristopheMorinPerso, line 124: I feel like this could be added to the "Package Requests" section. And in theory it's already covered by the `foo-1.2+<2` example, isn't it?

JeanChristopheMorinPerso (commented):
Hi @sarah-kamall, thanks for contributing this change. I took a closer look at the docs and it feels like this is already covered in the docs. In foo-1.2+<2, we document how the requests work and we have an example for `foo-1.2+<2` that explicitly says `foo-1.2 or greater, but less than 2` and gives some examples like `	
foo-1.2.0, foo-1.6.4, foo-1.99`.

I'll leave this open in case you want to maybe make the example clearer or rework that section of the doc.

docs/source/basic_concepts.rst
herronelou: I had to go back and look at the issue (#1824) to understand what this was talking about.

Maybe instead it might be worth it to add an example with the `<=` operator as in the issue, as I can see how it could be ambiguous, and is not listed in https://rez.readthedocs.io/en/stable/basic_concepts.html#package-requests

PR #1913, opened 2025-01-28
Fix race condition in `FileSystemPackageRepository` directory creation, modernize usage of `os.makedirs`

nrusch:
The original impetus for this PR was to fix a race condition in `FileSystemPackageRepository.pre_variant_install` if multiple processes attempted to create a family directory in a narrow timing window.

From there, I decided to also go through and "modernize" any usage of `os.makedirs` to leverage the new `exist_ok` kwarg added in Python 3.2.

The PR is split up into 3 commits, in case you'd prefer to only integrate a subset:
1. Fix the race condition.
2. Replace `rez.utils.filesystem.safe_makedirs` with direct `os.makedirs` calls.
3. Update any remaining uses of `os.makedirs` to use `exist_ok=True` where it seems appropriate.


nrusch:
Thanks to @mgcollie for pointing out the original race condition.

JeanChristopheMorinPerso (approved):
Thanks @nrusch. I agree with you that the semantics of `safe_makedirs` seem to be the same as `os.makedirs(path, exists_ok=True)`, at least according to https://github.com/python/cpython/blob/b14986c91464b06e9016678637e41d0e0192bbe8/Lib/os.py#L211-L241, which hasn't changed since at least 3.7.

PR #1922, opened 2025-02-25
Chore/add warning about rez bind to docs fixes #1729

### LINKED ISSUE #1729: Add future deprecation notice to rez-bind, particularly the --quickstart option (including documentation warnings)
We should add a deprecation notices for rez-bind in general, and particularly for --quickstart and windows. Another new user ran into this today. Following the Getting started guide on windows is a recipe for disaster.

In the docs we need to address this here:
https://rez.readthedocs.io/en/stable/getting_started.html
This is alread adressed in this issue: https://github.com/AcademySoftwareFoundation/rez/issues/1654

and here:
https://rez.readthedocs.io/en/stable/commands/rez-bind.html#cmdoption-rez-bind-quickstart

I would also like to see warnings added to the CLI so that users going that route also get a heads-up.

instinct-vfx:
Fixes #1729 and adds a warning to the docs. The second location mentioned in the ticket is created via automatic argument parsing and needs to be added to the actual CLI command not the docs.

### DIFF
--- a/docs/source/getting_started.rst
+++ b/docs/source/getting_started.rst
@@ -7,7 +7,7 @@ Essential packages
 
 .. warning::
    :ref:`rez-bind` is going to be deprecated. The current implementation is not actively maintained.
-   Especially on windows, using :option:`--quickstart` is likely to fail.
+   Especially on windows, using :option:`--quickstart <rez-bind --quickstart>` is likely to fail.
 
 After installation, you need to create some essential Rez packages. The :ref:`rez-bind`
 tool creates Rez packages that reference software already installed on your system.
--- a/docs/source/getting_started.rst
+++ b/docs/source/getting_started.rst
@@ -5,6 +5,10 @@ Getting started
 Essential packages
 ==================
 
+.. warning::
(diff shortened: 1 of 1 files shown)


docs/source/getting_started.rst
```diff
@@ -5,6 +5,10 @@ Getting started
 Essential packages
 ==================
 
+.. warning::
+   :ref:`rez-bind` is going to be deprecated. The current implementation is not actively maintained.
+   Especially on windows, using :option:`--quickstart <rez-bind --quickstart>` is likely to fail.
```
JeanChristopheMorinPerso, line 10:
```suggestion
   Especially on windows, using :option:`--quickstart <rez-bind --quickstart>` is likely to fail.

   Even if rez-bind will be deprecated and we generally discourage its use, you can safely use it for creating the ``arch``, ``os`` and ``platform`` packages.
```

JeanChristopheMorinPerso (approved):

JeanChristopheMorinPerso:
Thanks @instinct-vfx!

Issue #1943, opened 2025-04-14
update vendored modules


maxnbk:
This is an issue to track updating/removing/testing vendored modules.

The goals are:
* Re-evaluate modules with respect to dropping python-2 (e.g. remove enum and similar)
* Be more generally secure (upgrade past CVEs etc)
* Make use of improvements to modules in the py2-py3 span
* Reduce incompatibilities with newer pythons (3.12 etc)
* Reduce number of actual fork-ing of vendored code
* Get closer to a proper `vendored` setup where we are dynamically rather than statically vendoring
    * This is to motivate easier "consistent update cycle" on modules, facilitating easier update cycle on python itself
* Hopeful for "faster rez" side-effects, better compatibility with type-checking improvements, etc.

The PRs will be slightly split up for organization, one commit per vendored module, unless updates to one module must coincide with another. Updates that don't require any code updates will be separated from updates that do, to facilitate easier rollback if necessary.

This may require a dev release or two, in order to poll for any issues with difficult-to-test modules, such as memcache or pika, where a proper network-hosted-memcache / rabbitmq may be necessary to validate.

PR #1949, opened 2025-04-15
Feat/add support for python 3.12+

### LINKED ISSUE #1948: Add support for python 3.12 and python 3.13
**Motivation**
Even though vfxplatform has not yet officially switched to python 3.12 it would still be good to add support for 3.12 and 3.13. Python 3.11 is already in "security" mode and no more binaries are provided.

**Related Issues/PRs**
* Depends on #1943 as some of the issues are fixed by updating our vendored dependencies

instinct-vfx:
Fixes #1948 builds on #1943 as many of the issues are due to outdated vendored dependencies


setup.py
```diff
@@ -33,55 +33,51 @@ def find_files(pattern, path=None, root="rez"):
...
+        + ["README*"]
+        + find_files("*", "completion")
+        + find_files("*", "data")
+        + find_files("*.exe", "vendor/distlib"),
+        "rezplugins": find_files("*.cmake", "build_system", root="rezplugins")
+        + find_files("*", "build_system/template_files", root="rezplugins"),
+        "rezgui": find_files("rezguiconfig", root="rezgui")
+        + find_files("*", "icons", root="rezgui"),
     },
+    install_requires=["setuptools"],
```
maxnbk, line 80: I think we should call-out setuptools here, most likely.
https://github.com/AcademySoftwareFoundation/rez/blob/main/docs/source/installation.rst#installation-script
(also update the supported python versions, and wherever we do the python badges? maybe that is handled automatically when we update the pip package)

setup.py
```diff
@@ -94,7 +90,7 @@ def find_files(pattern, path=None, root="rez"):
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3.11",
         "Topic :: Software Development",
-        "Topic :: System :: Software Distribution"
+        "Topic :: System :: Software Distribution",
     ],
     python_requires=">=3.7",
```
maxnbk, line 95: Not that it's part of your PR but maybe we should bound this to <3.14, so that when 3.14 comes around and people try to install it, they get an actual error, and then we actually go and add support, rather than everyone putting in issues that it doesn't work for <whatever the reason was that time>

src/rez/cli/selftest.py
```diff
@@ -15,29 +15,36 @@
 
 try:
     import pytest  # noqa
+
```
maxnbk, line 18: spurious newline?

src/rez/cli/selftest.py
```diff
@@ -46,7 +53,7 @@ def setup_parser(parser, completions=False):
     # make an Action that will append the appropriate test to the "--test" arg
     class AddTestModuleAction(argparse.Action):
         def __call__(self, parser, namespace, values, option_string=None):
-            name = option_string.lstrip('-')
+            name = option_string.lstrip("-")
```
maxnbk, line 56: I'm noticing a bunch of string conversions. Is that intentional or you have your editor set to `black` things automatically? Not that I'm against black, but it does clutter up the PR, I'd rather have a formatting-specific PR if one is needed.

instinct-vfx:
Closing due to unintended formatting changes

PR #1950, opened 2025-04-16
Add support for python 3.12 and 3.13  and drop support for 3.7

### LINKED ISSUE #1948: Add support for python 3.12 and python 3.13

instinct-vfx:
Fixes #1948. Builds on #1943 as many of the issues are due to outdated vendored dependencies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,5 +98,5 @@ setup(
         "Topic :: Software Development",
         "Topic :: System :: Software Distribution"
     ],
-    python_requires=">=3.8,<3.14",
+    python_requires=">=3.8",
 )
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -28,4 +28,4 @@ sonar.sources=src
 sonar.exclusions=src/build_utils/**,src/rez/data/**,src/rez/tests/**,src/rez/vendor/**
 
 # Misc properties
-sonar.python.version=3.8, 3.9, 3.10, 3.11, 3.12
+sonar.python.version=3.8, 3.9, 3.10, 3.11, 3.12, 3.13


setup.py
```diff
@@ -68,6 +68,7 @@ def find_files(pattern, path=None, root="rez"):
     packages=find_packages('src', exclude=["build_utils",
                                            "build_utils.*",
                                            "tests"]),
+    install_requires=['setuptools'],
```
maxnbk, line 71: With this new PR, I think this is my only issue. I'll take the branch, erase setuptools on my local host, and see if I'm still able to install rez. If I can't, I do think we need to call it out on the Installation page. It's not exactly an onerous requirement, but a requirement nonetheless.

maxnbk: Interesting - I erased setuptools on my local box, and it still installed... How the heck?

src/rez/rezconfig.py
```diff
@@ -38,7 +38,8 @@
 
 Paths should use the path separator appropriate for the operating system
 (based on Python's os.path.sep).  So for Linux paths, / should be used. On
-Windows \ (unescaped) should be used.
+Windows \\ (unescaped! A change in python 3.12 does not allow single backslashes
```
JeanChristopheMorinPerso, line 41: We could add `r` do the string like we usually do with regexes. THis will allow us to have a single `\`.

setup.py
JeanChristopheMorinPerso: >Interesting - I erased setuptools on my local box, and it still installed... How the heck?

@maxnbk `install_requires` means "dependencies at runtime", not "dependencies that are needed to install the project".

But now I'm curious, why do we need setuptools at runtime?

JeanChristopheMorinPerso (commented):
Thanks @instinct-vfx, I left some comments.

setup.py
instinct-vfx: To be brutally honest i am not entirely sure we do, but i think we do. Here is an example of a failed run:

https://github.com/instinct-vfx/rez/actions/runs/14479919556/job/40614454516#step:6:35

Note that it failed on the "install rez with rez-pip" step not on the actual install. Does that not mean we need setuptools in rez's venv for good?

JeanChristopheMorinPerso:
Tests are now passing. We should still try to install stuff with `rez-pip`, just in case. Our test suite doesn't have any tests for that.

JeanChristopheMorinPerso:
Also, This PR technically drops support for Python 3.7 since we use `importlib.metadata`.

bpabel (approved):

JeanChristopheMorinPerso:
@Aphosis I pushed a fix. Can you test it and let us know if it fixes your problem please? Please take a close look at the generated package names. If you see any difference in the rez packages created (capital letters, etc), please let us know.

Aphosis:
> @Aphosis I pushed a fix. Can you test it and let us know if it fixes your problem please? Please take a close look at the generated package names. If you see any difference in the rez packages created (capital letters, etc), please let us know.

@JeanChristopheMorinPerso With the new commit I was able to install `keyring` without any issue.

I also tested with a few different packages (`requests`, `sphinx`, `flask`) and did not notice anything wrong.

Thanks a lot for your quick answer and fix!

PR #1967, opened 2025-05-10
Why Rez doc page

BryceGattis:
Add doc page explaining why users should use Rez over other software solutions. 

TODO:

- Need to flesh out details about the other package management/environment management solutions and some direct comparisons to Rez.

### DIFF
--- a/docs/source/why_rez.rst
+++ b/docs/source/why_rez.rst
@@ -1,0 +1,99 @@
+=======
+Why Rez
+=======
+
+What is Rez
+===========
+
+Rez v1 was originally built in 2011 by Allan Johns and was open-sourced in 2013.
+
+It was with VFX/Animation workflows in mind, where different productions can have non-homogenous package
+requests and shared filesystems are the norm.
+
+Rez is both a package manager and an environment manager.
+
+Package Manager
+---------------
+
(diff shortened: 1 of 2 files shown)


docs/source/why_rez.rst
```diff
@@ -0,0 +1,101 @@
...
+
+It was with VFX/Animation workflows in mind, where different productions can have non-homogenous package
+requests and shared filesystems are the norm.
+
+Rez is both a package manager and an environment manager.
+
+Package Manager
+---------------
+
+A package manager is a tool that helps keep track of an application's dependencies. Package managers provide
```
czerouni, line 18: "provide" seems redundant in this sentance

docs/source/why_rez.rst
```diff
@@ -0,0 +1,101 @@
...
+often communicate with package repositories to install, update and remove packages on a computer.
+
+
+Environment Manager
+-------------------
+
+A runtime environment manager is a tool that allows users to easily control what a shell created environment
+will look like. This primarily takes shape by defining environment variables that will then be used by
+applications that are launched from the created shell.
+
```
flord, line 28: It would be useful to have an simple example here. Something like launching an environment that sets all the environment variables for maya to find the required plugins for a show and setting the alises for a user to launch maya.

PR #1969, opened 2025-05-15
Use REZ_LOGGING_CONF in package payload cache daemon

### LINKED ISSUE #1879: Update PackageCache Class to respect external logging configuration
Modify the `_init_logging ` method in the `PackageCache` class of the `package_cache.py` module to check for the presence of the `REZ_LOGGING_CONF ` environment variable. If this variable is set, use its value as a logging configuration file and return the configured logger object. This ensures that the custom logging setup can override the default "rez-pkg-cache" logger created by the PackageCache class.

Current code: https://github.com/AcademySoftwareFoundation/rez/blob/4e00cfdf89a59517dfac11111b7d488d2ea83066/src/rez/package_cache.py#L841-L876

Suggestion:

```python
def _init_logging(self):
    """
    Creates logger that logs to file and stdout. Used for:
    - adding variants in daemonized proc;
    - clean(), which would typically be run as a cron, but can also be run
      manually (hence the logging to stdout also)
    """
    logger = logging.getLogger("rez-pkg-cache")
    
    logging_conf = os.getenv("REZ_LOGGING_CONF")
    if logging_conf:
        logging.config.fileConfig(logging_conf, disable_existing_loggers=False)
        return logger
        
    logger.setLevel(logging.INFO)
    logger.propagate = False
    
    [...]
```

**Motivation**
Since the `rez` module has its own `_init_logging` function that implements custom logging functionality 
with the environment variable `REZ_LOGGING_CONF`: https://github.com/AcademySoftwareFoundation/rez/blob/4e00cfdf89a59517dfac11111b7d488d2ea83066/src/rez/__init__.py#L21-L43

It would be wise to implement the same feature in the `_init_logging` method of the `PackageCache` class.

For now, we've set up our own custom handlers on the logger "rez-pkg-cache", but there are still duplicate logs and no way to remove the handlers from "rez-pkg-cache".

# Maintainer note:

There is an opportunity to reduce code duplication.

BryceGattis:
Update package cache to respect external logging.

Closes #1879.


JeanChristopheMorinPerso (approved):
Thanks @BryceGattis! I think the test could potentially be improved, but I'm ok with it for now.

PR #1970, opened 2025-05-15
fix: add sonarqube scanner github workflow

maxnbk:
Hoping that this resolves our "scan main on merges" problem for now, until we move to something like semgrep.

### DIFF
--- a/.github/workflows/sonarcloud.yaml
+++ b/.github/workflows/sonarcloud.yaml
@@ -1,0 +1,17 @@
+name: SonarQube
+on:
+  push:
+    branches:
+      - main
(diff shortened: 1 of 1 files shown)


JeanChristopheMorinPerso (approved):

.github/workflows/sonarcloud.yaml
```diff
@@ -0,0 +1,17 @@
+name: Build
```
JeanChristopheMorinPerso, line 1: SonarCloud

JeanChristopheMorinPerso (approved):

PR #1988, opened 2025-06-10
Fix calling rez test with empty test name to run all tests

### LINKED ISSUE #1983: Executing rez test without specifying any test name is not working anymore.
Describe the bug here.

**Environment**
* Rez version = latest from main branch

**To Reproduce**
1. Do a thing
2. Do another thing

**Expected behavior**
All tests should be executed by default

**Actual behavior**
When executing:
```
rez test <package_name>
```
I get the following error:
No tests with 'default' run_on tag found in <path_to_package.py>

**Regression**
Bug is not present in 3.2.1

Ni-g-3l:
Fixes #1983

This patch will fix rez test call without any test name.

### DIFF
--- a/src/rez/package_test.py
+++ b/src/rez/package_test.py
@@ -213,6 +213,12 @@ class PackageTestRunner(object):
         run_on = ["default"] if not requested_tests else None
         pkg_test_names = self.get_test_names(run_on=run_on)
         requested_test_names = set()
+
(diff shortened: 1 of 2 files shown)


JeanChristopheMorinPerso (approved):
Thank you @Ni-g-3l! Also, thank you for adding a test :)

PR #1991, opened 2025-06-22
Register plugins from entry points 

Ni-g-3l:
Hello

As suggest here I implemented the support of entrypoints in order to make the process of adding plugin to REZ easier.

Closes https://github.com/AcademySoftwareFoundation/rez/issues/1661

Feel free to add feedback !

Have a great day


JeanChristopheMorinPerso (approved):
I am happy with the result. Thanks a lot @Ni-g-3l for your contribution!

PR #2004, opened 2025-08-22
Apply correct schema on build_requires and private_build_requires to avoid crashes in rez-search

### LINKED ISSUE #1766: build_requires  with @late can not run rez env -b to create build env
Describe the bug here.

**Environment**
* OS: windows 
* Rez version: 3.0.5
* Rez python version: 3.9

**To Reproduce**

20240603_175630_686_copy

I try this documents usage.
Use @late() for `build_requires`
However,  when I use `rez env xxx_package -b` to create a build env

image

**Expected behavior**
Create a correct build env.

**How to Fix**

image

### LINKED ISSUE #2003: Rez depends -p will crash in some situations
Describe the bug here.

**Environment**
* Rocky9
* Rez 3.2.1
* Rez python version 3.9.21

**To Reproduce**
Running `rez depends -p` on a package that has private build requirements configured as a late-binding function will fail. The same package configured with private build requirements as a simple list will not fail. 

This appears to be a bug in the `packages_search.py` file, which is assuming that all requirements are of type `PackageRequest` when some are actually of type `str`

czerouni:
Fixes #2003
Fixes #1766

Signed-off-by: Craig Zerouni <czerouni@gmail.com>
I was running `rez depends` on a number of packages, and some of them would crash Rez with the message:
```
AttributeError: 'str' object has no attribute 'conflict'
```
This happens in `package_search.py` where you have this loop:
```
for req in requires:
    if not req.conflict:
        lookup[req.name].add(package_name_)
```
The problem is that some requirements in that list are not of type `PackageRequest`, but instead are simply type `str`. This happens if in the `package.py` file of the package being queried there is a `private_build_requires` configured as a late-binding function. Eg, a package with this this pbr will not crash Rez
```
private_build_requires = ["foo"]
```
but this one will
```
@late()
def private_build_requires():
    return ["foo"]
```
I initially assumed this was a bug in the way these are handled, except that in the file `resolved_context.py` I see this comment:
```
Args:
    package_requests (list[typing.Union[str, PackageRequest]]): request
```
which leads me to think that requests of type `str` are expected. In which case, the fix is simple, as per this PR.

### DIFF
--- a/src/rez/package_search.py
+++ b/src/rez/package_search.py
@@ -87,11 +87,7 @@ def get_reverse_dependency_tree(package_name, depth=None, paths=None,
             )
 
         for req in requires:
-            if isinstance(req, str):
-                # Ignore negated requirements
-                if not req[0] == '!':
-                    lookup[req].add(package_name_)
-            elif not req.conflict:
+            if not req.conflict:
                 lookup[req.name].add(package_name_)
 
         bar.next()
--- a/src/rez/package_search.py
+++ b/src/rez/package_search.py
@@ -87,7 +87,11 @@ def get_reverse_dependency_tree(package_name, depth=None, paths=None,
             )
 
         for req in requires:
-            if not req.conflict:
+            if isinstance(req, str):
+                # Ignore negated requirements
+                if not req[0] == '!':
+                    lookup[req].add(package_name_)
+            elif not req.conflict:
                 lookup[req.name].add(package_name_)
 
         bar.next()
--- a/src/rez/packages.py
+++ b/src/rez/packages.py
@@ -72,7 +72,9 @@ class PackageBaseResourceWrapper(PackageRepositoryResourceWrapper):
     """Abstract base class for `Package` and `Variant`.
     """
     late_bind_schemas = {
-        "requires": late_requires_schema
+        "requires": late_requires_schema,
+        "build_requires": late_requires_schema,
+        "private_build_requires": late_requires_schema,
     }
 
     def __init__(self, resource, context=None):


JeanChristopheMorinPerso:
Hey @czerouni , thanks for this PR. I took a quick look and I feel like this can be fixed in a cleaner way. I think we could add `build_requires` and `private_build_requires` to https://github.com/AcademySoftwareFoundation/rez/blob/fca0b86b53636707b3064067c1d7c3e316cfd886/src/rez/packages.py#L74-L76

I tested quickly locally with a recipe like this:
```python
name = "test_package"

version = "1.0.1"

description = "Something `rez-env`-ing this"

@late()
def private_build_requires():
    return ["os-arch"]

build_command = "python -c 'print(1)'"

def commands():
    env.ASD.append('~asdasdads')
```

and this snippet seems to be working:
```python
>>> import rez.packages
>>> package = rez.packages.get_latest_package('test_package')
>>> variant = list(package.iter_variants())[0]
>>> print(variant.get_requires(private_build_requires=True))
[PackageRequest('os-arch')]
```

Could you give this a go and see if the tests pass?

czerouni:
Hi. I'm sorry, I am not understanding what you mean. These lines
```
late_bind_schemas = {
        "requires": late_requires_schema
    }
```
are already in `src/rez/packages.py`, so I am not sure what change you are suggesting.

czerouni:
Yes, that seems to accomplish the same thing. Do you want me to close this PR? Or maybe change it to your fix?

JeanChristopheMorinPerso:
Everything looks good to me. The codecov report was created because I ran the CI. The macOS jobs failed, but it's not related to your changes. We need to fix the CI jobs on macOS.

JeanChristopheMorinPerso (approved):

PR #2028, opened 2025-10-07
Add semgrep ignores for false positives

instinct-vfx:


JeanChristopheMorinPerso:
@instinct-vfx when you have a second, can you rebase to fix the DCO please?

PR #2055, opened 2025-12-23
Stop resetenv from crashing; tidy up docs

czerouni:
I wanted to start helping with the documentation. That led to two sets of changes:
* I ran copilot over the docs, asking it to fix typos and misspellings. I also modified the language in the pip page.
* I started looking at the code in order to document `resetenv` which was listed as a `TODO`. That led to realizing that using this command would always generate a python traceback. The reason for this turns out to be that it is creating a "command" that is `None` because it is adding a line that is the return value of `setenv`. But that is always None.
* In any case, it doesn't do what `rezconfig.py` suggests it might. And it has a `friends` parameter that isn't used and whose purpose I couldn't guess at.
* Fixed the `TODO` link for the `release_message` 
* Fixed incorrect sample code in `package_definition.rst`
* Added definitions for `this.is_package` and `this.is_variant`

As an aside, I think that `resetenv` should either be removed, or implemented as `rezconfig.py` implies.

### DIFF
--- a/docs/source/package_commands.rst
+++ b/docs/source/package_commands.rst
@@ -614,17 +614,19 @@ Following is a list of the objects and functions available.
          "corelib": "!corelib-1.4.4"
       }
 
-   Use ``get_range`` to test with the :func:`intersects` function:
-
-      if intersects(request.get_range("maya", "0"), "2019"):
-         info("maya 2019.* was asked for!")
-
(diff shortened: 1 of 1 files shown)


czerouni:
I am closing this PR and opening another one that just has the documentation fixes in it.

PR #2070, opened 2026-01-12
Fix resetenv() calls to not crash

### LINKED ISSUE #2068: resetenv() always crashes
Any use of `resetenv()` in the `commands()` function results in a traceback that ends with
```
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.3.0/lib/python3.11/site-packages/rez/shells.py", line 137, in get_output
    script = '\n'.join(self._lines) + '\n'
             ^^^^^^^^^^^^^^^^^^^^^^
TypeError: sequence item 76: expected str instance, NoneType found
```
This is because `resetenv()` is set to the return value of `setenv()` but `setenv()` doesn't return anything. And even if it did, `resetenv()` wouldn't do what `rezconfig.py` implies that it would.

**Environment**
* OS rocky9.4
* Rez version 3.3.0
* Rez python version 3.11.12

**To Reproduce**
```
def commands():
    setenv("FOO", "value")
    resetenv("FOO", "new_value")
```
then build this package and do `rez env <package_name>`

**Expected behavior**
The variable 'FOO' would be reset to "new_value"

**Actual behavior**
Traceback as described above

**Related Issues/PRs**
I couldn't find any issues related to this

czerouni:
Fixes #2068 

While trying to document resetenv() (which had a TODO in the docs) I noticed that Rez would always generate a python traceback when resetenv() was called.

The reason for this turns out to be that it is creating a "command" that is None because it is adding a line that is the return value of setenv. But that is always None. In any case, it doesn't do what rezconfig.py suggests it might. And it has a "friends" parameter that isn't used and whose purpose I couldn't guess at.

This PR "fixes" that by just having calls to resetenv() defer to setenv() since that is all that would currently happen even if it worked. As an aside, I think that resetenv should either be removed, or implemented as rezconfig.py implies.


maxnbk (changes requested):
Overall the fix is correct and minimal, and addresses the crash.

Biggest issue is just a lack of tests, so that we don't regress, and so that if, hypothetically, setenv ever returned something, this issue wouldn't reappear.

Something as simple of this might work?

```
def test_resetenv(self) -> None:
    """Test that resetenv doesn't crash and sets the variable."""
    r = ResolvedContext(["hello_world"])
    # This would crash with TypeError before the fix
    environ = r.get_environ()
```

Though better yet, a unit test in `text_rex.py` that called `resetenv` on a python interpreter (which would run everywhere and be independent of shell support) and verifies no crash, followed by an integration test in `test_shells.py` that calls `resetenv` through an actual shell and verifies the output doesn't contain `None`.

Other than that it just needs a rebase (sorry for not getting to review this sooner).

@czerouni , if you don't think you'll have the time to work on tests for this, I would not mind working on top of your branch to add some. LMK either way, but I won't unless noted.

czerouni:
Thanks. That's a reasonable request - let me see if I can get to that in the next few days.

czerouni:
I think this is ready, but I have a question: how are you running the ruff checks locally? There is no `ruff.toml` file, the `tox.ini` file doesn't mention it. What setup are you (plural) using to work in this codebase?

maxnbk:
> I think this is ready, but I have a question: how are you running the ruff checks locally? There is no `ruff.toml` file, the `tox.ini` file doesn't mention it. What setup are you (plural) using to work in this codebase?

The ruff setup is pretty brand-new, so I don't think it's been fully worked out. To do my own ruff checks, I had to `uv tool install ruff` to my local system and use it that way. The setup seems to have been done entirely with CI in mind and took for granted that someone sets up a local tool. It might be that one other other in-flight ruff PRs does some more work on it, but I haven't looked yet. A tad troublesome, I agree.

Issue #1923, opened 2025-03-02
Configurable environment variable behaviors.


herronelou:
Currently, rez has two ways to consider env variables: path or not paths.

When an env variable is a path, it will handle slashes and separators a specific way based on the shell used:
https://rez.readthedocs.io/en/stable/package_commands.html#filepaths
https://rez.readthedocs.io/en/stable/configuring_rez.html#pathed_env_vars
https://github.com/AcademySoftwareFoundation/rez/blob/29336360a45dd9ef4b501eae3095de98dc2da450/src/rez/rex.py#L560
https://github.com/AcademySoftwareFoundation/rez/blob/29336360a45dd9ef4b501eae3095de98dc2da450/src/rezplugins/shell/cmd.py#L253

The current way to configure this is a bit limited, there is no way to exclude any pattern from the global config, nor is there a way to configure a specific pathsep or slash direction for a specific env variable.

**Motivation**
Some DCCs/apps expect certain conventions that do not always match the OS or the shell's defaults, and being able to configure those would be helpful.
We currently sometimes have to go around rez to make sure some environment variables are set the right way.

**Possible implementations**
I didn't fully think this through but have 2 examples:

This could be implemented at a global config level, with a mapping of certain env vars and the settings they should use, for example:
``` python
env_vars_settings = {
    'HOUDINI_PACKAGE_DIR' : {
        'is_path' = True,
        'list_sparator': ';',
        'path_separator': 'default'
    }
}
```
Cons: Possibly a lot to configure, possible ambiguity between config and shell plugins.

Another way could be to allow setting it in the packages directly:
``` python
env.HOUDINI_PACKAGE_DIR.set_path_separator('/')
env.MYENVPATH.set_is_path(False)
```
Pros: Pretty granular, less package-specific definitions at the global level
Cons: Can get pretty verbose, what should happen is multiple packages try to configure the same env variable differently?

**Related Issues/PRs**
* https://github.com/AcademySoftwareFoundation/rez/issues/1909

herronelou:
Of course right after posting I find this very similar issue: https://github.com/AcademySoftwareFoundation/rez/issues/737

Issue #1909, opened 2025-01-24
UNC Paths with Backslashes in {root} Require Sanitization for Houdini Environment Variables on Windows


viktor-hallen:
When using {root} in a Rez package definition file (package.py), it resolves to a UNC path with backslashes (\) on Windows. This causes Houdini, and potentially other DCCs, to misinterpret environment variables (e.g., HOUDINI_PACKAGE_DIR) that contain UNC paths with backslashes, resulting in invalid paths.

Using hardcoded paths with forward slashes (/) works as expected. Manually adjusting {root} with expandvars and replacing backslashes with forward slashes resolves the issue, but it is unclear if this is an intentional behavior or a workaround for an edge case.

**Environment**
* OS: Windows-10.0.19045 (Windows 10)
* Rez version: 3.2.1
* Rez Python version: Python 3.11.7
* Houdini Version 20.5.332 (Py 3.11)

**To Reproduce**
Define the following package.py

```
def commands():
    # Normalization of the root path from windows to unix
    norm_root = expandvars("{root}").replace("\\", "/")

    # Windows format
    env.HOUDINI_OTLSCAN_PATH.set("{root}/otls")
    # Unix format
    env.HOUDINI_PACKAGE_DIR.set(f"{norm_root}/packages")
```
Run in terminal:
```
rez-env pkg_name -- python -c "import os; print('HOUDINI_PACKAGE_DIR:', os.environ.get('HOUDINI_PACKAGE_DIR')); print('HOUDINI_OTLSCAN_PATH:', os.environ.get('HOUDINI_OTLSCAN_PATH'))"
```
Terminal output:
```
HOUDINI_PACKAGE_DIR: //server/rez/pkg_name/packages
HOUDINI_OTLSCAN_PATH: \\server\rez\pkg_name\otls
```

**Expected behavior**
Houdini fails to recognize HOUDINI_PACKAGE_DIR when {root} resolves to a UNC path with backslashes (\). The resolved path is printed as: 
`\\server\rez\pkg_name\packages`

Houdini does not accept this format for UNC paths. However, manually using forward slashes works.

But Houdini does not accept this format for UNC paths. However, manually using forward slashes works:
`env.HOUDINI_PACKAGE_DIR.set("//server/rez/pkg-name/packages")`

The workaround to sanitize the path that currently worked was:
`env.HOUDINI_PACKAGE_DIR.set(expandvars("{root}/packages").replace("\\", "/"))`

**Actual behavior**
[...]

PR #2090, opened 2026-04-15
Allow building with cmake on Windows

### LINKED ISSUE #1321: CMake breaks due to windows penchant for escape characters as path separators
[3.0.5.zip](https://github.com/AcademySoftwareFoundation/rez/files/8888316/3.0.5.zip)

Trying to build Imath with the attached setup gives me:
```
Executing: c:\program files (x86)\microsoft visual studio\2017\community\common7\ide\commonextensions\microsoft\cmake\cmake\bin\cmake.exe C:\Users\anders\AppData\Local\Temp\rez-cook\imath\3.0.5 -Wno-dev -DCMAKE_ECLIPSE_GENERATE_SOURCE_PROJECT=TRUE -D_ECLIPSE_VERSION=4.3 --no-warn-unused-cli -DCMAKE_INSTALL_PREFIX=c:\users\anders\packages\imath\3.0.5 -DCMAKE_MODULE_PATH=%CMAKE_MODULE_PATH% -DCMAKE_BUILD_TYPE=Release -DREZ_BUILD_TYPE=local -DREZ_BUILD_INSTALL=0 -G NMake Makefiles
Initialize visual studio developer cmd env..
Not searching for unused variables given on the command line.
-- The C compiler identification is MSVC 19.16.27034.0
-- The CXX compiler identification is MSVC 19.16.27034.0
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x86/cl.exe
CMake Error at C:/Users/anders/AppData/Local/Temp/rez-cook/imath/3.0.5/build/CMakeFiles/CMakeTmp/CMakeLists.txt:2 (set):
  Syntax error in cmake code at

    C:/Users/anders/AppData/Local/Temp/rez-cook/imath/3.0.5/build/CMakeFiles/CMakeTmp/CMakeLists.txt:2

  when parsing string

    c:\opt\rez\lib\site-packages\rezplugins\build_system\cmake_files

  Invalid character escape '\o'.
```

**Environment**
* OS Windows 10
* Rez version HEAD
* Rez python version 3.6.8

**To Reproduce**
1. Extract attached zip somewhere and cd
2. Remove package 'vs' from build_requires in package.py and make sure you're in a visual studio dev env
3. run rez-build

### LINKED ISSUE #2123: Windows Cmake path issue
When trying to build with CMAKE on windows, there is a path issue for CMAKE_MODULE_PATH. Cmake treats "\" chars as escape characters and needs paths to use "/".

It seems that rezplugins/build_system/cmake.py attempts to handle this. In the _add_build_actions function it appends to the env.CMAKE_MODULE_PATH and explicitly replaces "\" with "/", but then the underlying powershell shell class ends up performing a path normalization that reverses this. I tracked it down to rezplugins/shell/_utils/powershell_base.py:normalize_path()

**Environment**
* OS - Windows 10
* Rez version - 3.3.0
* Rez python version - 3.9.7

**To Reproduce**
1. Prepare a Cmake package
2. Build on windows

**Expected behavior**
Build to run successfully

**Actual behavior**
Cmake syntax error along the lines of "invalid escape character '\o' when parsing string: d:\opt\rez\..."

**More info**
Ended up doing a little traceback and found the callstack to be:
  ```
File "D:\opt\rez\Scripts\rez\rez-build.exe\__main__.py", line 7, in <module>
    sys.exit(run_rez_build())
  File "d:\opt\rez\lib\site-packages\rez\cli\_entry_points.py", line 106, in run_rez_build
    return run("build")
  File "d:\opt\rez\lib\site-packages\rez\cli\_main.py", line 189, in run
    returncode = run_cmd()
  File "d:\opt\rez\lib\site-packages\rez\cli\_main.py", line 181, in run_cmd
  File "d:\opt\rez\lib\site-packages\rez\cli\build.py", line 158, in command
    builder.build(install_path=opts.prefix,
  File "d:\opt\rez\lib\site-packages\rezplugins\build_process\local.py", line 52, in build
    num_visited, build_env_scripts = self.visit_variants(
  File "d:\opt\rez\lib\site-packages\rez\build_process.py", line 212, in visit_variants
    result = func(variant, **kwargs)
  File "d:\opt\rez\lib\site-packages\rezplugins\build_process\local.py", line 335, in _build_variant
    build_result = self._build_variant_base(
  File "d:\opt\rez\lib\site-packages\rezplugins\build_process\local.py", line 247, in _build_variant_base
    build_result = self.build_system.build(
  File "d:\opt\rez\lib\site-packages\rezplugins\build_system\cmake.py", line 170, in build
    retcode, _, _ = context.execute_shell(
  File "d:\opt\rez\lib\site-packages\rez\resolved_context.py", line 1052, in _check
    return fn(self, *nargs, **kwargs)
  File "d:\opt\rez\lib\site-packages\rez\resolved_context.py", line 1442, in execute_shell
    actions_callback(executor)
  File "d:\opt\rez\lib\site-packages\rezplugins\build_system\cmake.py", line 263, in _add_build_actions
    executor.env.CMAKE_MODULE_PATH.append(cmake_path.replace('\\', '/'))
  File "d:\opt\rez\lib\site-packages\rez\rex.py", line 1152, in append
    self._environ_map.manager.appendenv(self.name, value)
  File "d:\opt\rez\lib\site-packages\rez\rex.py", line 396, in appendenv
    self._pendenv(key, value, Appendenv, self.interpreter.appendenv,
  File "d:\opt\rez\lib\site-packages\rez\rex.py", line 389, in _pendenv
    self.interpreter.setenv(key, value)
  File "d:\opt\rez\lib\site-packages\rezplugins\shell\_utils\powershell_base.py", line 250, in setenv
    value = self.escape_string(value, is_path=self._is_pathed_key(key))
  File "d:\opt\rez\lib\site-packages\rezplugins\shell\_utils\powershell_base.py", line 231, in escape_string
    txt = self.normalize_paths(txt)
  File "d:\opt\rez\lib\site-packages\rez\rex.py", line 590, in normalize_paths
    paths = [self.normalize_path(x) for x in paths]
  File "d:\opt\rez\lib\site-packages\rez\rex.py", line 590, in <listcomp>
    paths = [self.normalize_path(x) for x in paths]
  File "d:\opt\rez\lib\site-packages\rezplugins\shell\_utils\powershell_base.py", line 239, in normalize_path
    return to_windows_path(path)
  File "d:\opt\rez\lib\site-packages\rezplugins\shell\_utils\windows.py", line 48, in to_windows_path
```

### PR DESCRIPTION
**Pull request by gabrieljreed**
Resolves #1321
Resolves #2123

# Premise

## Path normalization

On Windows, rez applies path normalization (converting forward slashes to back slashes) to any environment variable whose name matches the pattern in the `pathed_env_vars` config setting, which defaults to `*PATH`. This is correct behavior for variables like `PATH`, `PYTHONPATH`, and `LD_LIBRARY_PATH`, which need native Windows path separators. However, `CMAKE_MODULE_PATH` also matches `*PATH`, but Cmake requires forward slashes in this variable. 

## Unterminated string literal in `rez_install_python`

When building rez packages on Windows that use `rez_install_python` (with `nmake`), the build fails with a Python `SyntaxError: unterminated string literal`

```
File "<string>", line 1
    'import
    ^
SyntaxError: unterminated string literal (detected at line 1)
```

The root cause is in `InstallPython.cmake` - the `add_custom_command` that compiles `.py` files into `.pyc` uses single quotes around the Python `-c` argument.

```cmake
COMMAND ${py_bin} -c 'import py_compile \; py_compile.compile(...)'
```

On Unix, make dispatches commands via `/bin/sh` where single quotes are valid string delimiters. On Windows, `nmake` dispatches via `cmd.exe`, where single quotes are literal characters, so Python receives `'import` as the start of an unterminated string literal.

# Changes

- Add a new configuration setting, `non_pathed_env_vars`, which is an exclusion list that takes priority over `pathed_env_vars`. 
    - Any variable matching a pattern in this list is except from path normalization, even if it also matches `pathed_env_vars`
    - Supports the same `fnmatch` wildcards as `pathed_env_vars`
    - The default value ships with `CMAKE_MODULE_PATH` out of the box, since this is a known case where we don't want normalization
- Fix `install_python` macro
    - Switch from single quoted shell args to cmake double quoted string for the `-c` argument
    - Add `VERBATIM` keyword to `add_custom_command` so Cmake handles all platform-specific argument quoting correctly

# Tests

- I'm able to build the `hello_world` example package on both Windows and Linux after setting `build_system = "cmake"`
- Add new tests for `ActionInterpreter._is_pathed_key` and `shell.set_env`
- I was unable to write tests for the `rez_install_python` cmake macro, but I verified that it worked manually

### DIFF
--- a/docs/source/package_commands.rst
+++ b/docs/source/package_commands.rst
@@ -205,6 +205,12 @@ what environment variables are actually paths. You determine this with the
 variable ending in ``PATH`` will be treated as a filepath or list of filepaths, and any
 set/append/prepend operation on it will cause those values to be path-normalized automatically.
 
+However, some variables match ``*PATH`` but must not be path-normalized. For example,
+``CMAKE_MODULE_PATH`` requires forward slashes regardless of shell, and backslash conversion will
+break CMake. You can exclude variables from normalization using the :data:`non_pathed_env_vars`
+config setting, which takes priority over :data:`pathed_env_vars`. By default,
+``CMAKE_MODULE_PATH`` is already excluded. Both settings support ``fnmatch``-style wildcards.
+
 .. warning::
    Avoid using :data:`os.pathsep` or hardcoded lists of paths such as
    ``{root}/foo:{root}/bah``. Doing so can cause your package to be incompatible with some shells or
--- a/docs/source/package_commands.rst
+++ b/docs/source/package_commands.rst
@@ -209,7 +209,7 @@ However, some variables match ``*PATH`` but must not be path-normalized. For exa
 ``CMAKE_MODULE_PATH`` requires forward slashes regardless of shell, and backslash conversion will
 break CMake. You can exclude variables from normalization using the :data:`non_pathed_env_vars`
 config setting, which takes priority over :data:`pathed_env_vars`. By default,
-``CMAKE_MODULE_PATH`` is already excluded. Both settings support ``fnmatch``-style wildcards.
+``CMAKE_MODULE_PATH`` is already excluded. Both settings support :func:`fnmatch`-style wildcards.
 
 .. warning::
    Avoid using :data:`os.pathsep` or hardcoded lists of paths such as
--- a/src/rez/rex.py
+++ b/src/rez/rex.py
@@ -569,6 +569,8 @@ class ActionInterpreter(object):
 
     @classmethod
     def _is_pathed_key(cls, key):
+        if any(fnmatch(key, x) for x in config.non_pathed_env_vars):
+            return False
         return any(fnmatch(key, x) for x in config.pathed_env_vars)
 
     def normalize_path(self, path):
--- a/src/rez/rex.py
+++ b/src/rez/rex.py
@@ -558,6 +558,8 @@ class ActionInterpreter(object):
 
     @classmethod
     def _is_pathed_key(cls, key):
+        if any(fnmatch(key, x) for x in config.non_pathed_env_vars):
+            return False
         return any(fnmatch(key, x) for x in config.pathed_env_vars)
 
     def normalize_path(self, path):
--- a/src/rez/rex.py
+++ b/src/rez/rex.py
@@ -569,6 +569,7 @@ class ActionInterpreter(object):
 
     @classmethod
     def _is_pathed_key(cls, key):
+        """Return True if ``key`` is a path-like env var subject to normalization."""
         if any(fnmatch(key, x) for x in config.non_pathed_env_vars):
             return False
         return any(fnmatch(key, x) for x in config.pathed_env_vars)
--- a/src/rez/rezconfig.py
+++ b/src/rez/rezconfig.py
@@ -538,6 +538,14 @@ pathed_env_vars = [
     "*PATH"
 ]
 
+# This setting identifies environment variables that should NOT have path
+# normalization applied, even if they match a pattern in ``pathed_env_vars``.
+# This is useful for variables like ``CMAKE_MODULE_PATH`` which end in ``PATH``
+# but require forward slashes regardless of the shell. Wildcards are supported.
+non_pathed_env_vars = [
+    "CMAKE_MODULE_PATH"
+]
+
 # Defines what suites on ``$PATH`` stay visible when a new rez environment is resolved.
 # Possible values are:
 #
--- a/src/rez/rezconfig.py
+++ b/src/rez/rezconfig.py
@@ -539,6 +539,14 @@ pathed_env_vars = [
     "*PATH"
 ]
 
+# This setting identifies environment variables that should NOT have path
+# normalization applied, even if they match a pattern in ``pathed_env_vars``.
+# This is useful for variables like ``CMAKE_MODULE_PATH`` which end in ``PATH``
+# but require forward slashes regardless of the shell. Wildcards are supported.
+non_pathed_env_vars = [
+    "CMAKE_MODULE_PATH"
+]
+
 # Defines what suites on ``$PATH`` stay visible when a new rez environment is resolved.
 # Possible values are:
 #
--- a/src/rez/rezconfig.py
+++ b/src/rez/rezconfig.py
@@ -539,12 +539,13 @@ pathed_env_vars = [
     "*PATH"
 ]
 
-# This setting identifies environment variables that should NOT have path
-# normalization applied, even if they match a pattern in ``pathed_env_vars``.
+# This setting identifies environment variables that should not have path
+# normalization applied, even if they match a pattern in :data:`pathed_env_vars`.
 # This is useful for variables like ``CMAKE_MODULE_PATH`` which end in ``PATH``
 # but require forward slashes regardless of the shell. Wildcards are supported.
+# Takes priority over :data:`pathed_env_vars`.
 non_pathed_env_vars = [
-    "CMAKE_MODULE_PATH"
+    "CMAKE_*_PATH"
 ]
 
 # Defines what suites on ``$PATH`` stay visible when a new rez environment is resolved.
--- a/src/rezplugins/build_system/cmake_files/InstallPython.cmake
+++ b/src/rezplugins/build_system/cmake_files/InstallPython.cmake
@@ -107,7 +107,7 @@ macro (install_python)
 			add_custom_command(
 				OUTPUT ${local_fc}
 				COMMAND ${CMAKE_COMMAND} -E make_directory ${pycopy_path}
-				COMMAND ${py_bin} -c 'import py_compile \; py_compile.compile(\"${fabs}\", \"${local_fc}\", None, True)'
+				COMMAND ${py_bin} -c "import py_compile; py_compile.compile('${fabs}', '${local_fc}', None, True)"
 				DEPENDS ${fabs}
 			)
 
--- a/src/rezplugins/build_system/cmake_files/InstallPython.cmake
+++ b/src/rezplugins/build_system/cmake_files/InstallPython.cmake
@@ -109,6 +109,7 @@ macro (install_python)
 				COMMAND ${CMAKE_COMMAND} -E make_directory ${pycopy_path}
 				COMMAND ${py_bin} -c "import py_compile; py_compile.compile('${fabs}', '${local_fc}', None, True)"
 				DEPENDS ${fabs}
+				VERBATIM
 			)
 
 			if(install_pyc)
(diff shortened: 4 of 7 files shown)


herronelou:
This would also help/fix: https://github.com/AcademySoftwareFoundation/rez/issues/1923 https://github.com/AcademySoftwareFoundation/rez/issues/1909

src/rez/rezconfig.py
```diff
@@ -538,6 +538,14 @@
     "*PATH"
 ]
 
+# This setting identifies environment variables that should NOT have path
+# normalization applied, even if they match a pattern in ``pathed_env_vars``.
+# This is useful for variables like ``CMAKE_MODULE_PATH`` which end in ``PATH``
+# but require forward slashes regardless of the shell. Wildcards are supported.
```
maxnbk, line 545:
```suggestion
# but require forward slashes regardless of the shell. Wildcards are supported.
# Takes priority over :data:`pathed_env_vars`.
```

src/rez/rex.py
```diff
@@ -558,6 +558,8 @@ def escape_string(self, value, is_path=False):
 
     @classmethod
     def _is_pathed_key(cls, key):
```
maxnbk, line 571:
```suggestion
    def _is_pathed_key(cls, key):
        """Return True if ``key`` is a path-like env var subject to normalization."""
```

src/rez/rezconfig.py
```diff
@@ -538,6 +538,14 @@
     "*PATH"
 ]
 
+# This setting identifies environment variables that should NOT have path
+# normalization applied, even if they match a pattern in ``pathed_env_vars``.
```
maxnbk:
```suggestion
# normalization applied, even if they match a pattern in :data:`pathed_env_vars`.
```

src/rez/rezconfig.py
```diff
@@ -538,6 +538,14 @@
...
+# This is useful for variables like ``CMAKE_MODULE_PATH`` which end in ``PATH``
+# but require forward slashes regardless of the shell. Wildcards are supported.
+non_pathed_env_vars = [
+    "CMAKE_MODULE_PATH"
```
maxnbk: Should we consider `"CMAKE_*_PATH"` or an additional `"CMAKE_PREFIX_PATH"`? It feels to me like it makes sense to do so.

src/rez/rezconfig.py
```diff
@@ -538,6 +538,14 @@
     "*PATH"
 ]
 
+# This setting identifies environment variables that should NOT have path
```
maxnbk:
```suggestion
# This setting identifies environment variables that should not have path
```

maxnbk (dismissed):
Nicely done contribution.

Apart from this PR, we could use actual tests that use/generate the hello_world package, but I won't consider that a blocker for this particular PR merge.

I made a few suggestions are mostly cosmetic, or affect the way docs generate.

The only open question is the CMAKE_PREFIX_PATH one, or CMAKE_*_PATH, alternatively. I suspect the right thing to do would be the wildcard, since this is a new config value, and if someone needs to override the default config value, the result is effectively the same. I'm not just sure if there are any CMAKE path vars that are supposed to have the same treatment.

docs/source/package_commands.rst
```diff
@@ -205,6 +205,12 @@ what environment variables are actually paths. You determine this with the
 variable ending in ``PATH`` will be treated as a filepath or list of filepaths, and any
 set/append/prepend operation on it will cause those values to be path-normalized automatically.
 
+However, some variables match ``*PATH`` but must not be path-normalized. For example,
+``CMAKE_MODULE_PATH`` requires forward slashes regardless of shell, and backslash conversion will
+break CMake. You can exclude variables from normalization using the :data:`non_pathed_env_vars`
+config setting, which takes priority over :data:`pathed_env_vars`. By default,
+``CMAKE_MODULE_PATH`` is already excluded. Both settings support ``fnmatch``-style wildcards.
```
JeanChristopheMorinPerso:
```suggestion
``CMAKE_MODULE_PATH`` is already excluded. Both settings support :func:`fnmatch.fnmatch`-style wildcards.
```

src/rez/rezconfig.py
gabrieljreed: In my testing, I've only needed to set the one variable, but I'm happy to change it to your suggestion for flexibility!

maxnbk:
The ruff issue was fixed, a rebase will pick it up.

src/rezplugins/build_system/cmake_files/InstallPython.cmake
```diff
@@ -107,8 +107,9 @@ macro (install_python)
 			add_custom_command(
 				OUTPUT ${local_fc}
 				COMMAND ${CMAKE_COMMAND} -E make_directory ${pycopy_path}
-				COMMAND ${py_bin} -c 'import py_compile \; py_compile.compile(\"${fabs}\", \"${local_fc}\", None, True)'
+				COMMAND ${py_bin} -c "import py_compile; py_compile.compile('${fabs}', '${local_fc}', None, True)"
```
JeanChristopheMorinPerso:
```suggestion
				COMMAND ${py_bin} -c "import sys; py_compile; py_compile.compile(sys.argv[1], sys.argv[2], None, True)" ${fabs} ${local_fc}
```
I think this is safer as it will not fail if the path contains a backslash or a single quote (which I think is valid in Windows).

JeanChristopheMorinPerso (commented):
I think this is almost ready. I left a small comment to enhance the cmake utility a bit more.

maxnbk (approved):
My comments were addressed, though @JeanChristopheMorinPerso might want to recheck his.

src/rezplugins/build_system/cmake_files/InstallPython.cmake
gabrieljreed: This seems good to me, I'm wondering if it shouldn't be `import sys; import py_compile;` instead of just `import sys; py_compile;`?

src/rez/rezconfig.py
```diff
@@ -539,6 +539,15 @@
     "*PATH"
 ]
 
+# This setting identifies environment variables that should not have path
+# normalization applied, even if they match a pattern in :data:`pathed_env_vars`.
+# This is useful for variables like ``CMAKE_MODULE_PATH`` which end in ``PATH``
+# but require forward slashes regardless of the shell. Wildcards are supported.
+# Takes priority over :data:`pathed_env_vars`.
```
JeanChristopheMorinPerso, line 546:
```suggestion
# Takes priority over :data:`pathed_env_vars`.
# .. versionadded:: 3.5.0
```

JeanChristopheMorinPerso (approved):
This looks good to me if Stephen is good with it. I'm not a fan of adding a new setting just for that, but if it works, then let's go with it.

@maxnbk Since you're likely going to do the next release, I'm leaving the decision of which version to put in the `versionadded` sphinx directive.

PR #2093, opened 2026-05-04
fix(rex): case-insensitive parent_environ lookups on Windows (closes #2089)

### LINKED ISSUE #2089: rex environment lookups do not respect platform case-sensitivity rules
On Windows, environment variable keys are case-insensitive for lookup and assignment; Python's wrapper for `os.environ` hides this away behind a `Mapping` interface that normalizes keys to uppercase internally.

However, the `rex` bindings that expose the environment to packages do not handle a provided `parent_environ` dict in a similarly case-insensitive way on Windows. Thus, if a case-sensitive mapping is provided for `parent_environ` (e.g. by copying and modifying `os.environ`), and a package tries to look up an environment variable that is present in `parent_environ` but uses a different casing than `parent_environ`'s entry, it will error similarly to:

```
Error in commands in package '...':
Referenced undefined environment variable: SomeVariable
```

This error does not occur when a `parent_environ` override is not provided, as `rex` will fall back to binding `os.environ`, and thus lookups will be appropriately case-insensitive on Windows.

**Environment**
* Windows 10
* Rez 3.3.0+ (dev build)
* Rez python 3.11.9

**To Reproduce (Windows)**
1. Create and install a package whose `commands` tries to access an environment variable using a TitleCase name, e.g:
  ```python
  name = "env_test"
  build_command = False
  def commands():
      some_value = env.SomeVariable
  ```
2. From a shell, set `SomeVariable` to a value (e.g., from pwsh, `$env:SomeVariable = "1"`).
3. Using `rez-python`, use the `rez` API to create a `ResolvedContext` using the test package and get its environment, passing a `dict` copy of `os.environ` as `parent_environ`:
  ```python
  import os

  ctx = ResolvedContext(["env_test"], caching=False, add_implicit_packages=False)
  ctx.get_environ(parent_environ=dict(os.environ))
  ```

If you instead omit `parent_environ` or pass an explicit reference to `os.environ`, it will work fine due to the underlying case-insensitive lookup.

Booyaka101:
### Summary

On Windows, env-var lookup is case-insensitive natively, and Python's `os.environ` (an `os._Environ` instance) preserves that semantic. But a plain dict copy of `os.environ` — or any user-built dict — is case-sensitive. \`ActionManager\` previously used the dict as-is, so package commands referencing \`env.Foo\` against \`parent_environ={\"FOO\": \"...\"}\` failed with:

\`\`\`
Referenced undefined environment variable: Foo
\`\`\`

Reported in #2089 by @nrusch with a clean repro.

### Fix

Add a small \`CaseInsensitiveEnvironDict\` wrapper in [\`src/rez/utils/data_utils.py\`](src/rez/utils/data_utils.py) and apply it in [\`ActionManager.__init__\`](src/rez/rex.py) when \`parent_environ\` is provided on Windows. The wrapper:

- Stores keys in their original casing (\`__iter__\`, \`__repr__\` see them unchanged) but indexes case-folded for lookup.
- Resolves duplicate-case keys last-write-wins, matching how \`os.environ\` itself collapses case on Windows.
- Is a strict superset of \`MutableMapping\` so existing call sites (\`__getitem__\`, \`__contains__\`, \`get\`, \`keys\`, plus the \`expandvars(text, environ)\` lookups) all continue to work without modification.

The Windows branch is the only new path:

\`\`\`python
if parent_environ is None:
    self.parent_environ = os.environ        # already case-insensitive on Windows
elif platform_.name == \"windows\" and not isinstance(
    parent_environ, CaseInsensitiveEnvironDict
):
    self.parent_environ = CaseInsensitiveEnvironDict(parent_environ)
else:
    self.parent_environ = parent_environ
\`\`\`

### Non-Windows is a strict no-op

\`platform_.name == \"windows\"\` gates the wrapper. On Linux/macOS the assignment reduces byte-for-byte to the pre-fix logic (\`os.environ\` if None, otherwise pass through unchanged).

### Tests

\`test_parent_environ_case_insensitive_on_windows\` in [\`src/rez/tests/test_rex.py\`](src/rez/tests/test_rex.py):

- Direct lookup: dict has only one casing, every variant resolves.
- \`__contains__\` positive + negative.
- End-to-end rex code: \`getenv(\"SOMEVARIABLE\")\` against \`{\"SomeVariable\": \"covfefe\"}\` followed by \`setenv(\"RESULT\", v)\` — asserts the resulting environ contains the right value.

The test skips cleanly on non-Windows.

### Verification

| Platform | State | Result |
|---|---|---|
| Windows 11 + Python 3.11.9 | Pre-fix | new test FAILS at \`test_rex.py:572\` with \`KeyError: 'SOMEVARIABLE'\` — proves the bug is exercised |
| Windows 11 + Python 3.11.9 | Post-fix | new test PASSES + full 16/16 \`test_rex.py\` suite green |
| Linux (\`python:3.11-slim\` Docker) | Post-fix | 15 passed + 1 skipped (Windows-only test), zero regressions |

### Out of scope

- The fix only addresses the \`ActionManager\` lookup path. Other call sites in rez that consume user-provided env dicts (e.g. \`Popen\` invocations) preserve the existing case-sensitive behavior, since on Windows the underlying \`CreateProcess\` call uppercases the keys it actually applies.
- No public API change; \`parent_environ\`'s type annotation (\`Mapping[str, str] | None\`) is unchanged.

### DIFF
--- a/src/rez/rex.py
+++ b/src/rez/rex.py
@@ -23,7 +23,7 @@ from rez.util import shlex_join, is_non_string_iterable
 from rez.utils import reraise
 from rez.utils.execution import Popen
 from rez.utils.sourcecode import SourceCode, SourceCodeError
-from rez.utils.data_utils import AttrDictWrapper
+from rez.utils.data_utils import AttrDictWrapper, CaseInsensitiveEnvironDict
 from rez.utils.formatting import expandvars
 from rez.utils.platform_ import platform_
 
@@ -200,7 +200,19 @@ class ActionManager(object):
         '''
         self.interpreter = interpreter
         self.verbose = verbose
-        self.parent_environ = os.environ if parent_environ is None else parent_environ
+        if parent_environ is None:
+            # os.environ is already case-insensitive on Windows (Python's
+            # os._Environ normalizes keys internally), so leave it alone.
+            self.parent_environ = os.environ
+        elif platform_.name == "windows" and not isinstance(
+            parent_environ, CaseInsensitiveEnvironDict
+        ):
+            # User-provided dicts (e.g. dict(os.environ)) are case-sensitive.
+            # Wrap so env.SomeVariable lookups respect Windows native env
+            # semantics. See AcademySoftwareFoundation/rez#2089.
+            self.parent_environ = CaseInsensitiveEnvironDict(parent_environ)
+        else:
+            self.parent_environ = parent_environ
         self.parent_variables = True if parent_variables is True \
             else set(parent_variables or [])
         self.environ = {}
--- a/src/rez/tests/test_rex.py
+++ b/src/rez/tests/test_rex.py
@@ -547,6 +547,45 @@ class TestRex(TestBase):
         self.assertRaises(RuntimeError,  # no default
                           intersects, ephemerals.get_range("foo.bar"), "0")
 
+    def test_parent_environ_case_insensitive_on_windows(self) -> None:
+        """Regression test for #2089.
+
+        On Windows, env-var lookup is case-insensitive natively, and Python's
+        os.environ preserves that. But a plain dict copy of os.environ (or any
+        user-built dict) is case-sensitive, so passing one as parent_environ
+        used to break rex lookups whenever the casing in the dict differed
+        from the casing referenced in the package commands. The fix wraps
+        any user-provided parent_environ in a case-insensitive view on
+        Windows so env.SomeVariable resolves regardless of how the dict was
+        populated.
+
+        Skipped on non-Windows platforms (the wrapper is a no-op there).
+        """
+        from rez.utils.platform_ import platform_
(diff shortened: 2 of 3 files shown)


Booyaka101:
Closing this — I didn't realize ahead of time that ASF rez requires both an EasyCLA (Linux Foundation) signature and DCO sign-off on every commit, which I'm not in a position to provide right now.

The diagnosis and fix here are sound and 4-way verified locally (FAIL→PASS unit test on Windows, full `test_rex.py` green on both Windows and Linux Docker). If anyone with an existing ASF/LF CLA wants to take the diff forward, the branch is at https://github.com/Booyaka101/rez/tree/fix/rex-parent-environ-case-insensitive-2089 — feel free to cherry-pick (with your own DCO sign-off) and credit @nrusch for the report.

Apologies for the noise.

PR #2098, opened 2026-05-07
Fix rex parent_environ case sensitivity on Windows

### LINKED ISSUE #2089: rex environment lookups do not respect platform case-sensitivity rules

thc1006:
Fixes #2089

On Windows, environment variable keys are case-insensitive natively. `os.environ` preserves that contract through Python's `os._Environ` wrapper, but a plain `dict` copy of it (or any user-built dict) does not. Until now, `ActionManager` consumed whatever `Mapping` the caller passed as-is, so a package whose `commands()` referenced `env.SomeVariable` against a `parent_environ` that held the same variable under a different casing would raise:

```
Error in commands in package '...':
Referenced undefined environment variable: SomeVariable
```

Issue #2089 has a clean reproducer that follows exactly this shape: `dict(os.environ)` is passed to `ResolvedContext.get_environ`, and a package looks up `env.SomeVariable` while the dict only has `SOMEVARIABLE`.

The fix wraps a caller-supplied `parent_environ` in a small read-only, case-normalized snapshot (`_CaseInsensitiveEnvironProxy`), gated on `platform_.name == "windows"`. `os.environ` and `None` are used as-is (they are already case-insensitive on Windows), so their identity and liveness are preserved. On Linux and macOS a caller-supplied mapping retains the existing identity passthrough. Keys are upper-cased once at construction, i.e. a point-in-time snapshot, which matches how `parent_environ` is fixed for an executor's lifetime; iteration matches what `os.environ` yields on Windows, and when the input contains keys that differ only by case the last one encountered wins, mirroring how `os.environ` collapses duplicates. The proxy also exposes `.copy()` (returning a plain dict with upper-cased keys) for callers that expect a dict-like `parent_environ`. The `not isinstance(...)` guard in the gate keeps wrapping idempotent.

Tests:
- `test_case_insensitive_environ_proxy` unit-tests the proxy directly and runs on every platform, so the core normalization logic has real CI coverage: get/contains across casings, missing-key `KeyError`, `.copy()`, iteration/length, and last-write-wins collisions.
- `test_parent_environ_case_insensitive_on_windows` (Windows-only via `@unittest.skipIf`) drives the real package-commands entrypoint — `env.<MixedCase>` — through both `execute_function` and `execute_code` via the `_test` helper, alongside direct `parent_environ[...]` lookups across casings.
- `test_parent_environ_identity_passthrough_on_non_windows` (non-Windows) asserts `ex.manager.parent_environ is env` to guard against accidental wrapping.

Verified locally on Linux: `python -m unittest rez.tests.test_rex` → 17 passed + 1 Windows-only skip. The full `rez-selftest`, `flake8`, and `mypy` matrix runs in CI.

Scope: this makes lookups against a caller-supplied `parent_environ` case-insensitive on Windows — the exact #2089 reproducer. Variables set *within* rex code (`ActionManager.environ`) are stored verbatim and are not case-folded; that is a separate, pre-existing gap and intentionally out of scope here — tracked in #2164.

Reported by @nrusch. Independent diagnosis previously posted in #2093 (closed because the author could not satisfy the EasyCLA/DCO requirement, not for technical reasons).

Disclosure: Claude Code (Opus 4.8) used for adversarial review, edge-case analysis, test design, and implementation of the review feedback.

### DIFF
--- a/src/rez/rex.py
+++ b/src/rez/rex.py
@@ -189,8 +189,8 @@ class ActionManager(object):
         parent_environ: environment to execute the actions within. If None,
             defaults to the current environment. On Windows, a caller-supplied
             mapping is wrapped in a read-only, case-insensitive snapshot so that
-            env-var lookups match native Windows semantics regardless of key
-            casing (see #2089); `os.environ` and None are used as-is.
+            lookups against it are case-insensitive, matching how ``os.environ``
+            behaves natively (see #2089); `os.environ` and None are used as-is.
         parent_variables: List of variables to append/prepend to, rather than
             overwriting on first reference. If this is set to True instead of a
             list, all variables are treated as parent variables.
@@ -218,6 +218,9 @@ class ActionManager(object):
             self.parent_environ = parent_environ
         self.parent_variables = True if parent_variables is True \
             else set(parent_variables or [])
+        # Variables set during rex execution. Keys are stored verbatim and are
+        # not case-folded on Windows (unlike the parent_environ snapshot above);
+        # that broader normalization is a separate concern, tracked in #2164.
         self.environ = {}
         self.formatter = formatter or str
         self.actions = []
@@ -466,8 +469,9 @@ class _CaseInsensitiveEnvironProxy(Mapping):
     """Read-only, case-insensitive snapshot of an env-var mapping.
 
     Used by `ActionManager` to wrap a caller-supplied `parent_environ`
-    on Windows so rex lookups match native Windows env-var semantics
-    regardless of the casing used to populate the input dict.
+    on Windows so lookups against it are case-insensitive regardless of
+    the casing used to populate the input dict, matching how os.environ
+    behaves natively on Windows.
 
     Keys are normalized (upper-cased) once at construction, so this is a
     point-in-time snapshot: later mutations to the source mapping are not
--- a/src/rez/rex.py
+++ b/src/rez/rex.py
@@ -187,7 +187,10 @@ class ActionManager(object):
         interpreter: string or `ActionInterpreter`
             the interpreter to use when executing rex actions
         parent_environ: environment to execute the actions within. If None,
-            defaults to the current environment.
+            defaults to the current environment. On Windows, a caller-supplied
+            mapping is wrapped in a read-only, case-insensitive snapshot so that
+            env-var lookups match native Windows semantics regardless of key
+            casing (see #2089); `os.environ` and None are used as-is.
         parent_variables: List of variables to append/prepend to, rather than
             overwriting on first reference. If this is set to True instead of a
             list, all variables are treated as parent variables.
@@ -204,7 +207,9 @@ class ActionManager(object):
         # behaves that way but a plain dict supplied by the caller does not,
         # so wrap it on the Windows path only. See #2089.
         self.parent_environ: Mapping[str, str]
-        if parent_environ is None:
+        if parent_environ is None or parent_environ is os.environ:
+            # os.environ is already case-insensitive on Windows, so use it
+            # as-is (preserving liveness and identity); None means "current".
             self.parent_environ = os.environ
         elif platform_.name == "windows" and not isinstance(
                 parent_environ, _CaseInsensitiveEnvironProxy):
@@ -458,11 +463,16 @@ class ActionManager(object):
 
 
 class _CaseInsensitiveEnvironProxy(Mapping):
-    """Read-only case-insensitive view over an env-var mapping.
+    """Read-only, case-insensitive snapshot of an env-var mapping.
 
     Used by `ActionManager` to wrap a caller-supplied `parent_environ`
     on Windows so rex lookups match native Windows env-var semantics
     regardless of the casing used to populate the input dict.
+
+    Keys are normalized (upper-cased) once at construction, so this is a
+    point-in-time snapshot: later mutations to the source mapping are not
+    reflected. That matches how `parent_environ` is used in practice (fixed
+    for the lifetime of an executor).
 
     Kept private to this module on purpose; promote to a shared util
     only when a second caller needs it.
@@ -477,13 +487,19 @@ class _CaseInsensitiveEnvironProxy(Mapping):
         return self._data[key.upper()]
 
     def __contains__(self, key):
-        return isinstance(key, str) and key.upper() in self._data
+        return key.upper() in self._data
 
     def __iter__(self):
         return iter(self._data)
 
     def __len__(self):
         return len(self._data)
+
+    def copy(self):
+        # Return a plain dict with normalized (upper-cased) keys, mirroring
+        # os.environ.copy() on Windows. Defensive: lets callers that expect a
+        # dict-like parent_environ (e.g. `self.parent_environ.copy()`) work.
+        return dict(self._data)
 
 
 #===============================================================================
--- a/src/rez/rex.py
+++ b/src/rez/rex.py
@@ -12,8 +12,8 @@ from fnmatch import fnmatch
 from enum import Enum
 from contextlib import contextmanager
 from string import Formatter
-from collections.abc import MutableMapping
-from typing import Any, Iterable, Mapping
+from collections.abc import Mapping, MutableMapping
+from typing import Any, Iterable
 
 from rez.system import system
 from rez.config import config
@@ -200,7 +200,17 @@ class ActionManager(object):
         '''
         self.interpreter = interpreter
         self.verbose = verbose
-        self.parent_environ = os.environ if parent_environ is None else parent_environ
+        # On Windows env-var keys are case-insensitive; os.environ already
+        # behaves that way but a plain dict supplied by the caller does not,
+        # so wrap it on the Windows path only. See #2089.
+        self.parent_environ: Mapping[str, str]
+        if parent_environ is None:
+            self.parent_environ = os.environ
+        elif platform_.name == "windows" and not isinstance(
+                parent_environ, _CaseInsensitiveEnvironProxy):
+            self.parent_environ = _CaseInsensitiveEnvironProxy(parent_environ)
+        else:
+            self.parent_environ = parent_environ
         self.parent_variables = True if parent_variables is True \
             else set(parent_variables or [])
         self.environ = {}
@@ -445,6 +455,35 @@ class ActionManager(object):
 
     def _keytoken(self, key):
         return self.interpreter.get_key_token(key)
+
+
+class _CaseInsensitiveEnvironProxy(Mapping):
+    """Read-only case-insensitive view over an env-var mapping.
+
+    Used by `ActionManager` to wrap a caller-supplied `parent_environ`
+    on Windows so rex lookups match native Windows env-var semantics
+    regardless of the casing used to populate the input dict.
+
+    Kept private to this module on purpose; promote to a shared util
+    only when a second caller needs it.
+    """
+    def __init__(self, data):
+        # keys are upper-cased once on the way in; same shape as
+        # os.environ on Windows. Last write wins on case collisions,
+        # which mirrors how os.environ resolves them too.
+        self._data = {k.upper(): v for k, v in data.items()}
+
+    def __getitem__(self, key):
+        return self._data[key.upper()]
+
+    def __contains__(self, key):
+        return isinstance(key, str) and key.upper() in self._data
+
+    def __iter__(self):
+        return iter(self._data)
+
+    def __len__(self):
+        return len(self._data)
 
 
 #===============================================================================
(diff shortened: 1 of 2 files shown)


thc1006:
Closing+reopening to retrigger Read the Docs (initial build failed during clone, 2s duration, generic clone error — fork branch is public and accessible).

thc1006:
Status note for reviewers:

- The six core GitHub Actions workflows (`tests`, `flake8`, `mypy`, `copyright`, `installation`, `benchmark`) are queued in `action_required` state, awaiting maintainer approval per the first-time-contributor policy. Nothing to do on my side.
- The Read the Docs failure is `git fetch origin pull/2098/head` returning `fatal: couldn't find remote ref pull/2098/head`. The GitHub git API for this PR currently exposes only `refs/pull/2098/merge`; `refs/pull/2098/head` is missing from `matching-refs/pull/2098`, while other open PRs (e.g. #2090, #2091) have both refs as expected. An amend force-push was attempted to nudge GitHub into recreating the head ref but it has not reappeared. This appears to be a GitHub-side propagation glitch independent of the diff. Approving the workflows may incidentally fix it; if not, an RTD admin rebuild should clear it.
- DCO and EasyCLA are both green.

Local pre-push validation: `pytest src/rez/tests/test_rex.py` is green (15 pass + 1 Windows-only skip); full `rez-selftest` on a hermetic Python 3.11.14 container reports `Ran 259 tests in 24.997s, OK (skipped=10)`; `flake8 src/rez/rex.py src/rez/tests/test_rex.py` is clean; `mypy | mypy-baseline filter --ignore-categories=note --allow-unsynced` reports no new violations above the pinned baseline.

src/rez/tests/test_rex.py
```diff
@@ -547,6 +547,39 @@ def test_intersects_ephemerals(self) -> None:
...
+        """Regression for #2089.
+
+        On Windows env vars are case-insensitive; os.environ already
+        behaves that way, but a plain dict copy of it doesn't. Passing
+        such a dict as parent_environ used to break env.<MixedCase>
+        lookups inside commands(). We now wrap on the Windows side.
+        """
+        from rez.utils.platform_ import platform_
+        if platform_.name != "windows":
+            self.skipTest("Windows-only behaviour")
```
maxnbk: Let's use the @unittest.skipIf decorator instead of the in-test skip, to prevent the test from even firing on non-windows hosts.

src/rez/rex.py
```diff
@@ -447,6 +457,35 @@ def _keytoken(self, key):
...
+        # keys are upper-cased once on the way in; same shape as
+        # os.environ on Windows. Last write wins on case collisions,
+        # which mirrors how os.environ resolves them too.
+        self._data = {k.upper(): v for k, v in data.items()}
+
+    def __getitem__(self, key):
+        return self._data[key.upper()]
+
+    def __contains__(self, key):
+        return isinstance(key, str) and key.upper() in self._data
```
maxnbk: I don't believe the `isinstance(key, str)` check is necessary even if harmless. All callers use string keys. The guard means that `non_string_key in proxy` returns `False` instead of raising `TypeError` (which is what a normal dict does).

maxnbk (changes requested):
Apologies for taking a long time to get to this review.

I've commented on a few specific items.

More generally, `_CaseInsensitiveEnvironProxy` could have a `.copy()` method added, which would be defensive against future code calling `self.parent_environ.copy()` to return a plain `dict` with uppercased keys.

`ActionManager.__init__` docstring could document the Windows wrapping behavior.

And last, I feel it would be useful to consider a non-Windows identity passthrough test. i.e. assert `ex.manager.parent_environ is env` on non-Windows to guard against accidental wrapping. With that in mind, the test skip could be avoided and just assert differently for different host types, or be a different test skipping windows hosts, either way.

If you are using an LLM to produce this PR, please add a line to the PR description disclosing the usage, i.e. "Disclosure: <model> used for <task> (whichever things it was used for.)

src/rez/tests/test_rex.py
thc1006: Done in 1c60859e — switched to `@unittest.skipIf(platform_.name != "windows", ...)` with `platform_` imported at module level. I also added `test_parent_environ_identity_passthrough_on_non_windows` (asserts `parent_environ is env` on non-Windows) and a platform-independent `test_case_insensitive_environ_proxy`, so the core normalization logic is exercised on Linux/macOS CI rather than skipped everywhere but Windows.

src/rez/rex.py
thc1006: Done in 1c60859e — dropped the guard; `__contains__` is now just `return key.upper() in self._data`. Agreed the type contract is `Mapping[str, str]`, so non-str keys are out of contract anyway.

maxnbk (changes requested):
Tests failed, I believe it is because rex `env` is shadowing your variables. I am guessing something like this might fix it. GitHub UI is misbehaving so I wasn't able to apply it as a suggestion to click apply for:
```
         """Regression for #2089.
         ...
         """
-        env = {"SomeVariable": "covfefe"}
-        ex = self._create_executor(env)
+        parent_env = {"SomeVariable": "covfefe"}
+        ex = self._create_executor(parent_env)

         # any casing of the key resolves
         self.assertEqual(ex.manager.parent_environ["SomeVariable"], "covfefe")
         ...

         # end-to-end via the real package-commands entrypoint
         def _rex() -> None:
             env.RESULT = env.SOMEVARIABLE  # noqa: F821 - rex builtin

         self._test(
             func=_rex,
-            env={"SomeVariable": "covfefe"},
+            env=parent_env,
             expected_actions=[Setenv("RESULT", "covfefe")],
             expected_output={"RESULT": "covfefe"},
         )
```
(It may be wise to check the tests for any similar pattern and update those just on the off-chance someone introduces rex in any of those tests in the future and the same gotcha resurfaces too)

thc1006:
Thanks for the diagnosis — that was exactly it. `execute_function` rebuilds the function with `closure=func.__closure__`, so once the test declared a local `env` dict the nested `_rex` captured it as a closure cell (`LOAD_DEREF`) and shadowed the rex `env` builtin injected into globals — `env.SOMEVARIABLE` then hit a plain `dict`. `execute_code` was fine (fresh compile from source), which is why only the function path failed.

Fixed in 1c3299ea: renamed the caller dict to `parent_env` in both new tests, so `_rex`'s `env` is a plain global again and resolves to the rex builtin. Added a short comment on the test flagging the gotcha.

Per your future-proofing note, I audited the rest of `test_rex.py`: no other test declares a local `env` — they all pass `env=` as a kwarg to `_test`, so none can shadow the builtin.

I also closed the codecov gap on the Windows-only branch (it's skipped on the Linux/macOS jobs):
- `test_parent_environ_os_environ_passthrough` — covers the `parent_environ is os.environ` identity branch.
- `test_parent_environ_case_insensitive_windows_mocked` — mocks `platform_.name` so the case-insensitive wrapping branch and the full `env.<MixedCase>` end-to-end run on every host. As a bonus that validates the Windows logic off-Windows, which is what would have caught this shadowing bug locally.

maxnbk (approved):
It LGTM now. Thank you for making the requested changes @thc1006 . I marked Approve (from me) and put this into "Next" milestone for merge in the next release. No further action needed unless some other reviewer requests changes.

src/rez/rex.py
```diff
@@ -447,6 +465,47 @@ def _keytoken(self, key):
         return self.interpreter.get_key_token(key)
 
 
+class _CaseInsensitiveEnvironProxy(Mapping):
+    """Read-only, case-insensitive snapshot of an env-var mapping.
+
+    Used by `ActionManager` to wrap a caller-supplied `parent_environ`
```
JeanChristopheMorinPerso:
```suggestion
    Used by :class:`ActionManager` to wrap a caller-supplied ``parent_environ``
```

src/rez/rex.py
```diff
@@ -187,7 +187,10 @@ def __init__(self, interpreter: ActionInterpreter,
         interpreter: string or `ActionInterpreter`
             the interpreter to use when executing rex actions
         parent_environ: environment to execute the actions within. If None,
-            defaults to the current environment.
+            defaults to the current environment. On Windows, a caller-supplied
+            mapping is wrapped in a read-only, case-insensitive snapshot so that
+            lookups against it are case-insensitive, matching how ``os.environ``
+            behaves natively (see #2089); `os.environ` and None are used as-is.
```
JeanChristopheMorinPerso:
```suggestion
            lookups against it are case-insensitive, matching how :data:`os.environ`
            behaves natively (see #2089); :data:`os.environ` and None are used as-is.
```

src/rez/rex.py
```diff
@@ -447,6 +465,47 @@ def _keytoken(self, key):
         return self.interpreter.get_key_token(key)
 
 
+class _CaseInsensitiveEnvironProxy(Mapping):
+    """Read-only, case-insensitive snapshot of an env-var mapping.
+
+    Used by `ActionManager` to wrap a caller-supplied `parent_environ`
+    on Windows so lookups against it are case-insensitive regardless of
+    the casing used to populate the input dict, matching how os.environ
```
JeanChristopheMorinPerso:
```suggestion
    the casing used to populate the input dict, matching how :data:`os.environ`
```

src/rez/rex.py
```diff
@@ -447,6 +465,47 @@ def _keytoken(self, key):
...
+    behaves natively on Windows.
+
+    Keys are normalized (upper-cased) once at construction, so this is a
+    point-in-time snapshot: later mutations to the source mapping are not
+    reflected. That matches how `parent_environ` is used in practice (fixed
```
JeanChristopheMorinPerso:
```suggestion
    reflected. That matches how ``parent_environ`` is used in practice (fixed
```

src/rez/rex.py
```diff
@@ -447,6 +465,47 @@ def _keytoken(self, key):
         return self.interpreter.get_key_token(key)
 
 
+class _CaseInsensitiveEnvironProxy(Mapping):
```
JeanChristopheMorinPerso, line 468:
```suggestion
class _CaseInsensitiveEnvironProxy(Mapping[str, str]):
```
?

thc1006: tl;dr: happy to take the `Mapping[str, str]` change, but it should use `typing.Mapping` to stay 3.8-safe. Separately, the failing Ruff check looks like a pre-existing issue on `main`, not something in this PR.

On the base class: `Mapping` here is imported from `collections.abc`, and using one of those as a subscripted base (`class X(Mapping[str, str])`) only works on Python 3.9+. On 3.8 it raises `TypeError` at import, and we still support 3.8 (`python_requires >= 3.8`). The other parametrized bases in the repo, like `AttrDictWrapper(MutableMapping[str, Any])` and `PackageOrderList(List[...])`, go through `typing` for that reason. I can switch this to `from typing import Mapping` and apply `Mapping[str, str]`, or leave it bare, whichever you prefer.

On the Ruff failure: it's `ISC004` at `src/rez/utils/platform_.py:524` (the PowerShell command list), which this branch doesn't touch. It reproduces on a clean `main` (`4946f9de`) with ruff 0.16.0, the newest release matching `>= 0.15.0` that `ruff-action` pulls in. So it reads as the new ISC004 rule catching an existing line rather than anything from here. I'm happy to fold the one-line fix into this PR if that unblocks things, otherwise it may fit better as its own change on `main`.

Thanks for the sphinx pass too, the `:data:`/`:class:` refs read a lot cleaner.

maxnbk:
The ruff issue is fixed on main, so a rebase will pick it up and validate that no new ruff issues pop up.

PR #2103, opened 2026-05-14
Add support for "--variant -1" style syntax for rez-build

sanikache:
This pull request adds support for specifying negative variant indices (e.g., --variants -1) in rez build and copy workflows. 
Negative indices are now accepted to select variants counting backward from the end, mirroring Python’s negative indexing convention.

### DIFF
--- a/src/rez/build_process.py
+++ b/src/rez/build_process.py
@@ -210,12 +210,15 @@ class BuildProcessHelper(BuildProcess):
                        **kwargs) -> tuple[int, list[str | None]]:
         """Iterate over variants and call a function on each."""
         if variants:
-            present_variants = range(self.package.num_variants)
-            invalid_variants = set(variants) - set(present_variants)
+            n = self.package.num_variants
+            # Valid indices: 0..n-1 and their negative equivalents -n..-1
+            present_variants = set(range(-n, n))
+            invalid_variants = sorted(set(variants) - present_variants)
             if invalid_variants:
                 raise BuildError(
                     "The package does not contain the variants: %s"
-                    % ", ".join(str(x) for x in sorted(invalid_variants)))
+                    % ", ".join(str(v) for v in invalid_variants))
+            variants = sorted({v % n for v in variants})
 
         # iterate over variants
         results = []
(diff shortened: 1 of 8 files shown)


maxnbk:
One failing test in the jobs, that I see at a glance (I pulled linux 3.10 as a sample)
```
FAILED installdir/lib/python3.10/site-packages/rez/tests/test_context.py::TestContext::test_bundled - rez.exceptions.PackageCopyError: The package does not contain the variants: None
```

sanikache:
Taking a look . .

sanikache:
It's the change in `package_copy.py` that's breaking `test_context.py::TestContext::test_bundled `. It will fail for packages with no declared variants. Pushing a commit for that shortly.

sanikache:
It looks like there is no test for copy_package on a non-variants package. I added one to cover that.

maxnbk:
Overall I'm happy, I would ask the question if maybe, since the build_process and package_copy is _almost_ the same, if there is a sane/easy way to extract, put the variant-counting/set/invalid logic in a shared location, and call it, but the code is subtly different so perhaps it would not be completely simple? Happy to discuss.

sanikache:
For sure, let me take a look . .

sanikache:
What do you think about adding a `resolve_variant_indices` function in `package_copy` and then have `build_process` import `package_copy`. It has no circular import risk.

maxnbk:
Bit subjective... I think if it were me, I'd probably look at utils.py, or files in rez.utils , and pull it from one such place, but I don't think in there is any reason to choose package_copy over build_process and vice versa, probably just whichever has less likelihood to be brought in by the other, so your discretion is fine I would think.

sanikache:
Oh!, `utils.py` it is. `rez.utils.data_utils.py` maybe?  but I like `util.py`. It feel a little awkward having build_process import package_copy now that I think about it.

maxnbk (approved):
Personally I like it. I would like one other 👍  but I don't see anything structural to change, maybe just nitpicks if anything.

JeanChristopheMorinPerso (approved):
Thank you! Also, thank you for adding tests!

PR #1856, opened 2024-10-15
fix: realpath on windows changed behavior in python 3.8 to expand mapped drive letters to their UNC path names

cfxegbert:
Fixes https://github.com/AcademySoftwareFoundation/rez/issues/1438

`os.path.realpath` on Python 3.8 Windows changed.  When resolving mapped drive realpath returns the UNC path without the drive letter.

This pull restores the Python 3.7 Windows behavior of `realpath` being a wrapper around `abspath`.  The behavior can be changed with the rezconfig setting `windows_unc_path`

Issue #1438, opened 2023-02-02
Environment resolution uses UNC paths with Python 3.10, 3.7 was using mounted network drives.


KelSolaar:
Hello,

As we are trying to update our Rez version from **2.91**, I noted that the resolution now uses UNC paths:

### Rez v2.112.0
```
C:\Users\tmansencal>rez-env dcraw

You are now in a rez-configured environment.

resolved by tmansencal@***, on Fri Feb 03 08:48:25 2023, using Rez v2.112.0

requested packages:
dcraw
~platform==windows           (implicit)
~arch==AMD64                 (implicit)
~os==windows-10.0.***.SP0  (implicit)

resolved packages:
arch-AMD64        c:\***\packages\***\arch\AMD64
dcraw-9.27        \\***\***\packages\thirdparty\dcraw\9.27\platform-windows\arch-AMD64
platform-windows  c:\***\packages\***\platform\windows
```

### Rez v2.2.91.0
```
C:\Users\tmansencal>rez-env dcraw

You are now in a rez-configured environment.

resolved by tmansencal@***, on Fri Feb 03 08:53:09 2023, using Rez v2.91.0

requested packages:
dcraw
~platform==windows           (implicit)
~arch==AMD64                 (implicit)
~os==windows-10.0.***.SP0  (implicit)

resolved packages:
arch-AMD64        c:\***\packages\***\arch\AMD64
dcraw-9.27        s:\packages\thirdparty\dcraw\9.27\platform-windows\arch-AMD64
platform-windows  c:\***\packages\***\platform\windows
```

**Environment**
* OS: Window 10
* Rez version (eg "2.100.0"): v2.112.0
* Rez python version (output of "rez-python --version"): Python 3.10.9

I do think that this is strictly caused by the Python difference but would be keen to know if anyone has seen that already before I start digging.

**To Reproduce**

1. Build Rez with Python 3.7, resolve environment from mounted drive.
2. Build Rez with Python 3.10, resolve environment from mounted drive.
3. Compare the resolved package paths.

**Expected behavior**
The mounted drive package path should be used as UNC paths wreck havoc in many CMD scripts.

**Actual behavior**
The UNC package path is used instead.

flord:
I would argue this should be a config. Mapped drive letters can wreck havoc in other setups, like the one we have here. We only use powershell where UNC works fine.

KelSolaar:
[...]

PR #2104, opened 2026-05-14
fix: windows drive letters remap to unc paths

### LINKED ISSUE #2045: Windows: Repository mismatch with mounted network drives
I am experiencing the same issue reported in https://github.com/AcademySoftwareFoundation/rez/pull/957 where my mounted network drive is causing a location mismatch error when using Rez on Windows.

**Repro:**

```
any(p.is_local for p in ResolvedContext.get_current().resolved_packages)
```

**Error:**

```
  File "<stdin>", line 1, in <genexpr>
  File "n:\software\rez\os-windows-10.0.19045\packages\rez\3.2.1\platform-windows\arch-AMD64\os-windows-10.0\rez\utils\data_utils.py", line 244, in __get__
    result = self.func(instance)
             ^^^^^^^^^^^^^^^^^^^
  File "n:\software\rez\os-windows-10.0.19045\packages\rez\3.2.1\platform-windows\arch-AMD64\os-windows-10.0\rez\packages.py", line 108, in is_local
    self.config.local_packages_path)
    ^^^^^^^^^^^
  File "n:\software\rez\os-windows-10.0.19045\packages\rez\3.2.1\platform-windows\arch-AMD64\os-windows-10.0\rez\packages.py", line 102, in config
    return self.resource.config or config
           ^^^^^^^^^^^^^^^^^^^^
  File "n:\software\rez\os-windows-10.0.19045\packages\rez\3.2.1\platform-windows\arch-AMD64\os-windows-10.0\rez\utils\data_utils.py", line 555, in func
    value = getattr(self.wrapped, key, None)
                    ^^^^^^^^^^^^
  File "n:\software\rez\os-windows-10.0.19045\packages\rez\3.2.1\platform-windows\arch-AMD64\os-windows-10.0\rez\package_resources.py", line 519, in wrapped
    return self.parent
           ^^^^^^^^^^^
  File "n:\software\rez\os-windows-10.0.19045\packages\rez\3.2.1\platform-windows\arch-AMD64\os-windows-10.0\rez\utils\data_utils.py", line 244, in __get__
    result = self.func(instance)
             ^^^^^^^^^^^^^^^^^^^
  File "n:\software\rez\os-windows-10.0.19045\packages\rez\3.2.1\platform-windows\arch-AMD64\os-windows-10.0\rezplugins\package_repository\filesystem.py", line 272, in parent
    package = self._repository.get_resource(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "n:\software\rez\os-windows-10.0.19045\packages\rez\3.2.1\platform-windows\arch-AMD64\os-windows-10.0\rez\package_repository.py", line 465, in get_resource
    handle = self.make_resource_handle(resource_key, **variables)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "n:\software\rez\os-windows-10.0.19045\packages\rez\3.2.1\platform-windows\arch-AMD64\os-windows-10.0\rez\package_repository.py", line 443, in make_resource_handle  
    raise ResourceError("location mismatch - requested %r, repository "
rez.exceptions.ResourceError: location mismatch - requested 'n:\\network\\rez\\os-windows-10.0.19045\\packages', repository location is '\\\\mount-location\\software\\rez\\os-windows-10.0.19045\\packages'
```

It seems the fix for this to use `canonical_path` proposed in https://github.com/AcademySoftwareFoundation/rez/pull/957/files was rolled back in https://github.com/AcademySoftwareFoundation/rez/commit/72d37712cc43f7a4fa3a14383c22b4f1a903de78.    Is there any workaround I can use for this?

maxnbk:
Closes #2045 and #1438 .
Partial but not complete fix for #1909 (Intentional, I feel that the work that was occurring in the gitbash PR would ultimately resolve this behavior?) .

Explanations:

`os.path.realpath` on Python 3.8+ Windows silently expands mapped drive letters to their underlying UNC paths. Because `canonical_path` calls `realpath`, `FileSystemPackageRepository` was storing a UNC `self.location` even when the caller supplied a drive-letter path. This caused `make_resource_handle`, which does a raw string comparison, to raise `ResourceError` on every drive-letter handle against a UNC-stored repository.

The fix:

On Windows, `canonical_path` now uses `os.path.abspath` instead of `os.path.realpath`. This preserves path style (drive-letter stays drive-letter, UNC stays UNC) and restores the pre-3.8 normalization behavior without a network-resolution side-effect. A `make_resource_handle` override is also added to the `filesystem` plugin to handle residual case or separator mismatches via `canonical_path`.

Extension of work:

A `resolve_links_on_windows` config flag (default `False`) has been added for sites that need some form of symlink/junction resolution on Windows. When enabled, a purpose-built `_windows_realpath` walks components of the path using `os.readlink`, resolving real symlinks without actually touching the drive root, in order to avoid the UNC-expansion side-effect entirely. Long-path support (`\\?\` prefix handling) is included, with the relevant test skipped on hosts that do not have `LongPathsEnabled=1` in the registry. Note: I'm aware that we need to move off of the winreg module, but since we have it for now, might as well use it.

Testing:

This was tested against an actual Drive-letter-map-with-matching-UNC-network-share .

==== EDIT ====

This PR has been extended with a broader `real_path` migration.

`real_path()` helper - A new utility function in `filesystem.py` that replaces `os.path.realpath` at all call sites where path-style stability is required. 
On Windows, it is delegated to `os.path.abspath` to preserve the drive-letter style.
On all other platforms, it is delegated to `os.path.realpath` to preserve symlink resolution.
Importantly, no config-flag is required - the correct behavior is derived from the OS, not user preference. `canonical_path` from the original PR slice now calls `real_path` internally.

Call-site migration - All non-test `os.path.realpath` code that touched stored, serialized, or compared paths have been updated:
- `filesystem.py` - `canonical_path`, `is_subdirectory`
- `resolved_context.py` - `_adjust_variant_for_bundling`
- `serialise.py` - `open_file_for_write`, `load_from_file`
- `suite.py` - `Suite.save`, `Suite.load`
- `system.p`y - `rez_bin_path`
- `rezplugins/build_system/cmake.py` - also corrects a pre-existing dead-code bug (`os.path.abspath` used as a truthiness check where `os.path.isabs` was intended)
- `utils/py_dist.py` - egg-to-rez file path comparisons and copy destinations

Relationship to prior work - Several earlier attempts addressed this problem class but were either incomplete/partial, lacked tests, or relied on opt-in config flags:
  - #957 - first fix for the make_resource_handle location mismatch; was reverted. This PR restores the intent of that fix.
  - #1543 - introduced use_canonical_path config flag (default True); only patched FileSystemPackageRepository.__init__ and required users to opt out. Superseded by this work.
  - #1856 - introduced a real_path helper gated on a windows_unc_path config flag; covered three call sites (canonical_path, _adjust_variant_for_bundling, rez_bin_path). The approach here is the same in spirit but unconditional and covers all remaining call sites. Superseded by this work.

==== END EDIT ====

==== EDIT 2 (2026-07-17) ====

This PR has been further extended with two enhancements to `_windows_realpath` (the opt-in `resolve_links_on_windows` code path):

Intermediate-directory symlink re-walk - The original component walk resolved symlinks only on the final path component. If an intermediate directory was itself a symlink (e.g., `/a/b/target` where `/a` → `/x`), the intermediate symlink was not resolved. The walk has been converted from a linear `for` loop to a component-stack `while` loop: when a symlink resolves to an absolute target, the target's components are pushed back onto the front of the stack so that any symlinks in the target's prefix are re-walked transitively. A `total_depth` budget of 40 hops guards against symlink loops (the same limit `os.path.realpath` uses) and is enforced at both the inner resolution loop and the re-walk point.

Junction detection on Python 3.8–3.11 - `os.path.islink()` does not detect Windows junctions (mount points) on any Python version; it checks `IO_REPARSE_TAG_SYMLINK` only. `os.path.isjunction()` was added in Python 3.12 but is unavailable on 3.8–3.11, which rez still supports. A `_is_link_or_junction` helper now provides three-tier detection: `os.path.islink` → `os.path.isjunction` (3.12+) -> `st_reparse_tag == 0xA0000003` fallback (the `IO_REPARSE_TAG_MOUNT_POINT` value from the Windows SDK, exposed via `os.lstat().st_reparse_tag` on Python 3.8+). This allows `_windows_realpath` to resolve junctions via `os.readlink` on all supported Python versions. Note: `os.readlink` on a junction returns the substitute name with a `\??\` device-namespace prefix, which CPython internally converts to `\\?\` - the existing extended-length prefix stripping handles this correctly.

==== END EDIT 2 ====

==== EDIT 3 2026-07-29 ====

Three rounds of adversarial self-review were performed against the branch, with findings verified against the codebase. Many valid findings were addressed, and some deliberate non-fixes are documented below with rationale.

Addressed:
* `serialize.py` file_cache key mismatch. `open_file_for_write` used `os-path.realpath(real_path(filepath))` for the cache key, which on Windows expands driver letters to UNC. `load_from_file` used `real_path(filepath)` , preserving the drive letter. Write stored under `\\server\share\foo.py`, while read looked it up as `N:\foo.py` , creating a cache miss. Fixed by using `real_path` for the cache key (matching the read side) and `resolve_path` separately for the physical write target.
* `Suite.save()` re-introduced UNC expansion. `os.path.realpath(path)` was used before `safe_rmtree`, expanding `N:\suites\mysuite` to `\\nas\studio\stuites\mysuite`. Replaced with a new `resolve_path()` function (see below).
* `test_save_through_symlink_uses_samefile` would fail on Windows CI. On Windows default config, `resolve_path` returns `abspath` (no symlink resolution), so `rmtree_arg` would be the unresolved alias path. Fixed by mocking `rez.suite.resolve_path` to return the symlink target, so the test verifies rmtree-call-wiring identically on all platforms.
* `_add_extended_length_prefix` didn't guard against already-prefixed paths. Calling it on `\\?\C:\foo` would produce `\\?UNC\?\C:\foo`. Added a case-sensitive already-prefixed guard.
* `samefile` fallback did filesystem IO on every mismatch. A 50-package resolve with legacy UNC handle would do 150 network stats. Added class-level `_location_aliases` memoization on `FileSystemPackageRepository`. The first mismatch still hits, but subsequent lookups for the same path-pair are fast.
* To the above, stale `False` results in the memoization cache. Only `True` results are cached. A `False` (genuine mismatch) is re-checked every time, so a repo moved away and back in the same process is detected correctly.
* Useless test assertion: `os.path.islink(alias)` was always `True` (the test created the symlink), so the `or` short-circuited. Replaced with an `assertNotIn("alias_suite", rmtree_arg)` / `assertIn("actual_suite", rmtree_arg)`.
* `resolve_path` imported locally inside `open_file_for_write`. Moved to module-level alongside `real_path`.
* Misleading `Suite.save()` comment, claiming that symlink resolution happens, but on Windows default config, it doesn't. Updated to accurately describe the Windows default-config behavior.

New function: `resolve_path()`
* This was added to `utils.filesystem.py` to fill the gap between `real_path` (no symlink resolution) and `os.path.realpath` (UNC expansion on Windows). This resolves symlinks in a form-preserving way. It uses `_windows_realpath` when `resolve_links_on_windows` is configured `true`, otherwise `abspath`. Used by `serialize.py` (write target) and `suite.py` (rmtree target).

Identified but not addressed:
* `is_subdirectory` / `relpath` case inconsistency: `is_subdirectory` now canonicalizes (lowercases on Windows), but the subsequent `os.path.relpath` in `_adjust_variant_for_bundling` uses `real_path` (preserves case). If paths differ only in case, `is_subdirectory` returns `True` but `relpath` could produce a different result. Rationale: Same-case is the norm in practice. The inconsistency only manifests with mixed-case paths that are otherwise identical. Not worth the complexity of threading canonical paths through relpath at this time. Recommend addressing only if a real-world case mismatch is reported.
* Windows default config behavior with `resolve_path` does not resolve symlinks. When `resolve_links_on_windows=False` (default), `resolve_path` returns `abspath` (no symlink resolution). This means `Suite.save()` through a symlink alias will sever the symlink, and `_windows_realpath` long-path probe prefixing only helps when the flag is enabled. Rationale: This is the inherent tradeoff of the opt-in flag design. The alternative, always resolving symlinks on Windows, is exactly the behavior `os.path.realpath` provides, which this PR exists to avoid. The comment in `Suite.save()` now accurately documents this. Recommend leaving as-is. Users who need symlink resolution on Windows should enable the `resolve_links_on_windows` flag, and after this behavior is proven out, the flag could be removed and behavior defaulted-on. Documented in the config-setting comment.
* `probe` vs `candidate` asymmetry for long paths. `probe` is prefixed for filesystem IO. `candidate`, used for `os.path.join` and `os.path.dirname`, is not. Rationale: `os.path.join` and `dirname` are pure string operations with no filesystem access, so the unprefixed `candidate` works fine. No current code does filesystem IO on `candidate` directly. Recommend leaving as-is. If future code adds filesystem IO on `candidate`, prefix it at that point.
* `make_resource_handle` can't bridge drive-letter <-> UNC paths. The override canonicalizes via `canonical_path`, which preserves form. A drive-letter `N:\Packages` and UNC `\\nas\studio\Packages` canonicalize to different strings. Rationale: `get_resource_from_handle` (the resolution side) has the same `samefile` fallback that covers this. Handle creation is less critical. If the handle is created in the wrong form, the fallback catches it on resolution. Recommend leaving as acceptable for now. If handle-creation mismatches become a problem, add a `samefile` check to `make_resource_handle` as well.

New behaviors have reasonably comprehensive test coverage, and some test fixes are genuine test fixes for previous behavior that was effectively self-asserting.

==== END EDIT 3 ====

Disclosure:

This PR was AI-assisted with Claude Code, Sonnet 4.6, primarily for diagnostics, logic-tracing, documentation and edge-case verification.

This PR was AI-assisted with GLM-5.2, primarily for python-version-specific research and edge-case detection.

### DIFF
--- a/src/rez/rezconfig.py
+++ b/src/rez/rezconfig.py
@@ -125,6 +125,26 @@ package_definition_build_python_paths = []
 # For further information, see :ref:`package-definition-sharing-code`.
 package_definition_python_path = None
 
+# On Windows, whether to resolve symbolic links and junction points when
+# normalising filesystem paths (primarily inside ``canonical_path``).
+#
+# When ``False`` (default), rez uses ``os.path.abspath``, which normalises
+# separators and ``.``/``..`` components without following symlinks and without
+# expanding mapped drive letters to their UNC equivalents.  This preserves the
+# path style supplied by the caller (drive-letter input, drive-letter output,
+# UNC input, UNC output), effectively defaulting to pre-Python-3.8 behaviour of
+# ``os.path.realpath`` on Windows.
+#
+# When ``True``, rez performs a component-by-component walk using
+# ``os.path.islink`` / ``os.readlink``.  This resolves actual symlinks and
+# junction points without the drive-letter-to-UNC side-effect that
+# ``os.path.realpath`` introduced in Python 3.8.  Useful when package
+# repositories are accessed through directory symlinks or junctions.
+#
+# This setting is a no-op on non-Windows platforms, which always resolve
+# symlinks via ``os.path.realpath``.
+resolve_links_on_windows = False
+
 
 ###############################################################################
 # Extensions
--- a/src/rez/rezconfig.py
+++ b/src/rez/rezconfig.py
@@ -130,15 +130,15 @@ package_definition_python_path = None
 #
 # When ``False`` (default), rez uses ``os.path.abspath``, which normalises
 # separators and ``.``/``..`` components without following symlinks and without
-# expanding mapped drive letters to their UNC equivalents.  This preserves the
+# expanding mapped drive letters to their UNC equivalents. This preserves the
 # path style supplied by the caller (drive-letter input, drive-letter output,
 # UNC input, UNC output), effectively defaulting to pre-Python-3.8 behaviour of
 # ``os.path.realpath`` on Windows.
 #
 # When ``True``, rez performs a component-by-component walk using
-# ``os.path.islink`` / ``os.readlink``.  This resolves actual symlinks and
+# ``os.path.islink`` / ``os.readlink``. This resolves actual symlinks and
 # junction points without the drive-letter-to-UNC side-effect that
-# ``os.path.realpath`` introduced in Python 3.8.  Useful when package
+# ``os.path.realpath`` introduced in Python 3.8. Useful when package
 # repositories are accessed through directory symlinks or junctions.
 #
 # This setting is a no-op on non-Windows platforms, which always resolve
--- a/src/rez/tests/test_utils_filesystem.py
+++ b/src/rez/tests/test_utils_filesystem.py
@@ -168,6 +168,63 @@ class TestCanonicalPathIdempotency(TestBase, TempdirMixin):
         self.assertEqual(once, twice)
 
 
+class TestRealPath(TestBase, TempdirMixin):
+    """Cross-platform regression guards for real_path().
+
+    real_path() is the form-stable absolute-path helper for file I/O and
+    path storage.  Key contracts verified here:
+
+    - relative paths become absolute
+    - case is never lowercased (unlike canonical_path on case-insensitive FSes)
+    - on non-Windows, symlinks are resolved (delegates to os.path.realpath)
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        TempdirMixin.setUpClass()
+
+    @classmethod
+    def tearDownClass(cls):
+        super().tearDownClass()
+        TempdirMixin.tearDownClass()
+
+    def test_result_is_absolute(self):
+        self.assertTrue(os.path.isabs(real_path(self.root)))
+
+    def test_preserves_case(self):
+        """real_path() must not lowercase path components.
+
+        canonical_path() lowercases on case-insensitive filesystems;
+        real_path() must not, so stored paths are not corrupted for
+        case-sensitive consumers (e.g. a Linux NFS mount from Windows).
+        """
+        mixed = os.path.join(self.root, "FooBar")
+        os.makedirs(mixed)
+        result = real_path(mixed)
+        self.assertIn(
+            "FooBar", result,
+            "real_path() lowercased a path component: %r" % result,
+        )
+
+    @unittest.skipIf(
+        platform_.name == "windows",
+        "Symlink resolution via realpath is non-Windows only.",
+    )
+    def test_resolves_symlinks_on_non_windows(self):
+        """On non-Windows real_path() resolves symlinks (os.path.realpath).
+
+        Guards against real_path() being switched to abspath on all
+        platforms, which would silently break symlink resolution on
+        Linux/macOS.
+        """
+        target = os.path.join(self.root, "real_target")
+        link = os.path.join(self.root, "sym_link")
+        os.makedirs(target)
+        os.symlink(target, link)
+        self.assertEqual(real_path(link), real_path(target))
+
+
 # Simulate py3.8+ Windows os.path.realpath: N:\ expands to \\nas\studio\
 _MOCK_DRIVE_TO_UNC = {"n": "\\\\nas\\studio"}
 
--- a/src/rez/tests/test_utils_filesystem.py
+++ b/src/rez/tests/test_utils_filesystem.py
@@ -461,3 +461,129 @@ def _multi_drive_unc_realpath(path):
         if drive in _MOCK_MULTI_DRIVE_TO_UNC:
             return _MOCK_MULTI_DRIVE_TO_UNC[drive] + norm[2:]
     return path
+
+
+@unittest.skipIf(
+    platform_.name != "windows",
+    "Windows-specific real_path() call-site regression tests",
+)
+class TestRealPathCallSitesWindowsUNC(TestBase):
+    """tests that document where os.path.realpath must be replaced with
+    real_path() to avoid drive-letter -> UNC expansion on Windows.
+
+    Each test demonstrates a concrete failure against direct use of
+    os.path.realpath. They represent call sites that need real_path
+    """
+
+    def test_bundle_relpath_does_not_raise_on_different_unc_roots(self):
+        """os.path.relpath raises ValueError when realpath expands two drive-letter
+        paths to different UNC roots.
+
+        Mirrors the pattern in resolved_context.py::
+            relpath(realpath(repo_path), realpath(bundle_path))
+
+        This test currently FAILS because os.path.realpath expands N:\\ ->
+        \\\\nas\\studio\\ and M:\\ -> \\\\backup\\bundles\\, making relpath
+        raise ValueError (can't make-relative across different UNC roots).
+        Fix: replace both realpath() calls with real_path().
+        """
+        repo_path = "N:\\packages\\mypkg"
+        bundle_path = "M:\\bundles\\mybundle"
+
+        with unittest.mock.patch("os.path.realpath", side_effect=_multi_drive_unc_realpath):
+            try:
+                os.path.relpath(
+                    os.path.realpath(repo_path),
+                    os.path.realpath(bundle_path),
+                )
+            except ValueError:
+                self.fail(
+                    "os.path.realpath expanded drive-letter paths to different "
+                    "UNC roots, causing os.path.relpath to raise ValueError. "
+                    "Replace os.path.realpath with real_path() at this call site."
+                )
+
+    def test_suite_save_does_not_raise_suiteerror_when_saving_over_loaded_suite(self):
+        """Suite.save() raises SuiteError when the save path differs from
+        suite.load_path due to UNC expansion.
+
+        Suite.save() calls os.path.realpath(path) and compares it to
+        os.path.realpath(self.load_path). When both were originally the
+        same drive-letter path (e.g. N:\\suites\\mysuite), py3.8+ realpath
+        expands each independently to the same UNC path, so they still
+        compare equal and no error is raised - on the first call.
+
+        The real failure occurs when the suite was loaded from a drive-letter
+        path but is saved to what looks like the same path: the UNC expansion
+        changes the string, so subsequent code that compares the stored
+        load_path against the UNC-expanded save path will see a mismatch.
+
+        This test documents the fragility. Fix: replace os.path.realpath
+        with real_path() in Suite.save() and Suite.load().
+        """
+        from rez.suite import Suite, SuiteError
+
+        drive_path = "N:\\suites\\mysuite"
+        suite = Suite()
+        suite.load_path = drive_path
+
+        patch_context_names = unittest.mock.patch.object(
+            Suite, "context_names",
+            new_callable=unittest.mock.PropertyMock,
+            return_value=[],
+        )
+        with unittest.mock.patch("os.path.realpath", side_effect=_unc_expanding_realpath), \
+             unittest.mock.patch("os.path.exists", return_value=True), \
+             unittest.mock.patch("os.makedirs") as mock_makedirs, \
+             unittest.mock.patch("shutil.rmtree") as mock_rmtree, \
+             unittest.mock.patch("builtins.open", unittest.mock.mock_open()), \
+             patch_context_names, \
+             unittest.mock.patch.object(suite, "to_dict", return_value={}), \
+             unittest.mock.patch.object(suite, "get_tools", return_value={}):
+            try:
+                suite.save(drive_path)
+            except SuiteError as e:
+                self.fail("Suite.save() raised SuiteError: %s" % e)
+
+            # Reachable only after the fix - verify the save actually ran:
+            # existing dir must be cleared before re-creating
+            mock_rmtree.assert_called_once()
+            # suite directory must be (re)created
+            mock_makedirs.assert_called()
+
+    def test_suite_load_stores_drive_letter_load_path_not_unc(self):
+        """Suite.load() must store load_path in drive-letter form, not UNC.
+
+        Currently FAILS because Suite.load() calls os.path.realpath(path)
+        at line 531, which on py3.8+ Windows silently expands a mapped
+        drive-letter path to its UNC equivalent.
+
+        A UNC-form load_path causes Suite.save() to raise SuiteError when
+        the caller saves back to the same drive-letter path, because
+        ``os.path.realpath(drive_letter_save_path)`` returns the same UNC
+        string while ``self.load_path`` is also UNC - but if load_path was
+        set directly in drive-letter form (e.g. from user config or a
+        partially-fixed code path), the string comparison fails.
+
+        Fix: replace os.path.realpath with real_path() in Suite.load().
+        """
+        from rez.suite import Suite
+
+        drive_path = "N:\\suites\\mysuite"
+
+        with unittest.mock.patch("os.path.realpath", side_effect=_unc_expanding_realpath), \
+             unittest.mock.patch("os.path.exists", return_value=True), \
+             unittest.mock.patch("os.path.isfile", return_value=True), \
+             unittest.mock.patch("builtins.open", unittest.mock.mock_open(read_data="{}")), \
+             unittest.mock.patch.object(Suite, "from_dict", return_value=Suite()):
+            s = Suite.load(drive_path)
+
+        self.assertFalse(
+            s.load_path.startswith("\\\\"),
+            "Suite.load_path was UNC-expanded: %r. "
+            "Replace os.path.realpath with real_path() in Suite.load()." % s.load_path,
+        )
+        self.assertTrue(
+            s.load_path.lower().startswith("n:\\"),
+            "Expected drive-letter form load_path, got: %r" % s.load_path,
+        )
--- a/src/rez/tests/test_utils_filesystem.py
+++ b/src/rez/tests/test_utils_filesystem.py
@@ -365,3 +365,22 @@ class TestWindowsRealpathInternals(TestBase):
                 result = _windows_realpath("C:\\loop")
 
         self.assertIsInstance(result, str)
+
+
+# Map two different drive letters to two different UNC roots.
+# This simulates the py3.8+ realpath expansion that triggers the ValueError
+# in os.path.relpath when the two paths end up on different UNC servers.
+_MOCK_MULTI_DRIVE_TO_UNC = {
+    "n": "\\\\nas\\studio",
+    "m": "\\\\backup\\bundles",
+}
+
+
+def _multi_drive_unc_realpath(path):
+    """Expand two different drive letters to two different UNC roots."""
+    norm = path.replace("/", "\\")
+    if len(norm) >= 2 and norm[1] == ":":
+        drive = norm[0].lower()
+        if drive in _MOCK_MULTI_DRIVE_TO_UNC:
+            return _MOCK_MULTI_DRIVE_TO_UNC[drive] + norm[2:]
+    return path
--- a/src/rez/tests/test_utils_filesystem.py
+++ b/src/rez/tests/test_utils_filesystem.py
@@ -14,7 +14,7 @@ import unittest.mock
 from rez.tests.util import TestBase
 from rez.tests.util import TempdirMixin
 from rez.utils import filesystem
-from rez.utils.filesystem import canonical_path
+from rez.utils.filesystem import canonical_path, _windows_realpath
 from rez.utils.platform_ import platform_
 
 
@@ -198,6 +198,24 @@ class TestCanonicalPathWindowsFormPreservation(TestBase):
         )
 
 
+def _windows_longpaths_enabled() -> bool:
+    """Return True if the Windows LongPathsEnabled registry key is set.
+
+    When False, the Win32 API cannot open paths longer than MAX_PATH (260
+    chars) without an explicit ``\\?\\`` extended-length prefix.
+    """
+    try:
+        import winreg
+        with winreg.OpenKey(
+            winreg.HKEY_LOCAL_MACHINE,
+            r"SYSTEM\CurrentControlSet\Control\FileSystem",
+        ) as key:
+            value, _ = winreg.QueryValueEx(key, "LongPathsEnabled")
+            return bool(value)
+    except OSError:
+        return False
+
+
 @unittest.skipIf(
     platform_.name != "windows",
     "Windows symlink/junction resolution tests are Windows-only.",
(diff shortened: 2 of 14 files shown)


JeanChristopheMorinPerso:
@maxnbk What's the relationship to https://github.com/AcademySoftwareFoundation/rez/pull/1856?

During the TSC, @instinct-vfx said that the expected default behavior is that the style should stay consistent based on how the repo was configured?

cfxegbert:
Can you please look at https://github.com/AcademySoftwareFoundation/rez/pull/1856.   There are a few other places that realpath needs to be replaced.

maxnbk:
I'll take a look and see what else needs resolving. I may have to add some more tests to find all the edge-cases. Goal is to incorporate all solutions from all mismatched PRs, as I think there were approximately 3 all trying to solve the same problem with partial/incomplete solutions, and nobody got all corners.

Evidently that extends to me, so I'll re-evaluate and see what I need to do to extend.
Might be this weekend though. :)

cfxegbert:
@JeanChristopheMorinPerso Looks like this didn't get merged into the 3.4.0.  What can we do to get this out there?

JeanChristopheMorinPerso:
Hey @cfxegbert, this was intentional. We had a few more questions to answer and decided to push this to a follow up release. We are planning to do a 3.4.1 release soon-ish to get this PR out.

When I reviewed the PR, I noticed that `os.path.islink` doesn't do what we thought it would on windows. At least I think.

herronelou:
adding comment here too for visibility, but we've been running rez 3.3 with a patch of #1856 in production since January and that has solved the issues we'd been having since changing python versions.

src/rez/tests/test_utils_filesystem.py
```diff
@@ -125,3 +128,1085 @@ def test_safe_rmtree_with_other_error_and_apple_double(self) -> None:
...
+        This is a regression guard: if the resolved_context.py call site is
+        fixed by substituting canonical_path() (which lowercases on Windows),
+        the stored relative path will have wrong capitalisation on Linux NFS
+        mounts that are case-sensitive.
+
+        real_path() preserves the original casing, so it is the correct fix.
+        """
+        from rez.utils.platform_ import Platform
+
+        class _CaseInsensitivePlatform(Platform):
```
JeanChristopheMorinPerso, line 807: Relatively minor: We use mocks in other places. I think I prefer the subclassing approach though.

src/rez/tests/test_utils_filesystem.py
```diff
@@ -125,3 +128,1085 @@ def test_safe_rmtree_with_other_error_and_apple_double(self) -> None:
...
+        path components stored in bundle files.
+
+        This is a regression guard: if the resolved_context.py call site is
+        fixed by substituting canonical_path() (which lowercases on Windows),
+        the stored relative path will have wrong capitalisation on Linux NFS
+        mounts that are case-sensitive.
+
+        real_path() preserves the original casing, so it is the correct fix.
+        """
+        from rez.utils.platform_ import Platform
```
JeanChristopheMorinPerso, line 805: I think we can move this import at the top level

src/rez/tests/test_utils_filesystem.py
```diff
@@ -125,3 +128,1085 @@ def test_safe_rmtree_with_other_error_and_apple_double(self) -> None:
...
+
+    def test_bundle_relpath_does_not_raise_on_different_unc_roots(self):
+        """_adjust_variant_for_bundling must not raise ValueError when
+        os.path.realpath would have expanded drive-letter paths to UNC form.
+
+        Exercises the real call site in resolved_context.py. is_subdirectory
+        is mocked True to force the relpath branch. With real_path() (abspath)
+        the drive-letter form is preserved and os.path.relpath succeeds.
+        """
+        from rez.resolved_context import ResolvedContext
```
JeanChristopheMorinPerso, line 762: Move at the top

src/rez/tests/test_utils_filesystem.py
```diff
@@ -125,3 +128,1085 @@ def test_safe_rmtree_with_other_error_and_apple_double(self) -> None:
...
+
+        The real failure occurs when the suite was loaded from a drive-letter
+        path but is saved to what looks like the same path: the UNC expansion
+        changes the string, so subsequent code that compares the stored
+        load_path against the UNC-expanded save path will see a mismatch.
+
+        This test documents the fragility. Fix: replace os.path.realpath
+        with real_path() in Suite.save() and Suite.load().
+        """
+        from rez.suite import Suite, SuiteError
```
JeanChristopheMorinPerso, line 861: Move at the top

src/rez/tests/test_utils_filesystem.py
```diff
@@ -125,3 +128,1085 @@ def test_safe_rmtree_with_other_error_and_apple_double(self) -> None:
...
+        drive_path = "N:\\suites\\mysuite"
+        suite = Suite()
+        suite.load_path = drive_path
+
+        patch_context_names = unittest.mock.patch.object(
+            Suite, "context_names",
+            new_callable=unittest.mock.PropertyMock,
+            return_value=[],
+        )
+        with unittest.mock.patch("os.path.realpath", side_effect=_unc_expanding_realpath), \
```
JeanChristopheMorinPerso, line 872: This is a lock of mocking... It seems fragile. Is there a way to get this tested using less mocks?

src/rez/tests/test_utils_filesystem.py
```diff
@@ -125,3 +128,1085 @@ def test_safe_rmtree_with_other_error_and_apple_double(self) -> None:
...
+        A UNC-form load_path causes Suite.save() to raise SuiteError when
+        the caller saves back to the same drive-letter path, because
+        ``os.path.realpath(drive_letter_save_path)`` returns the same UNC
+        string while ``self.load_path`` is also UNC - but if load_path was
+        set directly in drive-letter form (e.g. from user config or a
+        partially-fixed code path), the string comparison fails.
+
+        Fix: replace os.path.realpath with real_path() in Suite.load().
+        """
+        from rez.suite import Suite
```
JeanChristopheMorinPerso, line 907: Move at the top

src/rez/tests/test_utils_filesystem.py
```diff
@@ -125,3 +128,1085 @@ def test_safe_rmtree_with_other_error_and_apple_double(self) -> None:
...
+
+        drive_path = "N:\\suites\\mysuite"
+
+        with unittest.mock.patch("os.path.realpath", side_effect=_unc_expanding_realpath), \
```
JeanChristopheMorinPerso, line 911: Same comment regarding too much mocks here.

src/rez/tests/test_utils_filesystem.py
```diff
@@ -125,3 +128,1085 @@ def test_safe_rmtree_with_other_error_and_apple_double(self) -> None:
...
+             unittest.mock.patch("os.path.samefile", return_value=False):
+            with self.assertRaises(SuiteError):
+                suite.save(other)
+
+
+@unittest.skipIf(
+    sys.platform == "win32",
+    "Test creates real symlinks; Windows requires admin or developer mode",
+)
+class TestGetResourceFromHandleSamefileFallback(TestBase, TempdirMixin):
```
JeanChristopheMorinPerso, line 1042: Should this go in a module dedicated to testing the filesystem repo?

src/rez/tests/test_utils_filesystem.py
```diff
@@ -125,3 +128,1085 @@ def test_safe_rmtree_with_other_error_and_apple_double(self) -> None:
...
+        if drive in _MOCK_MULTI_DRIVE_TO_UNC:
+            return _MOCK_MULTI_DRIVE_TO_UNC[drive] + norm[2:]
+    return path
+
+
+@unittest.skipIf(
+    platform_.name != "windows",
+    "Windows-specific real_path() call-site regression tests",
+)
+class TestRealPathCallSitesWindowsUNC(TestBase):
```
JeanChristopheMorinPerso, line 746: Should this go in a dedicated `test_resolved_context` module?

src/rez/utils/filesystem.py
```diff
@@ -513,15 +519,190 @@ def to_posixpath(path: str):
     return posixpath.sep.join(path.split(ntpath.sep))
 
 
+_WINDOWS_MAX_PATH = 259  # Win32 MAX_PATH minus the null terminator
+
+# Reparse tag constant for Windows junctions (mount points).  Defined in
+# ntifs.h as IO_REPARSE_TAG_MOUNT_POINT.  We hardcode it because it is not
+# exposed in Python's stat module on all versions.
+_IO_REPARSE_TAG_MOUNT_POINT = 0xA0000003
```
JeanChristopheMorinPerso, line 527: https://docs.python.org/3/library/stat.html#stat.IO_REPARSE_TAG_MOUNT_POINT says that `stat.IO_REPARSE_TAG_MOUNT_POINT` is available since 3.8 and I do see it in the latest 3.8 tag on GH: https://github.com/python/cpython/blob/39b2f82717a69dde7212bc39b673b0f55c99e6a3/Modules/_stat.c#L599-L600

src/rez/utils/filesystem.py
```diff
@@ -513,15 +519,190 @@ def to_posixpath(path: str):
...
+    ``os.path.islink`` detects symlinks on all Python versions but does NOT
+    detect junctions on any version (it checks for
+    ``IO_REPARSE_TAG_SYMLINK`` only).  ``os.path.isjunction`` (3.12+) covers
+    junctions; for 3.8-3.11 we fall back to checking ``st_reparse_tag`` on
+    ``os.lstat``.
+    """
+    if os.path.islink(path):
+        return True
+    # os.path.isjunction was added in Python 3.12.
+    isjunction = getattr(os.path, "isjunction", None)
```
JeanChristopheMorinPerso, line 568: I'm starting to wonder if we should maybe gate this with `if sys.version_info[0:2]`. This would allow us ot easily spot that we can remove it once we drop support for python 3.11 (which is far from today, but still).

src/rez/utils/filesystem.py
```diff
@@ -513,15 +519,190 @@ def to_posixpath(path: str):
...
+    if isjunction is not None:
+        return isjunction(path)
+    # Python 3.8-3.11: detect junctions via the reparse tag on lstat.
+    # st_reparse_tag is non-zero when the path is a reparse point; we
+    # check for the mount-point tag specifically.
+    try:
+        st = os.lstat(path)
+    except OSError:
+        return False
+    tag = getattr(st, "st_reparse_tag", 0)
```
JeanChristopheMorinPerso, line 578: If I'm not mistaken, `st_reparse_tag` is available since 3.8: https://docs.python.org/3/library/os.html#os.stat_result.st_reparse_tag. Is there a reason to buard the access using `getattr`?

src/rez/utils/filesystem.py
```diff
@@ -513,15 +519,190 @@ def to_posixpath(path: str):
...
+            root = t_root
+            result = t_root
+            components = new_parts + components
+            continue
+
+        result = candidate
+    return result
+
+
 def canonical_path(path: str, platform=None):
```
JeanChristopheMorinPerso, line 685: We could take the occasion and add a type hint to `platform` (and remove the types from the docstring).

src/rez/utils/filesystem.py
```diff
@@ -513,15 +519,190 @@ def to_posixpath(path: str):
...
+    external tool, or store in a bundle/context that may be read on
+    another platform - the lowercasing it applies on case-insensitive
+    filesystems mutates the string and can break case-sensitive consumers
+    (e.g. a Linux NFS mount accessed from a Windows client).  Use
+    :func:`real_path` for those cases instead.
 
     Args:
         path (str): Filepath being formatted
-        platform (rez.utils.platform\_.Platform): Indicates platform path is being
+        platform (rez.utils.platform_.Platform): Indicates platform path is being
```
JeanChristopheMorinPerso, line 705:
```suggestion
        platform: Indicates platform path is being
```

src/rez/utils/filesystem.py
```diff
@@ -513,15 +519,190 @@ def to_posixpath(path: str):
...
+
+    Do not use this when you need a path to open a file, pass to an
+    external tool, or store in a bundle/context that may be read on
+    another platform - the lowercasing it applies on case-insensitive
+    filesystems mutates the string and can break case-sensitive consumers
+    (e.g. a Linux NFS mount accessed from a Windows client).  Use
+    :func:`real_path` for those cases instead.
 
     Args:
         path (str): Filepath being formatted
```
JeanChristopheMorinPerso, line 704:
```suggestion
        path: Filepath being formatted
```
Sphinx will be able to add the type hint in the docs without us having to add in in the docstring (as long as the function has proper type hints).

src/rez/utils/filesystem.py
```diff
@@ -351,9 +352,14 @@ def make_tmp_name(name):
 
 
 def is_subdirectory(path_a, path_b) -> bool:
-    """Returns True if `path_a` is a subdirectory of `path_b`."""
-    path_a = os.path.realpath(path_a)
-    path_b = os.path.realpath(path_b)
+    """Returns True if `path_a` is a subdirectory of `path_b`.
```
JeanChristopheMorinPerso, line 355:
```suggestion
    """Returns True if ``path_a`` is a subdirectory of ``path_b``.
```

src/rez/utils/filesystem.py
```diff
@@ -351,9 +352,14 @@ def make_tmp_name(name):
...
+
+    Both paths are canonicalised (symlinks resolved on non-Windows, and on
+    Windows when resolve_links_on_windows is enabled) so that containment
```
JeanChristopheMorinPerso, line 358:
```suggestion
    Windows when :data:`resolve_links_on_windows` is enabled) so that containment
```

src/rez/utils/filesystem.py
```diff
@@ -530,14 +711,83 @@ def canonical_path(path: str, platform=None):
...
+def resolve_path(path: str) -> str:
+    r"""Resolve symlinks in a form-preserving way.
+
+    Like :func:`real_path`, but also resolves symlinks on Windows without
+    expanding drive letters to UNC paths. Use this when you need the
+    physical target of a path (e.g. before ``rmtree`` to avoid severing a
+    symlink) but must not lose the drive-letter form.
+
+    On non-Windows this is identical to ``os.path.realpath``.
+    On Windows with ``resolve_links_on_windows`` enabled, uses
```
JeanChristopheMorinPerso, line 779:
```suggestion
    On Windows with :data:`resolve_links_on_windows` enabled, uses
```

src/rez/utils/filesystem.py
```diff
@@ -530,14 +711,83 @@ def canonical_path(path: str, platform=None):
...
+    :func:`_windows_realpath`. Otherwise, uses ``os.path.abspath``
```
JeanChristopheMorinPerso, line 780:
```suggestion
    ``_windows_realpath``. Otherwise, uses ``os.path.abspath``
```
This will not work as the function is not exported and sphinx might even complain about it.

WHich means that we might need a better way to phrase this, because I'm not sure we should be referencing private functions in public function docstrings.

src/rez/rezconfig.py
```diff
@@ -125,6 +125,26 @@
 # For further information, see :ref:`package-definition-sharing-code`.
 package_definition_python_path = None
 
+# On Windows, whether to resolve symbolic links and junction points when
+# normalising filesystem paths (primarily inside ``canonical_path``).
```
JeanChristopheMorinPerso, line 129:
```suggestion
# normalising filesystem paths (primarily inside :func:`canonical_path`).
```
(or something like that. I don't remember off the top of my head if you need to use the full absolute function name (`rez.utils.filesystem.canonical_path`) or if sphinx will figure it out by itself.

src/rez/rezconfig.py
```diff
@@ -125,6 +125,26 @@
...
+#
+# When ``False`` (default), rez uses ``os.path.abspath``, which normalises
```
JeanChristopheMorinPerso, line 131: Do we need to mention `os.path.abspath` here since you already explain what it does? (but I would keep the last reference to it at the end of the paragraph as that will help the users know when to use it or not.

src/rez/rezconfig.py
```diff
@@ -125,6 +125,26 @@
...
+# ``os.path.realpath`` on Windows.
+#
+# When ``True``, rez performs a component-by-component walk using
+# ``os.path.islink`` / ``os.readlink``. This resolves actual symlinks and
+# junction points without the drive-letter-to-UNC side-effect that
+# ``os.path.realpath`` introduced in Python 3.8. Useful when package
+# repositories are accessed through directory symlinks or junctions.
+#
+# This setting is a no-op on non-Windows platforms, which always resolve
+# symlinks via ``os.path.realpath``.
```
JeanChristopheMorinPerso, line 145:
```suggestion
# symlinks via ``os.path.realpath``.
#
# .. versionadded:: 3.5.0
```
(or whatever the next version will be)

(I'll push a PR to add tests that will catch this problem moving forward).

JeanChristopheMorinPerso (commented):
Generally, I think it looks good. I've not sent my agent again as it seems already in good shape. The only real concern I have is performance, particularly in more complex cases where we will do a lot of "add unc, remove unc, add unc again", etc, but I'm willing to get this merged and see how things go. We can always improve the performance if it becomes a bottleneck. I also see that you've added some caching already, which would hopefully help a bit.

The only reason why I didn't approve yet are:
* `.. versionadded:: 3.5.0` missing in new setting doc
* getattr calls that could maybe be replaced with `if sys.version_info...`
* `IO_REPARSE_TAG_MOUNT_POINT` constant.

JeanChristopheMorinPerso:
Oh, we shouold also think about semgrep rules to protect us from reintroducing problems later. Something like this?

```yaml
rules:
  - id: no-os-path-realpath
    languages:
      - python
    severity: HIGH
    message: |-
      Do not use `os.path.realpath`. Use `rez.utils.filesystem.real_path`
      instead. On Windows, `os.path.realpath` (since Python 3.8) resolves mapped
      drive letters to their UNC equivalents and follows junctions/symlinks,
      which can break path comparisons and break callers that expect
      drive-letter paths. `rez.utils.filesystem.real_path` honours the
      `resolve_links_on_windows` config setting and preserves the path style
      supplied by the caller. See
      https://github.com/AcademySoftwareFoundation/rez/pull/2104.

      If a specific call is intentional (e.g. inside the `real_path`
      implementation itself), suppress it inline with a trailing `# nosemgrep:
      no-os-path-realpath` comment and a short justification.
    pattern: os.path.realpath($...ARGS)
    fix: rez.utils.filesystem.real_path($...ARGS)
    paths:
      exclude:
        - "/src/rez/vendor/**"
        - /install.py
        - /release-rez.py
        - /setup.py
    metadata:
      category: portability
```

PR #2107, opened 2026-05-14
fix: windows case sensitivity in pkg cache variants uri

### LINKED ISSUE #2101: rez-pkg-cache fails on Windows for packages with upper-case characters
On Windows, the `rez.packages.get_package_from_uri()` function returns None for any packages that have one or more uppercase characters in either their package or version name (when using the `FilesystemPackageRepository`. This also affects the `get_variant_from_uri()` method from the same module that calls it as well. 

From what I can tell, the culprit is here where the URI path is lower-cased before all the subsequent comparison checks, but the subsequent checks for package and version through `.get_package()` are checking against the non-normalised versions, so they don't match:
https://github.com/AcademySoftwareFoundation/rez/blob/d415b96538f574871f2e1082e0b7df411eb0bade/src/rezplugins/package_repository/filesystem.py#L604

This was introduced a few years in the following PR, which was also to address issues around finding variants on Windows so I'm not sure what reverting this change could break:
https://github.com/AcademySoftwareFoundation/rez/pull/1011

This leads to any variants on these packages not being able to be added to the package cache manually using `rez-pkg-cache`, as it will give the following error:
```shell
$ rez-pkg-cache -a "\\storage\pipeline\rez\packages\prod\ext\arch\AMD64\package.py[0]"
No such variant: \\storage\pipeline\rez\packages\prod\ext\arch\AMD64\package.py[0]
```

This seems to still work for the normal `rez-env` case as you already have the `Variant` objects from the resolve - the issue here is getting from the URI input to the actual `Variant` API object.

**Environment**
* OS: Windows 11 25H2
* Rez version: 3.1.1
* Rez python version: 3.10.10

### API

1. On Windows, find a package that has at least one uppercase character in either the package name of version, e.g (`Pyside6` or `arch-AMD64`) and get the full URI for it:
```powershell
> rez-view arch-AMD64
URI:
\\storage\pipeline\rez\packages\prod\ext\arch\AMD64\package.py
...
```
2. Open up a Rez Python shell and run the following snippet, which will show that the API is not able to identify that package by it's URI:
```python
>>> import rez.packages
>>> uri = r"\\storage\pipeline\rez\packages\prod\ext\arch\AMD64\package.py"
>>> print(rez.packages.get_package_from_uri(uri))
```
If you repeat the test with a package version that has no uppercase characters in either package name or version, a `Variant` object should be returned.

### CLI

1. On Windows, find a variant that has at least one uppercase character in either the package name of version, e.g (`Pyside6` or `arch-AMD64`) and get the full URI for it:
```powershell
> rez-search arch-AMD64 -t variant -f "{uri}"
\\storage\pipeline\rez\packages\prod\ext\arch\AMD64\package.py[]
```
2. Try to add that URI to the package cache:
```powershell
> rez-pkg-cache -a "\\storage\pipeline\rez\packages\prod\ext\arch\AMD64\package.py[]"
16:49:43 INFO     Adding variant '\\\\storage\\pipeline\\rez\\packages\\prod\\ext\\arch\\AMD64\\package.py' to package cache at C:\...\rez\cache\:
No such variant: \\storage\pipeline\rez\packages\prod\ext\arch\AMD64\package.py
```

**Expected behavior**

The `rez-pkg-cache` command would finish successfully and the `rez.packages.get_variant_from_uri()` method would return a `Variant` object.

**Actual behavior**
Adding the package with an uppercase character to the package cache fails and does not return via the `rez.packages.get_variant_from_uri()` method.

maxnbk:
Closes #2101 .

Explanation:

`get_package_from_uri` in the filesystem package repository applied `os.path.normcase` to the entire URI before slicing out the package name and version. On Windows, `normcase` lowercases the whole string, silently corrupting mixed-case names (`FooBar` -> `foobar`) and versions (`1.0B` -> `1.0b`) before they were used for lookup, causing both to return `None`.

Fix:

We scope the `normcase` usage to the `is_subdirectory` boundary check only. The subsequent name/version extraction slices from the original URI, preserving caller-supplied casing.

Disclosure:

This PR was AI-assisted with Claude Code, Sonnet 4.6, primarily for diagnostics, logic-tracing, documentation and edge-case verification.

### DIFF
--- a/src/rez/tests/test_package_repository.py
+++ b/src/rez/tests/test_package_repository.py
@@ -11,6 +11,7 @@ from rezplugins.package_repository import filesystem
 from rez.packages import create_package
 from rez.tests.util import TestBase, TempdirMixin
 from rez.utils.platform_ import platform_
+from rez.version import Version
 
 
 class TestFilesystemPackageRepository(TestBase, TempdirMixin):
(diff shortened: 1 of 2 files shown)


JeanChristopheMorinPerso (approved):
LGTM!

PR #2109, opened 2026-05-15
Fix PowerShell command failure status

### LINKED ISSUE #2073: bug on catching exitcode for powershell commands
Hey there,
I was looking at the rez code to have a reference for interacting with power-shell from Python.

At some point I copied this snippet of code which allow to catch the lastexit code:

https://github.com/AcademySoftwareFoundation/rez/blob/78e26236cdd65bc549db6b4e5782c70fe86ed5db/src/rezplugins/shell/_utils/powershell_base.py#L163-L170

However on my side it was not working as expected. For example, if you call a command that don't exist before it, the code behave as it was successfull.

From what I found it seems that it's become in that case we should rely on `$?`, however at the time of inspection, it has already been overwritten by the `Test-Path` command we call just before. So it's always `True`.

I manage to fix it by storing it's value in a variable before calling `Test-Path`

```powershell
$_op_success = $?
if ((Test-Path variable:LASTEXITCODE) -and $LASTEXITCODE) {
  exit $LASTEXITCODE
}
if (! $_op_success) {
  exit 1
}
```

Keep in mind that none of this was tested with rez so I might be mistaking.
Liam.

snoopuppy582:
## Summary
- preserve `$?` before checking `LASTEXITCODE` in generated PowerShell startup scripts
- return exit code 1 when a PowerShell command fails before setting `LASTEXITCODE`
- add a Windows regression test for missing-command failures in both `powershell` and `pwsh`

## Test plan
- `pytest src/rez/tests/test_shells.py::TestShells::test_powershell_command_not_found_returncode -q`
- `pytest src/rez/tests/test_shells.py::TestShells::test_pwsh_lastexitcode_gui -q`
- `flake8 src/rezplugins/shell/_utils/powershell_base.py src/rez/tests/test_shells.py`
- manual `execute_shell` check: missing command returns 1 for `powershell` and `pwsh`; `hello_world -q -r 66` still returns 66 for both

Closes #2073

### DIFF
--- a/src/rez/tests/test_shells.py
+++ b/src/rez/tests/test_shells.py
@@ -286,6 +286,24 @@ class TestShells(TestBase, TempdirMixin):
                 # should bubble up through the rez shell as a 0 exit status.
                 self.assertEqual(p.returncode, 0)
 
+    @unittest.skipIf(platform_.name != "windows", "PowerShell return-code handling is Windows-specific")
+    def test_powershell_command_not_found_returncode(self) -> None:
+        shells = [shell for shell in ("powershell", "pwsh") if shell in get_available_shells()]
+        if not shells:
+            self.skipTest("PowerShell unavailable or disabled")
+
+        r = self._create_context([])
+        for shell in shells:
+            with self.subTest(shell=shell):
+                with r.execute_shell(
+                    shell=shell,
+                    command="NoSuchRezCommand2073",
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE,
+                ) as p:
+                    p.wait()
+                self.assertEqual(p.returncode, 1)
+
     @per_available_shell()
     def test_norc(self, shell) -> None:
         sh = create_shell(shell)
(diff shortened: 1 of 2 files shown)


maxnbk:
@snoopuppy582 was this issue tackled as part of devdays, or independently? Thank you.

snoopuppy582:
This was independent. I saw #2073, reproduced the PowerShell status issue locally on Windows, and opened the PR with the focused regression test for that case. Thanks for checking.

JeanChristopheMorinPerso (approved):
Thanks for fixing this bug and adding a test!

src/rez/tests/test_shells.py
```diff
@@ -286,6 +286,24 @@ def actions_callback(ex):
                 # should bubble up through the rez shell as a 0 exit status.
                 self.assertEqual(p.returncode, 0)
 
+    @unittest.skipIf(platform_.name != "windows", "PowerShell return-code handling is Windows-specific")
+    def test_powershell_command_not_found_returncode(self) -> None:
+        shells = [shell for shell in ("powershell", "pwsh") if shell in get_available_shells()]
+        if not shells:
+            self.skipTest("PowerShell unavailable or disabled")
```
JeanChristopheMorinPerso, line 293:
```suggestion
    @unittest.skipIf("pwsh" not in get_available_shells(), "PowerShell unavailable or disabled")
    @unittest.skipIf("powershell" not in get_available_shells(), "PowerShell unavailable or disabled")
    def test_powershell_command_not_found_returncode(self) -> None:
```

JeanChristopheMorinPerso: Actually, this is not better. Sorry. Ignore this comment. I would like to eventually revisit the `per_available_shell` decorator to better handle this case But let's push this to later.

PR #2111, opened 2026-05-18
Fix for zsh-specific env loading behavior

### LINKED ISSUE #2058: zsh unable to load rez context properl
When executing `rez-env my_package` in zsh rez is unable to load context of package even though some sort of child shell is created. The same thing works however when using bash. More in thread on slack: https://academysoftwarefdn.slack.com/archives/C0321B828FM/p1767119865724849

**Environment**
- MacOS
- "3.3.0"
- 3.11

**To Reproduce**
1. `cd ~`
2. `touch .zshenv`
3. `rez-env`

**Expected behavior**
Rez load created context

**Actual behavior**
New shell created but nothing happens.

derrickauyoung:
## Original Issue

Fixed #2058 

## Root Cause

zsh runs .zshenv before .zshrc, so the export HOME=/real/home inside the generated .zshenv triggered before rez's .zshrc could be sourced, bypassing the context

## Fix

`ZDOTDIR` is the correct zsh-native mechanism for redirecting dot file loading, and doesn't require touching HOME at all

- Updated `shells.py` UnixShell class to implement default HOME environment var setting for the HOME env var hijack
- Updated `rezplugins/shells/zsh.py` class to re-implement default methods to utilize ZDOTDIR instead

Thanks to @JeanChristopheMorinPerso and @legomir for their initial diagnosis and background info on this problem.

### DIFF
--- a/src/rez/shells.py
+++ b/src/rez/shells.py
@@ -497,15 +497,16 @@ class UnixShell(Shell):
                             files_ = [file_]
 
                         ex = _create_ex()
-                        ex.setenv('HOME', os.environ.get('HOME', ''))
+                        self._write_startup_env(ex)
                         _record_shell(ex, files=files_, bind_rez=bind_rez,
                                       print_msg=bind_rez)
                         _write_shell(ex, os.path.basename(file_))
 
                     if config.debug("shell_startup"):
-                        print_debug("Setting $HOME for new shell to %s" % tmpdir)
+                        print_debug("Setting $%s for new shell to %s"
+                                    % (self._startup_env_var(), tmpdir))
 
-                    executor.setenv("HOME", tmpdir)
+                    executor.setenv(self._startup_env_var(), tmpdir)
 
                     # keep history
                     if self.histfile and self.histvar:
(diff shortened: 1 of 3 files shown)


src/rez/shells.py
```diff
@@ -556,6 +557,23 @@ def _create_ex() -> RexExecutor:
...
         return p
 
+    def _startup_env_var(self) -> str:
+        """Environment variable used to redirect shell startup file loading to tmpdir.
+        Shells that support dedicated dotfile directory var (e.g., zsh's ZDOTDIR)
+        should override this and _write_startup_env methods.
+        """
+        return "HOME"
+
+    def _write_startup_env(self, ex) -> None:
```
JeanChristopheMorinPerso:
```suggestion
    def _write_startup_env(self, ex: RexExecutor) -> None:
```

src/rez/shells.py
```diff
@@ -556,6 +557,23 @@ def _create_ex() -> RexExecutor:
...
+        """Write env setup into each generated startup file.
+        Default sets HOME to real users's home directory so the user's real
+        dot files can be sourced from within the generated files.
+
+        Args:
+            ex (RexExecutor): shell executor
```
JeanChristopheMorinPerso:
```suggestion
            ex: shell executor
```

src/rez/shells.py
```diff
@@ -556,6 +557,23 @@ def _create_ex() -> RexExecutor:
...
+
+    def _write_startup_env(self, ex: RexExecutor) -> None:
+        """Write env setup into each generated startup file.
+        Default sets HOME to real users's home directory so the user's real
+        dot files can be sourced from within the generated files.
+
+        Args:
+            ex: shell executor
+        """
+        ex.setenv("HOME", os.environ.get("HOME", ""))
```
JeanChristopheMorinPerso:
```suggestion
        startup_env_var = self._startup_env_var()
        ex.setenv(startup_env_var, os.environ.get(startup_env_var, ""))
```

JeanChristopheMorinPerso (approved):
I think this looks good. I added a test and it seems to be working. Thanks!

PR #2117, opened 2026-05-21
feat: add ephemerals to the list of objects available in late functions

### LINKED ISSUE #1739: Ephemerals should be available in late binding functions
Late binding functions have access to `implicits` in the function but not to `ephemerals`.

**Motivation**
To get to an ephemeral in a late binding function I have to use `context.resolved_ephemerals` which is a list.  I have to search through the list manually to get the ephemeral.  Having the EphemeralBinding object will allow me to easily call `get_range` instead of searching the list.

cfxegbert:
Late binding functions have direct access to `implicits`.  To make late binding functions more consistent add `ephemerals`.

Closes: #1739

### DIFF
--- a/docs/source/package_definition.rst
+++ b/docs/source/package_definition.rst
@@ -293,7 +293,7 @@ is ``True``:
 * **testing**: see :rex:attr:`testing`;
 * **request**: see :rex:attr:`request`;
 * **implicits**: see :rex:attr:`implicits`;
-* **ephemerals**: see :attr:`ephemerals`.
+* **ephemerals**: see :rex:attr:`ephemerals`.
 
 The following objects are available in **all** cases:
 
--- a/docs/source/package_definition.rst
+++ b/docs/source/package_definition.rst
@@ -292,7 +292,8 @@ is ``True``:
 * **building**: see :rex:attr:`building`;
 * **testing**: see :rex:attr:`testing`;
 * **request**: see :rex:attr:`request`;
-* **implicits**: see :rex:attr:`implicits`.
+* **implicits**: see :rex:attr:`implicits`;
+* **ephemerals**: see :attr:`ephemerals`.
 
 The following objects are available in **all** cases:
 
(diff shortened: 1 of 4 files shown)


docs/source/package_definition.rst
```diff
@@ -292,7 +292,8 @@ is ``True``:
 * **building**: see :rex:attr:`building`;
 * **testing**: see :rex:attr:`testing`;
 * **request**: see :rex:attr:`request`;
-* **implicits**: see :rex:attr:`implicits`.
+* **implicits**: see :rex:attr:`implicits`;
+* **ephemerals**: see :attr:`ephemerals`.
```
JeanChristopheMorinPerso:
```suggestion
* **ephemerals**: see :rex:attr:`ephemerals`.
```

src/rez/packages.py
```diff
@@ -185,6 +186,7 @@ def _eval_late_binding(self, sourcecode: SourceCode[T]) -> T:
         else:
             g["in_context"] = lambda: True
             g["context"] = self.context
+            g["ephemerals"] = EphemeralsBinding(self.context.resolved_ephemerals or [])
```
JeanChristopheMorinPerso: Should we move this into `self.context._get_pre_resolve_bindings()`?

cfxegbert: Moved it to `_get_pre_resolve_bindings`

src/rez/resolved_context.py
```diff
@@ -2049,6 +2049,7 @@ def _get_pre_resolve_bindings(self) -> dict:
                 "testing": self.testing,
                 "request": RequirementsBinding(self._package_requests),
                 "implicits": RequirementsBinding(self.implicit_packages),
+                "ephemerals": EphemeralsBinding(self.resolved_ephemerals or []),
```
cfxegbert, line 2052: Everywhere we use to property `resolved_ephemerals` we always or it with an empty list to make sure it is not None.  Should I update resolved_ephemerals to return an empty list or is it part of the public interface we should not touch?

JeanChristopheMorinPerso: Let's not change this. `self.resolved_packages` behaves the same and changing `self.resolved_ephemerals` would create an inconsistency.

JeanChristopheMorinPerso (approved):

PR #1645, opened 2024-02-14
Add support for virtualenv in installer

Pixel-Minions:
Fixes #1641

Add support for virtualenv in installer

PR #2119, opened 2026-05-23
fix: cleanup legacy code

cfxegbert:
* `raw_input` was changed to `input` in Python 3
* Unused code in the filesystem utilities was using Python 2 Unicode functions that no longer exists.  Python 3 strings are now Unicode.


install.py
```diff
@@ -12,13 +12,7 @@
 import sys
 import shutil
 import subprocess
-
-USE_VIRTUALENV = False
```
JeanChristopheMorinPerso, line 16: This is not a leftover from Python 2. See https://github.com/AcademySoftwareFoundation/rez/pull/1645. Support for virtualenv is a feature we want to keep.

src/rez/utils/filesystem.py
```diff
@@ -514,147 +513,6 @@ def canonical_path(path: str, platform=None):
     return path
 
 
-def encode_filesystem_name(input_str: str):
```
JeanChristopheMorinPerso, line 542: These are not necessarily python 2 related from what I see, but they are unused, so it's good to remove them. From what I see, they stopped being used 11 years ago. See https://sourcegraph.com/search?q=context%3Aglobal+repo%3Agithub.com%2Facademysoftwarefoundation%2Frez+test_encode_decode+or+encode_filesystem_name+or+decode_filesystem_name&patternType=keyword&sm=0&__cc=1&df=%5B%22type%22%2C%22Diffs%22%2C%22type%3Adiff%22%5D.

JeanChristopheMorinPerso (changes requested):

install.py
cfxegbert: venv became part of the standard library in Python 3.3.  The rez documentation says the earliest we support is Python 3.8.  I removed the commit but do we want to keep it long term if we only support 3.8?

cfxegbert: Reading https://virtualenv.pypa.io/en/latest/explanation.html.  We should probably keep it but I don't think the test for venv would ever fail.  Maybe a command line option?

JeanChristopheMorinPerso: We have users don't depend on some recent python interpreters that don't have it. So we have to keep support for virtualenv.

JeanChristopheMorinPerso (approved):
Thanks!

PR #2134, opened 2026-06-15
Add the option to compile key modules to C using mypyc

chadrik:
Here is an interesting payoff from my recent type annotations PR that was merged:  we can now use `mypyc` to compile key libraries into C for added performance. 

Running the benchmarks locally I see a speedup of about ~2.6x. 

Notes:
- Successfully building rez as a C-library requires a [custom build of mypyc](https://github.com/chadrik/mypy/tree/rez-mypyc-fixes) that fixes several issues encountered during this project.
- Most of the changes in this PR are safe typing changes that could be split off into a separate PR to release first
- This PR changed the lint job to use the latest mypy because it needs to match the version used for building with mypyc, and there are a number of fixes in the newer versions.  The primary ramification of this is that mypy won't flag errors if the code uses language features that are not compatible with python 3.8 or 3.9.  I _think_ these same issues should be flagged by flake8, though.   The other effect is that mypyc can't compile rez for python 3.8 or 3.9. 
- One fundamental change is that the schemas are now used to generate accessor functions in codegen step, rather than using dynamic lookups -- this is a prerequisite for fast performance. 
- I have some additional modifications that eke out some more performance, but they're a bit uglier, so I will save them for a followup

Obviously, if this gets released we'll need a plan for gradual deployment/adoption in order to release something this radical.  Gradual adoption is one of the big advantages of using something like mypyc over rewriting specific modules in rust:  the business logic and API interface remains the same between pure python and C-extension, compilation is just a installation setting.  Here are some thoughts:

- We should scour the diff looking for edits that might introduce breaking changes.  I believe they're minimal, but we should document them all.  
- Any initial release of this should be optional at first.  Advanced users who build their own rez can begin opting into it, while tests continue to enforce that it works.


src/rez/build_process.py
```diff
@@ -404,10 +404,11 @@ def get_changelog(self) -> str | None:
             previous_revision = None
 
         changelog = None
-        with self.repo_operation():
-            changelog = self.vcs.get_changelog(
-                previous_revision,
-                max_revisions=config.max_package_changelog_revisions)
+        if self.vcs:
```
chadrik, line 411: TODO:  revisit why this is necessary or safe.   Start by removing the change and rebuilding and running the self test.

src/rez/package_filter.py
```diff
@@ -7,16 +7,54 @@
...
 
 if TYPE_CHECKING:
     from typing import Self
 
+import functools
+from typing import Callable, Generic, TypeVar
+T = TypeVar("T")
+TypeT = TypeVar("TypeT", bound=type)
+
+class cached_class_property(Generic[T]):
```
chadrik: TODO:  remove this unused experiment

src/rez/package_filter.py
```diff
@@ -325,6 +374,8 @@ def singleton(cls) -> PackageFilterList:
         """
         return cls.from_pod(config.package_filter)
 
+    # singleton = cached_class_property(_singleton)
```
chadrik: TODO: cleanup comment

src/rez/package_filter.py
```diff
@@ -316,7 +365,7 @@ def __str__(self) -> str:
         filters = sorted(self.filters, key=lambda x: (x.cost, str(x)))
         return str(tuple(filters))
 
-    @cached_class_property
+    @classmethod
```
chadrik: This is a breaking change:  accessing the singleton is now a method instead of a property.   I could not create a recipe for `cached_class_property` that worked in mypyc. 

TODO:  Try again to make a `cached_class_property` that works with mypy.  It's been at least a year since I last experimented with this, and there have been a number of fixes to mypyc since then.

src/rez/package_order_list.py
```diff
@@ -0,0 +1,95 @@
...
+    @classmethod
+    def from_pod(cls, data: list[dict[str, Any]]) -> PackageOrderList:
+        flist = PackageOrderList()
+        for dict_ in data:
+            f = from_pod(dict_)
+            flist.append(f)
+        return flist
+
+    @cached_class_property
+    def singleton(cls) -> PackageOrderList:
```
chadrik, line 35: FIXME:  PackageOrderList still implements `singleton` as a property because it's pure python, unlike `PackageFilterList` which switched to a method.   Either get the property working or switch all to method for consistency.

src/rez/packages.py
```diff
@@ -338,8 +383,132 @@ def get_variant(self, index: int | None = None) -> Variant | None:
...
+    def variants(self) -> list[list[PackageRequest]] | None:
+        return self._wrap_forwarded('variants', self.wrapped.variants)
+
+    @property
+    def vcs(self) -> str | None:
+        return self._wrap_forwarded('vcs', self.wrapped.vcs)
+
+    # FIXME: manually corrected
+    @property
+    def version(self) -> Version:
```
chadrik, line 505: FIXME: This auto-generated accessor was manually corrected.  First revisit and understand why this was changed, then address this in `data_utils`.

src/rez/packages.py
```diff
@@ -937,7 +1235,23 @@ def get_completions(prefix: str, paths: list[str] | None = None, family_only: bo
...
+
+@overload
+def get_latest_package(name: str, *, range_: VersionRange | None = None,
+                       paths: list[str] | None = None,
+                       error: Literal[False] | bool = False) -> Package | None:
+    pass
+
+
+def get_latest_package(name: str, *, range_: VersionRange | None = None,
+                       paths: list[str] | None = None,
```
chadrik, line 1254: This is a breaking change: `range_`, `paths`, and `error` are now keyword-only arguments.  You can no longer pass values by position.  This was done because the alternative is a _very_ long list of overloads covering all the permutations of positional and keyword scenarios.

src/rez/resolver.py
```diff
@@ -43,8 +46,8 @@ class ResolverStatus(Enum):
     failed = ("The resolve is not possible.", )
     aborted = ("The resolve was stopped by the user (via callback).", )
 
-    def __init__(self, description) -> None:
-        self.description = description
+    # def __init__(self, description: str) -> None:
```
chadrik: FIXME: look into implementing `description` as a property that looks up the value.

src/rez/solver.py
```diff
@@ -86,9 +87,6 @@ class SolverStatus(Enum):
     cyclic = ("The solve contains a cycle.", )
     unsolved = ("The solve has started, but is not yet solved.", )
 
-    def __init__(self, description):
-        self.description = description
```
chadrik, line 90: FIXME:  try to implement `description` as a property

src/rez/util.py
```diff
@@ -156,7 +156,7 @@ def find_last_sublist(list_, sublist):
     return None
 
 
-@atexit.register
+# @atexit.register
```
chadrik: FIXME: restore this

src/rezplugins/package_repository/filesystem.py
```diff
@@ -538,62 +600,76 @@ def __init__(self, location: str, resource_pool: ResourcePool, disable_memcache:
...
-    @pool_memcached_connections
-    def iter_package_families(self) -> Iterator[PackageFamilyResource]:
+    # @pool_memcached_connections
+    def iter_package_families(self) -> Iterator[FileSystemPackageFamilyResource | FileSystemCombinedPackageFamilyResource]:
         for family in self.get_families():
             yield family
 
-    @pool_memcached_connections
-    def iter_packages(self, package_family_resource: PackageFamilyResource) -> Iterator[Package]:
+    # @pool_memcached_connections
```
chadrik: FIXME:  restore `pool_memcached_connections` and determine why I removed it:  possibly it was unable to be compiled by mypyc so I bypassed it.

src/rez/package_filter.py
chadrik: apparently the existing approach works with newer versions of mypy.  breaking change removed.

Issue #1500, opened 2023-06-19
Unable to use package cache from within a rez-env 


herronelou:
It seems impossible to cache packages locally when resolving the context via API rather than via the command line `rez env`.

Rez does a couple of checks here:
https://github.com/AcademySoftwareFoundation/rez/blob/da139d305b5d276777b471a3dfcff532e60ca528/src/rez/package_cache.py#L381-L384
and here:
https://github.com/AcademySoftwareFoundation/rez/blob/master/src/rez/resolved_context.py#L1836
which are preventing caching to happen unless in a production rez install.

As far as I know, it's impossible to import a production rez install from within rez, as `rez env python rez` followed by launching python and importing rez always imports a non-production install.

Our artists are never using rez-env directly.
They call a script that does `rez-env launcher -- launcher` and opens a GUI, in which they can click the software to launch and we have pre-defined package requests to load for each software, which we resolve by making a `ResolvedContext` object. Since network performance can greatly vary based on load, we'd like to enable local package caching, but it seems impossible without getting rid of our launcher (not happening) or re-implementing part of it.

See conversation with @JeanChristopheMorinPerso on slack for more context: https://academysoftwarefdn.slack.com/archives/C0321B828FM/p1687202355451219

**Environment**
* OS Windows 10

JeanChristopheMorinPerso:
The first thing that comes to mind is to make the rez rez package created with [install.py --as-rez-package](https://github.com/AcademySoftwareFoundation/rez/blob/master/install.py) a prod ready install. Well need to remove the console scripts installed or make them conditional. We don't need the console scripts since they'll always be available from the rez install anyway. There could be an argument made that someone would want to use the CLI from a specific version from the rez rez package, but I don't think we necessarily want to support nested installs, so it shouldn't be supported and shouldn't cause issues.

herronelou:
> There could be an argument made that someone would want to use the CLI from a specific version from the rez rez package,

This is not our case, if we really want to use a different rez install we can always change which one is in our original PATH before getting into rez

PR #2136, opened 2026-06-16
Check production install only if async is needed

doruchan:
Hello, 

This is spawn from https://github.com/AcademySoftwareFoundation/rez/issues/1500.

In the `PackageCache.add_variants` method there's a check for verify that rez is installed as a "production install". 
I reckon this check only needs to happen if we want to cache the packages in async mode, and it should not matter if we are caching them in the same thread.

Thank you!

### DIFF
--- a/src/rez/package_cache.py
+++ b/src/rez/package_cache.py
@@ -506,7 +506,7 @@ class PackageCache(object):
         # a prod install. On non-windows we could fork instead, but there would
         # remain no good solution on windows.
         #
-        if not system.is_production_rez_install:
+        if package_cache_async and not system.is_production_rez_install:
             raise PackageCacheError(
                 "PackageCache.add_variants is only supported in a "
                 "production rez installation."


manuelkoester:
👍 We have a similar fix on our end, but including the same check additionally at https://github.com/AcademySoftwareFoundation/rez/blob/3.4.0/src/rez/resolved_context.py#L1938

maxnbk:
I've reviewed this pair of PRs (this one, and the one @manuelkoester produced in #2151 ), and judged manuel's to be a strict superset that covers more code, which turns out to be a good idea because of the resolved guard in resolved_context.py.

I'm therefore going to close this PR in favor of @manuelkoester 's, and move review over to that thread. Thank you for the submission anyway, @doruchan , there's not per-se anything wrong with it, but we would likely have asked for tests anyway and manuel's has them, so it saves a back-and-forth loop.

doruchan:
Thank you Max, no worries I am glad you are merging @manuelkoester! Thank you!

Issue #1100, opened 2021-06-21
add provides functionality


nerdvegas:
Snippet from https://github.com/nerdvegas/rez/issues/673:

In VFX there are often cases where a DCC ships with its own copies of other libraries (python is a classic example). Having rez resolve an environment that includes such a DCC, but also includes a standard installation of such a library, can be problematic.

To help manage this situation, it would be useful if a package could describe that it also provides another package as part of its own installation. This is another case where ephemerals come to the rescue - along with some changes to the solver.

To indicate that a package provides its own copy of another package, you would include it as a package feature:

requires = [
    ".{this.name}.feature.provides.python-2.7.3"
]
When the solver encounters a "provides" ephemeral, it would use it in lieu of its possibly already-resolved, real package counterpart. If a provided package ends up in the resolve, it plays no part in configuring the environment (the responsibility for that is left to the package - typically a DCC - doing the providing).

Note that the distinction between (eg) python-2.7.3 and python==2.7.3 as a provides statement is important. Whereas python-2.7.3 would take precedence over any python package within that range (including for eg python-2.7.3.1), python==2.7.3 would provide for that exact version only. If python-2.7.3.1 ended up in the resolve in this case, it would cause a conflict.

We will also need to consider what it means if >1 package provide the same package. Perhaps a configurable setting would cause a warning or error; if a warning, perhaps the first package encountered wins.

bfloch:
We are having active discussions about this very problem.
Is it something you foresee being addressed anytime soon?

nerdvegas:
I'm working on it as we speak!

The basic idea:
* A request for '.provides-foo-1' implies that foo-1 is already provided by
something
  * if a pkg specifies it as a requirement, the pkg means that it's
providing it (eg python within dcc)
  * if specified by user (ie in rez-env call) or as an implicit - this
means that the system itself provides it.
    * for eg, user may not want to have python itself under rez control and
instead just use system installation
* If a package is already found during a solve, but hits up on an
equivalent .provides later, that .provides is ignored and a warning is
printed.

That last point is important. The thing is, a resolve may pull in say,
foo-1 early in the solve, and later another pkg pulls in a .provides.foo-1.
But by then the damage is done - the resolve has been influenced by foo-1
in a way that can't be reversed (eg foo-1's requirements are already pulled
in). It also wouldn't be helpful nor intuitive to treat this as a conflict
and backtrack the solve either. Perhaps this behaviour should be
configurable though, I'm open to ideas.

[...]

PR #2137, opened 2026-06-20
Feature/issue 1100 provides

manuelkoester:
# `.provides.*` ephemerals

* Provides ticket: [#1100](https://github.com/AcademySoftwareFoundation/rez/issues/1100)
* Allan's initial proposal/notes: [REP-002.013](https://github.com/AcademySoftwareFoundation/rez/issues/673)

The goal of the provides feature is to be able to describe one package being bundled or `provided` by another. Therefore if a `provides` package is encountered it would be resolved & used in lieu of its possibly real package counterpart

This is especially usefull for things like DCCs which tend to ship a lot of their libraries/interpreters bundled as part of themselves. Houdini example: https://www.sidefx.com/docs/houdini/licenses/index.html
Before this there was no clean way to tell rez "this app already has python, do not pull in the actual python package"

I've opted to implement this feature via ephemerals and not an explicit package attribute like `provides = [..]` as with ephemerals it direclty works for variants and it also doesn't need an additional concept in the solver

To add a provided package to a package you add the provided package to the requirements and prefix it with `.provides.`

## How it behaves

A package that bundles its own python:

```python
name = 'maya'

requires = [
    '.provides.python-2.7.11'
]
```

Any request for the provided package that intersects the provided range gets satisfied by the provider.  
The real package is not resolved, doesn't show up in the resolved list, and doesn't even need to exist in a repo.  
The `.provides.*` ephemeral itself stays visible in the resolve:

```bash
$ rez-env maya my_python_tool
...
resolved packages:
.provides.python-2.7.11  (provided)
maya-2020                /pkgs/maya/2020
my_python_tool-1.0.0     /pkgs/my_python_tool/1.0.0
```

A request outside the provided range is a conflict, not a second python:

```bash
$ rez-env maya python-3
The following package conflicts occurred: (.provides.python-2.7.11 <--!--> python-3)
```

These are just normal package requests, so users can pass them too:  
`rez-env '.provides.python-3.7' my_tool` ~ "python is on the system"  

For more examples take a look at the tests in [test_solver.py](src/rez/tests/test_solver.py)

## Open questions

I'm not sure how we should handle multiple providers for teh same package. Currently it does resolve and just lets all providers provide the package (test covered in [test_solver.py](src/rez/tests/test_solver.py))

I'm also generally interested if people think this is the way to go about it

## Additional Ref / discussion dump

* https://github.com/AcademySoftwareFoundation/rez/blob/bb5ce448260da7c3d800e71325d4c985e0a4b769/ASWF/TSC/meeting-notes/2023-01-19.md?plain=1#L201
* https://academysoftwarefdn.slack.com/archives/C0321B828FM/p1749023896496059
* https://academysoftwarefdn.slack.com/archives/C0321B828FM/p1689373444632239
* https://academysoftwarefdn.slack.com/archives/C0321B828FM/p1711581154305699
* https://academysoftwarefdn.slack.com/archives/C0321B828FM/p1730978346899429
* https://academysoftwarefdn.slack.com/archives/C0321B828FM/p1756441918721399
* https://academysoftwarefdn.slack.com/archives/C0321B828FM/p1695335908531119

### DIFF
--- a/.github/workflows/installation.yaml
+++ b/.github/workflows/installation.yaml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
+        os: ['ubuntu-latest']
         python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
         method: ['install' ,'pip']
 
@@ -46,24 +46,6 @@ jobs:
           method: pip
           REZ_SET_PATH_COMMAND: 'export PATH=${PATH}:/opt/rez/bin PYTHONPATH=${PYTHONPATH}:/opt/rez'
           REZ_INSTALL_COMMAND: pip install --target /opt/rez .
-        # macOS
-        - os: macos-latest
-          method: install
-          REZ_SET_PATH_COMMAND: 'export PATH=${PATH}:~/rez/bin/rez'
-          REZ_INSTALL_COMMAND: python ./install.py ~/rez
-        - os: macos-latest
-          method: pip
-          REZ_SET_PATH_COMMAND: 'export PATH="$PATH:~/rez/bin" PYTHONPATH=$PYTHONPATH:$HOME/rez'
-          REZ_INSTALL_COMMAND: pip install --target ~/rez .
-        # windows
-        - os: windows-latest
-          method: install
-          REZ_SET_PATH_COMMAND: '$env:PATH="$env:PATH;C:\ProgramData\rez\Scripts\rez"'
-          REZ_INSTALL_COMMAND: python ./install.py C:\ProgramData\rez
-        - os: windows-latest
-          method: pip
-          REZ_SET_PATH_COMMAND: '[System.Environment]::SetEnvironmentVariable("PATH","$env:PATH;C:\ProgramData\rez\bin"); $env:PYTHONPATH="$env:PYTHONPATH;C:\ProgramData\rez"'
-          REZ_INSTALL_COMMAND: pip install --target C:\ProgramData\rez .
 
     steps:
     - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
--- a/.github/workflows/installation.yaml
+++ b/.github/workflows/installation.yaml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ['ubuntu-latest']
+        os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
         python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
         method: ['install' ,'pip']
 
@@ -46,6 +46,24 @@ jobs:
           method: pip
           REZ_SET_PATH_COMMAND: 'export PATH=${PATH}:/opt/rez/bin PYTHONPATH=${PYTHONPATH}:/opt/rez'
           REZ_INSTALL_COMMAND: pip install --target /opt/rez .
+        # macOS
+        - os: macos-latest
+          method: install
+          REZ_SET_PATH_COMMAND: 'export PATH=${PATH}:~/rez/bin/rez'
+          REZ_INSTALL_COMMAND: python ./install.py ~/rez
+        - os: macos-latest
+          method: pip
+          REZ_SET_PATH_COMMAND: 'export PATH="$PATH:~/rez/bin" PYTHONPATH=$PYTHONPATH:$HOME/rez'
+          REZ_INSTALL_COMMAND: pip install --target ~/rez .
+        # windows
+        - os: windows-latest
+          method: install
+          REZ_SET_PATH_COMMAND: '$env:PATH="$env:PATH;C:\ProgramData\rez\Scripts\rez"'
+          REZ_INSTALL_COMMAND: python ./install.py C:\ProgramData\rez
+        - os: windows-latest
+          method: pip
+          REZ_SET_PATH_COMMAND: '[System.Environment]::SetEnvironmentVariable("PATH","$env:PATH;C:\ProgramData\rez\bin"); $env:PYTHONPATH="$env:PYTHONPATH;C:\ProgramData\rez"'
+          REZ_INSTALL_COMMAND: pip install --target C:\ProgramData\rez .
 
     steps:
     - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -37,12 +37,18 @@ jobs:
 
     strategy:
       matrix:
-        os: ['ubuntu-latest']
+        os: ['macos-latest', 'ubuntu-latest', 'windows-latest']
         python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
         include:
+        - os: macos-latest
+          shells: 'sh,csh,bash,tcsh,zsh,pwsh'
+          rez-path: /installdir/bin/rez
         - os: ubuntu-latest
           shells: 'sh,csh,bash,tcsh,zsh,pwsh'
           rez-path: /installdir/bin/rez
+        - os: windows-latest
+          shells: 'cmd,pwsh,gitbash'
+          rez-path: \installdir\Scripts\rez
 
       fail-fast: false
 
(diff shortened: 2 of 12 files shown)


Priiam:
Great changes, I would very much benefit from this feature being made official!

I have one apprehension here, it is that my studio and other studios already use the ".provides..." ephemeral with hardcoded logic in `commands()`  to skip the part we don't want to load into the environment. If this PR was to be merged, the new feature would break my existing packages (while rez usually avoids releasing breaking changes)

I don't know if this alternative solution would be better or worse, but I imagine this could be a separate package.py attribute named `provides`, that would not overlap with existing ephemerals:


```python
name = "mypackage"

version = "1.2.3"

requires = ["dep2", "dep3"]

provides = ["dep1-1.1.1"]
```

On the behaviour side, I would personally prefer that the "provided" packages still be loaded from Rez, and that only a segment of code in the commands() function be ignored to avoid overtaking the provided components. For example, if my main package provides "git", I still have two valuable bits of information coming from my own "git" Rez package:

 - The environment overrides that I've set to configure git
 - The CPP files to be found by other packages when building from sources

So it would be good to have my "git" package loaded in, and just skip the part where git is added to the environment.

I'll add a message on ASWF's slack to see if people have any preference to use ephemerals or a separate attribute. I'm sure there are plenty of good reasons to prefer each option.

manuelkoester:
> I have one apprehension here, it is that my studio and other studios already use the ".provides..."

Mhh yea i was thinking about that potentially being a problem.. I guess we have a couple of options

1) Gate the feature behind a rezconfig flag  
My preferred path as its easy and people can opt in/out it at any point

2) Major bump rez  
Not great for many reasons

3) A separate package attribute like `provides = [".."]`.  
I'll go into it below

---- 

> but I imagine this could be a separate package.py attribute named provides

I (and others like Thorsten) had the same instict initially, but it has some downsides (see below). Allan also had an initial comment about it [here](https://github.com/AcademySoftwareFoundation/rez/issues/1100#issuecomment-868011275)

JC also thought about it and actually recommended adding it mostly as syntactic sugar.  
So `provides = ["python"]` behind the scenes results in ~`requires += [".provides.python"]`.  
I'm personally not super opposed, but I think it can add some confusion on requiring the `.provides` prefix in one place, but skipping it in another. So personally I would prefer staying explicit without the sugar.

**Variant support**  
By using ephemerals we get support for different **provides** variants for free. Think of an application supporting multiple python versions during a transition period. Houdini usually actually does this:  
* houdini-py310-20.5.684-win64-vc143.exe
* houdini-20.5.684-win64-vc143.exe

```python
name = "houdini"
version = "20.5.684"
variants = [
   [".provides.python-3.10"],  # maybe some more py3.10 specific provides
   [".provides.python-3.11"],  # maybe some more py3.11 specific provides
]
requires = [".provides.PySide6", ".provides.qtpy", ".provides.usd", ".provides.numpy"]
# provides = ["python-3.10"]  # not directly clear how to best combine variants with provides
```

**Additional solver complexity**
If we opted for adding a "properly" different package attribute into the solver it would increase in complexity and potentially solve speed quite a bit. Also I am personally not confident in making these changes. 

---- 

> I would personally prefer that the "provided" packages still be loaded from Rez, and that only a segment of code in the commands() function be ignored

In my mind if one package provides another, it's also responsible for providing the needed functionality of the package it provides. Otherwise it would not actually be "fully" provided. This MR follows that mentality and *completely* excludes the provided package from the resolve (and therefore also it's package commands).

So while i get the use case, i think loading the package also can't be the default behaviour for a couple of technical reasons

**The provided package intentionally does not have to exist**  
Houdini bundles `qtpy-2.4.3` and there may be no such rez package at all (see `test_17_provides_nonexistent_package`). And even when the family exists, which version do we load? Let's say `mypackage` bundles `git-2.41.0`, the repo only has `git-2.39.1` and `git-2.40.2`

**"Skip a segment of commands()" has no defined rules**  
commands() is arbitrary python, rez can't know which lines should be skipped. Depending on needs it could be the env vars or any other execution

**The loaded metadata can be actively wrong**  
On your CPP example: the external git package would put *its* `{root}/include` on CPATH, not the copy bundled in the provider. Other packages would compile against headers that don't match the binary on PATH -> silent version/ABI skew

---

For your git example there are a couple of ways the provider can live up to that responsibility:

* One way is to mirror the functionality; `mypackage` could do the same git config modifications that the git package itself does.  
* Another solution could be to separate that git config logic into its own package `git_config` that is then both required by `mypackage` and `git`  
* Or one could use one of the various current `provides` workarounds like a `GIT_NOOP` env var. (inspired by Thorsten's solution here: https://academysoftwarefdn.slack.com/archives/C0321B828FM/p1749023896496059). This doesn't actually exclude the git package from the resolve, so you have full control over what should be provided and what shouldn't

So for custom scenarios, where you'd still need **something** from the package that is being provided, you'll also need additonal steps while using `.provides` feature or you can use one of the current custom configurable solutions

I see your point though where this could be better in the future. Maybe could have a separate package attribute `on_provided`, that only runs when the package is provided. But after a quick think, it definitely comes with downsides that I personally would not like to address in this PR.

Priiam:
That makes a lot of sense, yes, and I think that if the feature is gated under a rezonfig.py attribute, then it is up to the user to define the behaviour they want. Sounds safe enough to me. 

Thanks for taking the time to answer thoroughly!

maxnbk:
> So it would be good to have my "git" package loaded in, and just skip the part where git is added to the environment.
> 
> I'll add a message on ASWF's slack to see if people have any preference to use ephemerals or a separate attribute. I'm sure there are plenty of good reasons to prefer each option.

Point 1: Personally, I don't think this would be possible from the reasons that Manuel stated and what the original design called for, however, I have a sidebar idea that might be worth entertaining. This would require, I believe, an additional (but minor) phase to the solver, where an otherwise 'final' solve could evaluate any `provides`-ed packages in the solve, find if or whether any existing package-versions happen to still fit that solve (solver order still wins), and perhaps any of those package-versions could be added as, syntax made up on the spot, REZ_SHADOWED_<BLAH> variables, which could be evaluated by some other processes and used in a manner external to the solve, or via some post_commands() hook of some final package or similar. Or perhaps a package attribute allows some aspect of those packages to still 'fire' some code even if they aren't truly in the solve. But I think this angle is overall a low priority compared to the value of the feature as it stands.

Point 2: I believe that whether they are expressed via a separate attribute of not, the intention of ephemerals was ultimately to be used as the basis for `provides`. Whether it has some sugar or not, I'm not really pick about at this stage, but just clarifying that I believed this to have been the original intent.

(I am still reading and working through thoughts on this, but wanted to post that much for now. Huge props to you for working on this, @manuelkoester !)

Priiam:
> any of those package-versions could be added as, syntax made up on the spot, REZ_SHADOWED_ variables, which could be evaluated by some other processes and used in a manner external to the solve, or via some post_commands() hook of some final package or similar. Or perhaps a package attribute allows some aspect of those packages to still 'fire' some code even if they aren't truly in the solve. But I think this angle is overall a low priority compared to the value of the feature as it stands.

Agreed. It is very useful but not the core of the feature here.

sanikache:
Thanks for taking this on, @manuelkoester! I was testing your changes and ran into a couple of unexpected behaviors when a weak package reference and an anti-package are included in the package list. I was expecting the solve to fail but it's not. 
Could you please check and confirm what I am seeing?

example1 `reqs = [Requirement("pydcc-1"), Requirement("!python")]`
```
>>> from rez.version import Requirement
>>> from rez.solver import Solver
>>> import os
>>> 
>>> packages_path = [os.path.join("src", "rez", "data", "tests", "solver", "packages")]
>>> 
>>> reqs = [Requirement("pydcc-1"), Requirement("!python")]
>>> 
>>> s = Solver(reqs, packages_path, optimised=True, verbosity=0)
>>> s.solve()
>>> 
>>> print("status:", s.status)
status: SolverStatus.solved
```

example2: `reqs = [Requirement("pydcc-1"), Requirement("~python-2.7")]`
```
>>> from rez.solver import Solver
>>> import os
>>> 
>>> packages_path = [os.path.join("src", "rez", "data", "tests", "solver", "packages")]
>>> 
>>> reqs = [Requirement("pydcc-1"), Requirement("~python-2.7")]
>>> 
>>> s = Solver(reqs, packages_path, optimised=True, verbosity=0)
>>> s.solve()
>>> 
>>> print("status:", s.status)
status: SolverStatus.solved
```

manuelkoester:
Hey @sanikache ! Thanks for giving this a try and finding these bugs. I didn't really think about weak or anti-package requirements..

I did check on my side and I'm seeing these fail too. I'll first add tests for them and then try to fix them.  I guess there'll need to be another section dedicated to specifically these types of requirements in combination with provides

manuelkoester:
I've added a step in the provides section in the solver that should cover your cases @sanikache and tests similar to your examples are green. 

Thanks for catching that bit

Issue #2120, opened 2026-05-25
Migrate from flake8 to ruff for linting and formatting


cfxegbert:
For linting the code base move from flake8 to ruff.

**Motivation**
* **Performance**: Written in Rust, Ruff is often 10–100x faster than traditional Python-based linters like Flake8. This makes it fast enough to run on every file save, providing near-instant feedback during development.
* **Tool consolidation**: Ruff serves as a "Swiss Army knife," replacing not just Flake8 but also isort, pydocstyle, pyupgrade, autoflake, and dozens of popular Flake8 plugins.
* **Built-in auto-fixing**: Ruff can automatically fix many violations, such as removing unused imports or upgrading syntax, using the --fix flag.
* **Code formatting**: Ruff can also be used for code formatting that matches the formatting from black
* **Doc string linting**: Ensure our code documentation is using one style of documentation strings.

**Related Issues/PRs**
* #51
* #1584

**Implementation**
* Create `ruff.toml` configuration that lints the code base without any changes to the code base.
* Add GitHub action to add ruff linting to workflow
* Incrementally enable more ruff rules.  Only enable a few rules at a time to limit the size of the pull requests to facilitate code reviews.  As mentioned in #1584 use `.git-blame-ignore-revs` to not cause large formatting changes to not muddle up the blame history.  Repeat until we feel we have implemented all the rules that we feel are necessary.
* Disable current flake8 GitHub action in workflow.
* Enable ruff formatting.

cfxegbert:
**Rule sets**

This comment will be kept up-to-date with the current rule sets that have been implemented.

|                    | Rule set | Name                                                                                   | Description                                        | Notes                                                                                          |
|:------------------:|----------|----------------------------------------------------------------------------------------|----------------------------------------------------|------------------------------------------------------------------------------------------------|
|        N/A         | AIR      | [Airflow](https://pypi.org/project/apache-airflow/)                                    |                                                    |                                                                                                |
|                    | ERA      | [Eradicate](https://pypi.org/project/eradicate/)                                       | Commented out code                                 |                                                                                                |
|        N/A         | FAST     | [FastAPI](https://pypi.org/project/fastapi/)                                           |                                                    |                                                                                                |
| :heavy_check_mark: | YTT      | [flake8-2020](https://pypi.org/project/flake8-2020/)                                   | Misuse of sys.version or sys.version_info          |                                                                                                |
[...]

PR #2139, opened 2026-06-26
refactor(ruff): remove ruff per file exemptions (#2120)

cfxegbert:
## Branch summary: `refactor/2120-remove-ruff-per-file-exemptions`

**13 commits** ahead of `main` (issue #2120). **16 files changed**, +55 / −84 lines.

### Goal

Remove most Ruff **per-file-ignores** and fix the underlying lint issues in code instead of suppressing them globally per file.

### Config & CI

- **`ruff.toml`**: Dropped ~20 per-file ignore entries (tests, bind, resolver, CLI, rezgui, install/setup scripts, etc.). Only one remains:
  - `src/rezplugins/release_vcs/git.py` — **F821** (undefined `git` in `export()`; uses the `git` Python package, which isn’t installed — noted as a known bug used by `rez-diff` / `diff_packages.py`)
- **`ruff.toml` + `.github/workflows/ruff.yaml`**: Exclude **`src/support/`** from Ruff entirely.
- **`src/rez/tests/ruff.toml`** (new): Extends root config and treats **rex builtins** (`env`, `command`, etc.) as known names in tests.

### Code fixes (replacing suppressions)

| Area | Change |
|------|--------|
| **Type comparisons (E721)** | `package_order.py`, `rezgui/objects/Config.py` — use `type(x) is type(y)` |
| **String concatenation (ISC002)** | `cli/_main.py` — f-strings instead of implicit concatenation |
| **Unused import (F401)** | `ContextToolsWidget.py` — removed unused `app` import; `install.py` — removed unused `virtualenv` import |
| **Lambda assignment (E731)** | `TimestampWidget.py` — inline lambda |
| **Undefined names (F821)** | `resolver.py` — TYPE_CHECKING imports; `bind/_pymodule.py` — inline `# noqa: F821` on `env` |
| **Ambiguous class (E742)** | `test_formatter.py` — inline `# noqa: E742` on class `I` |
| **Import order (E402)** | `install.py`, `setup.py`, `release-rez.py` — inline `# noqa: E402` |
| **Datetime (DTZ011)** | `release-rez.py` — inline `# noqa: DTZ011` on `date.today()` |
| **Git VCS** | `git.py` — added missing `retain_cwd` import (used by `export()`, which `diff_packages.py` calls) |

### Not in committed changes

**`diff_packages.py`** is unchanged on this branch; it’s relevant context because it calls `GitReleaseVCS.export()`, which is why the `git.py` F821 exemption remains.

I used some help from an LLM to create the summary for this PR

### DIFF
--- a/.github/workflows/ruff.yaml
+++ b/.github/workflows/ruff.yaml
@@ -13,6 +13,7 @@ on:
       - '!src/rez/utils/_version.py'
       - '!src/rez/vendor/**'
       - '!src/rez/data/**'
+      - '!src/support/**'
   push:
     paths:
       - '.github/workflows/ruff.yaml'
@@ -26,6 +27,7 @@ on:
       - '!src/rez/utils/_version.py'
       - '!src/rez/vendor/**'
       - '!src/rez/data/**'
+      - '!src/support/**'
 
 permissions:
   contents: read
--- a/install.py
+++ b/install.py
@@ -18,7 +18,6 @@ try:
     import venv
 except ImportError:
     USE_VIRTUALENV = True
-    import virtualenv
 
 
 source_path = os.path.dirname(os.path.realpath(__file__))
@@ -29,7 +28,7 @@ sys.path.insert(0, src_path)
 # though rez is not yet built.
 #
 from rez.utils._version import _rez_version  # noqa: E402
-from rez.utils.filesystem import safe_rmtree
+from rez.utils.filesystem import safe_rmtree  # noqa: E402
 from rez.utils.which import which  # noqa: E402
 from rez.cli._entry_points import get_specifications  # noqa: E402
 from rez.vendor.distlib.scripts import ScriptMaker  # noqa: E402
--- a/src/rez/bind/_pymodule.py
+++ b/src/rez/bind/_pymodule.py
@@ -21,12 +21,12 @@ import os.path
 
 
 def commands() -> None:
-    env.PYTHONPATH.append('{this.root}/python')
+    env.PYTHONPATH.append('{this.root}/python')  # noqa: F821
 
 
 def commands_with_bin() -> None:
-    env.PYTHONPATH.append('{this.root}/python')
-    env.PATH.append('{this.root}/bin')
+    env.PYTHONPATH.append('{this.root}/python')  # noqa: F821
+    env.PATH.append('{this.root}/bin')  # noqa: F821
 
 
 def copy_module(name, destpath):
--- a/src/rez/package_order.py
+++ b/src/rez/package_order.py
@@ -193,7 +193,9 @@ class PackageOrder(object):
         raise NotImplementedError
 
     def __eq__(self, other):
-        return type(self) == type(other) and str(self) == str(other)  # noqa: E721
+        if isinstance(other, type(self)):
+            return str(self) == str(other)
+        return NotImplemented
 
     def __ne__(self, other) -> bool:
         return not self == other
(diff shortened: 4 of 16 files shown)


.github/workflows/ruff.yaml
```diff
@@ -13,6 +13,7 @@ on:
       - '!src/rez/utils/_version.py'
       - '!src/rez/vendor/**'
       - '!src/rez/data/**'
+      - '!src/support/**'
```
JeanChristopheMorinPerso, line 16: Umm, support is something we would want to lint eventually...

src/rez/bind/_pymodule.py
```diff
@@ -21,12 +21,12 @@
 
 
 def commands() -> None:
-    env.PYTHONPATH.append('{this.root}/python')
+    env.PYTHONPATH.append('{this.root}/python')  # noqa: F821
```
JeanChristopheMorinPerso, line 24: I would prefer if we use qualified names instead of IDs.

src/rez/tests/ruff.toml
```diff
@@ -0,0 +1,20 @@
+extend = "../../../ruff.toml"
+
+# Treat rex as builtin 
+builtins = [
```
JeanChristopheMorinPerso, line 4: This feels "risky" to me. Can we instead use https://github.com/AcademySoftwareFoundation/rez/blob/main/src/rez/utils/lint_helper.py where these are referenced?

src/rez/package_order.py
```diff
@@ -193,7 +193,9 @@ def __str__(self) -> str:
         raise NotImplementedError
 
     def __eq__(self, other):
-        return type(self) == type(other) and str(self) == str(other)  # noqa: E721
+        if isinstance(other, type(self)):
```
JeanChristopheMorinPerso, line 196: Are we 100% sure that this will not create a problem? This is different than `type() == type()`. `isinstance` will check if an object is an instance of a class or a subclass.

src/rez/package_order.py
```diff
@@ -220,7 +222,7 @@ def __str__(self) -> str:
         return "{}"
 
     def __eq__(self, other):
-        return type(self) == type(other)  # noqa: E721
+        return type(self) is type(other)
```
JeanChristopheMorinPerso, line 225: Why do we use `is` here while we are using `isinstance` higher?

src/rez/package_order.py
```diff
@@ -193,7 +193,9 @@ def __str__(self) -> str:
         raise NotImplementedError
 
     def __eq__(self, other):
-        return type(self) == type(other) and str(self) == str(other)  # noqa: E721
+        if isinstance(other, type(self)):
+            return str(self) == str(other)
+        return NotImplemented
```
JeanChristopheMorinPerso, line 198: Why return NotImplemented here and not in other classes?

install.py
```diff
@@ -18,7 +18,6 @@
     import venv
 except ImportError:
     USE_VIRTUALENV = True
-    import virtualenv
```
JeanChristopheMorinPerso, line 21: Should we keep this just to make sure that there's an import failure if it's not installed?

JeanChristopheMorinPerso (commented):
Let a few comments. Let me know what you think

src/rez/bind/_pymodule.py
JeanChristopheMorinPerso: 0.16.0 was just released and they support that: https://docs.astral.sh/ruff/linter/#comments

Issue #2092, opened 2026-05-03
Increase test coverage


JeanChristopheMorinPerso:
Rez currently has a lot of tests, but our test coverage is only 60% at the time of writing this. Coverage can be visualized at https://app.codecov.io/gh/AcademySoftwareFoundation/rez/tree/main/.

The goal here is to add tests to cover critical pieces and help us reach at least 80% coverage.

Critical parts that need immediate attention, in no particular order:
* [src/rezplugins/release_vcs](tree/main/src/rezplugins/release_vcs) (particularly the git plugin)
* [src/rezplugins/release_hook](tree/main/src/rezplugins/release_hook) (this currently has no tests).
* [src/rez/package_cache.py](tree/main/src/rez/package_cache.py) (some tests, but could have more)
* [src/rez/package_search.py](tree/main/src/rez/package_search.py) (no tests)
* [src/rez/package_test.py](tree/main/src/rez/package_test.py) (has a few tests, but we need way more)
* [src/rez/resolved_context.py](tree/main/src/rez/resolved_context.py) (this is probably the part that needs the most attention with its 56% coverage)
* [src/rez/utils](tree/main/src/rez/utils) (these will be easy wins because most of the code under that package is relatively easy to test)

Also critical, but much harder to test:
* [src/rez/resolver.py](tree/main/src/rez/resolver.py) and [src/rez/solver.py](tree/main/src/rez/solver.py) (this is the brain. We have some solver tests, but we need way more. Before tackling this, please reach out to the TSC to make a plan)
* [src/rez/cli](tree/main/src/rez/cli) (this is the CLI, which was not particularly designed to be tested, not impossible, but also not always easy to test)

chinmaychahar:
Hey @JeanChristopheMorinPerso , I'd like to give this task a try, make an implementation plan and increase the code cov. Could it be assigned to me?

JeanChristopheMorinPerso:
Hi @chinmaychahar, thanks for showing interest in helping us increase code coverage! Because this issue is mostly an umbrella, I would prefer not to assign it to anyone.

Which part are you interested in? Tell us here and that should be enough for tracking who is working on what.

chinmaychahar:
I can start by working on the utils dir for now

ahlbertng:
Hey @JeanChristopheMorinPerso , I'd like to work on src/rez/package_search.py. i am a beginner contributor, starting there since it has no test coverage yet.

PR #2149, opened 2026-07-08
Add tests for package_search  (#2092)

ahlbertng:
## Description

This PR adds unit tests for `src/rez/package_search.py`, which currently has no test coverage (as noted in #2092).

Addresses part of #2092.

## What's tested

 `get_reverse_dependency_tree()`
  Correctly builds a reverse dependency tree for a package with known direct dependents
   Respects the `depth` argument (e.g. `depth=0` returns only the root package)
 `ResourceSearcher.search()`
   Finds all versions of a package family
   Returns only the latest version when `latest=True`
 `ResourceSearchResultFormatter`
   Default (no `output_format`) output correctly shows a package's qualified name
 `ResourceSearchResult`
   Correctly stores `resource`, `resource_type`, and `validation_error` (including default `None`)

## Notes

Tests reuse the existing `solver/packages` test data already used by `test_packages.py`, so no new test fixtures were added.
`get_plugins()` is not yet covered, no existing test data currently includes packages with `plugin_for` set. Open to guidance on whether to add new fixture packages for this, or if there's a preferred approach.

### DIFF
--- a/src/rez/tests/test_package_search.py
+++ b/src/rez/tests/test_package_search.py
@@ -8,7 +8,7 @@ test package_search module
 import unittest
 
 from rez.package_search import get_reverse_dependency_tree, get_plugins, \
-    ResourceSearcher
+    ResourceSearcher, ResourceSearchResultFormatter, ResourceSearchResult
 from rez.tests.util import TestBase, TempdirMixin
 from rez.version import Version
 
@@ -59,6 +59,28 @@ class TestPackageSearch(TestBase, TempdirMixin):
         self.assertEqual(resource_type, "package")
         self.assertEqual(len(results), 1)
         self.assertEqual(results[0].resource.version, Version("3"))
+    def test_formatter_default_output(self) -> None:
+        """test ResourceSearchResultFormatter with default (no output_format) settings"""
+        searcher = ResourceSearcher(package_paths=[self.solver_packages_path], latest=True)
+        resource_type, results = searcher.search("pydad")
+
+        formatter = ResourceSearchResultFormatter()
+        formatted_lines = formatter.format_search_results(results)
+
+        # default output just shows the qualified name
+        self.assertEqual(len(formatted_lines), 1)
+        text, color = formatted_lines[0]
+        self.assertEqual(text, "pydad-3")
+    def test_resource_search_result(self) -> None:
+        """test ResourceSearchResult stores its attributes correctly"""
+        result = ResourceSearchResult("fake_resource", "package")
+
+        self.assertEqual(result.resource, "fake_resource")
+        self.assertEqual(result.resource_type, "package")
+        self.assertIsNone(result.validation_error)
+
+        result_with_error = ResourceSearchResult("fake_resource", "package", "some error")
+        self.assertEqual(result_with_error.validation_error, "some error")
 
 if __name__ == '__main__':
     unittest.main()
--- a/src/rez/tests/test_package_search.py
+++ b/src/rez/tests/test_package_search.py
@@ -7,8 +7,12 @@ test package_search module
 """
 import unittest
 
-from rez.package_search import get_reverse_dependency_tree, get_plugins, \
-    ResourceSearcher, ResourceSearchResultFormatter, ResourceSearchResult
+from rez.package_search import (
+    get_reverse_dependency_tree,
+    ResourceSearcher,
+    ResourceSearchResultFormatter,
+    ResourceSearchResult
+)
 from rez.tests.util import TestBase, TempdirMixin
 from rez.version import Version
 
--- a/src/rez/tests/test_package_search.py
+++ b/src/rez/tests/test_package_search.py
@@ -1,0 +1,46 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Contributors to the Rez Project
+
+
+"""
+test package_search module
+"""
+import unittest
+
+from rez.package_search import get_reverse_dependency_tree, get_plugins, \
+    ResourceSearcher
+from rez.tests.util import TestBase, TempdirMixin
+
+
+class TestPackageSearch(TestBase, TempdirMixin):
+    @classmethod
+    def setUpClass(cls) -> None:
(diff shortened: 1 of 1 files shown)


src/rez/tests/test_package_search.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def test_resource_searcher_latest_only(self) -> None:
+        """test ResourceSearcher with latest=True returns only newest version"""
+        searcher = ResourceSearcher(package_paths=[self.solver_packages_path], latest=True)
+        resource_type, results = searcher.search("pydad")
+
+        self.assertEqual(resource_type, "package")
+        self.assertEqual(len(results), 1)
+        self.assertEqual(results[0].resource.version, Version("3"))
+    def test_formatter_default_output(self) -> None:
```
maxnbk, line 72: Could use a newline between this test-def and the prior line, this will likely fail future-enforced linter rules (E305)

src/rez/tests/test_package_search.py
```diff
@@ -0,0 +1,90 @@
...
+        resource_type, results = searcher.search("pydad")
+
+        formatter = ResourceSearchResultFormatter()
+        formatted_lines = formatter.format_search_results(results)
+
+        # default output just shows the qualified name
+        self.assertEqual(len(formatted_lines), 1)
+        text, color = formatted_lines[0]
+        self.assertEqual(text, "pydad-3")
+    def test_resource_search_result(self) -> None:
```
maxnbk, line 85: (same, could use a newline between this and the line above)

src/rez/tests/test_package_search.py
```diff
@@ -0,0 +1,90 @@
...
+
+        cls.settings = dict(
+            packages_path=[cls.solver_packages_path],
+        )
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        TempdirMixin.tearDownClass()
+
+    def test_reverse_dependency_tree(self) -> None:
```
maxnbk, line 35: `get_reverse_dependency_tree` tests don't verify the graph (g). The function returns a 2-tuple `(pkgs_list, g)` where `g` is a `digraph`. The tests unpack both but only assert on `pkgs_list`. The graph is arguably the more valuable output - it encodes the edge relationships. A test verifying `g.nodes()` and `g.edges()` would be more meaningful than just checking the list-of-lists.

src/rez/tests/test_package_search.py
```diff
@@ -0,0 +1,90 @@
...
+        """test get_reverse_dependency_tree finds direct and indirect parents"""
+        pkgs_list, g = get_reverse_dependency_tree("pymum")
+
+        self.assertEqual(pkgs_list[0], ["pymum"])
+        self.assertEqual(pkgs_list[1], ["pydad", "pyson"])
```
maxnbk: The assertion `pkgs_list[1] == ["pydad", "pyson"]` depends on the sorted order of the working set in `package_search.py` `(sorted(list(working_set_)))`. That's correct today, but the test doesn't communicate why those two packages appear - it's because `pymum-3` is required by `pydad-3` and `pyson-2`, and the function only looks at the latest version of each family. If someone adds a new package that requires `pymum` to the fixture, or changes which version is latest, the assertion breaks with no obvious connection to the underlying logic. A set-based assertion (`self.assertEqual(set(pkgs_list[1]), {"pydad", "pyson"})`) would be both more robust and more honest about what the test is actually verifying - that the direct dependents are found - without coupling to sort order.

maxnbk (commented):
Thank you for the submission of this new set of tests. We really appreciate any added coverage we can squeeze into the repo at this point, especially for such un-tested areas.

I've issued some specific comments attached to areas near the code that they pertain to. These are the main comments I'm making that would flip my recommendation from "don't merge" to "do merge". I made some specific line comments, but the real issue is that the current flake8 linter job fails, so just address those items.

I would express that the largest useful gaps to continue covering, whether they are a part of this PR, or you or anyone else tackles with additional work, would be these:

(Please comment, if you intend to go further with the coverage and aim to address these as well, or if you want to keep this PR self-contained just to the current change-set)

---

test the `validate=True` path. The `ResourceSearcher.search()` method has a substantial validate code path with error handling for `ResourceContentError`. No tests exercise this yet, and is arguably the most complex portion of the module.

test for glob patterns. `ResourceSearcher.search()` uses `fnmatch.fnmatch` on the name pattern. A test like `searcher.search("py*")` would verify the glob codepaths, which is a core feature of the searcher.

test for `family` resource type. When no version range is specified and multiple families match, the searcher returns `resource_type="family"`. The current tests only exercise the `package` resource type.

src/rez/tests/test_package_search.py
ahlbertng: Fixed. Added the missing `blank line`

src/rez/tests/test_package_search.py
ahlbertng: fixed

src/rez/tests/test_package_search.py
ahlbertng: added assertions on g.nodes() and g.edges() to verify the actual graph structure, not just the flattened list.

src/rez/tests/test_package_search.py
ahlbertng: switched to a set comparison so the test verifies the direct dependents are found, without depending on sort order.

ahlbertng:
Thanks for the detailed feedback! Fixed the flake8 issues (E302/E305 blank line spacing). It should be clean now.

ahlbertng:
I'd like to keep this PR focused on the current change set for now, since I'm still learning the code base. I am happy to open a follow up PR tackling the validate=True path, glob pattern tests, and family resource type tests next, once this one's merged

maxnbk:
If you wish, a rebase will clear the `ruff` CI error (fixed on main), otherwise I believe this is fine to merge.

maxnbk (approved):
Rebase to clear CI would be nice but otherwise good to merge.

PR #2150, opened 2026-07-09
Don't assume package repositories are on the filesystem

manuelkoester:
Hey!

We found during our tests that some rez internals assume that the package repository lives on the filesystem. 
There are very likely more of these assumptions scattered around, but the fixes in this PR are the ones we found blocking from doing any package_cache/rez release + remote repo work (without larger rez internal restructuring). The introduced `VariantResource.install` should likely make future fixes easier, too

I also added tests covering some cases around this topic for the package cache

Cheers!

### DIFF
--- a/src/rez/package_cache.py
+++ b/src/rez/package_cache.py
@@ -33,7 +33,10 @@ from rez.packages import get_variant, Variant
 from rez.system import system
 from rez.utils.filesystem import rename
 
-from typing import Iterable, Iterator
+from typing import Iterable, Iterator, cast, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from rez.package_resources import VariantResource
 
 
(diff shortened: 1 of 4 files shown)


src/rez/package_cache.py
```diff
@@ -410,7 +416,7 @@ def _while_copying() -> None:
         th.start()
 
         try:
-            shutil.copytree(variant_root, rootpath)
+            cast("VariantResource", variant.resource).install(rootpath)
```
manuelkoester, line 419: This actually seems nicer/cleaner: https://github.com/AcademySoftwareFoundation/rez/pull/1993/changes#diff-0127bbf5ad451b2bd74550607c4e46dbb9c6c427545dda044ed246ac13df50c1R408

PR #2151, opened 2026-07-09
Only async package caching needs prod install

manuelkoester:
Hey!

Basically the same as https://github.com/AcademySoftwareFoundation/rez/pull/2136, but does the same check in another needed location & adds some tests!

Relates to, but doesn't ffix #1500.

Cheers,
Manuel

### DIFF
--- a/src/rez/package_cache.py
+++ b/src/rez/package_cache.py
@@ -501,15 +501,16 @@ class PackageCache(object):
         """Add the given variants to the package payload cache.
         """
 
-        # A prod install is necessary because add_variants works by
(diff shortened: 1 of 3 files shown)


maxnbk:
I've reviewed this and found no substantive issues with the logic and exposed bits and bobs. I'm approving this and put it into the "Next" release milestone. @manuelkoester I'm not bothered by the one-liner coverage gap because the value of the fix supercedes the coverage stat, but if you want to find an easy way to cover it, I wouldn't complain.

maxnbk (approved):

manuelkoester:
Cheers for the review! Added the ~same tests for the other line 🫡

PR #2156, opened 2026-07-19
fix: base26 identifier ordering and input validation, with unit tests

darrenhuai:
rez.utils.base26 had no tests at all - get_next_base26 (the letter-increment logic for variant shortlink IDs) and create_unique_base26_symlink were both uncovered.

Writing the tests turned up two bugs, so this isn't tests-only any more - it also touches src/rez/utils/base26.py. Both were caught by @JeanChristopheMorinPerso reviewing the first round of tests:

- create_unique_base26_symlink picked the highest existing symlink with `max(names)`, which is plain lexicographic, so 'z' sorted above 'aa'. Once a directory rolls past 'z' the function keeps choosing 'z' as the previous ID and retrying the already-taken 'aa', burning all 10 retries and failing with "there was too much contention" - a real contention error for what is actually an ordering mistake. Now sorted with `key=lambda x: (len(x), x)`.
- get_next_base26 validated with `re.match`, and `$` matches before a trailing newline, so 'a\n' passed and incremented into 'a\x0b'. Now uses `fullmatch`.

Tests for get_next_base26 cover the increment/rollover cases directly (a->b, z->aa, az->ba, zz->aaa, etc) and invalid input. For create_unique_base26_symlink I mocked the filesystem calls (os.listdir, os.symlink, find_matching_symlink) instead of creating real symlinks, since that needs elevated privileges on Windows without Developer Mode - this still covers the existing-match short circuit, picking the next id after the highest existing one, the EEXIST retry-on-race path, non-EEXIST errors propagating, and giving up after too much contention.

Each fix has a regression test: a mixed-length listdir ('a', 'b', 'c', 'z', 'aa') that has to resolve to 'ab', and a subTest loop over the invalid inputs so a failure points at the specific bad value.

Ran the full suite locally (332 passed), flake8 and ruff clean on the new file.

Refs #2092

---

Disclosure: Claude (Sonnet 5) was used to assist in writing the test code and drafting this commit message and PR description.

### DIFF
--- a/src/rez/tests/test_utils_base26.py
+++ b/src/rez/tests/test_utils_base26.py
@@ -1,0 +1,112 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Contributors to the Rez Project
+
+
+"""
+test rez.utils.base26
+"""
+import errno
+import os.path
+from unittest.mock import patch
+
+from rez.utils.base26 import create_unique_base26_symlink, get_next_base26
+from rez.tests.util import TestBase
+
+
+class TestGetNextBase26(TestBase):
+    def test_no_prev_returns_a(self) -> None:
+        self.assertEqual(get_next_base26(None), 'a')
+        self.assertEqual(get_next_base26(''), 'a')
+
+    def test_increments_last_letter(self) -> None:
+        self.assertEqual(get_next_base26('a'), 'b')
+        self.assertEqual(get_next_base26('y'), 'z')
+        self.assertEqual(get_next_base26('ab'), 'ac')
+
+    def test_rolls_over_to_new_letter(self) -> None:
+        self.assertEqual(get_next_base26('z'), 'aa')
+        self.assertEqual(get_next_base26('az'), 'ba')
+        self.assertEqual(get_next_base26('zz'), 'aaa')
+
+    def test_rolls_over_multiple_letters(self) -> None:
+        self.assertEqual(get_next_base26('azz'), 'baa')
+
+    def test_invalid_input_raises(self) -> None:
+        for bad in ('A', 'a1', 'a-b', ' a', 'a '):
+            with self.assertRaises(ValueError):
+                get_next_base26(bad)
+
+
+class TestCreateUniqueBase26Symlink(TestBase):
+    """These mock the filesystem boundary (os.listdir/os.path.islink/os.symlink)
+    rather than creating real symlinks, since symlink creation requires elevated
+    privileges on some platforms (e.g. Windows without Developer Mode).
+    """
+
+    def test_returns_existing_symlink_if_already_pointing_at_source(self) -> None:
+        with patch(
+            "rez.utils.base26.find_matching_symlink", return_value="c"
+        ), patch("os.symlink") as symlink:
+            result = create_unique_base26_symlink("/pkgs", "/source/1.0")
+
+        self.assertEqual(result, os.path.join("/pkgs", "c"))
+        symlink.assert_not_called()
+
+    def test_creates_first_symlink_when_none_exist(self) -> None:
+        with patch(
+            "rez.utils.base26.find_matching_symlink", return_value=None
+        ), patch("os.listdir", return_value=[]), patch(
+            "os.symlink"
+        ) as symlink:
+            result = create_unique_base26_symlink("/pkgs", "/source/1.0")
+
+        symlink.assert_called_once_with("/source/1.0", result)
+        self.assertTrue(result.endswith('a'))
+
+    def test_creates_symlink_after_highest_existing(self) -> None:
+        with patch(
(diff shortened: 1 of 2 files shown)


maxnbk:
Hi @darrenhuai , thank you for the testing contribs.
Were these (this and previous PR) generated using LLMs?
If so, we need a disclosure in the PR title stating the LLM used and what they were used for.
This is in response to ASWF TAC discussions and will be added to our CONTRIBUTING.md shortly.
Thank you.

darrenhuai:
Yes, both PRs were AI-assisted. Claude assisted in writing the test code, and drafting the commit messages and PR descriptions. I reviewed the diffs and the test/lint output before pushing each one and made the call on what to ship. Happy to add a disclosure to the PR title, is there a format you'd prefer, since the policy isn't finalized yet? Thank you!

maxnbk:
@darrenhuai Something as simple as what I'm putting on my PRs is fine for now, like this, in the PR description (not the title):

`Disclosure: GLM-5.2 was used to assist in diagnostics, docstring, and test-writing.` Or something to that effect.

Just to explain (as a point of reference), we're not barring LLM-generated code, it's just so that we the maintainers can:
1. Review MRs better, because humans and LLMs tend to make different kinds of mistakes that can require different attitudes when evaluating
2. Understand better how much LLM usage is helping devs/maintainers work on the project.

darrenhuai:
perfect thank you! will do so

maxnbk (approved):

src/rez/tests/test_utils_base26.py
```diff
@@ -0,0 +1,112 @@
...
+            "rez.utils.base26.find_matching_symlink", return_value=None
+        ), patch(
+            "os.listdir", return_value=['a', 'b', 'c']
+        ), patch(
+            "os.path.islink", return_value=True
+        ), patch("os.symlink") as symlink:
+            result = create_unique_base26_symlink("/pkgs", "/source/2.0")
+
+        symlink.assert_called_once_with("/source/2.0", result)
+        self.assertTrue(result.endswith('d'))
```
JeanChristopheMorinPerso, line 78:
```suggestion
            "os.listdir", return_value=['a', 'b', 'c', 'z', 'aa']
        ), patch(
            "os.path.islink", return_value=True
        ), patch("os.symlink") as symlink:
            result = create_unique_base26_symlink("/pkgs", "/source/2.0")

        symlink.assert_called_once_with("/source/2.0", result)
        self.assertTrue(result.endswith('ab'))
```
Doing this will highlight that there's actually a bug in `create_unique_base26_symlink`. `max(names)` should be replaced with `max(names, key=lambda x: (len(x), x))`.

That's because:
```python
>>> max(['z', 'aa'])
'z'
```

In other words, the tests have full coverage purely in terms of code, but not in terms of logic.

src/rez/tests/test_utils_base26.py
```diff
@@ -0,0 +1,112 @@
...
+        self.assertEqual(get_next_base26('az'), 'ba')
+        self.assertEqual(get_next_base26('zz'), 'aaa')
+
+    def test_rolls_over_multiple_letters(self) -> None:
+        self.assertEqual(get_next_base26('azz'), 'baa')
+
+    def test_invalid_input_raises(self) -> None:
+        for bad in ('A', 'a1', 'a-b', ' a', 'a '):
+            with self.assertRaises(ValueError):
+                get_next_base26(bad)
```
JeanChristopheMorinPerso:
```suggestion
        for bad in ('A', 'a1', 'a-b', ' a', 'a '):
            with self.subTest(item=bad):
                with self.assertRaises(ValueError):
                    get_next_base26(bad)
```
This will more clearly show that we are testing multiple things.

ALso, note how there's a but in the `get_next_base26` where it accepts a trailing `\n`. We should replace `match` with `fullmatch`.

JeanChristopheMorinPerso (commented):
I threw a coding agent at your PR and it found two bugs in `base26.py` that would be very easy to fix in this PR if you want. I don't consider these blocking as they are probably corner cases.

darrenhuai:
@JeanChristopheMorinPerso both fixes are in as of 20e08b4 - apologies for pushing them without saying anything.

`get_next_base26` now uses `fullmatch` instead of `match`, so a trailing newline is rejected rather than silently accepted. And `create_unique_base26_symlink` picks the previous name with `max(names, key=lambda x: (len(x), x))`, so 'aa' correctly counts as higher than 'z'. I took your test suggestions as well - the mixed-length listdir case is in as its own regression test, and the invalid-input loop uses `subTest` so a failure points at the specific bad value.

Worth flagging that this makes the PR no longer tests-only: it now touches `src/rez/utils/base26.py`. @maxnbk your approval predates that commit, so you may want another look at it.

Suite and linters were clean when I pushed.

PR #2157, opened 2026-07-19
test: add unit tests for package_search.py

darrenhuai:
package_search.py didn't have any tests - ResourceSearcher and ResourceSearchResultFormatter were both completely uncovered. Added tests for the search behavior (family/package/variant resolution, glob matching, version ranges, latest-only, after_time/before_time filtering) and the formatter (default qualified-name output, abbreviation expansion, ambiguous abbreviations left as-is, splitting on the literal newline token in output_format).

Reused the existing fixture packages under data/tests and the TestBase/TempdirMixin setup already used elsewhere in the suite, so no new test data needed.

Ran the full suite locally (321 passed), flake8 and ruff are both clean on the new file.

Refs #2092

### DIFF
--- a/src/rez/tests/test_package_search.py
+++ b/src/rez/tests/test_package_search.py
@@ -1,0 +1,200 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Contributors to the Rez Project
+
+
+"""
+test package searching (ResourceSearcher, ResourceSearchResultFormatter)
+"""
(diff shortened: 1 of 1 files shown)


darrenhuai:
Closing - this duplicates #2149, which is already open and in review for the same file. Opened by mistake while I was in the middle of switching to a different part of #2092. Sorry for the noise.

PR #2158, opened 2026-07-20
Fix late-bound requires attributes returning None (#2153)

### LINKED ISSUE #2153: Rez 3.4.0 regression with private_build_requires()
Describe the bug here.

**Environment**
* rocky 9.5 linux
* Rez 3.4.0
* Rez python 3.11

**To Reproduce**
Add these lines in a `package.py` file
```
@late()
def private_build_requires():
    pass
```
then `rez build -ci`

**Expected behavior**
This is a regression. In Rez 3.3.0, this has no effect. In Rez 3.4.0, this generates a traceback.

The setting *is* wrong - the function should return a list, not None. But it does not fail in 3.3.0, but does in 3.4.0. 

I am not sure which version is the bug, but I think the correct behavior would be a warning that the value was not set properly, and then ignore it and continue.

Also, this only seems to happen with a late-binding function. If I just have this:

```
private_build_requires = None
```
It "works" (does not crash) in Rez 3.4.0

maxnbk:
Closes #2153 

3.4.0: PR #2004 added `build_requires` and `private_build_requires` to `late_bind_schemas` to fix a bug in `rez-search` causing it to crash. However, `late_requires_schema` only accepted lists. So, a `@late()` function returning `None`, for example, when there are no private build requirements, would raise `SchemaError: None should be instance of <class 'list'>` instead of being treated as an empty list. In 3.3.0, the value passed through unchecked and the `or []` in `get_requires()` handled it.

The fix was performed in two layers. Only one is necessary, but the second is a safety-guard for other schemas to produce warnings rather than crash.

Layer one: `late_requires_schema` (`package_resources.py`) - modified in place to accept `None` and normalize to `[]`.
Layer two: `_wrap_forwarded` (`packages.py`) - The defensive check: if any late-bound attribute's schema validation fails, and the value is `None`, warn, and treat as unset, rather than crashing. Protects attributes beyond the requires-family (`tools`, `cachable`, etc.) that have their own schemas.

To test, a new fixture `late_binding_none` with `@late()` functions returning `None` for `requires`, `build_requires`, and `private_build_requires`. All three assert that they are normalized to `[]`.

An additional test exercises the Layer two fix by patching the schema to intentionally hit the otherwise invalid path.

Disclosure: GLM-5.2 was used to assist in diagnostics, docstring, and test-writing.

### DIFF
--- a/src/rez/data/tests/packages/py_packages/late_binding_invalid/1.0/package.py
+++ b/src/rez/data/tests/packages/py_packages/late_binding_invalid/1.0/package.py
@@ -1,0 +1,14 @@
+name = "late_binding_invalid"
+
+version = "1.0"
+
+
+@late()
+def requires():
+    # Returns a non-None value that will fail schema validation when a strict
+    # schema is applied. Used to test the re-raise path in _wrap_forwarded.
+    return 42
+
+
+def commands() -> None:
+    pass
--- a/src/rez/data/tests/packages/py_packages/late_binding_none/1.0/package.py
+++ b/src/rez/data/tests/packages/py_packages/late_binding_none/1.0/package.py
@@ -1,0 +1,22 @@
+name = "late_binding_none"
+
+version = "1.0"
+
+
+@late()
+def requires():
+    pass
+
+
+@late()
+def build_requires():
+    pass
+
+
+@late()
+def private_build_requires():
+    pass
+
+
+def commands() -> None:
+    pass
--- a/src/rez/package_resources.py
+++ b/src/rez/package_resources.py
@@ -72,9 +72,12 @@ def late_bound(schema):
 
 
 # used when 'requires' is late bound
-late_requires_schema = Schema([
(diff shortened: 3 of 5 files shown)


JeanChristopheMorinPerso:
@maxnbk That's an oversight from me, thanks for fixing! I never thought someone would return None in a late bound function. Do you think this is something we should discourage this practice? If so, we should add a note in the docs and also in the warning about that I think as to make it more actionable.

Thoughts?

maxnbk:
> @maxnbk That's an oversight from me, thanks for fixing! I never thought someone would return None in a late bound function. Do you think this is something we should discourage this practice? If so, we should add a note in the docs and also in the warning about that I think as to make it more actionable.
> 
> Thoughts?

I see it as valid from a package construction and debugging practice, I don't have strong feelings about it otherwise. I think this is just a case of "cover our bases one way or the other", either by explicitly schema-fying the None away, or by allowing it but letting it no-op.

PR #2165, opened 2026-07-21
Draft: feat: package orderers as configurable plugins with config-arg-passing

maxnbk:
** WIP, DO NOT REVIEW YET :) **

# Summary

Rebases and fixes #1787 by @cfxegbert 
Moves the 5 built-in package-orderers from `src/rez/package_order.py` to `src/rezplugins/package_order/` as proper rez plugins.
Adds runtime configuration support via `REZ_PACKAGE_ORDERERS_JSON` and cache invalidation, creating a platform for PRs #1706 + #1709 to land in a usable manner with minimal changes.

# What changed (on top of the PR #1787 rebase+fixing)

1. Moves PackageOrderers into plugins.
2. Registers `PackageOrderPluginType` in `plugin_managers.py`
3. Replaces #1787 's function-based backward-compat shims with PEP-562 `__getattr__` lazy aliases, so `isinstance`, `from_pod`, `type(x) is` checks all work against the imported class names, while avoiding a circular import that module-level aliases would cause (plugin files import from package_order.py, which would call `_find_orderer` -> plugin system at load time)
4. Keeps `register_orderer()` and the `_orderers` dict as a means to subclass/apply orderers programmatically.
5. ruff-compatible reformatting of touched files ends up affecting many lines of code generically. It may be easiest to review commit by commit for this reason.

# Runtime configuration of Package Orderers

1. Documents that `REZ_PACKAGE_ORDERERS_JSON` works end-to-end using the existing `_JSON` env-var configuration pathway that then feeds into `config.package_orderers` -> `PackageOrderList.singleton` -> `from_pod()`
2. Adds `PackageOrderList.clear_singleton_cache()` classmethod to invalidate the cached singleton, so runtime config changes can take effect without restarting, or API-based-usage of the feature can be correctly managed.

# Documentation

1. Adds `REZ_PACKAGE_ORDERERS_JSON` usage section to `docs/source/package_orderers.rst`
2. Updates custom orderers section to recommend the plugin approach over `register_orderer()`

# Design Notes

1. The original plan was `SortedOrder = _find_orderer('sorted')` at module level, but this causes a circular import (`package_order.py` -> `_find_orderer()` -> plugin system -> `rezplugins/package_order/sorted.py` -> `from rez.package_order import PackageOrder` -> still importing). The `__getattr__` approach defers resolution to the first attribute access, breaking the cycle.
2. The locked config used by the test framework blocks env-var reads, so the runtime config tests simulate the `REZ_PACKAGE_ORDERERS_JSON` path by parsing JSON and using `config.override()` which is what the `_JSON` path does internally.
3. Needed a lazy class resolution to survive test_plugin_manager's plugin manager reset

# Testing

1. 46 orderer tests + 2 completion tests
2. Pre-flight tests were added _before_ the refactor to catch potential regressions: `from_pod` round-trip, `isinstance` against imported classes, `get_orderer` default fallback
4. New tests: plugin loading of all five built-in orderers, `to_pod` / `from_pod` round-trip through the plugin system, legacy `register_orderer` fallback, runtime config via `config.override` (simulating the `_JSON` path), cache invalidation, and a test-only kwarg-orderer that validates the pattern used in #1706 / #1709 (constructor kwargs configured at runtime).

# Credit

Original refactor by @cfxegbert . Rebased on to current main, conflicts resolved, backward-compat shims fixed, and runtime configuration support added by me.

# Follow-on work (not here)

1. Possibly perform an API-breaking change from usage of `PackageOrder` class names (and other namings/imports/etc) to `PackageOrderer` for making-more-sense?
2. PRs #1706 and #1709 could land as-is or could be migrated into `rezplugins` to join the rest.

Disclosure: This PR was assisted by GLM-5.2 for docstrings, documentation, and test generation.

### DIFF
--- a/docs/source/package_orderers.rst
+++ b/docs/source/package_orderers.rst
@@ -14,6 +14,35 @@ Configuration
 
 Package orderers can be configured in the ``rezconfig.py`` via the :data:`package_orderers` setting.
 
+Environment Variable Configuration
+==================================
+
+Package orderers can also be configured at runtime using the
+``REZ_PACKAGE_ORDERERS_JSON`` environment variable. The value must be a
+JSON-encoded list of orderer definitions, using the same structure as the
+:data:`package_orderers` setting:
+
+.. code-block:: bash
+
+   export REZ_PACKAGE_ORDERERS_JSON='[{"type": "per_family", "orderers": [{"packages": ["python"], "type": "version_split", "first_version": "2.7.16"}]}]'
+
+This is useful for configuring orderers with constructor arguments (such as
+prerelease risk tolerance or custom version priority ranges) without modifying
+``rezconfig.py``.
+
+.. note::
+
+   The environment variable is read on first access to the configuration.
+   If you need to change it at runtime via the API, clear both caches:
+
+   .. code-block:: python
+
+      from rez.config import config
+      from rez.package_order import PackageOrderList
+
+      config._uncache("package_orderers")
+      PackageOrderList.clear_singleton_cache()
+
 Types
 =====
 
@@ -188,9 +217,20 @@ package. You would use a :class:`rez.package_order.NullPackageOrder` in a :class
 Custom orderers
 ===============
 
-It is possible to create custom orderers using the API. This can be achieved
-by subclassing :class:`rez.package_order.PackageOrder` and implementing some mandatory
-methods. Once that's done, you need to register the orderer using :func:`rez.package_order.register_orderer`.
+It is possible to create custom orderers using the API. The preferred approach
+is to create a plugin in ``rezplugins/package_order/``. Alternatively, you can
+subclass :class:`rez.package_order.PackageOrder` and register the orderer using
+:func:`rez.package_order.register_orderer` (API-based registration).
+
+.. note::
+
+   The plugin system is preferred for persistent, facility-wide orderers.
+   Use ``register_orderer()`` for dynamic or programmatic registration.
+
+.. seealso::
+
+   For plugin-based orderer examples, see the built-in orderers in
+   ``src/rezplugins/package_order/``.
 
 .. note::
 
--- a/src/rez/package_order.py
+++ b/src/rez/package_order.py
@@ -9,10 +9,11 @@ from hashlib import sha1
 from typing import Any, Callable, Iterable, List, TYPE_CHECKING
 
 from rez.config import config
+from rez.exceptions import RezPluginError
 from rez.utils.data_utils import cached_class_property
 from rez.version import Version, VersionRange
-from rez.version._version import _Comparable, _ReversedComparable, _LowerBound, _UpperBound, _Bound
-from rez.packages import iter_packages, Package
+from rez.version._version import _Comparable, _LowerBound, _UpperBound, _Bound
+from rez.packages import Package
 from rez.utils.typing import SupportsLessThan
 
 if TYPE_CHECKING:
@@ -193,7 +194,7 @@ class PackageOrder(object):
         raise NotImplementedError
 
     def __eq__(self, other):
-        return type(self) == type(other) and str(self) == str(other)  # noqa: E721
+        return type(self) is type(other) and str(self) == str(other)
 
     def __ne__(self, other) -> bool:
         return not self == other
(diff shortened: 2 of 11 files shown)


src/rez/package_order.py
```diff
@@ -202,439 +191,55 @@ def __repr__(self) -> str:
...
-
-    @classmethod
-    def from_pod(cls, data: dict[str, Any]) -> Self:
-        return cls(
-            data["timestamp"],
-            rank=data.get("rank", 0),
-            packages=data.get("packages"),
-        )
+    if name in _LEGACY_ORDERER_NAMES:
+        return _find_orderer(_LEGACY_ORDERER_NAMES[name])
```
JeanChristopheMorinPerso, line 237: I think we should raise a deprecation warning and pave the road for removing this fallback in a few releases. We can use `rez.deprecations.warn`.

src/rez/tests/test_completion.py
```diff
@@ -5,6 +5,7 @@
 """
 test completions
 """
+
```
JeanChristopheMorinPerso, line 8: I would get rid of all the formatting changes if possible. We can make the formatting changes separately in another PR.

src/rez/package_order.py
```diff
@@ -202,439 +191,55 @@ def __repr__(self) -> str:
...
-       2.2.0
-
-    Notice that packages before ``T`` are preferred, followed by newer versions.
-    Newer versions are consumed in ascending order, except within rank (this is
-    why 2.1.1 is consumed before 2.1.0).
+    from rez.plugin_managers import plugin_manager
+
+    try:
+        return plugin_manager.get_plugin_class("package_order", name)
+    except RezPluginError:
```
JeanChristopheMorinPerso, line 209: Might be worth introducing a new exception class that subclasses `RezPluginError` to be more specific about what the error is, like `RezPluginNotFound` or something like that? Doing this wouldn't break user and it would improve our plugin system just a little bit.

src/rez/package_order.py
```diff
@@ -657,6 +262,19 @@ def singleton(cls) -> PackageOrderList:
...
         return cls.from_pod(config.package_orderers)
 
+    @classmethod
+    def clear_singleton_cache(cls) -> None:
+        """Clear the cached singleton so the next access re-reads config.
+
+        Use this when runtime configuration (e.g. REZ_PACKAGE_ORDERERS_JSON)
+        has changed and you want the new config to take effect. Note that
+        the config-level cache must also be cleared via
+        ``config._uncache("package_orderers")``.
```
JeanChristopheMorinPerso, line 272: >`config._uncache("package_orderers")`

This makes me feel uneasy as it's a private method. Also, why is this method needed? Are we doing something similar in other places and this just keeps things consistent? API users have access to `rez.config.config.copy/override/remove_override` to change the condig at runtime...

src/rez/package_order.py
```diff
@@ -744,28 +362,36 @@ def get_orderer(package_name: str, orderers: PackageOrderList | dict[str, Packag
         orderer = orderers.get(ALL_PACKAGES)
     if orderer is None:
         # default ordering is version descending
-        orderer = SortedOrder(descending=True)
+        sorted_order = _find_orderer("sorted")
```
JeanChristopheMorinPerso, line 365:
```suggestion
        sorted_order = _find_orderer("SortedOrder")
```
to avoid relying on the module getattr.

src/rez/package_order.py
```diff
@@ -744,28 +362,36 @@ def get_orderer(package_name: str, orderers: PackageOrderList | dict[str, Packag
...
+        orderer = sorted_order(descending=True)
     return orderer
 
 
+# Orderers registered at runtime via register_orderer(). This is the API-based
+# registration pathway, used as fallback by _find_orderer when an orderer is
+# not found in the plugin system.
+_orderers = {}
```
JeanChristopheMorinPerso, line 373: This is already defined at line 196.

src/rez/package_order.py
```diff
@@ -744,28 +362,36 @@ def get_orderer(package_name: str, orderers: PackageOrderList | dict[str, Packag
...
+    programmatic orderer registration. For persistent, facility-wide
+    orderers, prefer the plugin system (rezplugins/package_order/).
 
     Args:
-        cls (type[PackageOrder]): Package orderer class to register.
+        cls (type[PackageOrder]): Package order class to register.
 
     returns:
         bool: True if successfully registered, else False.
     """
```
JeanChristopheMorinPerso, line 392: We should add a deprecation warning to nudge users towards the plugins.

src/rez/plugin_managers.py
```diff
@@ -477,66 +464,72 @@ def create_instance(self, plugin_type: str, plugin_name, **instance_kwargs: Any)
...
 
 class CommandPluginType(RezPluginType):
-    """Support for different custom Rez applications/subcommands.
-    """
+    """Support for different custom Rez applications/subcommands."""
+
     type_name = "command"
 
 
+class PackageOrderPluginType(RezPluginType):
```
JeanChristopheMorinPerso, line 527: We need to update the plugin docs with this new plugin type: https://github.com/AcademySoftwareFoundation/rez/blob/main/docs/source/plugins.rst?plain=1#L14

src/rez/package_order.py
```diff
@@ -744,28 +362,36 @@ def get_orderer(package_name: str, orderers: PackageOrderList | dict[str, Packag
...
+# found by _find_orderer as a fallback when the plugin system does not have
+# a matching orderer. For filesystem-based, facility-wide orderers, prefer
+# creating a plugin in rezplugins/package_order/ instead.
 def register_orderer(cls: type[PackageOrder]) -> bool:
-    """Register an orderer
+    """Register an orderer.
+
+    This is the API-based registration pathway, useful for dynamic or
+    programmatic orderer registration. For persistent, facility-wide
+    orderers, prefer the plugin system (rezplugins/package_order/).
```
JeanChristopheMorinPerso, line 385:
```suggestion
    orderers, it is preferred to use :doc:`plugins <plugins>`.
```

docs/source/package_orderers.rst
```diff
@@ -14,6 +14,35 @@ Configuration
 
 Package orderers can be configured in the ``rezconfig.py`` via the :data:`package_orderers` setting.
 
+Environment Variable Configuration
```
JeanChristopheMorinPerso, line 17: I think it might be better if we document all the JSON environment variables. It feels weird that we would document this one specifically but not the others. Also, does the note about changing the config at runtime applicable to all settings? If so, it would be preferable to have this documented in the config docs.

JeanChristopheMorinPerso (commented):
I think I'm generally good with this but would like to see some changes before we merge this. I think it's going in the right direction. Note that this review was a manual review from me with no LLM assistance.

JeanChristopheMorinPerso:
Oh, I'm also wondering if we should deprecate `config.package_orderers` in favor of configuring them through the plugins?

docs/source/package_orderers.rst
JeanChristopheMorinPerso: Here's a PR that adds documentation for JSON env vars.

src/rez/plugin_managers.py
isohedronpipeline: Should the class be `PackageOrdererPluginType`?

PR #2169, opened 2026-07-25
fix(rex): don't expose failed env lookups in the env mapping

### LINKED ISSUE #2039: Accessing undefined env variable causes side-effect in unrelated package
Using `env.MY_VARIABLE` to access a non-existent variable in one package causes a `RexUndefinedVariableError`. This is expected behavior and can be caught in a try / except block. However, if another unrelated package's logic accesses the env's values using `env.values()` or similar afterwards, a second `RexUndefinedVariableError` is raised in that package due to the access to the variable in the first package. Order of operation is crucial here, as this obviously does not occur if `packageB` is resolved before `packageA`.

**Environment**
* Linux Rocky 9 / macOS 15.7.1
* Rez version 3.3.0 and older
* Rez Python version 3.11.13 and older

**To Reproduce**
1. Download the example [packages.zip](https://github.com/user-attachments/files/23263408/packages.zip) archive. 
2. Change directory into the extracted `packages` directory.
3. Execute: `REZ_PACKAGES_PATH=. rez-env packageA packageB`

**Expected behavior**
No error should be thrown during packageB's `commands` execution, even though a non-existent variable was accessed in packageA's `commands` function earlier. The environment should resolve as expected.

**Actual behavior**
The environment fails to resolve due to the error.

**Workaround**
Use `getenv("MY_VARIABLE")` instead of `env.MY_VARIABLE` to get the variable value and check for existence.

Sanjays2402:
Closes #2039

Accessing an undefined variable via `env.FOO` raises `RexUndefinedVariableError` (expected, catchable), but `EnvironmentDict.__getitem__` caches an `EnvironmentVariable` for that name anyway. A later `env.values()`/iteration — typically from an unrelated package resolved afterwards — resolves the cached name and raises a second `RexUndefinedVariableError`, failing the whole resolve.

The mapping methods now filter out names that are not actually defined, so a failed lookup leaves no trace; setting the variable later makes it visible again. New `test_11` in `test_rex.py` fails on `main` with the reported error and passes with the fix.

### DIFF
--- a/src/rez/rex.py
+++ b/src/rez/rex.py
@@ -1114,8 +1114,15 @@ class EnvironmentDict(MutableMapping):
         self._var_cache = dict((k, EnvironmentVariable(k, self))
                                for k in manager.parent_environ.keys())
 
+    def _defined_keys(self):
+        # A key is cached on first access, even if it was never defined (eg a
+        # package's commands() referencing a variable that does not exist).
+        # Such keys must not be exposed by the mapping interface, otherwise
+        # iterating the environment raises RexUndefinedVariableError.
+        return [k for k in self._var_cache if not self.manager.undefined(k)]
+
     def keys(self):
-        return self._var_cache.keys()
+        return dict.fromkeys(self._defined_keys()).keys()
 
     def __repr__(self) -> str:
         return '%s(%s)' % (self.__class__.__name__, str(self._var_cache))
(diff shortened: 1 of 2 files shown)


maxnbk:
Hello, thanks for the PR submission. We'd like to see CLA's signed before reviewing if possible, because if the CLA is unable to be signed, we are unable to accept the contribution. Hope you understand.

PR #2174, opened 2026-07-30
fix(ephemerals): don't cache empty ephemerals

### LINKED ISSUE #2173: Empty `ephemerals` inside `commands`
Hello ! We had an issue internally after witching to Rez 3.4.0, where sometimes the [`ephemerals` binding](https://rez.readthedocs.io/en/stable/package_commands.html#ephemerals) inside `commands` would be empty.

This was not the case for `context.resolved_ephemerals`, nor for the created `REZ_USED_EPH_RESOLVE` and `REZ_EPH_<PKG>_REQUEST` environment variables.

This does happen (or not) consistently for any given request, but does not happen for every request, if that makes sense.

One of the factors I could determine was if there was another late bound function in the package.

**Environment**
* OS: Windows
* Rez version: 3.4.0
* Rez python version: Python 3.13.8

**To Reproduce**

I wrote a small script to reproduce the issue.

The script creates a simple package that accesses `ephemerals` inside `commands` and has a late bound function, and marks whether it found a particular ephemeral.

It then resolves a context with this package and the expected ephemeral, and compares what was seen in `commands` with the actual resolution (the context `resolved_ephemerals` and environment variables injected by Rez on resolution).

You can run it with `rez python ./path/to/script.py`.

```python
import tempfile
from pathlib import Path

PACKAGE_PY = """
name = "foo"
version = "1.0"

@late()
def requires():
    return []

def commands():
    if "myeph" in ephemerals:
        env.EPH_SEEN_IN_COMMANDS = "1"
    else:
        env.EPH_SEEN_IN_COMMANDS = "0"
"""


def main():
    with tempfile.TemporaryDirectory(prefix="repro_case_empty_ephemerals_") as temp:
        package_repository = Path(temp)

        package_definition = package_repository / "foo" / "1.0" / "package.py"

        package_definition.parent.mkdir(parents=True, exist_ok=True)
        package_definition.write_text(PACKAGE_PY)

        context = ResolvedContext(
            ["foo", ".myeph-1"], package_paths=[str(package_repository)]
        )
        environ = context.get_environ()

        from_context = context.resolved_ephemerals
        from_env = environ.get("REZ_USED_EPH_RESOLVE")
        seen_in_commands = environ.get("EPH_SEEN_IN_COMMANDS")

        print(f"Ephemerals from `context.resolved_ephemerals`: {from_context}")
        print(f"Ephemerals from `REZ_USED_EPH_RESOLVE`: {from_env}")
        print(f"Ephemeral seen in `commands`: {seen_in_commands}")

        did_resolve = bool(from_context) and bool(from_env)

        if did_resolve and seen_in_commands != "1":
            print(
                "Ephemeral was resolved but absent from `ephemerals` inside `commands`"
            )


if __name__ == "__main__":
    main()
```

**Expected behavior**

An `ephemerals` binding matching `context.resolved_ephemerals` and the associated environment variables.

**Actual behavior**

Empty `ephemerals` binding.

**Related Issues/PRs**

* #1739
* #2117

**Regression**

The bug only affects the `ephemerals` binding, and is seemingly caused by the recent change introduced in #2117.

Internally, we found a few possible workarounds (such as using `context.resolved_ephemerals` in a late bound arbitrary attribute), and are working on a fix. It will probably be a patch to `rez.resolved_context.ResolvedContext._get_pre_resolve_bindings` that avoids caching `ephemerals` with other bindings, as ephemerals can be filled _during_ the resolution.

Aphosis:
Closes #2173 

Ephemerals were initially exposed to package commands as an `ephemerals` binding.

Since dcf77a26761627c02a671848943bf3d7c2d7df5c, they are also exposed to late-bound functions using the same strategy: reusing the existing list of resolved ephemerals.

The issue lies in the difference in timing between late-bound functions and package commands:

- Package commands are executed after the context have been resolved, so we get a properly resolved list of ephemerals
- Late-bound functions are executed _during_ a solve, which means we don't have access to the resolved ephemerals yet

The previous implementation did not crash when accessing `ephemerals` in late-bound functions: instead, it defaulted to an empty list.

This behavior leads to a second issue: ephemerals were added and **cached** as part of the list of pre-resolve bindings.

This can lead to the following scenario:

- A package late-bound function accesses the `ephemerals` binding
- Since the context is not resolve yet, no resolved ephemerals is found and an empty list is returned and cached
- Later, in the package commands, we also access `ephemerals`
- The cached empty list is returned, meaning that most queries will likely fail

This commit addresses these issues the following way:

- Move `ephemerals` out of pre-resolve bindings for commands, so that they are always retrieved from the resolved context
- For late-bound functions, use resolved ephemerals only if a context is available

### DIFF
--- a/src/rez/packages.py
+++ b/src/rez/packages.py
@@ -9,6 +9,7 @@ from rez.package_resources import PackageFamilyResource, PackageResource, \
     VariantResource, package_family_schema, package_schema, variant_schema, \
     package_release_keys, late_requires_schema
 from rez.package_serialise import dump_package_data
+from rez.rex_bindings import EphemeralsBinding
 from rez.utils import reraise
 from rez.utils.sourcecode import SourceCode
 from rez.utils.data_utils import cached_property
@@ -182,9 +183,11 @@ class PackageBaseResourceWrapper(PackageRepositoryResourceWrapper):
 
         if self.context is None:
             g["in_context"] = lambda: False
+            g["ephemerals"] = EphemeralsBinding([])
         else:
             g["in_context"] = lambda: True
             g["context"] = self.context
+            g["ephemerals"] = EphemeralsBinding(self.context.resolved_ephemerals or [])
 
             # 'request', 'system' etc
             bindings = self.context._get_pre_resolve_bindings()
--- a/src/rez/resolved_context.py
+++ b/src/rez/resolved_context.py
@@ -2049,7 +2049,6 @@ class ResolvedContext(object):
                 "testing": self.testing,
                 "request": RequirementsBinding(self._package_requests),
                 "implicits": RequirementsBinding(self.implicit_packages),
-                "ephemerals": EphemeralsBinding(self.resolved_ephemerals or []),
                 "intersects": intersects
             }
 
@@ -2135,6 +2134,7 @@ class ResolvedContext(object):
             executor.bind(k, v)
 
         executor.bind("resolve", VariantsBinding(variant_bindings))
+        executor.bind("ephemerals", EphemeralsBinding(self.resolved_ephemerals or []))
 
         #
         # -- apply each resolved package to the execution context
--- a/src/rez/tests/test_context.py
+++ b/src/rez/tests/test_context.py
@@ -8,16 +8,19 @@ test resolved contexts
 from rez.tests.util import restore_os_environ, restore_sys_path, TempdirMixin, \
     TestBase
 from rez.resolved_context import ResolvedContext
+from rez.resolver import ResolverStatus
 from rez.bundle_context import bundle_context
 from rez.bind import hello_world
 from rez.utils.platform_ import platform_
 from rez.utils.filesystem import is_subdirectory
+from rez.version import Requirement
 import unittest
 import subprocess
 import platform
 import shutil
 import os.path
 import os
+import textwrap
 
 
 class TestContext(TestBase, TempdirMixin):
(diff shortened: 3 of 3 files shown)


Aphosis:
I didn't add any test yet because I'm unsure where it should be added in the test suite, and how the test package should be created.

A good candidate seemed to be in the TestContext class: https://github.com/AcademySoftwareFoundation/rez/blob/aa10016d2d6eff2655001265cfe40855aa9f4e1d/src/rez/tests/test_context.py?plain=1#L23

Also, the test requires a package with specific attributes, and I'm not sure if the package should be created inline in the test or stored somewhere under `data/tests`.

maxnbk:
> A good candidate seemed to be in the TestContext class:

I would agree. The bug is about ephemerals being empty inside `commands()` during a `ResolvedContext` execution, which makes it a context execution test, rather than a rex binding test or a package attribute test. The existing methods in that class already test patterns like "resolve a context, execute commands, check the environ", which is exactly what needs doing to test this.

> Also, the test requires a package with specific attributes, and I'm not sure if the package should be created inline in the test or stored somewhere under `data/tests`.

IMHO ( @JeanChristopheMorinPerso might disagree), an inline-package created into a tempdir, resolving against it, and checking the environ is preferable to a fixture under `data/tests` because the pkgdef is right in the test, we can easily see what `late()` / `commands()` are set to, and what is being checked without digging into `data/tests`. The existing `late_binding` fixture is for testing package-attribute-access, not context-execution, so adding it there would probably blur the line between the package resource tests (`test_packages.py`) and context execution tests (`test_context.py`). The test also requires the specific ephemeral request, and so separating the pkgdef from the test would make the relationship between the two less clear.

A few specifics to help guide you to a working test that tests what you want it to test:
* Use self.root for the temp directory (TempdirMixin is being used so it's a managed temp dir that gets clean up automatically)
* Use `get_environ()` and not `execute_command()` to avoid annoying subprocess plumbing or `parent_environ` e.g. the way that `text_execute_command_testing_environ` does it.
* Assert both the truthiness of `r.resolved_ephemerals` and that the environ flag is `"1"` to catch both the case where the text fixture is wrong and no ephemeral resolves, and the actual bug itself
* Comment that the late function is there to trigger the bug, not to test the requires behavior.

I would also advise a second test that goes without the late-function as a control, which should also see the ephemeral, but is what proves the fix doesn't break the normal path while demonstrating that the late function was the trigger for the bug.

maxnbk (changes requested):
I am just marking this "Request Changes" so that review can be re-requested when updates are made, so that I'll see it again.

JeanChristopheMorinPerso:
>IMHO ( @JeanChristopheMorinPerso might disagree), an inline-package created into a tempdir, resolving against it, and checking the environ is preferable to a fixture under data/tests because the pkgdef is right in the test

FYI @maxnbk I'm usually in the camp of "no test data file and everything must be inline with the test to easily see what's being tested". Though, I know that this can sometimes incur some costs, but I don't think there's any costs involved here.

Aphosis:
Hi, and thanks for your answers! I'm resuming work on this and just added the test case, is there anything else worth adding (CHANGELOG.md maybe, or is that an automatic process) ?

EDIT: sorry for the multiple force-push, I fixed a formatting issue, updated the tests to take your suggestions into account, and signed-off the commits.

Aphosis:
Hi ! I'm not sure I re-requested a review correctly, so I'm commenting here again just in case. I added a test following your recommandations and this should now be ready for review.

JeanChristopheMorinPerso:
Hi @Aphosis, sorry for the delay. I took a closer look the other day and asked some feedback privately from other maintainers.

I'm kind of not sure how useful it is to have access to ephemerals in late bound functions. I think there's a use case, but it's quite narrow and there are a lot of corner cases that makes it a weird feature to support... From what I understand, ephemerals will only be available if the context is already solved (https://github.com/AcademySoftwareFoundation/rez/blob/dd5f7e323eb216afe24e15c282b86a8eaa41972a/src/rez/resolved_context.py#L377-L383) which is too late for late bound functions...

Basically, I'm not sure we should have merged https://github.com/AcademySoftwareFoundation/rez/pull/2117 which caused the bug you are trying to fix.

Aphosis:
Thanks for your answer, we are using this fix internally but we don't depend on the original feature itself (ephemerals binding in late bound functions), so either way is fine to me, although the original implementer of #2117 might rely on it now.

PR #2183, opened 2026-08-09
Add Sphinx domains for intersphinx consumers

JeanChristopheMorinPerso:
I introduced two new Sphinx domains in 3.4.0 without realizing that it's impossible to reference non-standard domains in other sphinx projects.

This PR creates a new Sphinx extension that our users can use to more easily reference our new domains using intersphinx.

Amp-Thread-ID: https://ampcode.com/threads/T-019fdefd-f7a9-75c8-8d27-ddb0ca62d6fe

### DIFF
--- a/src/rez/utils/sphinxext.py
+++ b/src/rez/utils/sphinxext.py
@@ -9,6 +9,8 @@ Consumers only need these domain names and the inherited Python roles for
 intersphinx resolution.
 """
 
+from rez.utils import _rez_version
+
 from sphinx.domains.python import PythonDomain
 
 __all__ = ["setup"]
@@ -34,6 +36,7 @@ def setup(app):
     app.add_domain(PkgDefDomain)
 
     return {
+        "version": _rez_version,
         "parallel_read_safe": True,
         "parallel_write_safe": True,
     }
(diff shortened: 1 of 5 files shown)


src/rez/utils/sphinxext.py
```diff
@@ -9,6 +9,8 @@
 intersphinx resolution.
 """
 
+from rez.utils import _rez_version
+
```
sanikache, line 13: This should be `from rez.utils._version import _rez_version` similar to https://github.com/AcademySoftwareFoundation/rez/blob/main/src/rez/__init__.py#L5

It is what's causing the following redthedocs build failure: https://app.readthedocs.org/projects/rez/builds/34187628/
```
sphinx.errors.ExtensionError: Could not import extension rez_sphinxext (exception: cannot import name '_rez_version' from 'rez.utils' (/home/docs/checkouts/readthedocs.org/user_builds/rez/checkouts/2183/src/rez/utils/__init__.py))
```

JeanChristopheMorinPerso: Good catch!

sanikache (approved):
LGTM.

PR #2187, opened 2026-08-15
Query Windows physical core count using `ctypes` and native Win32 API

nrusch:
# Overview

During the course of synchronizing some internal patches with the 3.4.0 release, I found that rez invocations were noticeably slower than they are using our current 3.3.0-based version. Some basic investigation showed that this is almost entirely the result of the new PowerShell-based CPU count query added in https://github.com/AcademySoftwareFoundation/rez/pull/2108.

In short, this method is extremely slow compared to the previous WMIC-based solution.

This PR introduces a new default code path for querying the physical core count that uses `ctypes` and the Win32 API to query the system directly.

# Performance comparison

To verify the difference, I extracted the current PowerShell- and WMIC-based methods for querying CPU count into a standalone file and ran them through the `timeit` module.

Here's what I get, running in Python 3.11 on Windows 11 Pro 25H2 (I still have WMIC installed):

### PowerShell

`1 loop, best of 5: 1.34 sec per loop`

Extremely slow, and it performs even worse in our work environment (averaging 2.5+ seconds per call).

### WMIC

`5 loops, best of 5: 53 msec per loop`

An easy order of magnitude faster, even though it still involves a process invocation.

### This PR (native Win32 wrapped in `ctypes`)

`10000 loops, best of 5: 33.6 usec per loop`

Let me know what you think.

### DIFF
--- a/src/rez/tests/test_utils.py
+++ b/src/rez/tests/test_utils.py
@@ -127,32 +127,64 @@ class TestWindowsPlatformCoreCounting(TestBase):
 
     # -- _physical_cores fallback chain ---------------------------------------
 
-    def test_physical_cores_prefers_powershell(self):
-        """powershell succeeds → wmic is never called."""
-        with unittest.mock.patch.object(platform_, '_physical_cores_from_powershell',
+    def test_physical_cores_prefers_native(self):
+        """If native succeeds, powershell and wmic are never called."""
+        with unittest.mock.patch.object(platform_, '_physical_cores_native',
+                                        return_value=4) as mock_native, \
+             unittest.mock.patch.object(platform_, '_physical_cores_from_powershell',
                                         return_value=4) as mock_ps, \
              unittest.mock.patch.object(platform_, '_physical_cores_from_wmic',
                                         return_value=4) as mock_wmic:
             result = platform_._physical_cores()
+            mock_native.assert_called_once()
+            mock_ps.assert_not_called()
+            mock_wmic.assert_not_called()
+            self.assertEqual(result, 4)
+
+    def test_physical_cores_powershell_fallback(self):
+        """If native fails and powershell succeeds, wmic is never called."""
+        with unittest.mock.patch.object(platform_, '_physical_cores_native',
+                                        return_value=None) as mock_native, \
+             unittest.mock.patch.object(platform_, '_physical_cores_from_powershell',
+                                        return_value=4) as mock_ps, \
+             unittest.mock.patch.object(platform_, '_physical_cores_from_wmic',
+                                        return_value=4) as mock_wmic:
+            result = platform_._physical_cores()
+            mock_native.assert_called_once()
             mock_ps.assert_called_once()
             mock_wmic.assert_not_called()
             self.assertEqual(result, 4)
 
     def test_physical_cores_falls_back_to_wmic(self):
         """powershell returns None (e.g. older Windows) → wmic result used."""
-        with unittest.mock.patch.object(platform_, '_physical_cores_from_powershell',
+        with unittest.mock.patch.object(platform_, '_physical_cores_native',
+                                        return_value=None), \
+             unittest.mock.patch.object(platform_, '_physical_cores_from_powershell',
                                         return_value=None), \
              unittest.mock.patch.object(platform_, '_physical_cores_from_wmic',
                                         return_value=4):
             self.assertEqual(platform_._physical_cores(), 4)
 
-    def test_physical_cores_both_fail_returns_none(self):
-        """Both helpers return None → _physical_cores returns None."""
-        with unittest.mock.patch.object(platform_, '_physical_cores_from_powershell',
+    def test_physical_cores_all_fail_returns_none(self):
+        """All helpers return None → _physical_cores returns None."""
+        with unittest.mock.patch.object(platform_, '_physical_cores_native',
+                                        return_value=None), \
+             unittest.mock.patch.object(platform_, '_physical_cores_from_powershell',
                                         return_value=None), \
              unittest.mock.patch.object(platform_, '_physical_cores_from_wmic',
                                         return_value=None):
             self.assertIsNone(platform_._physical_cores())
+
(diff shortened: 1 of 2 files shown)


maxnbk:
Very cool, thanks for this solution @nrusch . Might be tough to review since I doubt many of us have CTypes API experience, but if we can't figure out any problems, I think we will probably have to / want to merge this either way.

One thing I am wondering, or was wondering after I merged the previous powershell workaround for the deprecated WMIC methods is:

1. Is there any reason to, or not to, memoize the calls? It's hard for me to imagine situations where the underlying system could somehow change available core counts during a running rez process, other than cgroup-managed kubernetes-style pod/cluster setups, and even that, I don't know if that works that way precisely. (Though I also am not totally sure if there are any routes in rez that involve these calls being made more than once?)
2. The other windows-based (and I think all the others too) do some form of try/except. With your new method, if it threw an exception anywhere in that function, the other fallback options wouldn't even fire because it would raise an exception and break, defaulting back to undetected -> 1 core, unless I'm misreading something?

nrusch:
Hey @maxnbk , thanks for taking a look. Regarding your questions:

> 1. Is there any reason to, or not to, memoize the calls?

I had a similar thought as well, and I can't imagine why one wouldn't want to do this. Even if there are some ways to change the apparent core count of a system while processes are running (which seems extremely niche), I'm hard-pressed to imagine why `rez` would care...

I'm happy to add this if you think it's a reasonable step.

> 2. The other windows-based (and I think all the others too) do some form of try/except. With your new method, if it threw an exception anywhere in that function, the other fallback options wouldn't even fire because it would raise an exception and break, defaulting back to undetected -> 1 core, unless I'm misreading something?

You are not misreading. Win32 is a C API, so it cannot raise exceptions, and thus there are no exception semantics to propagate to Python.

However, I don't disagree with your overall instinct for caution here... I suspect it's possible there could be some temporary/edge-case conditions under which `ctypes` itself could raise an exception (separate from any of the Win32 interaction), so I'm happy to add in exception handling.

On a tangentially-related note, given that the method is already fairly long, I was thinking about establishing a new module for Windows-specific utilities, and then that method could import and call the function from there. I've also thought of adding `ctypes` implementations for some other system interaction patterns in `rez`, and that would likely end up being a natural home for any Windows-specific ones.

Any thoughts on that idea, or what the most logical module location/identity might be?

src/rez/utils/platform_.py
```diff
@@ -516,6 +516,105 @@ def symlink(self, source: str, link_name: str):
...
+        if not success:
+            return None
+
+        # Walk the buffer of `SYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX` struct instances and count the
+        # number of matching entries.
+        offset = 0
+        num_cpus = 0
+        buffer_start = ctypes.addressof(buffer)
+
+        while offset < buffer_size.value:
```
JeanChristopheMorinPerso: I don't know if this can ever happen, but if `buffer_size.value` is 0, this will loop forever.

src/rez/utils/platform_.py
```diff
@@ -516,6 +516,105 @@ def symlink(self, source: str, link_name: str):
     def _terminal_emulator_command(self) -> str:
         return "START"
 
+    def _physical_cores_native(self) -> int | None:
+        """Query the physical CPU count using `ctypes` wrappers for the Win32 API."""
+        import ctypes
+        from ctypes import wintypes
+
+        kernel32 = ctypes.WinDLL("kernel32", use_last_error=True)
```
JeanChristopheMorinPerso: I think I would put all of these under a try/except. I don't know if this and other call can fail, but let's protect ourselves, just in case.

JeanChristopheMorinPerso (changes requested):

src/rez/utils/platform_.py
nrusch: I don't think that's possible according to the parameters of the Win32 API, but I've added a check for a 0 value just to be safe.

src/rez/utils/platform_.py
nrusch: I've wrapped the entire function body in a `try-except`.

nrusch:
@maxnbk just to follow up on this particular question:

> Is there any reason to, or not to, memoize the calls?

It's worth noting that the value is ultimately cached at a "higher" level by the `Platform.physical_cores` accessor.

src/rez/tests/test_utils.py
```diff
@@ -127,33 +127,65 @@ def test_wmic_no_match_returns_none(self):
...
                                         return_value=None):
             self.assertIsNone(platform_._physical_cores())
 
+    # -- _physical_cores internal consistency ---------------------------------
+
+    def test_physical_cores_methods_same_result(self):
+        """All helpers that succeed return the same result."""
+        native_cores = platform_._physical_cores_native()
+        pwsh_cores = platform_._physical_cores_from_powershell()
+        wmic_cores = platform_._physical_cores_from_wmic()
```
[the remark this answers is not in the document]
nrusch: I can look at mocking the `ctypes` Win32 interface and the calls used by this method, but I think it would be quite a bit of test machinery. I'm not sure how thoroughly you'd like to exercise all the early-out cases here, but the juice may not be worth the squeeze given that the entire thing is in a `try-except` now.

@JeanChristopheMorinPerso let me know if you'd like me to go ahead with this.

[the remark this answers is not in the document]
JeanChristopheMorinPerso: I think we can ignore this.

JeanChristopheMorinPerso (approved):
LGTM, thanks for contributing this change!

nrusch:
Thanks @JeanChristopheMorinPerso. Just to check in, is there anything else that needs to be done before this is merged? Just asking because I'm hoping to be able to integrate it into our internal deployments, but I'd like to avoid having to do it via an internal patch at this point.

JeanChristopheMorinPerso:
There you go @nrusch

PR #2189, opened 2026-08-22
Improve configuration environment variable docs

JeanChristopheMorinPerso:
We've had a gap for years now in regards to config settings environment variables. I partly filled that gap when I introduced https://rez.readthedocs.io/en/stable/environment.html and the automatic documentation of env vars in settings (https://rez.readthedocs.io/en/stable/configuring_rez.html#envvar-REZ_CONTEXT_TMPDIR), but there was still a major missing piece.

Some of our settings can only be configured via an environment variable suffixed with `_JSON`.

This PR properly documents those and also documents that all settings be set via `REZ_XXX_JSON`. I also took the occasion to document the format of some env vars too.

With this, I think we can finally say that all our settings and env vars are properly documented. At least I think, until someone comes in to tell me that we missed something :smile:

Lastly, I also fixed an old bug that was triggered when a user was setting a setting via an env var but without the JSON suffix for a setting that only supports JSON. Rez was raising a fat from helpful `NotImplementedError`. With the fix, rez raises an actionable exception.

AI usage: I started with an LLM and then reworked the results a bit and continued to iterate with it after.

Amp-Thread-ID: https://ampcode.com/threads/T-01a02b3d-dd38-7533-abe8-4844dadda902

### DIFF
--- a/docs/rez_sphinxext.py
+++ b/docs/rez_sphinxext.py
@@ -4,6 +4,7 @@ import argparse
 
 import rez.cli._main
 import rez.cli._util
+import rez.config
 import rez.rezconfig
 import docutils.nodes
 import sphinx.util.nodes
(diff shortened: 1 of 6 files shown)


sanikache:
Is it worth adding a new test that guards against any Setting subclass referenced in config_schema that doesn't set `env_var_json_only = True` ?

JeanChristopheMorinPerso:
@sanikache I pushed a test that I think covers what you asked for. What do you think?

sanikache:
Looks great!

sanikache (approved):
LGTM.

PR #2194, opened 2026-09-06
Speed up plugin discovery

JeanChristopheMorinPerso:
Leaving as draft until I get the time to polish and fully review the tests and think about potential unexpected side effects.

## Summary

Plugin discovery was spending a surprising amount of time repeatedly scanning Python packages, installed entry-point metadata, and shared plugin configuration.

This changes the discovery path in three places:

- Uses an `os.scandir()` fast path when looking for top-level packages that may contain `rezplugins`.
- Discovers installed entry points once and shares the result across plugin types. The cache is invalidated when `sys.path` or its directory mtimes change.
- Loads `rezconfig` once per plugin directory instead of once per plugin module.

ZIP files and custom importers still fall back to `pkgutil`, so those discovery mechanisms continue to work as before.

## Performance

The benchmark repeatedly resets and loads all seven plugin types for 100 discovery cycles. It excludes the unrelated import optimizations investigated earlier.

| Measurement | Before | After | Improvement |
|---|---:|---:|---:|
| Overall plugin discovery/loading | 1.146 s | 548.2 ms | 2.09× faster |
| Top-level `rezplugins` discovery | 1.203 s | 297 ms | 4.05× faster |
| Entry-point discovery | 850 ms | 249 ms | 3.41× faster |
| Shared plugin config loading | 231 ms | 55 ms | 4.20× faster |

The individual timings are cumulative `cProfile` results over the repeated workload run using hyperfine, so they are not intended to add up to the `hyperfine` wall-clock result.

## Testing

- Added coverage ensuring entry-point metadata is discovered once across plugin types.
- Added coverage ensuring entry points are rediscovered after `sys.path` changes.
- Added coverage ensuring shared plugin configuration is loaded once per plugin path.
- Existing plugin-manager tests pass.
- Verified the optimized scanner returns the same ordered module set as `pkgutil.iter_modules()` in the benchmark environment.

## AI usage

This was mostly generated by an LLM with a lot of babysitting.

Amp-Thread-ID: https://ampcode.com/threads/T-01a0771d-16ef-7787-b60b-eb71b1cc8a4b (sorry in advanced for cursing a bit in that thread... The models were gaslighting me in multiple threads all trying to inline dozens of imports to find performance gains).

### DIFF
--- a/src/rez/plugin_managers.py
+++ b/src/rez/plugin_managers.py
@@ -15,6 +15,8 @@ from rez.utils.logging_ import print_debug, print_warning
 from rez.exceptions import RezPluginError
 from zipimport import zipimporter
 from typing import overload, Any, Literal, TypeVar, TYPE_CHECKING
+from importlib.machinery import FileFinder, all_suffixes
+from functools import lru_cache
 import pkgutil
 import os.path
 import sys
@@ -36,6 +38,25 @@ if TYPE_CHECKING:
 
 
 T = TypeVar("T")
+
+
+def _entry_points_cache_key():
+    """Describe import locations closely enough to detect installations."""
+    result = []
+    for path in sys.path:
+        path = path or os.getcwd()
+        try:
+            mtime = os.stat(path).st_mtime_ns
+        except OSError:
+            mtime = None
+        result.append((path, mtime))
+    return tuple(result)
+
+
+@lru_cache(maxsize=1)
+def _get_entry_points(_cache_key):
+    """Discover installed entry points once for all plugin types."""
+    return entry_points()
 
 
 # modified from pkgutil standard library:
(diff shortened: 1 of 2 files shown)


src/rez/plugin_managers.py
```diff
@@ -38,6 +40,25 @@
...
+        try:
+            mtime = os.stat(path).st_mtime_ns
+        except OSError:
+            mtime = None
+        result.append((path, mtime))
+    return tuple(result)
+
+
+@lru_cache(maxsize=1)
+def _get_entry_points(_cache_key):
```
JeanChristopheMorinPerso, line 57: This is such a weird pattern. I'll have to rewrite this into something that looks more sane.

src/rez/plugin_managers.py
```diff
@@ -38,6 +40,25 @@
 T = TypeVar("T")
 
 
+def _entry_points_cache_key():
```
JeanChristopheMorinPerso, line 43: Same here, weird pattern.

src/rez/plugin_managers.py
```diff
@@ -98,6 +119,83 @@ def append_if_valid(dir_) -> None:
 def uncache_rezplugins_module_paths(instance=None) -> None:
     instance = instance or plugin_manager
     cached_property.uncache(instance, "rezplugins_module_paths")  # type: ignore[attr-defined]
+    _get_entry_points.cache_clear()
+
+
+def _iter_top_level_modules():
```
JeanChristopheMorinPerso, line 125: I have to review the correctness of this function.

PR #2195, opened 2026-09-06
Uncached os.listdir() in _get_family() causes much slower resolves on Windows

mandeep:
Hi all,

I was seeing extremely slow resolve times on Windows, even with memcached showing a 99% hit rate. I caught that `os.listdir()` was being called for every package check. I changed this to use the decorated `_get_familydirs()` and noticed that resolve times were now instant. I'm new to Rez so I don't know if this is the correct way to handle this, but it seems to be working now as intended.

Thanks!
Mandeep

### DIFF
--- a/src/rezplugins/package_repository/filesystem.py
+++ b/src/rezplugins/package_repository/filesystem.py
@@ -1151,9 +1151,10 @@ class FileSystemPackageRepository(PackageRepository):
         is_valid_package_name(name, raise_error=True)
         if os.path.isdir(os.path.join(self.location, name)):
             # force case-sensitive match on pkg family dir, on case-insensitive platforms
-            if not platform_.has_case_sensitive_filesystem and \
-                    name not in os.listdir(self.location):
-                return None
+            if not platform_.has_case_sensitive_filesystem:
+                dirs = set(dir_name for (dir_name, ext) in self._get_family_dirs() if ext is None)
+                if name not in dirs:
+                    return None
 
             return self.get_resource(
                 FileSystemPackageFamilyResource.key,
@@ -1166,7 +1167,8 @@ class FileSystemPackageRepository(PackageRepository):
                 # force case-sensitive match on pkg filename, on case-insensitive platforms
                 if not platform_.has_case_sensitive_filesystem:
                     ext = os.path.splitext(filepath)[-1]
-                    if (name + ext) not in os.listdir(self.location):
+                    filenames = set("%s.%s" % (filename, extension) for filename, extension in self._get_family_dirs() if extension)
+                    if (name + ext) not in filenames:
                         return None
 
                 return self.get_resource(
--- a/src/rezplugins/package_repository/filesystem.py
+++ b/src/rezplugins/package_repository/filesystem.py
@@ -556,6 +556,14 @@ class FileSystemPackageRepository(PackageRepository):
             )
             self._get_version_dirs = decorator2(self._get_version_dirs)
 
+            decorator3 = memcached(
+                servers=config.memcached_uri if config.cache_listdir else None,
+                min_compress_len=config.memcached_listdir_min_compress_len,
+                key=self._get_dir_listing__key,
+                debug=config.debug_memcache
+            )
+            self._get_dir_listing = decorator3(self._get_dir_listing)
+
     def _uid(self) -> tuple:
         t = ["filesystem", self.location]
         if os.path.exists(self.location):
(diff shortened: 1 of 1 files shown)


JeanChristopheMorinPerso:
Hi @mandeep , thanks for opening this PR. Do you have more details on the performance slow down you were getting and what you get now with these changes? Extreme and massive are two big adjectives. Not saying that it wasn't the case, but I'd like to see some numbers to back this up and help us a bit.

When talking about performance, it's important to provide some numbers and more context to get an idea of the scale of the problem and changes. Things like what kind of performance you were getting before, and what you get now, how many directories are in your repos, how many repos, how many packages, type of filesystem, etc.

It will also help us determine if the problem is this or if it's related to something (like the cache for example).

Thanks!

mandeep:
Hi @JeanChristopheMorinPerso!

On Windows 11 with a network drive formatted to NTFS, resolves were taking 45-60 seconds for any given project. There are 10 or so repos with 250-350 packages in each. This setup uses memcached with a hit rate of 99%. With this change, resolves are now 1-3 seconds. Calling `os.listdir` across a network for each package is extremely expensive, even with a memcached setup since the cache is unused. 

If there's any command output or anything else you would like to see, please let me know!

src/rezplugins/package_repository/filesystem.py
```diff
@@ -1151,9 +1151,10 @@ def _get_family(self, name: str) -> PackageFamilyResource | None:
         is_valid_package_name(name, raise_error=True)
         if os.path.isdir(os.path.join(self.location, name)):
             # force case-sensitive match on pkg family dir, on case-insensitive platforms
-            if not platform_.has_case_sensitive_filesystem and \
-                    name not in os.listdir(self.location):
-                return None
+            if not platform_.has_case_sensitive_filesystem:
+                dirs = set(dir_name for (dir_name, ext) in self._get_family_dirs() if ext is None)
```
JeanChristopheMorinPerso: I've not yet spent enough time on this, but from a first look, `self._get_family_dirs()` calls `os.isdir` adn then `os.listdir`, and then for each dir returned by `os.listdir`, call `os.isdir`. This will result in a performance regression in theory... And `_get_family_dirs` is not cached, so I don't see how it can actually improve the performance.

JeanChristopheMorinPerso (commented):
I've only looked for 10 minutes, but so far, I'm not convinced that this will bring a performance improvement. In fact, I think this might actually create a performance regression. I think.

src/rezplugins/package_repository/filesystem.py
JeanChristopheMorinPerso: Spent a little bit more time on this. The performance regression will only be when caching is disabled, which is the default.

mandeep: I am new to the codebase, but I think this line uses memcached for directory lookup. https://github.com/AcademySoftwareFoundation/rez/blob/6c3d75415d066726856b42613710aafc025cfb15/src/rezplugins/package_repository/filesystem.py#L549

You're correct that this would be a regression when caching is disabled. Ideally, `os.listdir` would not be called per package.

maxnbk:
I'm 99% certain that this specific "fix" has come up before, and I do recall that it's fundamentally required for some mechanic reason in rez, but I'll need to hunt to come up with the reason why. That doesn't mean that opportunities for optimizing it, or perhaps finding creative ways to work around or configure-away the problem are without merit, but it does need scrutiny.

JeanChristopheMorinPerso:
@maxnbk https://github.com/AcademySoftwareFoundation/rez/pull/689?

mandeep:
Thanks @maxnbk. I agree that scrutiny is needed and hopefully we can work through a solution.

The fix here doesn't really alter much for those not using a cache. `os.listdir()` is still called to collect a list of directories. This will not change anything for case-insensitive filesystems I believe. The difference is that with this change, memcached can be used on Windows for this lookup. I believe this is already the case for case-sensitive filesystems. 

In #689 it appears that the author added the `os.listdir()` call but did not provide a cached solution.

src/rezplugins/package_repository/filesystem.py
mandeep: @JeanChristopheMorinPerso, I was thinking that since the extra `os.isdir` call will slow down the default case even more that we could just add a check to see if the cache is enabled. If it is enabled then it runs the new cached branch, and if it is disabled then we can just run the old `os.listdir` check as is. 

This can be achieved with a simple if statement or a more involved solution of a new decorator wrapped around listdir.

nrusch:
I've had a note (unfortunately low on my priority list) to look at overhauling the filesystem repository implementation with a focus on eliminating unnecessary filesystem operations. In particular, there are parts of the implementation where the number of filesystem-related operations scales exponentially with the number of package families and packages in a given family.

A central starting point for that effort would be to look at replacing most (if not all) of the bare `listdir`/`os.path` operations with `os.scandir`, since that provides a very fast path to eliminating system calls.

PR #2202, opened 2026-09-17
Fix AttributeError when using ++versions in a suite

### LINKED ISSUE #2046: _rez_fwd ++version causes exception with ++versions
When attempting to use ++versions on a suite-based tool, python crashes.

**Environment**
* OS linux_rocky9_x86_64
* Rez version 3.2.1
* Rez python version 3.9.21
* 
**To Reproduce**
1. Create a suite
2. Call a tool from that suite with "++versions" command line argument

**Expected behavior**
Get the list of versions.

**Actual behavior**
We get a backtrace:
```
usdcat ++versions
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/RELEASE/bin/rez/_rez_fwd", line 8, in <module>
    sys.exit(run_rez_fwd())
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/cli/_entry_points.py", line 92, in run_rez_fwd
    return run("forward")
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/cli/_main.py", line 189, in run
    returncode = run_cmd()
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/cli/_main.py", line 181, in run_cmd
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/cli/forward.py", line 73, in command
    target_func(*nargs, **kwargs)
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/suite.py", line 763, in _FWD__invoke_suite_tool_alias
    retcode = w.run(*(_cli_args or []))
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/wrapper.py", line 85, in run
    return self._run(prefix_char, args)
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/wrapper.py", line 155, in _run
    return self.print_package_versions()
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/wrapper.py", line 249, in print_package_versions
    rows.append((name, pkg.path, label))
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/packages.py", line 213, in __getattr__
    raise AttributeError("Package instance has no attribute '%s'" % name)
AttributeError: Package instance has no attribute 'path'
```
I am new to Rez source code, but the path variable does not appear to ever be set/exist within a Package object, though location does and maybe path should be derived from it?

ClementHector:
**Problem**
Running a suite-based tool with the ++versions argument crashes:
```
usdcat ++versions
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/RELEASE/bin/rez/_rez_fwd", line 8, in <module>
    sys.exit(run_rez_fwd())
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/cli/_entry_points.py", line 92, in run_rez_fwd
    return run("forward")
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/cli/_main.py", line 189, in run
    returncode = run_cmd()
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/cli/_main.py", line 181, in run_cmd
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/cli/forward.py", line 73, in command
    target_func(*nargs, **kwargs)
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/suite.py", line 763, in _FWD__invoke_suite_tool_alias
    retcode = w.run(*(_cli_args or []))
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/wrapper.py", line 85, in run
    return self._run(prefix_char, args)
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/wrapper.py", line 155, in _run
    return self.print_package_versions()
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/wrapper.py", line 249, in print_package_versions
    rows.append((name, pkg.path, label))
  File "/laika/dist/third_party/linux_rocky9_x86_64/rez/v3.2.1/lib64/python3.9/site-packages/rez/packages.py", line 213, in __getattr__
    raise AttributeError("Package instance has no attribute '%s'" % name)
AttributeError: Package instance has no attribute 'path'
```

**Fix**
Use `pkg.uri` instead of `pkg.path` when listing package versions. This is what the rest of the codebase already uses to display a package's location (rez-view, rez-help, rez-pkg-cache,...), so the output is consistent with other commands.

**Tests**
Added `test_print_package_versions in test_suites.py` 
It creates a suite from the `foo` test package, invokes `Wrapper.print_package_versions()` on the `fooer` tool, and asserts:
- the return code is 0
- exactly one row is printed
- the current version is marked with *
- the printed location is the package URI

The test fails with the `AttributeError` above without the fix and passes with it.

Example output after the fix
```
$ fooer ++versions
* foo-1.2.0  /path/to/packages/foo/1.2.0/package.py
  foo-1.1.0  /path/to/packages/foo/1.1.0/package.py
  foo-1.0.0  /path/to/packages/foo/1.0.0/package.py  (local)
```

Closes #2046


maxnbk:
Thanks for the submission, @ClementHector .

I verified locally on linux w/ py3.13. The new test passes, and fails with the earlier `wrapper.py` with the `AttributeError` you specified, so exactly the type of regression insurance we like to see.

The fix direction is right in my eyes. This could have been one plugin-boundary away from working-by-accident, which would have masked the issue.

One consistency note as I discovered it: For combined-type packages (ones that have a single `multi.yaml` file), `pkg.uri` reads like `.../multi.yaml<1.0>` - Angle-brackets and all - So the `++versions` column will show that form rather than a directory-path. That matches what `uri` means, and is arguably more informative (it indicates the condition of coming from a combined file), but it is a visible output format change for those users. I personally think it's fine. The alternative would be special-casing combined packages, which I don't think has a justification currently. Just noting it so that the format change is deliberate, rather than a rude surprise to be discovered later.

CI side, I note that the test fails on Windows due to the discrepancy in naming, where on windows, `create_forwarding_script` appends `.cmd` to the Windows filepath, so `suite.save()` creates `bin/fooer.cmd`, and the tests direct-construction of `suite_path` + `bin` + `fooer` misses it. The existing test suite avoids this by going through `PATH` (the shell resolves so the extensionless version). Yours is the first to construct a `Wrapper` from a saved suite directly, which is why it's the first to hit that issue.

In this case, I might suggest the uncomplicated-if-mildly-tedious fix:
```
tool_file = os.path.join(suite_path, "bin", "fooer")
if platform_.name == "windows":
    tool_file += ".cmd"
w = Wrapper(tool_file)
```

I expect the CI to go green with that, or a similar adjustment.

ClementHector:
Thanks for the thorough review and for verifying locally, @maxnbk.

I've pushed the .cmd adjustment you suggested to see what CI says. However, while looking into it, I found a second issue that I suspect will make the Windows test fail with a different error even with the right file path.

On Windows, create_executable_script writes a batch header at the top of the forwarding script (execution.py):
```
@echo off
_rez_fwd.exe %~dpnx0 %*
goto :eof
:: YAML
```
`forward.py` accounts for this and strips the first 4 lines before calling yaml.load when the file ends with .cmd. But `Wrapper.__init__` does not: it does `yaml.load(f.read())` on the raw file content. 
Since @ is a reserved indicator in YAML, PyYAML will raises.
```
cd src && python -c "from rez.vendor import yaml; yaml.load('@echo off\n_rez_fwd.exe\n...', Loader=yaml.FullLoader)"
```

maxnbk:
Makes sense @ClementHector . Seems to me like it should likely just do what `cli/forward.py` already does when parsing and mimic the strip more-or-less verbatim.

If this turns out to be useful in more than one test, I could see us splicing it out into `rez.utils.execution`, next to where the header is written, that both `forward.py` / `wrapper.py` call and would keep them from drifting apart. But the four-line strip in `Wrapper.__init__` alone would be enough to unblock for now.

So, for the simpler case, I'd propose something like:
```
if platform_.name == "windows" and yaml_file.lower().endswith(".cmd"):
    with open(yaml_file) as f:
        content = "\n".join(f.readlines()[4:])  # strip batch script
else:
    with open(yaml_file) as f:
        content = f.read()
```

If that fixes it for now, I'd maybe want to create a minor github issue for potentially doing what I mentioned above about unifying/portabilizing the wrap/strip calls with their own tests around them.

src/rez/tests/test_suites.py
```diff
@@ -141,6 +146,38 @@ def test_3(self) -> None:
...
+        Regression test for https://github.com/AcademySoftwareFoundation/rez/issues/2046
+        """
+        c_foo = ResolvedContext(["foo"])
+        s = Suite()
+        s.add_context("foo", c_foo)
+
+        suite_path = os.path.join(self.root, uuid.uuid4().hex)
+        s.save(suite_path)
+
+        tool_file = os.path.join(suite_path, "bin", "fooer")
```
JeanChristopheMorinPerso, line 161: Have you considered doing something similar to `test_executable` or maybe extending `test_executable` to call the tool with `++version`?

ClementHector: Thanks @JeanChristopheMorinPerso, that's a much better approach .
I hadn't considered it and it solves several problems at once.
I've rewritten the test to follow the same pattern as test_executable: save the suite, prepend bin/ and rez_bin_path to PATH, and run fooer `++versions` in a subprocess. 
I'd prefer to have a separate test for the `++versions` argument, it provides more clarity (and can be extended later).

ClementHector:
@maxnbk since this test no longer goes through `Wrapper.__init__`, the batch-header issue in `Wrapper.__init__` on Windows is now out of scope for this PR.

Issue #1961, opened 2025-05-02
remove rez-bind quickstart from docs


maxnbk:
rez 'getting started' should remove references to the rez-bind quickstart.

**Motivation**
We've been saying we want to deprecate/remove rez-bind for a long time, but we haven't fully committed to an alternate approach yet, since REP-002 is not yet completed. at the very least, quickstart, which does not work properly on windows, should be removed from the docs, and should follow the information provided in the below youtube video instead:

https://www.youtube.com/watch?v=pYOxL_MVCBQ

JeanChristopheMorinPerso:
Relates to #1654

georgeantonopoulos:
Hi, I would be interested in helping with this documentation cleanup issue. I'm new to ASWF contributions but have been using rez for a while, seems like a great one to start with. Is this still available to work on?

JeanChristopheMorinPerso:
Hi @georgeantonopoulos , yes, this is still available!

georgeantonopoulos:
Hi @JeanChristopheMorinPerso , i have a PR here regarding this:
[](https://github.com/AcademySoftwareFoundation/rez/pull/1982
)
Let me know if you need any changes.

PR #2203, opened 2026-09-17
Rewrite the getting started guide

### LINKED ISSUE #1654: Refresh "Getting Started" documentation
The getting started documentation at https://rez.readthedocs.io/en/stable/getting_started.html is getting old. It's still using `rez-bind --quickstart` which we would like to deprecate, the package build example uses the cmake integration, etc.

A good getting started page would instead:
* Explain how to create the `os`, `platform` and `arch` packages, with a note to mention that this is platform specific, etc.
* Explain how to bootstrap a python package. This will be hard, but we can do out best to point at resources, etc.
* Show a minimal package + build it.
* Show how to rez-env
* Clearly guide our users on what the next steps are (configuring `packages_path`, how to install packages from PyPI, etc).
* Demonstrably works on Linux, macOS and Windows, backed with CI to make sure that it continues to work as expected.

I am probably missing, but this would be a good starting point.

Requirements:
* https://www.bestpractices.dev/en/projects/8389?criteria_level=1#documentation_quick_start

ankurc:
Closes #1654

The page no longer mentions rez-bind --quickstart, which also addresses the narrower request in #1961.


maxnbk:
Cool PR @ankurc , I like the idea of this self-verifying setup via the github-actions.

Overall I feel like it's pretty close to mergeable. I have a few pieces of feedback, mostly nitpicky in substance.

The reference:
```
:data:`platform_map`
```
from `getting_started.rst:72` points at a setting documented via the generated `rez-config` directive, not an autodata target - This resolves fine in the non-notpicky RTD build but is another instance of the docs pattern of referencing config settings by `:data:` where the target is convention-dependent. Harmless at the moment, but consider a link to the `rezconfig.py` setting section instead.

I'm on linux at the moment, so I can't double check the windows behavior here, so grain of salt:

The Windows `hello` invocation diverges from the guide. The workflow itself special-cases Windows as it `cd`'s into the hello_world bin folder and runs `py hello`, whereas the guide claims the `hello` tool "is on your PATH" and shows plain `hello`. This implies the `tools` entry does not put `hello` on PATH the same way on Windows (maybe its .cmd or .exe?)
Recommend either a Windows-specific note in the guide, or checking whether `rez-env hello_world -- hello` truly works there.

ankurc:
Thanks @maxnbk, I worked on both issues.


**Windows**: your hunch was right, and it was neither `.cmd` nor `.exe`, it was nothing at all. `cmd.exe` only resolves names whose extension is in `PATHEXT`, so an extensionless `bin/hello` is invisible however good `PATH` is. The `tools` entry itself was fine.

`build.py` now emits a launcher that is both a batch file and a Python script:

```bat
@python -x "%~f0" %* & exit /b

from hello_world import hello
hello()
```

`-x` skips line one, the slot the shebang normally occupies, so the body is never duplicated.

One trap worth flagging: the binaries get chmod'd `0o550`, and Windows cannot delete a read-only file, so the replaced original needs `stat.S_IWRITE` first. Otherwise the build dies and the install path is silently empty.

Workflow now runs plain `rez-env hello_world -- hello` on all three platforms, green, and verified by hand on Windows. Took your guide-note suggestion too, since `hello.bat` is visible on disk.

maxnbk:
@ankurc Great, that .cmd crisscross shebang mechanic is a new for me, "TIL"! I'm a little sketched out by the weird windows permissions mechanic, but it feels more like Windows fault than ours. At worst maybe an extra comment or two to explain what is going on there in the build.py would be justified.

Overall I like where the PR is. I'm going to take my time to slowly peruse the real resulting documentation for a bit and look for anything that is strange or feels wrong but I feel like even if there are fixes to make, it puts things in better shape than they were, so it's hard not to want to merge it anyway.

ankurc:
Thanks [@maxnbk](https://github.com/maxnbk), yeah you're right, that bit reads like voodoo without context.. dropped comments in build.py for the three non-obvious bits:

1. why we even need a .bat (it's PATHEXT doing it, not PATH)
2. how that first line pulls double duty for cmd and python -x
3. why the original wants stat.S_IWRITE before os.remove

Take your time.. no rush on the docs read-through.

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
+
+The examples below install packages into the :data:`local package path
+<local_packages_path>`, which defaults to ``~/packages``. It is created when the
+first package is installed.
+
+Essential packages
+==================
+
+Rez models the current system as three packages: ``platform``, ``arch`` and
+``os``. They are added to every resolve as
```
JeanChristopheMorinPerso, line 34:
```suggestion
``os``. By default, they are added to every solve via
```

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
-=================
+   PACKAGE   URI
+   -------   ---
+   arch      /home/you/packages/arch/x86_64/package.py
+   os        /home/you/packages/os/rocky-9/package.py
+   platform  /home/you/packages/platform/linux/package.py
 
-Before building your first rez package, ensure that:
+.. warning::
+   :ref:`rez-bind` is going to be deprecated. The current implementation is not
```
JeanChristopheMorinPerso, line 57:
```suggestion
   :ref:`rez-bind` is deprecated. The current implementation is not
```

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
+   actively maintained, and most of its bind modules are likely to fail on
+   Windows.
 
-* The directory ``$HOME/packages`` exists and is writable;
-* The `cmake <https://cmake.org/>`_ tool is available.
+   Even though rez-bind will be deprecated and we generally discourage its use,
```
JeanChristopheMorinPerso, line 61:
```suggestion
   Even though rez-bind is deprecated and its usage is discouraged,
```

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
+   you can safely use it to create the ``arch``, ``os`` and ``platform``
+   packages.
 
-  .. note::
-      CMake is not a hard requirement to build packages in rez. This example just
-      happens to use CMake, but you can use anything you want in your packages.
+These three packages describe *this* machine. Run ``rez-bind os`` separately on
+every operating system and architecture that will use the package repository,
```
JeanChristopheMorinPerso, line 66:
```suggestion
every operating system and architecture where you wish to use rez,
```
I feel like this is more beginner friendly than talking about repositories.

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
+and do not copy the resulting packages between unlike platforms.
```
JeanChristopheMorinPerso, line 67: This last part seems unnecessary IMO. The getting started guide should be straight to the point and not talk about things the user shouldn't do. Generally speaking, users aren't supposed to copy packages around until much later in their rez journey and it's also highly discouraged to manually modify a repository. Saying "do not copy the resulting packages between unlike platforms" might make them think that their might be occasions where this is appropriate.

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
 
-The :ref:`rez-build` tool is used to build packages and install them locally (typically to ``$HOME/packages``).
+The ``os`` version is read from the running machine, so it can be more
+fine-grained than you want to target. Windows reports a build number such as
+``windows-10.0.19045``, and Linux distributions report point releases. Use
```
JeanChristopheMorinPerso, line 71: I would add an example for linux and remove "point releases" to make this more beginner friendly and visual.

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
+``platform_map`` in the :ref:`Package Resolution <config-package-resolution>`
+settings to collapse these onto the versions you actually support, before
```
JeanChristopheMorinPerso, line 73:
```suggestion
the :data:`platform_map` setting to collapse these onto the versions you actually support, before
```

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
 
-  .. note::
-      CMake is not a hard requirement to build packages in rez. This example just
-      happens to use CMake, but you can use anything you want in your packages.
+These three packages describe *this* machine. Run ``rez-bind os`` separately on
+every operating system and architecture that will use the package repository,
+and do not copy the resulting packages between unlike platforms.
 
-The :ref:`rez-build` tool is used to build packages and install them locally (typically to ``$HOME/packages``).
+The ``os`` version is read from the running machine, so it can be more
```
JeanChristopheMorinPerso, line 69: I would move this section at the complete end. A user that is getting started will likely not be interested in that right away. So I would move this into the "next steps" section.

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
-   Even if rez-bind will be deprecated and we generally discourage its use, you can safely use it for creating the ``arch``, ``os`` and ``platform`` packages.
+You need a working Rez installation. :doc:`installation` lists the supported
+Python versions and covers the install itself.
 
-After installation, you need to create some essential Rez packages. The :ref:`rez-bind`
-tool creates Rez packages that reference software already installed on your system.
-Use the :option:`--quickstart <rez-bind --quickstart>` argument to bind a set of standard packages:
+.. important::
+   Install Rez with the ``install.py`` script, not with ``pip install rez``. A
+   pip installation provides the Rez API, but its command line tools are not
```
JeanChristopheMorinPerso, line 16:
```suggestion
   pip installation provides the Rez API, but the command line tools are not
```

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
+   guaranteed to work, and the commands in this guide assume a production
```
JeanChristopheMorinPerso, line 17: A beginner might be confused by "production" here since it's the first time it's mentioned. They might not know that "install.py" is what creates a production isntall unless they actually open that install page.

If we take a step back, mentioning pip at all might be cnofusing. https://rez--2203.org.readthedocs.build/en/2203/installation.html says to `python install.py`. We might simplify all this by saying something like "Rez must be installed via the install script to function correctly".

Thoughts?

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
-The :ref:`rez-build` tool is used to build packages and install them locally (typically to ``$HOME/packages``).
+The ``os`` version is read from the running machine, so it can be more
+fine-grained than you want to target. Windows reports a build number such as
+``windows-10.0.19045``, and Linux distributions report point releases. Use
+``platform_map`` in the :ref:`Package Resolution <config-package-resolution>`
+settings to collapse these onto the versions you actually support, before
+creating these packages for a shared repository.
 
-Once you've done that, you can use them via :ref:`rez-env`, just like any other package:
+A Python package
```
JeanChristopheMorinPerso, line 76: I'm -1 on this section as it is right now. `rez-bind python` is so broken that we should avoid mentioning it in the getting started guide.

I would suggest two options:
* Make the guide use anything else than python as an example
* Create an example package that we will host either in the rez repo or in another repo that will contain a proper python package that can be used as a starting point. That package could use https://github.com/astral-sh/python-build-standalone to avoid needing to compile anything. I can volunteer to do that if you want.

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
+with a site-wide repository as-is.
 
-   Invoking cmake build system...
-   Executing: /usr/local/bin/cmake -d /home/ajohns/workspace/rez/example_packages/hello_world -Wno-dev -DCMAKE_ECLIPSE_GENERATE_SOURCE_PROJECT=TRUE -D_ECLIPSE_VERSION=4.3 --no-warn-unused-cli -DCMAKE_INSTALL_PREFIX=/home/ajohns/packages/hello_world/1.0.0 -DCMAKE_MODULE_PATH=${CMAKE_MODULE_PATH} -DCMAKE_BUILD_TYPE=Release -DREZ_BUILD_TYPE=local -DREZ_BUILD_INSTALL=1 -G Unix Makefiles
-   Not searching for unused variables given on the command line.
-   -- Could NOT find PkgConfig (missing:  PKG_CONFIG_EXECUTABLE)
-   -- Configuring done
-   -- Generating done
-   -- Build files have been written to: /home/ajohns/workspace/rez/example_packages/hello_world/build
+There is no single blessed way to package Python for Rez yet. In practice sites
```
JeanChristopheMorinPerso, line 100:
```suggestion
There is no single blessed way to package Python for rez yet. In practice sites
```
(we spell rez with a capital letter when it starts a sentence and lower case when in the middle of a sentence)

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
 
-   Executing: make -j4 install
-   [100%] Built target py
-   Install the project...
-   -- Install configuration: "Release"
-   -- Installing: /home/ajohns/packages/hello_world/1.0.0/./python/hello_world.py
-   -- Installing: /home/ajohns/packages/hello_world/1.0.0/./python/hello_world.pyc
-   -- Installing: /home/ajohns/packages/hello_world/1.0.0/./bin/hello
+The Rez source contains a small example package in
+``example_packages/hello_world``. Its ``package.py`` is short enough to read in
```
JeanChristopheMorinPerso, line 112:
```suggestion
`:gh-rez:`example_packages/hello_world`. It's package definition is short enough to read in
```

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
+full:
 
-   All 1 build(s) were successful.
+.. literalinclude:: ../../example_packages/hello_world/package.py
```
JeanChristopheMorinPerso, line 115: I'm wondering if we should provide a simpler example that doesn't require python. A less experienced user might struggle to understand what the build script does.

I'm thinking of something as simple as:
```python
name = "hello_world"

version = "1.0.0"

description = "This is your first package"

def commands():
    env.alias("hello-world", "echo 'Hello, World!'")
```

And then tell the user to not worry too much about this for now and to simply `rez-build --install` and `rez-env hello_world` and then run `hello-world`. This would be much more user friendly.

We can always have a second part to the guide (next steps for example) where we guide users through a more complex example and provide explanations for what all of these things do.

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
 
-You have just built your first package, and installed it to the :data:`local package path <local_packages_path>`,
-which defaults to (and is usually kept as) ``$HOME/packages``.
+It builds with a plain Python script, ``build.py``, which lives next to
+``package.py`` and is run by :pkgdef:attr:`build_command`. That means this
+example needs no build tooling beyond the ``python`` package you just created.
 
-For more information, see :doc:`building_packages`.
+Use :ref:`rez-build` to build it and install it to your
+:data:`local package path <local_packages_path>`:
```
JeanChristopheMorinPerso, line 123: I feel like mentioning local_packages_path might be too much for a getting started guide. A getting started guide should be light in details and serve as a way to get users going, confirm that the program works, show them the basic commands, etc. A second guide can start to explain what happened, how to configure things, etc.

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
+
+See :doc:`building_packages` for how the build environment is constructed, how
+to use CMake or another build system, and how to work in a local
+build-install-test cycle.
 
 Testing your package
 ====================
 
-You can use the :ref:`rez-env` tool to request a configured environment containing your package:
+Use :ref:`rez-env` to request a configured environment containing your package:
```
JeanChristopheMorinPerso, line 137:
```suggestion
Use :ref:`rez-env` to create an environment containing your package:
```
(You use a package request to create an environment). I'm being picky, but that's the first page after the install page, before a user has even read the basic concepts. We need to be consistent in how we call things.

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
+   platform-linux     /home/you/packages/platform/linux                                       (local)
+   python-3.9.16      /home/you/packages/python/3.9.16/platform-linux/arch-x86_64/os-rocky-9  (local)
 
-Now you are within the configured environment. The caret (``>``) prefixed to your prompt is a visual
-cue telling you that you're within a rez-configured subshell. Rez does not update the current environment,
-instead it configures a subshell and puts you within it.
+   > ]$
 
-Now you can run the ``hello`` tool in our ``hello_world`` package:
+The ``>`` prefix on your prompt means you are inside a Rez-configured subshell.
```
JeanChristopheMorinPerso, line 162:
```suggestion
The ``>`` prefix on your prompt means you are inside a rez environment.
```

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
+Every package is tagged ``(local)`` because they all came from your local
+package path. Notice that ``python`` was resolved even though you only asked for
+``hello_world``: Rez pulled it in because ``hello_world`` requires it, and it
```
JeanChristopheMorinPerso, line 165:
```suggestion
``hello_world``: rez pulled it in because ``hello_world`` requires it, and it
```

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
 
-.. code-block:: text
+.. note::
+   The paths and prompts shown here are from Linux. On Windows the same resolve
+   is written under ``%USERPROFILE%\packages``, and ``os`` will report a build
+   number such as ``windows-10.0.19045``.
 
-   > ]$ hello
-   Hello world!
+The ``hello_world`` package declares a :pkgdef:attr:`tools` entry, so its
```
JeanChristopheMorinPerso, line 174: That's not true in this specific context. `tools` only has an effect inside suites.

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
-   arch-x86_64        /home/ajohns/packages/arch/x86_64                                            (local)
-   hello_world-1.0.0  /home/ajohns/packages/hello_world/1.0.0                                      (local)
-   os-osx-10.11.5     /home/ajohns/packages/os/osx-10.11.5                                         (local)
-   platform-osx       /home/ajohns/packages/platform/osx                                           (local)
-   python-2.7.11      /home/ajohns/packages/python/2.7.11/platform-osx/arch-x86_64/os-osx-10.11.5  (local)
+   > ]$ hello
+   Hello world!
 
-To exit the configured environment, simply exit the shell using the ``exit`` command or :kbd:`Control-d`:
+Leave the environment with the ``exit`` command or :kbd:`Control-D`.
```
JeanChristopheMorinPerso, line 185: I feel like the previous sentence made more sense, though I'd remove `simply` from it.

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
-When you use this form, the shell is immediately exited after the command runs:
+You can also create a configured environment and run a single command inside it.
+The shell exits as soon as the command finishes, which is the form you want in
+scripts and automation:
 
 .. code-block:: text
 
    ]$ rez-env hello_world -- hello
    Hello world!
-   ]$ █
```
JeanChristopheMorinPerso, line 188: I would add it back, and opint the reader to the absence of `>`, showing that we are not in a rez environment.

docs/source/getting_started.rst
```diff
@@ -2,187 +2,209 @@
...
+
+Next steps
+==========
+
+* Read :doc:`configuring_rez`, and set :data:`packages_path` to the ordered list
```
JeanChristopheMorinPerso, line 199: The first step should be to read the "basic concept" page I think. Configuring should come later.

I think I would see this:
* Read more about the basic concepts
* Learn about releasing
* Deepen knowledge in how to build a package
* Learn about how to make PyPI packages available as rez packages

I wouldn't point to any technical reference (including package definition, package commands, etc). This way, we keep the "tutorial/guide" nature of this page. The other pages will lead the users to the appropriate reference pages in due time.

JeanChristopheMorinPerso (commented):
Thanks @ankurc! I have a few suggestions. I think we should simplify the guide more, remove extra things that a user can discover in other guides/pages and simplify the package example. When reading the getting started guide (even before you changes), I have the impression that there's just too much information.

Hopefully my comments are all easy to address. Let me know if you think otherwise or disagree with them.

